PR #9, opened 2019-08-14
AF-2090: Adding plugin conf to copy editor resources from maven

jesuino:
Adding a maven plugin to retrieve the editor from maven and paste the resources inside the extension resources.

This is the first implementation that could apply also for other plugins. This first commit only includes DMN.

To bring the resources to the app you must run `mvn clean install` in dir `kogito-tooling/packages/vscode-extension`.

After that you should be able to see dmn resources under `kogito-tooling/packages/vscode-extension/src/resources`.


jesuino:
@tiagobento @ederign would you please review?

domhanak:
@jesuino I kinda dislike the fact that we are introducing maven here. Would it be possible to do this without maven?

tiagobento:
@domhanak This project has a dependency on a Java artifact. By using Maven, we can isolate this dependency in a single file (the pom.xml) and keep things clear. Also, many of the things we need to do with this Java artifact (.war) like unzipping and copying certain resources from one place to another is already done well by Maven. It also has a nice `.m2` folder that caches artifacts.

What's your concern with bringing Maven into the build?

tiagobento:
@jesuino That's almost what I had in mind. I think the best place to put the pom.xml would be on `vscode-extension-pack-kogito`, since that's the module that packs de extension. 

Another thing is that I'd like to have it run automatically on one of the build steps. I think if you're running the extension on VSCode or running a build:prod, you should run the `mvn clean package` command. If you think there's another place where we should run it, please do add it to it.

jesuino:
Hello @tiagobento I moved the `pom.xml` to the right dir.

I am concerned that it does not have an `index.html`, but an `index.jsp`, and I am not copying it. I also added the build call (`mvn clean package`) to `build:prod` command.

@domhanak we need the WAR from maven repo, hence why we are using maven to download and retrieve the files we need to place in the plugins resources.

One note is that I had to update the `EditorEnvelopeController.test.tsx.snap` snapshot just to build it, but I didn't commit the changes, this probably should be handled in other PR.

Please let me know if you have any question.

domhanak:
@tiagobento @jesuino I understand that :) I am curious if there is a way to do this without maven efficiently. It is not common to mix maven in js/ts projects so I would like to at least discuss if this is the best way forward. :)

tiagobento:
@jesuino Thx! Another thing I discovered yesterday is that the module to use for the Kogito-enabled DMN editor is `kie-wb-common-dmn-webapp-kogito-runtime`, and this one has an `index.html` file. However, it's not in master yet, only in `DROOLS-3727`, so I think we'll have to wait a little bit to have a good version of this `pom.xml`.

packages/vscode-extension-pack-kogito/pom.xml
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>kogito-tooling-vscode-extension</artifactId>
+  <groupId>org.kogito.tooling</groupId>
+  <packaging>jar</packaging>
+  <version>0.0.0</version>
```
domhanak, line 10: Imho we need a proper versioning here. Let's align it with kogito-tooling version. It will be useful in future, we will know that for X.X.X version we used 7.26 version of editor and for X.X.X+1 we use 7.26.1 etc.

domhanak (changes requested):
LGTM, let's resolve the version pointed out and wait until DROOLS-3727 is merged.

packages/vscode-extension-pack-kogito/pom.xml
tiagobento: This is not going to generate any JARs, so there won't be anything published anywhere. That means that we won't be able to use this version information for anything. The version here won't do us any good :)

For us to discover which version of the BPMN editor was used in a specific version of the extension we should simply look at this file on our version control.

packages/vscode-extension-pack-kogito/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+
+  <properties>
+    <version.dmn.webapp>7.26.0-SNAPSHOT</version.dmn.webapp>
```
manstis, line 13: `7.26.0-SNAPSHOT` is very old and will not contain client-side marshalling.

porcelli:
I don't think it's acceptable to introduce maven into this repository. GWT based editors should publish to NPM.

ederign:
+1 for this Porcelli. I think our target should be publishing the dmn/bpmn editors on npm.

PR #11, opened 2019-08-22
AF-2164: Preparation for BPMN release

tiagobento:
As the client-side editors are coming, I'd like to change the structure a bit so that when they're here we can integrate them more easily.

@jesuino I created this 'package' called `unpacked-gwt-editors`, so that it is the one responsible for unpacking the WARs. The other modules depending on the DMN/BPMN editors will then point to this 'package'. I think it makes thinks simpler and more organized. When we have the definitive `pom.xml` we can just edit the `build:fast` script on the `packages/unpacked-gwt-editors/package.json` file.

@domhanak Could you please review that? I also changed the output path of the JUnit report to `target/` so that we don't commit it by accident.


packages/vscode-extension-pack-kogito-bpmn/tests-it/suite/index.ts
```diff
@@ -26,7 +26,7 @@ export function run(testsRoot: string, callback: (error: any, failures?: number)
     reporter: "mocha-jenkins-reporter",
     reporterOptions: {
       junit_report_name: "VSCode Extension Tests",
-      junit_report_path: "vscode-it-test-report.xml",
+      junit_report_path: "target/vscode-it-test-report.xml",
```
domhanak, line 29: good catch :)

tiagobento:
@domhanak Could you please review it asap? Also, I think there's something wrong with the CI since I see it green on Jenkins and yellow here :(

image

ederign (approved):

domhanak:
@tiagobento it was a Jenkins problem :) it needed restart and these checks got a little bit messed up

PR #12, opened 2019-08-23
AF-2119: Add PatternFly loading spinner

jschuler:
- Update webpack configs to pull in PatternFly
- Add the loading spinner (@srambach)
- This also sets up the project for more work coming in soon to include a PatternFly table/editor


packages/vscode-extension-pack-kogito/package.json
```diff
@@ -68,22 +68,28 @@
...
+    "webpack-merge": "^4.2.1"
   },
   "dependencies": {
     "appformer-js-core": "0.1.0",
     "appformer-js-microeditor-envelope": "0.1.0",
     "appformer-js-microeditor-envelope-protocol": "0.1.0",
-    "appformer-js-vscode-extension": "0.1.0"
+    "appformer-js-vscode-extension": "0.1.0",
+    "node-sass": "^4.12.0",
+    "sass-loader": "^7.3.1"
```
tiagobento: Do we need these as runtime dependencies or they're just used during compile time to process `scss` files?

base.webpack.config.js
```diff
@@ -0,0 +1,109 @@
...
+      {
+        test: /\.tsx?$/,
+        include: path.resolve(dirname, "src"),
+        use: [
+          {
+            loader: "ts-loader",
+            options: {
+              transpileOnly: false,
+              experimentalWatchApi: true,
+              configFile: path.resolve(dirname, "tsconfig.json")
```
tiagobento: What are those options `transpileOnly` and `experimentalWatchApi`?

packages/microeditor-envelope/package.json
```diff
@@ -11,12 +11,15 @@
...
+    "@patternfly/react-core": "^3.75.2",
     "@types/react": "16.8.8",
     "@types/react-dom": "16.8.2",
     "appformer-js-core": "0.1.0",
     "appformer-js-microeditor-envelope-protocol": "0.1.0",
+    "node-sass": "^4.12.0",
     "react": "16.8.0",
-    "react-dom": "16.8.4"
+    "react-dom": "16.8.4",
+    "sass-loader": "^7.3.1"
```
tiagobento: Do we need these (`node-sass` and `sass-loader`) as runtime dependencies or they're just used during compile time to process scss files?

packages/microeditor-envelope/src/LoadingScreen.tsx
```diff
@@ -36,14 +48,26 @@ export function LoadingScreen(props: { visible: boolean }) {
         <div
           style={{
             width: "100vw",
-            height: "100vh",
-            textAlign: "center",
-            backgroundColor: "#1e1e1e",
-            padding: "40px 0 0 0",
-            ...cssAnimation
+            height: "100vh"
           }}
```
tiagobento: Should this inline style be in a `.css` file?

packages/microeditor-envelope/src/index.ts
```diff
@@ -15,6 +15,8 @@
  */
 
 import * as ReactDOM from "react-dom";
+import "@patternfly/react-core/dist/styles/base.css";
+import "@patternfly/patternfly/patternfly-addons.scss";
```
tiagobento: That's odd that you have to import these on `index.ts`. Shouldn't these import statements be moved to where they're actually needed?

packages/vscode-extension-pack-kogito/src/styles/dark.scss
```diff
@@ -0,0 +1,13 @@
...
+
+:not(.vscode-dark) {
+  .pf-c-page {
+    --pf-c-page--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  }
+}
+
+.vscode-dark {
+  @extend .pf-t-dark;
+}
```
tiagobento: Now I get what you said about duplication on `dark.scss`. Since this is used by components on the `microeditor-envelope` package, shouldn't this file be there? This way you avoid the duplication between `vscode-extension-pack-kogito` and `vscode-extension-simple-react`

packages/vscode-extension-pack-kogito/src/webview/index.ts
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+import "@patternfly/react-core/dist/styles/base.css";
+import "@patternfly/patternfly/patternfly-addons.scss";
+import "../styles/dark.scss";
```
tiagobento: Same question as before. Shouldn't these imports be moved?

packages/vscode-extension-pack-simple-react/package.json
```diff
@@ -64,24 +64,36 @@
...
+    "@patternfly/patternfly": "^2.26.4",
+    "@patternfly/react-core": "^3.87.3",
+    "@patternfly/react-inline-edit-extension": "^2.10.5",
+    "@patternfly/react-table": "^2.17.5",
     "appformer-js-core": "0.1.0",
     "appformer-js-microeditor-envelope": "0.1.0",
     "appformer-js-microeditor-envelope-protocol": "0.1.0",
-    "appformer-js-vscode-extension": "0.1.0"
+    "appformer-js-vscode-extension": "0.1.0",
+    "jsonix": "^3.0.0"
```
tiagobento: `jsonix` and `@patternfly/*` dependencies are not needed here, are they? I think they come transitively from `microeditor-envelope`.

packages/vscode-extension-pack-simple-react/src/webview/index.ts
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+import "@patternfly/react-core/dist/styles/base.css";
+import "@patternfly/patternfly/patternfly-addons.scss";
+import "../styles/dark.scss";
```
tiagobento: Same as above.

packages/vscode-extension-pack-simple-react/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
 	"extends": "../../tsconfig.json",
 	"compilerOptions": {
-      "declaration": false
+			"declaration": false
```
tiagobento: Weird formatting? 😝

tiagobento (changes requested):
Thx for the PR @jschuler and @srambach! :tada:
I'm sure that waiting for editors to open will be less of a burden now with this awesome Loading Screen 😃 

I have a few comments inline, please go through them.

As a general comment I think there are a lot of structural changes to the project on this PR, and although I understand most of them are necessary to make CSS and other related files work, some could've been avoided.

I think it's too early on the project to commit to a `base.webpack.config.js` configuration file, since we're still figuring out what's gonna be for the different packages we have here. This compromises the individuality of packages and adds an extra level of indirection on the already dense Webpack configuration files.

I know you guys had the best intention on doing that, probably mostly for avoiding code duplication between projects, but I would like to keep the `webpack.config.js` files independent for each package for the time being.

I also added a few comments inline on dependencies that caught my attention but it would be nice if you could give the `package.json` files another look so that you can catch something I didn't see.

Hope there are more PRs coming in the future! 😄

packages/vscode-extension-pack-simple-react/package.json
porcelli: jsonix is only used by marshalling, don't need it here

packages/vscode-extension-pack-kogito/package.json
jschuler: should be devDependencies

base.webpack.config.js
jschuler: [transpileOnly](https://github.com/TypeStrong/ts-loader#transpileonly-boolean-defaultfalse) can speed up compilation if set to true. The value of `false` here is the default. I can take it out. 
[experimentalWatchApi](https://webpack.js.org/guides/build-performance/#typescript-loader) can "dramatically decreases the number of modules to be rebuilt on each iteration"

packages/microeditor-envelope/package.json
jschuler: should be dev, thanks :)

packages/microeditor-envelope/src/index.ts
jschuler: should be able to move this it into LoadingScreen component directly, once there are more components in this package then we have to move the css/scss up a level

packages/vscode-extension-pack-kogito/src/styles/dark.scss
jschuler: yes you're right i have moved it locally into microeditor-envelope package!

packages/vscode-extension-pack-kogito/.gitignore
```diff
@@ -3,4 +3,5 @@ node_modules
 .vscode-test/
 *.vsix
 dist
-*.dockerimage
\ No newline at end of file
+*.dockerimage
+vscode-it-test-report.xml
```
tiagobento: I addressed this on #11 by moving the report XML to `target/`, you can remove that from here.

tiagobento (commented):
Thx for being so fast @jschuler! Really appreciate that.

Minor nitpick inline on the `.gitignore` file and one more comment about the `webpack.config.js` files. 

I think we should configure the loaders for modules that need them. For example, the `microeditor-envelope-protocol` module does not have any `css/scss/.../etc` files, so I think we could leave the `package.json` and the `webpack.config.js` files of this package unchanged. Does that make sense?

packages/microeditor-envelope/src/LoadingScreen/LoadingScreen.tsx
```diff
@@ -33,17 +45,24 @@ export function LoadingScreen(props: { visible: boolean }) {
...
-            width: "100vw",
-            height: "100vh",
-            textAlign: "center",
-            backgroundColor: "#1e1e1e",
-            padding: "40px 0 0 0",
-            ...cssAnimation
-          }}
-        >
-          <span style={{ fontFamily: "Helvetica", color: "white", fontSize: "12pt" }}>Loading...</span>
+        <div className="loading-screen">
```
tiagobento: We should stil linclude the detructed cssAnimation variable in the style of this `div`. That's used to make a nice "fadeOut" animation when switching from the `LoadingScreen` to the underlying ready editor.

One thing that gave me better results too was to remove line 42 and add an `onTransitionEnd={() => setMustRender(false)}` on this `div`. Worked very well.

tiagobento:
@jschuler Do we still need the changes on `vscode-extension-pack-kogito` and on `vscode-extension-simple-react`? Since we moved the CSS files loading to `microeditor-envelope`, I guess we could remove these changes?

jschuler:
@tiagobento looks like we still need the webpack changes in `vscode-extension-simple-react` otherwise it complains, and looks like `vscode-extension-pack-kogito` is not in src anymore but is built?

jschuler:
Closing in favor of https://github.com/kiegroup/kogito-tooling/pull/19

PR #23, opened 2019-10-18
AF-2287: Incorporate DMN editor

tiagobento:
This PR features the configuration necessary to add support for DMN files.

It also adds a `webpack-dev-server` configuration to be able to locally test the Chrome Extension.
Added a fix to not duplicate `setContent` request on the visualization screen.


tiagobento:
@manstis Keep in mind that building this branch won't work, because we don't commit the editors themselves on this repo. I'm going to create a branch with the editors committed so you can test it better. In the case you wanna use _this_ branch, just copy the WAR contents to `packages/unpacked-gwt-editors/dmn` and you'll be good to go.

tiagobento:
@manstis https://github.com/tiagobento/kogito-tooling/tree/dmn-for-reals-PR-with-editor

packages/gwt-editors/src/GwtEditorWrapper.tsx
```diff
@@ -92,4 +97,10 @@ export class GwtEditorWrapper extends AppFormer.Editor {
       headerTable.remove();
     }
   }
+
+  private injectStyleToFixResponsivenessIssue_DROOLS_3995() {
+    const style = document.createElement("style");
+    style.textContent = ".list-view-pf-body { display: flex !important; }";
```
karreiro: @tiagobento Since it's applied globally for `.list-view-pf-body`, it can potentially change the behaviour in every use of `.list-view-pf-body`.

A safer possible alternative would be this:
```suggestion
    style.textContent = '[data-i18n-prefix="DataTypeListItemView."] .list-view-pf-body { display: flex !important; }';
```

karreiro (commented):
Thank you, @tiagobento. I'm glad the DMN editor is being incorporated.

I've noticed this PR into the [DROOLS-3995](https://issues.jboss.org/browse/DROOLS-3995), so I've left a minor comment.

packages/gwt-editors/src/GwtEditorWrapper.tsx
tiagobento: That's great, @karreiro! Thanks for the suggestion. To be even better, I think we could replace the string's double quotes by single quotes and avoid using the escaping slash :)

manstis (approved):
The DMN editor seems to work fine in both Chrome and VSCode.. so approving :-)

Thank-you @tiagobento

packages/gwt-editors/src/GwtEditorWrapper.tsx
karreiro: @tiagobento I've updated the _Suggested change_ following your suggestion ;-)

ederign (approved):

PR #25, opened 2019-10-22
Introduce GitHub Actions CI

tiagobento:
On this PR:

- Introducing GitHub Workflow for CI on PRs and on `master` branch. Currently running on Ubuntu and on macOS.
- Configure `chrome-extension` module to produce a .zip file with its bundled files so that we can download it directly from the Checks tab on PRs on the Artifacts link. 

NOTE: This Workflow produces complete VSCode and Chrome Extension artifacts. It uses the 0.2.0 tag as reference point to download the editors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
      - name: Fetch editors from last release
        uses: actions/checkout@v1
        with:
+         repository: kiegroup/kogito-tooling
          ref: 0.2.0
      
      - name: Upload DMN editor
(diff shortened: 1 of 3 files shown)


mareknovotny:
Do we will have 2 CI setups? look at https://github.com/kiegroup/kogito-tooling/blob/master/Jenkinsfile

tiagobento:
@mareknovotny For a while, yes. We want to try GitHub Actions as it shows very good potential. It brings the community closer to the project because everyone can see the results, as it's not hidden behind a VPN. And it's completely free for open source projects, like this one.

After some time with the two, we can see which one suits our goals better and choose. There's no harm in letting them both coexist for some time.

ederign (approved):

PR #34, opened 2019-11-05
Add XML formatting to DMN/BPMN editor outputs

### LINKED ISSUE #31: pretty print the xml edited by the kogito plugin so github diff view actually works
i.e. trying to use editor editing https://github.com/quarkusio/quarkus-quickstarts/blob/master/using-kogito/src/main/resources/org/acme/kogito/persons.bpmn2 you end up with a massive diff just to change the position of a box.

The actual diff is just a single parameter change.

if the plugin would pretty print the generated xml this would be  a much smoother experience.

tiagobento:
This PR partially fixes #31.

There's still the issue with the ids on the BPMN editor, but diffs are now better than a one-liner.

### DIFF
--- a/packages/chrome-extension/package.json
+++ b/packages/chrome-extension/package.json
@@ -27,8 +27,7 @@
     "test": "echo 'No tests to run.'",
     "build:fast": "rm -rf dist && webpack",
     "build": "yarn run lint && yarn test && yarn run build:fast",
-    "build:prod": "yarn run build --mode production --devtool none",
-    "serve-envelope": "webpack-dev-server"
+    "build:prod": "yarn run build --mode production --devtool none"
   },
   "babel": {
     "presets": [
--- a/packages/chrome-extension/src/app/components/common/KogitoEditorIframe.tsx
+++ b/packages/chrome-extension/src/app/components/common/KogitoEditorIframe.tsx
@@ -63,6 +63,10 @@ const RefForwardingKogitoEditorIframe: React.RefForwardingComponent<IsolatedEdit
             if (props.readonly) {
               return;
             }
+
+            //keep line breaks
+            content = content.split("\n").join("\\n");
+
             runScriptOnPage(
               `document.querySelector("${GITHUB_CODEMIRROR_EDITOR_SELECTOR}").CodeMirror.setValue('${content}')`
             );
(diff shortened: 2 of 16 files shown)


packages/chrome-extension/package.json
```diff
@@ -27,8 +27,7 @@
     "test": "echo 'No tests to run.'",
     "build:fast": "rm -rf dist && webpack",
     "build": "yarn run lint && yarn test && yarn run build:fast",
-    "build:prod": "yarn run build --mode production --devtool none",
-    "serve-envelope": "webpack-dev-server"
```
domhanak, line 31: is this line removed on purpose?

packages/chrome-extension/src/app/components/common/KogitoEditorIframe.tsx
```diff
@@ -63,6 +63,10 @@ const RefForwardingKogitoEditorIframe: React.RefForwardingComponent<IsolatedEdit
             if (props.readonly) {
               return;
             }
+
+            //keep line breaks
+            content = content.split("\n").join("\\n");
```
domhanak, line 68: just thinking out loud: Is this also windows compatible? I can give it a try later if needed.

packages/gwt-editors/src/DefaultXmlFormatter.ts
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { XmlFormatter } from "./XmlFormatter";
+import * as prettifyXml from "prettify-xml";
+
+export class DefaultXmlFormatter implements XmlFormatter {
+  public format(xml: string) {
+    return prettifyXml(xml, { indent: 2, newLine: "\n" });
```
domhanak: same question: Is this also windows compatible? I can give it a try later if needed.

domhanak (approved):
Looks good, couple of questions regarding Windows compatibility and one removed line. If the line was meant to be removed we need to update readme too.

domhanak:
jenkins retest this

packages/chrome-extension/package.json
tiagobento: Oops! Not at all :)

tiagobento: Oh, duh. Yes! Line 31 was removed on purpose. We now use this exact same command on chrome-extension-pack-kogito-kie-editors instead.

tiagobento:
@domhanak Thx for the review! I think the `\n`s will work on Windows, but it's worth the test :)

domhanak (approved):

tiagobento:
Tested on Windows. Works fine.

tiagobento:
Tested on Windows. Works fine.

ederign (approved):

PR #40, opened 2019-11-21
AF-2350: Resource Content API Implementation

jesuino:
PR for the following JIRAs:

AF-2350: Define Resource Content API (core-api)
AF-2351: Implement Resource Content Envelope integration ( microeditor-envelope-protocol and microeditor-envelope)
AF-2352: Create Resource Content VsCode implementation (vscode-extension)
AF-2315: Resource Content Service integration with Envelope API
AF-2355: Create Resource Content Chrome implementation (chrome-extension)

Related PRs:

https://github.com/kiegroup/appformer/pull/854
https://github.com/kiegroup/kie-wb-common/pull/3034


packages/chrome-extension/src/app/components/common/KogitoEditorIframe.tsx
```diff
@@ -35,6 +37,8 @@ const RefForwardingKogitoEditorIframe: React.RefForwardingComponent<IsolatedEdit
   props,
   forwardedRef
 ) => {
+  const githubAPI = useGitHubApi();
```
tiagobento: githubApi

packages/chrome-extension/src/app/components/common/KogitoEditorIframe.tsx
```diff
@@ -83,6 +87,18 @@ const RefForwardingKogitoEditorIframe: React.RefForwardingComponent<IsolatedEdit
           if (onEditorReady) {
             onEditorReady();
           }
+        },
+        receive_resourceContentRequest(uri: string) {
+          console.log(`Trying to read content from ${uri}`);
```
tiagobento: Change to `console.debug`

packages/microeditor-envelope-protocol/src/EnvelopeBusOuterMessageHandler.ts
```diff
@@ -26,6 +26,8 @@ export interface EnvelopeBusOuterMessageHandlerImpl {
   receive_contentResponse(content: string): void;
   receive_setContentError(errorMessage: string): void;
   receive_dirtyIndicatorChange(isDirty: boolean): void;
+  receive_resourceContentRequest(uri: string): void;
+  receive_resourceListRequest(pattern: string): void;
```
tiagobento: Rename to "globPattern"

packages/chrome-extension/src/app/components/common/ChromeResourceContentService.ts
```diff
@@ -0,0 +1,107 @@
...
+  public tree: GithubAsset[];
+}
+
+class RepoInfo {
+  public owner: string;
+  public repo: string;
+  public gitref: string;
+}
+
+export class ChromeResourceContentService implements ResourceContentService {
```
tiagobento: I would create a factory ( * ) for this Service with no parameters and transfer the Octokit from an instance to a parameter. We can't rely on it being an instance because it can't be "injected" (if you think in terms of DI). Octokit instances are volatile during runtime, so I'd rather have it passed as a first argument on the `read` and `list` methods.

packages/chrome-extension/src/app/components/common/KogitoEditorIframe.tsx
```diff
@@ -35,6 +37,8 @@ const RefForwardingKogitoEditorIframe: React.RefForwardingComponent<IsolatedEdit
   props,
   forwardedRef
 ) => {
+  const githubAPI = useGitHubApi();
+  const resourceContentService = new ChromeResourceContentService(githubAPI.octokit());
```
tiagobento: Here ( * )

tiagobento (changes requested):
Looks good overall! Left minor change requests inline.

packages/chrome-extension/src/app/components/common/ChromeResourceContentService.ts
```diff
@@ -0,0 +1,117 @@
...
+      .catch(e => emptyResponse); // TODO: better handle this
+  }
+
+  public list(pattern: string): Promise<ResourcesList> {
+    const repoInfo = this.repoInfo();
+    const emptyList = new ResourcesList(pattern, []);
+
+    if (!repoInfo) {
+      return Promise.resolve(emptyList);
+    }
```
tiagobento: I think a console.debug wouldn't hurt here. We never expect to not be able to parse the URL, but... :)

tiagobento: Actually, now that I think about it, I guess it's better to throw here. Since we're returning a Promise (and not a Thenable) we actually have an API that gives the ability for consumers to handle errors. If we don't have the "repoInfo", we can't proceed.

packages/chrome-extension/src/app/components/common/ChromeResourceContentService.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  private repoInfo(): RepoInfo | undefined {
+    const info = window.location.pathname.split('/');
+    if (info.length >= 4) {
+      return {
+        owner: info[1],
+        repo: info[2],
+        gitref: info[4]
+      };
+    }
```
tiagobento: If `info.lenth` is 4, we will have an indexing error here. Change to >= 5.

packages/chrome-extension/src/app/components/common/ChromeResourceContentService.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    return fetchFile(
+      this.octokit,
+      repoInfo.owner,
+      repoInfo.repo,
+      repoInfo.gitref,
+      uri
+    ).then(assetContent => new ResourceContent(uri, assetContent))
+      .catch(e => emptyResponse); // TODO: better handle this
+  }
```
tiagobento: You're never returning undefined here. The `emptyResponse` is a ResourceContent with no "content". I think we should consider throwing an exception  if we don't have a "repoInfo" and we should resolve to `undefined` if we end up on the `catch` block of fetchFile.

packages/chrome-extension/src/app/components/common/KogitoEditorIframe.tsx
```diff
@@ -83,6 +87,18 @@ const RefForwardingKogitoEditorIframe: React.RefForwardingComponent<IsolatedEdit
...
           }
+        },
+        receive_resourceContentRequest(uri: string) {
+          console.debug(`Trying to read content from ${uri}`);
+          resourceContentService.read(uri).then(r => {
+            self.respond_resourceContent(r!);
+          });
+        },
+        receive_readResourceContentError(errorMessage: string) {
+          console.log(`Error message retrieving a resource content ${errorMessage}`);
```
tiagobento: `console.debug` instead.

packages/microeditor-envelope/src/ResourceContentEditorCoordinator.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  public exposed(messageBus: EnvelopeBusInnerMessageHandler): ResourceContentEditorService {
+    const pendingResourceRequests = this.pendingResourceRequests;
+    const pendingResourceListRequests = this.pendingResourceListRequests;
+    return {
+      get(uri: string) {
+        messageBus.request_resourceContent(uri);
+        return new Promise((success, failure) => {
```
tiagobento: That's not a rule, but more like a convention. First parameter is `res` or `resolve` and second parameter is `rej` or `reject`. Please check for other Promises too.

packages/chrome-extension/src/app/components/common/ChromeResourceContentService.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+const resourceContentServiceFactory = {
+  create: (octokit : Octokit) => {
+
+    return new ChromeResourceContentService(octokit);
+  
+  }
+}
+
```
tiagobento: Weird formatting? Please make sure all the code is properly formatted.

tiagobento (changes requested):
More minor change requests inline. We're almost there!

packages/chrome-extension/src/app/components/common/GithubInfo.ts
```diff
@@ -0,0 +1,74 @@
...
+    return GitHubPageType.ANY;
+}
+
+function parseRepoInfo(): RepoInfo {
+    const info = window.location.pathname.split('/');
+    if (info.length >= 5) {
+        return {
+            owner: info[1],
+            repo: info[2],
+            gitref: info[4]
```
jomarko: Imagine url `https://github.com/kiegroup/kogito-tooling/pull/40/files`, could you please clarify what will be the values?
owner=keigroup, repo=kogito-tooling, gitref=40? if that is the case, shouldn't we rename `gitref` to `prref`

packages/core-api/src/content/ResourceContentService.ts
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ResourceContent } from "./ResourceContent";
+import { ResourcesList } from "./ResourcesList";
+
+export interface ResourceContentService {
+
+  read(uri: string): Promise<ResourceContent | undefined>;
```
jomarko: I think would be better if we align naming as for chrome extension so for VScode tooling. please use either **get** or **read**.

jomarko (commented):
I think we miss test that would demonstrate **get**/**read** can handle different kind of content in wid (empty, parameters missing, ....) and the **list** works for files with (white space in names, multibyte characters in names, ...)

packages/chrome-extension/src/app/components/common/GithubInfo.ts
jesuino: Hello @jomarko 

Thanks for your great question. The repo info will only be retrieved according to the page we are. For example, this page will not activate the extension, hence it will be ignored.

However, when we are in a BPMN file page or in a PR that contains BPMN files, the extension is activated and then we retrieve RepoInfo accordingly.

Tiago helped me to improve the code to make it clear and the next commit will clarify how it works.

packages/core-api/src/content/ResourceContentService.ts
jesuino: agreed, thanks!

packages/chrome-extension/src/app/components/common/GithubInfo.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    if (uriMatches(`.*/.*/pull/[0-9]+/commits.*`)) {
+        return GitHubPageType.PR;
+    }
+
+    return GitHubPageType.ANY;
+}
+
+export { RepoInfo, discoverCurrentGitHubPageType };
```
tiagobento: `uriMatches` and `discoverCurrentGitHubPageType` methods should go back to where they were and this file should become `RepoInfo.ts`

packages/chrome-extension/src/app/components/common/IsolatedEditorContext.ts
```diff
@@ -15,9 +15,11 @@
  */
 
 import * as React from "react";
+import {RepoInfo} from "./GithubInfo";
```
tiagobento: Format. Please make sure to format all the code.

packages/chrome-extension/src/app/components/common/KogitoEditorIframe.tsx
```diff
@@ -80,13 +86,23 @@ const RefForwardingKogitoEditorIframe: React.RefForwardingComponent<IsolatedEdit
         },
         receive_ready() {
           logger.log(`Editor is ready`);
-          if (onEditorReady) {
-            onEditorReady();
-          }
+          onEditorReady ?.();
```
tiagobento, line 89: Format.

packages/online-editor/src/editor/Editor.tsx
```diff
@@ -57,6 +58,14 @@ const RefForwardingEditor: React.RefForwardingComponent<EditorRef, Props> = (pro
...
       receive_ready() {
         console.info(`Editor is ready`);
+      },
+      receive_resourceContentRequest(uri: string) {
+        console.info(`Resource Content Request`);
+        self.respond_resourceContent(new ResourceContent(uri, undefined));
+      },
+      receive_resourceListRequest(globPattern: string) {
+        console.info(`Resource List Request`);
+        self.respond_resourceList(new ResourcesList(globPattern, []));
```
tiagobento: Please change to `console.debug`

tiagobento (changes requested):
Small changes requested inline. We're almost there!

tiagobento (approved):

PR #47, opened 2020-01-03
KOGITO-652: Context for editors

jesuino:
Creates context for editors where they can retrieve the editor channel using `window.envelope.editorContext.channel` and the channel value is a String which currently has three possible values:

* ONLINE
* VSCODE
* GITHUB 

Also pass the content path when calling `setContent`.

Part of an ensemble:

https://github.com/kiegroup/appformer/pull/873
https://github.com/kiegroup/kie-wb-common/pull/3099

### DIFF
--- a/packages/core-api/src/core/ChannelType.ts
+++ b/packages/core-api/src/core/ChannelType.ts
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *       http://www.apache.org/licenses/LICENSE-2.0
+ *        http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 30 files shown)


manstis (approved):

packages/core-api/src/core/ChannelType.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export enum ChannelType {
+    VSCODE = "VSCODE",
+    ONLINE = "ONLINE",
+    GITHUB = "GITHUB"
```
jomarko, line 20: Don't we need `DEFAULT` as we have in `appformer` part of the ensemble?

jomarko (commented):
Where is some test showing usage of the new path parameter please?

packages/core-api/src/core/ChannelType.ts
jesuino: In Kogito Tooling side we don't have default, we need to select a channel when starting the envelope

ederign (approved):

tiagobento (approved):

jomarko (approved):

PR #55, opened 2020-01-15
KOGITO-876: Fix issue when user was not logged in

tiagobento:


ederign (approved):

ederign:
Reproducible unlogged with the old version and fixed by this one. Thanks for the quick fix @tiagobento 
🔥

paulovmr (approved):

packages/chrome-extension/src/app/components/common/Main.tsx
```diff
@@ -37,6 +37,20 @@ export interface Globals {
...
+    <>
+      {githubApi.userIsLoggedIn() &&
+        ReactDOM.createPortal(
+          <KogitoMenu />,
+          kogitoMenuContainer(props.id, dependencies__.all.notificationIndicator()!.parentElement!)
+        )}
+    </>
+  );
+}
+
```
jomarko, line 53: As we implemented `useGithubApi` function thanks to adding `dependencies` into `GitHubContext` I think we should make `useGithubApi` more general, probably just returning `dependencies__.all.notificationIndicator()`. That would allow us remove importing `dependencies` from `Main` and rewrite this `KogitoMenuPortal` function as
```js
function KogitoMenuPortal(props: { id: string }) {
  const githubApi = useGitHubApi();

  return (
    <>
      {!!githubApi.generalUserIsLoggedIn() &&
        ReactDOM.createPortal(
          <KogitoMenu />,
          kogitoMenuContainer(props.id, githubApi.generalUserIsLoggedIn()!.parentElement!)
        )}
    </>
  );
}
```

WDYT?

jomarko (commented):
### Code review

packages/chrome-extension/src/app/components/common/Main.tsx
tiagobento: Thx for the review @jomarko. While your suggestion makes sense from a sort of "DRY" perspective, it unfortunately does not apply in this case. Explaining:

1. `GitHubContext` should provide a way for its consumers to know whether the user is logged in or not.
2. `Dependencies` holds functions which return references to DOM elements that are part of the GitHub UI. These elements are special to us since we use them to put our extension's components in place.
3. In this case, I'm using a DOM element from `Dependencies` to implement the `userIsLoggedIn` method on `GitHubContext`. Notice that I could've done it another way, but this one was stable and easy enough for me.
4. On `Main`, I'm using `Dependencies` to obtain a reference to a special DOM element, in this case, the `notificationIndicator`.
5. Implementing your suggestion would make `GitHubContext` the owner of a function that returns a reference to a special DOM element, and that actually _breaks_ encapsulation and separation of concerns, since that's `Dependencies` job.

jomarko: @tiagobento thank you for your explanation. I have to agree that making the `notificationIndicator` available in `GitHubContext` and reusing in `Main` would break principles of `encapsulation` and `single responsibility`.

However I still think more appropriate way would be create class with responsibility of getting reference to dom elements. That new class would be reused as in `GitHubContext` so in `Main` and we wouldn't repeat the code `dependencies__.all.notificationIndicator()`. I can imagine repeated code `dependencies__.all.notificationIndicator()` can be risky, if public github API will change (github developers decide to announce log in status differently). 

I will probably approve and not block this PR anymore, as there is nothing like perfect code, but would like to hear your opinion.

jomarko (approved):
@tiagobento thank you. I tried manually scenario described in [KOGITO-876](https://issues.redhat.com/browse/KOGITO-876) and it works. So approving.

However still would like to hear your opinion about my https://github.com/kiegroup/kogito-tooling/pull/55/files#r368844173 even if you have time to respond after merge.

tiagobento:
Thanks @jomarko! Yes, that's a good suggestion. Making the 'dependencies' methods into methods of a class is a good idea. Feel free to create a JIRA for it if you'd like, but I think this refactor will happen naturally in the near future.

PR #56, opened 2020-01-15
KOGITO-860: Add error alerts for when Online Editor fails to open a file URL

paulovmr:


tiagobento (approved):

ederign (approved):

packages/online-editor/src/index.tsx
```diff
@@ -38,21 +54,75 @@ if (urlParams.has("ext")) {
...
   );
 }
 
-if (!urlParams.has("ext") && !urlParams.has("file")) {
+function showResponseError(statusCode: number, description: string) {
   ReactDOM.render(
-    <App iframeTemplateRelativePath={"envelope/index.html"} file={EMPTY_FILE} readonly={false} external={false} />,
+    <div className={"kogito--alert-container"}>
+      <Alert
+        variant="danger"
```
jomarko: Can we replace "danger" with `{AlertVariant.danger}` ?

packages/online-editor/src/index.tsx
```diff
@@ -38,21 +54,75 @@ if (urlParams.has("ext")) {
...
+        title="An error happened while fetching your file"
+        action={<AlertActionLink onClick={goToHomePage}>Go to Home Page</AlertActionLink>}
+      >
+        <br />
+        <b>Error details: </b>
+        {statusCode}
+        {statusCode && description && " - "}
```
jomarko, line 95: Do we need to repeat `statusCode` twice? Once standalone and once in combination with `description`. I would expected the combination `statusCode` and `description` is enough. Do I miss something?

packages/online-editor/src/index.tsx
```diff
@@ -19,10 +19,26 @@ import * as ReactDOM from "react-dom";
...
 
 const urlParams = new URLSearchParams(window.location.search);
 
 if (urlParams.has("ext")) {
+  waitForEventWithFileData();
+} else if (urlParams.has("file")) {
+  openFileByUrl();
+} else {
+  openDefaultOnlineEditor();
+}
```
jomarko, line 32: I am sorry, as new member of team, could somehow provide three sample urls to check all three options?

jomarko (commented):
@paulovmr thank you for PR, some questions directly here and some also on [jira](https://issues.redhat.com/browse/KOGITO-860?focusedCommentId=13953114&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13953114)

packages/online-editor/src/index.tsx
paulovmr: For the first condition, you have to use the button "Open in Online Editor" in GitHub with the Chrome extension installed. An URL similar to the following will be opened, and the page will wait for an browser event fired by the extension to fill the editor's content: `http://localhost:9001/?ext#/editor/dmn`.

For the second condition, here is an URL: `http://localhost:9001/?file=https://raw.githubusercontent.com/ederign/KieDora/master/Drawing%201-3.dmn#/editor/bpmn`.

For the third condition, any URL that has neither `file` or `ext` path params, like: `http://localhost:9001`.

packages/online-editor/src/index.tsx
paulovmr: Done

packages/online-editor/src/index.tsx
paulovmr: In line 95, `statusCode` and `description` are only conditions, that return `true` if they are defined. So " - " is only written if both are defined.

jomarko (approved):
thank you @paulovmr

PR #58, opened 2020-01-23
KOGITO-734: Remove scroll from pages on both Chrome and Safari

tiagobento:


paulovmr (approved):

manstis:
I thought you was only targeting Chrome?

If you're targeting Safari too; then surely we'd need to cover that permutation in the GWT editors?

tiagobento:
@manstis I thought Chrome and Safari shared the same 'target' on GWT compilation.. isn't that the case? If it's not, then I think we should have another permutation. cc @ederign @paulovmr

manstis:
@tiagobento Yeah my mistake.. looks like Chrome uses `safari` in it's User Agent.

ederign (approved):

jomarko (approved):
Works fine! Thank you @tiagobento

PR #61, opened 2020-02-06
KOGITO-930: Write unit tests for Chrome Extension components

tiagobento:
On this PR:
- Introduced [React Testing Library](https://github.com/testing-library/react-testing-library) (RTL) to test React components. It's approach is similar to enzyme except for the fact that it encourages [black-box tests](https://en.wikipedia.org/wiki/Black-box_testing). It also provides a very easy way of interacting with the DOM to simulate user actions. It works really well with React Hooks!
- Refactored `dependencies.ts` to the `Dependencies` class. Components now use GitHub's distinct DOM elements through the Global context.
- Updated React and ReactDOM versions to 16.12.0
- Code formats and small overall cleanups
- Not all tests were created, so I left placeholders for the future. The tests I created were selected to serve as a guideline for the next persons writing RTL tests.


tiagobento:
@jomarko @tomasdavidorg I requested reviews from both of you because I know that Tomas has recently started working with testing kogito-tooling, but it's not imperative that both of you review this PR. Please feel free to coordinate amongst yourselves who can review that. Thx! 😊

tomasdavidorg:
Hi @tiagobento, thanks for letting me know. Next week I am on PTO so I won't have time to take a look. I hope @jomarko find some time for this PR. Thanks.

packages/chrome-extension/src/__tests__/app/Dependencies.test.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("open github and test dependencies", () => {
+  test("test 1", async () => {
+    expect(true).toBeTruthy();
```
jomarko, line 19: Most tests from **packages/chrome-extension/src/__tests__/app** are like this, is it expected? do we have a jira for future work?

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorApp.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { usingTestingGitHubContext, usingTestingGlobalContext } from "../../../testing_utils";
+import { removeAllChildren } from "../../../../app/utils";
+
+beforeAll(() => {
+  chrome.extension = {
+    getURL: jest.fn((path: string) => {
+      return `chrome-testing://${path}`;
+    })
+  } as any;
+});
```
jomarko, line 29: Could you please help me understand this?

I thought we mock behavior of `getURL` in the context of `chrome.extension` but can not find usage of that method.

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorApp.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+            toolbarContainer={newDivOnBody()}
+            iframeContainer={newDivOnBody()}
+            githubTextEditorToReplace={newDivOnBody()}
+            fileInfo={testFileInfo}
+          />
+        ).wrapper
+      ).wrapper
+    );
+
+    expect(document.body).toMatchSnapshot();
```
jomarko, line 88: Does this mean we render totally same both in read and non readonly mode?

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorApp.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+            iframeContainer={newDivOnBody()}
+            githubTextEditorToReplace={newDivOnBody()}
+            fileInfo={testFileInfo}
+          />
+        ).wrapper
+      ).wrapper
+    );
+
+    fireEvent.click(screen.getByTestId("go-fullscreen-button"));
+    expect(document.body).toMatchSnapshot();
```
jomarko, line 110: General question about `expect(document.body).toMatchSnapshot();`

When is a snapshot captured? I am asking because we use the snippet `expect(document.body).toMatchSnapshot();` in all tests, what kind of makes me think we have still the same screen == in full screen, when click to open online editor.

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorToolbar.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+    ).toMatchSnapshot();
+  });
+
+  test("readonly false | textMode false | textModeEnabled true", () => {
+    expect(
+      render(
+        <SingleEditorToolbar
+          readonly={true}
+          textMode={false}
+          textModeEnabled={true}
```
jomarko, line 73: Shouldn't be also false?

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorToolbar.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  test("see as source", () => {
+    const onSeeAsSource = jest.fn();
+    const onSeeAsDiagram = jest.fn();
+
+    const component = render(
+      <SingleEditorToolbar
+        readonly={false}
+        textMode={false}
+        textModeEnabled={true}
```
jomarko, line 92: Same question as above

packages/chrome-extension/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import { GitHubPageType } from "../app/github/GitHubPageType";
+
+function setWindowLocationPathname(pathname: string) {
+  window = Object.create(window);
+  delete window.location;
+  window.location = { pathname: pathname } as any;
+}
+
+describe("extractFileInfoFromUrl", () => {
+  test("", async () => {
```
jomarko: can we name the test?

jomarko (commented):
Some questions, otherwise looks good

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorApp.test.tsx
tiagobento: Sure. `chrome` is a global variable and is only available at runtime, just like "provided" dependencies on the Java world. Here we mock one of its properties because it's used on `DefaultChromeRouter`.

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorApp.test.tsx
tiagobento: No. Snapshots are separate for individual tests, so `expect(document.body).toMatchSnapshot();` actually means: Verify that the document.body for this test is exactly the same of the snapshot for this test.

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorApp.test.tsx
tiagobento: Snapshots are captured when you first create the test. It's the programmers job to verify that the first snapshot is good. After that, tests exist basically to ensure that the renderings don't change. If they do and that's expected, we should delete the snapshots or update them with `jest -U`

tiagobento: See explanation on the comment above as well :)

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorToolbar.test.tsx
tiagobento: No, they're different test cases.

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorToolbar.test.tsx
tiagobento: No, they're different test cases.

packages/chrome-extension/src/__tests__/app/Dependencies.test.ts
tiagobento: Yes, I left them there as placeholders for tests that need to be implemented. I'll create a follow-up JIRA to track the remaining work.

packages/chrome-extension/src/__tests__/index.test.ts
tiagobento: Sure!

packages/chrome-extension/src/__tests__/app/components/single/SingleEditorToolbar.test.tsx
tiagobento: No, they're different test cases.

paulovmr (approved):

tiagobento:
@jomarko Please tell me if it's okay to merge.

jomarko (approved):
@tiagobento ok, thank you, I think we can merge them

PR #62, opened 2020-02-06
KOGITO-736: Adding links to external editor

jesuino:
Adding links to external online editor.

This is how it looks like:

image


jesuino:
This feature in action:

final_online_editor_link

packages/chrome-extension/src/app/components/tree/externalEditorLinkManager.tsx
```diff
@@ -0,0 +1,87 @@
...
+function getLinkContainers(): HTMLElement[] {
+    return Array.from(document.querySelectorAll(LINK_CONTAINER_SELECTOR));
+}
+function getLinkToAsset(parent: HTMLElement): (HTMLAnchorElement | null) {
+    return parent.querySelector(LINK_SELECTOR);
+}
+
+function getLinkParentId(linkId: string) {
+    return "external_link_" + linkId;
+}
```
tiagobento: There's a lot going on here. This class like the other equivalents is responsible for assembling a React application. To do that, it has to interact with known DOM elements provided by GitHub. 

What I want to change on this file:

1. Any dependency on GItHub's DOM elements should be part of the Dependencies class.
2. The mutation observer should be used with a 'useEffect' like on the PR screen
3. There should only be one call to ReactDOM.render, so you'll have to use React portals to render multiple components inside different DOM trees.

If you have any questions please talk to me directly.

tiagobento (changes requested):
Thanks for the quick PR @jesuino! I left a comment inline but I'd like to highlight the importance of proper code formatting. Looks like there are some unformatted code left.

kywalker-rh:
This is really great! I think a better icon for this use would be the external link icon (<ExternalLinkAltIcon />) since it takes you to a separate webpage.

jesuino:
Thank you, @kywalker-rh I changed the icon to arrow

image

jesuino:
Sometimes looks like there's some sort of cache on github DOM that does not allow the extension to work at first time a repo is rendered. I will rebase this PR and check the issue with more details

cache_issue

packages/chrome-extension/src/app/components/tree/FileTreeWithExternalLink.tsx
```diff
@@ -0,0 +1,66 @@
...
+            return (
+              <OpenExternalEditorButton
+                id={id}
+                href={createTargetUrl(container, props.externalEditorManager)}
+                container={container}
+              />
+            );
+          }
+        })}
+    </>
```
tiagobento: Too much going on on the JSX part here. Please extract code that's not related to the UI to a useMemo() or something that you think fits. I'd recommend first deciding which fileLinkContainers are eligible for the "Open on external editor" link and then just render the `OpenExternalEditorButton` component for them.

packages/chrome-extension/src/app/components/tree/FileTreeWithExternalLink.tsx
```diff
@@ -0,0 +1,66 @@
...
+function isBlob(container: HTMLElement) {
+  const fileLink = treeView.fileLinkTarget(container);
+  return fileLink && fileLink.pathname.split("/")[3] === "blob";
+}
+function createTargetUrl(container: HTMLElement, externalEditorManager?: ExternalEditorManager): string {
+  const fileLink = treeView.fileLinkTarget(container)!;
+  const split = fileLink.pathname.split("/");
+  split.splice(0, 1);
+  split.splice(2, 1);
+  const linkToOpen = split.join("/");
```
tiagobento: These lines could become a separate method with a unit test associated.

packages/chrome-extension/src/app/components/tree/FileTreeWithExternalLink.tsx
```diff
@@ -0,0 +1,66 @@
...
+            );
+          }
+        })}
+    </>
+  );
+}
+
+function isBlob(container: HTMLElement) {
+  const fileLink = treeView.fileLinkTarget(container);
+  return fileLink && fileLink.pathname.split("/")[3] === "blob";
```
tiagobento: You can use optional chaining here: `return treeView.fileLinkTarget(container)?.pathname.split("/")[3] === "blob";`

packages/chrome-extension/src/app/components/tree/OpenExternalEditorButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <div id={props.id} className="float-right">
+          <a href={props.href} target="blank" title="Open in Online Editor">
+            <ArrowIcon />
+          </a>
+        </div>,
+        props.container
+      )}
+    </>
+  );
+}
```
tiagobento: To avoid passing the container so much deep into the UI tree, you can leave the Portal outside of this component and let its parent decide where to render it ;)

packages/chrome-extension/src/app/components/tree/externalEditorLinkManager.tsx
```diff
@@ -0,0 +1,42 @@
...
+        editorIndexPath={args.editorIndexPath}
+        resourceContentServiceFactory={args.resourceContentServiceFactory}
+        externalEditorManager={args.externalEditorManager}
+      >
+        <FileTreeWithExternalLink router={args.router} externalEditorManager={args.externalEditorManager} />
+      </Main>,
+      createAndGetMainContainer(args.id, dependencies__.all.body())
+    );
+  }
+}
```
tiagobento: This looks almost perfect. Remember the problem you were having with the links not appearing sometimes when the page opens? I think you could take a look at the `PrEditorsApp`. There I configured a state that has the elements which have to be 'Kogitoized" (😛). The initial state looks at what the page has at the moment the Chrome Extension renders, but then the MutationObserver chnages this state with the new elements that come later into the page, so by changing this state, React helps us to render the missing elements. All that is necessary because GitHub apparently renders the page incrementally, so we have to take that into consideration. Hope that helps!

tiagobento (changes requested):
Changes requested inline 😊

packages/chrome-extension/src/app/components/tree/FileTreeWithExternalLink.tsx
jesuino: I removed the need to filter blob by selecting the blob links only

packages/chrome-extension/src/app/components/tree/OpenExternalEditorButton.tsx
jesuino: Ok, I am creating the portal in the parent component.

jesuino:
Hello @tiagobento  - thank you so much for your review. I made changes to the code and also effectively observed the files container for changes.

Thanks!

packages/chrome-extension/src/app/components/tree/FileTreeWithExternalLink.tsx
```diff
@@ -0,0 +1,91 @@
...
+function htmlElementChangeListener(action: () => void, target: HTMLElement) {
+  useEffect(() => {
+    const elementObserver = new MutationObserver(e => action());
+    elementObserver.observe(target, {
+      childList: true,
+      subtree: true
+    });
+    return () => elementObserver.disconnect();
+  });
+}
```
tiagobento: This is what's called a "Custom Hook". Custom Hooks are very much encouraged on React, but there's a convention to follow. You should always start the hook's with a 'use' word. Also, I would invert the parameters order 😝

packages/chrome-extension/src/app/components/tree/FileTreeWithExternalLink.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export function FileTreeWithExternalLink() {
+  const { externalEditorManager, router, dependencies } = useGlobals();
+
+  const filteredLinks = useCallback(
+    () =>
+      dependencies.treeView
+        .linksToFiles()
+        .filter(fileLink => router.getLanguageData(extractOpenFileExtension(fileLink.href) as any))
+        .filter(fileLink => !document.getElementById(externalLinkId(fileLink))),
```
tiagobento: This could be an external function on this same file. No need for it to be inside the component, right?

packages/chrome-extension/src/app/components/tree/FileTreeWithExternalLink.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+function htmlElementChangeListener(action: () => void, target: HTMLElement) {
+  useEffect(() => {
+    const elementObserver = new MutationObserver(e => action());
+    elementObserver.observe(target, {
+      childList: true,
+      subtree: true
+    });
+    return () => elementObserver.disconnect();
+  });
```
tiagobento: I think you forgot the empty dependencies `[]`. The way it is now, every time this component is rendered, this MutationObserver gets destroyed and re-created.

tiagobento (changes requested):
Almost there! Minor comments inline.

jesuino:
ops, added my changes to the last commit by mistake, but the changes are in https://github.com/kiegroup/kogito-tooling/pull/62/commits/76fc0774c82a3bfdd3cff0078e5d2088d92a68a4

would you please review?

tiagobento (approved):

PR #65, opened 2020-02-12
KOGITO-861: Add an input to paste a file URL in the Online Editor home page

jesuino:
KOGITO-861: Add an input to paste a file URL in the Online Editor home page

Validates and redirect users to edit a valid file URL from online editor home page.

Happy path:

success_path

If user tries to use invalid URL we make three validations:

* Check if it is a valid URL;
* If it has an extension, which is important for us to redirect to the correct editor;
* Finally check if we support that extension.

You can see this in action in gif below

invalid_URLs

Notice I am changing the app location and skipping the react router so the file can be correctly fetched by index.tsx


kywalker-rh:
Looks great @jesuino!

srambach:
LGTM! 👍

srambach (approved):

packages/online-editor/src/App.tsx
```diff
@@ -88,7 +88,7 @@ export function App(props: Props) {
             <EditorPage onFileNameChanged={onFileNameChanged} />
           </Route>
           <Route exact={true} path={routes.home.url({})}>
-            <HomePage onFileOpened={onFileOpened} />
+            <HomePage router={onlineEditorRouter} onFileOpened={onFileOpened} />
```
paulovmr: @jesuino you can access the router in the HomePage.tsx by using `context.router`.

jesuino: fixed in https://github.com/kiegroup/kogito-tooling/pull/65/commits/2d2f38a527dde0b7fbb8d2d86f016141cd2fa759

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -128,8 +135,37 @@ export function HomePage(props: Props) {
...
+      setIsValidFileUrl(false);
+      return;
+    }
+    if (!context.router.getLanguageData(fileType)) {
+      setInvalidFileText(`File type ${fileType} is not supported.`);
+      setIsValidFileUrl(false);
+      return;
+    }
+
+    window.location.href = `?file=${inputFileUrl}#/editor/${fileType}`;
```
tiagobento: Hum..... this looks like something that we should be making with the help of our `Router`.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -56,6 +59,10 @@ export function HomePage(props: Props) {
   const uploadInputRef = useRef<HTMLInputElement>(null);
   const uploadBoxRef = useRef<HTMLDivElement>(null);
 
+  const [inputFileUrl, setInputFileUrl] = useState("");
+  const [isValidFileUrl, setIsValidFileUrl] = useState(false);
+  const [invalidFileText, setInvalidFileText] = useState("");
```
tiagobento: These two states behave very similarly. I think you could create an enum describing the state of the input (e.g. BLANK, VALID, INVALID_URL, INVALID_EXTENSION) and derive the message from there. This way you have only one state variable and your component is less complex. Also, it's easier to test, since we're deriving the message from the input state. Of course you can also derive the 'isValidFileUrl' from this enum.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -128,8 +135,37 @@ export function HomePage(props: Props) {
...
+  const inputFileChanged = (fileUrl: string) => {
+    setInputFileUrl(fileUrl);
+    try {
+      new URL(fileUrl);
+      setIsValidFileUrl(true);
+    } catch (e) {
+      setIsValidFileUrl(false);
+      setInvalidFileText("Enter a valid URL");
+    }
+  };
```
tiagobento: Remember to use the Hooks! Here's a good case for 'useCallback' with `[]` as its dependencies. This prevents React from creating a new function every time and saves unnecessary renders :)

tiagobento (changes requested):
Changes requested inline 😊

jesuino:
@tiagobento I had to make a big commit to re-organize the code and put editors and home page behind the router, I also had to organize error messages and created an error page. Would you please review?

I still need to work on small details, but it is working now.

Thanks!

jesuino:
Hello @tiagobento , as we talked I created KOGITO-1202 to focus on routes changes and decided to keep window.location change for now, please see https://github.com/kiegroup/kogito-tooling/pull/65/commits/3549d78769337eb447dc52ea732de31645e1f5ec and let me know your thoughts

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -310,34 +362,40 @@ export function HomePage(props: Props) {
...
+              <Title headingLevel="h2" size="2xl">
+                Import source code
+              </Title>
+            </CardHeader>
+            <CardBody isFilled={false}>Paste a URL to a source code link (GitHub, Dropbox, etc.)</CardBody>
+            <CardBody isFilled={true}>
+              <FormGroup
+                label="URL"
+                fieldId="url-text-input"
+                isValid={inputFileUrlState === InputFileUrlState.VALID}
```
srambach: Although I see the invalid helper text show up, I don't see the form element displaying as invalid (red underline, red !) This is keyed off of the aria-invalid attribute, which is still showing as false (i.e. the field is marked as valid)

srambach:
@jesuino This looks really nice. Just one other question - it looks like the enter key doesn't open the given URL. Would that be possible to implement?

jesuino:
@srambach Thank you so much for your input
The reason why enter was not working and the invalid field message was not red is because the form group was not wrapped by a form element. It should all work now (see below). 

@tiagobento would you please review my last commit? Thanks!

This is how it looks like now:

form_validation_improvement

srambach:
Can a blank input not be marked as invalid? So it only turns red/shows the icon when the field loses focus? That way we aren't showing them an error if they aren't even trying to import from a url, and not when they are still trying to type it in.
(maybe it's that way, it's hard to see from the video)
Also, thanks for fixing the error display :)

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -128,8 +135,68 @@ export function HomePage(props: Props) {
...
+      }
+
+      if (!context.router.getLanguageData(fileType)) {
+        setInputFileUrlState(InputFileUrlState.INVALID_EXTENSION);
+        return false;
+      }
+      setInputFileUrlState(InputFileUrlState.VALID);
+    } catch (e) {
+      setInputFileUrlState(InputFileUrlState.INVALID_URL);
+    }
```
tiagobento: I believe the exception happens on the `new URL` call, right? So the try catch block should be only there. We don't want to catch other exceptions.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -128,8 +135,68 @@ export function HomePage(props: Props) {
...
+      default:
+        return "";
+    }
+  }, [inputFileUrlState]);
+
+  const externalFileFormSubmit = useCallback(
+    e => {
+      e.preventDefault();
+      e.stopPropagation();
+      return openFile();
```
tiagobento: Shouldn't it be just `openFile()`?

tiagobento (changes requested):
Just one change request and a question inline. Looking good @jesuino!

jesuino:
@srambach I had the exactly same question, I was expecting your opnion on this, thanks :-) 

@tiagobento  I made the changes

page_last_test

srambach:
@jesuino Great! If I clear the field and leave, it stays invalid. Any chance of clearing the invalid state in that case? Otherwise, it looks great!
Enter key working <3

jesuino:
Hello Sarah,

Thanks for the suggestion. I did implement this behavior in latest version.

Please let me know if you have any question. Thanks!

tiagobento (approved):
I'm approving that but I'd like to let @srambach give the final approve. @ederign Please do merge it once @srambach approves it. Thanks @jesuino!

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -310,34 +386,41 @@ export function HomePage(props: Props) {
...
+            </CardHeader>
+            <CardBody isFilled={false}>Paste a URL to a source code link (GitHub, Dropbox, etc.)</CardBody>
+            <CardBody isFilled={true}>
+              <Form onSubmit={externalFileFormSubmit} disabled={!validatedInputUrl}>
                 <FormGroup
                   label="URL"
                   fieldId="url-text-input"
-                  helperText="http://"
+                  isValid={validatedInputUrl}
+                  helperTextInvalid={messageForState}
```
srambach, line 403: It looks like without a `helperText` prop, when the field is valid, there is an a11y violation because there is no div matching the `aria-describedby` id. I _think_ the "http://" text could be put as the `helperText` and the 3 error states would provide `helperTextInvalid`.

srambach: Looks like just adding a prop `helperText="http://"` will fix the a11y violation. 👍

jesuino: Thank you so much again, Sarah, I added the missing property

srambach (approved):
Looks great!! 👍

PR #71, opened 2020-03-03
KOGITO-779: Exclude html files from webapps

romartin:
Hey @manstis @tiagobento 

Thanks for pointing this out! Excluded `html` files from the webapps...

Assembly with:
https://github.com/kiegroup/kie-wb-common/pull/3203
https://github.com/kiegroup/kie-wb-common/pull/3205

### DIFF
--- a/packages/kie-bc-editors-unpacked/pom.xml
+++ b/packages/kie-bc-editors-unpacked/pom.xml
@@ -71,6 +71,7 @@
                         </goals>
                         <configuration>
                             <excludes>WEB-INF/</excludes>
+                            <excludes>**/*.html</excludes>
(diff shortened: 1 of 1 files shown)


tiagobento (approved):

manstis (commented):
Won't this also exclude `index.html` that we need?

tiagobento:
@manstis We don't need the `index.html` to open the editor. From what I know you guys use the `index.html` to test it during development.

manstis (approved):

romartin:
yep @manstis , as @tiagobento said, I noticed the `html` files are not being used by the plugins, but were also deployed, so removed too!!!

PR #72, opened 2020-03-05
KOGITO-883: VSCode Integration: Make undo/redo work with existing VSCode commands

pefernan:
Hey @tiagobento those are the changes, just for you to take a look.

### DIFF
--- a/packages/vscode-extension/src/KogitoActiveEditor.ts
+++ b/packages/vscode-extension/src/KogitoActiveEditor.ts
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { KogitoEditor } from "./KogitoEditor";
+import { Disposable } from "vscode";
+import * as vscode from "vscode";
+
+export class KogitoActiveEditor {
+  private static readonly UNDO_COMMAND_ID = "undo";
+  private static readonly REDO_COMMAND_ID = "redo";
+
+  public readonly editor: KogitoEditor;
+
+  private readonly undoCommand: KogitoActiveEditorCommand;
+  private readonly reCommand: KogitoActiveEditorCommand;
+
+  constructor(editor: KogitoEditor) {
+    this.editor = editor;
+
+    this.undoCommand = new KogitoActiveEditorCommand(KogitoActiveEditor.UNDO_COMMAND_ID, () => {
+      editor.notifyUndo();
+    });
+    this.reCommand = new KogitoActiveEditorCommand(KogitoActiveEditor.REDO_COMMAND_ID, () => {
+      editor.notifyRedo();
+    });
+  }
+
+  public dispose() {
+    this.undoCommand.dispose();
+    this.reCommand.dispose();
+  }
+}
+
+class KogitoActiveEditorCommand {
+  private readonly id: string;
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/KogitoActiveEditor.ts
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+import { KogitoEditor } from "./KogitoEditor";
+import { Disposable } from "vscode";
+import * as vscode from "vscode";
+
+export class KogitoActiveEditor {
+  private static readonly UNDO_COMMAND_ID = "undo";
+  private static readonly REDO_COMMAND_ID = "redo";
```
jomarko: Question from newbie, shouldn't we use "more unique" command ids? Won't this conflict with something else in VS code? I mean, won't this conflict with something else in VS code? For example undo/redo will work in BPMN/DMN but stop to work in other VScode editors/extensions?

jomarko (commented):
One question here

packages/vscode-extension/src/KogitoActiveEditor.ts
tiagobento: Hi @jomarko, these ids are defined by VSCode. It's a little bit odd, I know, but take a look here: https://code.visualstudio.com/docs/getstarted/keybindings#_basic-editing.

tiagobento: Can you make those be inline strings, @pefernan? I feel like we're never going to reuse that and having it directly inline gives a better idea that this string is not defined by us.

packages/vscode-extension/src/KogitoActiveEditor.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  public readonly editor: KogitoEditor;
+
+  private readonly undoCommand: KogitoActiveEditorCommand;
+  private readonly reCommand: KogitoActiveEditorCommand;
```
tiagobento: Rename to `redoCommand`?

packages/vscode-extension/src/KogitoActiveEditor.ts
```diff
@@ -0,0 +1,60 @@
...
+  constructor(editor: KogitoEditor) {
+    this.editor = editor;
+
+    this.undoCommand = new KogitoActiveEditorCommand(KogitoActiveEditor.UNDO_COMMAND_ID, () => {
+      editor.notifyUndo();
+    });
+    this.reCommand = new KogitoActiveEditorCommand(KogitoActiveEditor.REDO_COMMAND_ID, () => {
+      editor.notifyRedo();
+    });
+  }
```
tiagobento: You can remove the extra braces on the last parameter, I think it will make the code look better.

packages/vscode-extension/src/KogitoEditor.ts
```diff
@@ -182,6 +178,24 @@ export class KogitoEditor {
     this.panel.reveal(this.viewColumn(), true);
   }
 
+  public notifyUndo() {
+    this.executeUndRedo(() => this.envelopeBusOuterMessageHandler.notify_editorUndo());
+  }
+
+  public notifyRedo() {
+    this.executeUndRedo(() => this.envelopeBusOuterMessageHandler.notify_editorRedo());
+  }
```
tiagobento: You can inline those?

packages/vscode-extension/src/KogitoEditor.ts
```diff
@@ -182,6 +178,24 @@ export class KogitoEditor {
...
+
+  private executeUndRedo(runnable: () => void) {
+    if(this.enabledUndoRedo) {
+      runnable();
+      this.enabledUndoRedo = false;
+      setTimeout(()=> {
+        this.enabledUndoRedo = true;
+      }, 100);
```
tiagobento: Extra braces also not needed.

tiagobento (changes requested):
Looks good to me! Requested some minor changes inline. Thanks @pefernan!

packages/vscode-extension/src/KogitoActiveEditor.ts
pefernan: yea... those ID's are the ones vscode defines for the undo/redo commands. Other commands have more complex Id's, maybe they weren't so creative when defining these ones ;)

tiagobento (approved):

jomarko (approved):
Seems working fine in VS code.

PR #77, opened 2020-03-10
Release 0.2.8

paulovmr:


ederign (approved):

packages/kie-bc-editors-unpacked/pom.xml
```diff
@@ -71,7 +71,6 @@
                         </goals>
                         <configuration>
                             <excludes>WEB-INF/</excludes>
-                            <excludes>**/*.html</excludes>
```
manstis, line 74: Out of interest can you please explain why these are added back?

@tiagobento advised `.html` pages were not needed for `kogito` however I suspect we now need `index.html` for the JS library injection (DMN's Documentation, FEEL expression editing and ACE editor issue reported recently - "3-) Document tab is not working (works on VSCode)" in Eders email).

Also, please note the editors' WARs have two `.html` pages per editor (`index.html` and `test.html`). 

Please see https://github.com/kiegroup/kie-wb-common/pull/3205#pullrequestreview-367844499 (PR pending merge) and https://github.com/kiegroup/kie-wb-common/pull/3203/files#diff-aa2fb706735c6976d355059a43f82c93 (PR merged). I suspect we'd still want to exclude `test.html`.

@romartin FYI.

tiagobento: Hi @manstis. We had to remove that because the WEB-INF directly was not being excluded too. At the time of the release, we only thought of reverting it, since we identified that it was the cause of the WEB-INF directory not being ignored.

Now, looking at the documentation of the `maven-dependency-plugin`, I know that the way to add other patterns to ignore is by separating them with commas. See https://maven.apache.org/plugins/maven-dependency-plugin/unpack-mojo.html#excludes

We continue to not need the index.html file and, by design, there are no way we could ever need it. However, it may be the case that additional JS/CSS files are created, in that case, we need to update the [GwtEditorRoutes](https://github.com/kiegroup/kogito-tooling/blob/master/packages/kie-bc-editors/src/GwtEditorRoutes.ts) file to include the files for the specific file extensions.

Hope that's helpful and thanks for taking the time to look at the release commit!

ederign: @tiagobento could you please create a jira for this?

manstis: Thank-you for the explanation @tiagobento (I like to understand things).

romartin: good to know, thanks guys! :+1:

tiagobento: @ederign https://issues.redhat.com/browse/KOGITO-1430

PR #83, opened 2020-03-12
KOGITO-1427: VSCode Integration: Redo only executed once (new VSCode API (1.43.0))

pefernan:
@tiagobento could you please take a look?

### DIFF
--- a/packages/microeditor-envelope/src/api/stateControl/KogitoCommandRegistry.ts
+++ b/packages/microeditor-envelope/src/api/stateControl/KogitoCommandRegistry.ts
@@ -29,6 +29,7 @@ export interface KogitoCommandRegistry<T> {
   isEmpty(): boolean;
   getCommands(): T[];
   clear(): void;
+  setMaxSize(size: number): void;
 }
 
 export class DefaultKogitoCommandRegistry<T> implements KogitoCommandRegistry<T> {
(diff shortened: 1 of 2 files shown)


tiagobento:
HI @pefernan. Thx for the PR! Looks fine to me. Would it be possible to use the same commit on `master`?

packages/microeditor-envelope/src/api/stateControl/KogitoCommandRegistry.ts
```diff
@@ -29,20 +29,30 @@ export interface KogitoCommandRegistry<T> {
...
   clear(): void;
+  setMaxSize(size: number): void;
 }
 
 export class DefaultKogitoCommandRegistry<T> implements KogitoCommandRegistry<T> {
   private readonly messageBus: EnvelopeBusInnerMessageHandler;
 
   private maxStackSize = 200;
   private commands: Array<KogitoCommand<T>> = [];
+  private _removedCommands: string[] = [];
```
tiagobento, line 40: Why the underscore?

tiagobento: Maybe we should rename that to 'undoneCommands'? Since they come from an 'undo' action?

tiagobento (approved):
LGTM. Comments inline.

PR #86, opened 2020-03-16
KOGITO-766: Kogito Keyboard Shortcuts API

tiagobento:
On this PR:
- Created and exposed Keyboard Shortcuts API for editors to use inside Kogito channels
- Created an overlay to display active keyboard shortcuts. The experience is the same across all Kogito channels (see Video below)
- Unified PatternFly configuration across all channels, since now the envelope needs a "full" version of PatternFly inside it. Build configuration files are less complex and we're now sure that everyone is using the same PatternFly loaders. Artifact size reduced a little bit too.
- Exposed a new "context" variable with the current Operating System

Video: https://issues.redhat.com/secure/attachment/12467822/Screen%20Recording%202020-03-16%20at%2012.59.11.mov

**VSIX for testing:** https://drive.google.com/file/d/1pKGSpn6xHm3sll8TH0XOc3fp6FMHjkN6/view?usp=sharing

Related PRs:
https://github.com/kiegroup/lienzo-core/pull/102
https://github.com/kiegroup/lienzo-tests/pull/84
https://github.com/kiegroup/appformer/pull/922
https://github.com/kiegroup/kie-wb-common/pull/3221
https://github.com/kiegroup/kogito-tooling/pull/86


packages/microeditor-envelope/src/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
```diff
@@ -84,7 +84,7 @@ export function KeyBindingsHelpOverlay(props: { keyboardShortcuts: KeyboardShort
         <KeyboardIcon />
       </div>
 
-      <Modal title={"Keyboard shortcuts"} isOpen={showing} isLarge={true} onClose={toggle}>
+      <Modal title={"Keyboard shortcuts"} isOpen={showing} width={"60%"} onClose={toggle}>
```
srambach: Does `isSmall` give you a size you like rather than using `width`? The advantage of that is that it uses a max-width rather than width, so that at a smaller screen width, it's not 60% of the narrow window.

tiagobento: @srambach `isSmall` made the modal too narrow, and Liz thought `isLarge` made it too wide, so we kinda agreed at 60%. See https://issues.redhat.com/browse/KOGITO-1437

packages/microeditor-envelope/src/api/keyboardShortcuts/DefaultKeyboardShortcutsService.ts
```diff
@@ -0,0 +1,243 @@
...
+    if (e.metaKey) {
+      pressedKeySet.add(ModKeys.META);
+    }
+    if (e.altKey) {
+      pressedKeySet.add(ModKeys.ALT);
+    }
+    if (e.shiftKey) {
+      pressedKeySet.add(ModKeys.SHIFT);
+    }
+    if (Array.from(MODIFIER_KEY_NAMES.keys()).indexOf(e.code) === -1) {
```
karreiro, line 227: I think the `KeyboardEvent.code` has an odd behavior on different keyboard layouts.

> **Warning:** This ignores the user's keyboard layout, so that if the user presses the key at the "Y" position in a QWERTY keyboard layout (near the middle of the row above the home row), this will always return "KeyY", even if the user has a QWERTZ keyboard (which would mean the user expects a "Z" and all the other properties would indicate a "Z") or a Dvorak keyboard layout (where the user would expect an "F").
> Source: [https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent)

Also, [caniuse.com](https://caniuse.com) seems to prefer [KeyboardEvent.key](https://caniuse.com/#search=KeyboardEvent.key) more than [KeyboardEvent.code](https://caniuse.com/#search=KeyboardEvent.code).

Does the `KeyboardEvent.code` have any benefits over the `KeyboardEvent.key`?

Thanks, and great stuff! :-)

karreiro (commented):
Thank you, @tiagobento. This ensemble brings great value for BC/Kogito editors.

I just notice a minor thing and I've left an inline comment. Please let me know wdyt :-)

packages/microeditor-envelope/src/api/keyboardShortcuts/DefaultKeyboardShortcutsService.ts
tiagobento: That's a good catch, @karreiro. I'm not aware of any advantages of one over the other, but I *think* I saw that KeyboardEvent.key was deprecated somewhere? Have to double-check. Thanks for the pointer anyway, it's always good to have a fresh pair of eyes looking at our codebase! I'll get back to you as soon as I can :)

karreiro: Thank-you, @tiagobento!

`KeyboardEvent.key` is not deprecated, the `KeyboardEvent.keyCode` is the deprecated one :-)

ederign (changes requested):
@tiagobento can you please improve the test coverage for this?

tiagobento:
@ederign Sure. I'll do when I resume this task.

packages/microeditor-envelope/src/api/keyboardShortcuts/DefaultKeyboardShortcutsService.ts
tiagobento: @karreiro As we discussed earlier today, we can't use KeyboardEvent.key because it does not suit our needs of shortcuts we need to register (e.g. `Alt + E` and `Alt + I` both produce a "Dead" code on macOS, which can result in unexpected behaviour when registering these shortcuts).

tiagobento:
Resuming this PR, I'll start writing tests.

paulovmr (approved):

ederign (approved):

karreiro (approved):
Thank you, @tiagobento.

ederign:
@jomarko @karreiro can you please take a look at this?

jomarko:
### Manual Check
This list is just for my easier tracking progress
- [x] VS Code
- [x] Online Editor
- [x] Desktop
- [x] Github Extension

### Code review
Is pending

jomarko:
### Manual review - 01
See pr-review-popups attached to https://issues.redhat.com/browse/KOGITO-766. There is minor issue with shortcuts panel when other popup is shown.

Firstly search overlays shortcuts panel and secondly shortcuts panel can be shown when dmn constraint popup active. 

It is minor issue. I would like to ask what is your opinion. Should I create some ticket for it or there is chance ~0 prevent these overlays.

jomarko:
### Manual review - 02
We have set DMN specific shortcuts. In DMN grids we invoke context menu with Ctrl + Space. In DMN data types we can edit, save, insert field data type by keyboard. Those Shortcuts are not listed. I expect it is because we register them differently than using new API. Do we plan to unify registration?

For more details see pr-review-dmn-specific attached to https://issues.redhat.com/browse/KOGITO-766

jomarko:
### Manual review - 03 
Shouldn't we list also shortcuts below? I am not sure as they kind of conflict what is generally available in applications:
- Undo/Redo
- Save
- Zoom

jomarko:
### Manual review - 04
Similar to 01, however this is on desktop app. When shortcuts panel is active it is in foreground and the editor is in background. however save and close buttons are still usable. Not sure if this is real issue so asking what the rest of the team thinks about it.

Screenshot from 2020-05-21 12-04-38

jomarko:
### Manual review - 05
Just question. The shortcuts icon is in left down corner of screen, not left down corner of  canvas of online editor. Is that OK?
Screenshot from 2020-05-21 12-27-44

jomarko:
### Manual review - 06
We miss shortcuts panel in chrome github extension
Screenshot from 2020-05-21 13-21-46

packages/chrome-extension-pack-kogito-kie-editors/src/envelope/index.ts
```diff
@@ -30,5 +30,5 @@ MicroEditorEnvelope.init({
     }
   },
   editorFactory: new GwtEditorWrapperFactory(gwtAppFormerApi, new DefaultXmlFormatter()),
-  editorContext : { channel: ChannelType.GITHUB }
+  editorContext: { channel: ChannelType.GITHUB, operatingSystem: getOperatingSystem() }
```
jomarko, line 33: I looked on `getOperatingSystem()` impl. we have 3 constants while appformer codebase have 4 constans, DEFAULT is additional there, shouldn't we align?

packages/microeditor-envelope/src/EditorEnvelopeView.tsx
```diff
@@ -59,6 +67,7 @@ export class EditorEnvelopeView extends React.Component<Props, State> {
   public render() {
     return (
       <>
+        {!this.state.loading && <KeyBindingsHelpOverlay keyboardShortcuts={this.props.keyboardShortcuts} context={this.props.context}/>}
```
jomarko, line 70: is it ok we check `this.state.loading` even before we render `LoadingScreenPortal()` ?

packages/microeditor-envelope/src/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+export function KeyBindingsHelpOverlay(props: { keyboardShortcuts: KeyboardShortcutsApi; context: EditorContext }) {
+  const [showing, setShowing] = useState(false);
+
+  const toggle = useCallback(() => {
+    setShowing(!showing);
+  }, [showing]);
+
+  const keyBindings = useMemo(() => {
+    return removeDuplicatesByAttr(props.keyboardShortcuts.registered(), "combination")
```
jomarko, line 43: What duplicates we need to remove? isn't that sign we register same things multiple times? If that is the case doesn't decrease performance such duplicit registration?

packages/microeditor-envelope/src/__tests__/EditorEnvelopeController.test.tsx
```diff
@@ -89,7 +90,12 @@ afterEach(() => {
 });
 
 async function startController() {
-  await controller.start(envelopeContainer);
+  const context = { channel: ChannelType.VSCODE, operatingSystem: OperatingSystem.WINDOWS };
+  await controller.start({
+    container: envelopeContainer,
+    keyboardShortcuts: new DefaultKeyboardShortcutsService(context),
+    context: context
```
jomarko, line 97: Commenting here. I think I didn't found place in codebase where we would obtain some data from context. Seems to me we are just passing it through classes. Could you please double check? My point is we shouldn't add it as `start` argument if it is not needed in final.

packages/microeditor-envelope/src/__tests__/EditorEnvelopeView.test.tsx
```diff
@@ -18,15 +18,23 @@ import * as React from "react";
...
   let view: EditorEnvelopeView;
+  const context = { channel: ChannelType.VSCODE, operatingSystem: OperatingSystem.WINDOWS };
   const render = shallow(
-    <EditorEnvelopeView exposing={self => (view = self)} loadingScreenContainer={loadingScreenContainer} />
+    <EditorEnvelopeView
+      keyboardShortcuts={new DefaultKeyboardShortcutsService(context)}
+      context={context}
+      exposing={self => (view = self)}
+      loadingScreenContainer={loadingScreenContainer}
+    />
```
jomarko, line 37: So we do not test the icon of shortcuts panel was added?

jomarko:
### Manual review - 07
Do we plan some ordering in the shortcuts panel? Just asking. As there is no option to search in panel.

packages/microeditor-envelope/src/__tests__/KeyBindingsHelpOverlay.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import * as React from "react";
+import { KeyBindingsHelpOverlay } from "../KeyBindingsHelpOverlay";
+import { DefaultKeyboardShortcutsService } from "../api/keyboardShortcuts";
+import { ChannelType, OperatingSystem } from "@kogito-tooling/core-api";
+import { fireEvent, render } from "@testing-library/react";
+
+describe("KeyBindingsHelpOverlay", () => {
+  test("snapshot", async () => {
```
jomarko: could we improve the name of this test please?

packages/microeditor-envelope/src/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+      .filter(k => !k.opts?.hidden)
+      .map(k => {
+        return {
+          combination: handleMacOsCombination(k.combination, props.context),
+          category: k.label.split("|")[0]?.trim(),
+          label: k.label.split("|")[1]?.trim()
```
jomarko, line 49: I think we should document somewhere that registered label should contain '|'. Or ts doesn't thrown and exception if spit doesn't find such char?

packages/microeditor-envelope/src/__tests__/KeyBindingsHelpOverlay.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+import * as React from "react";
+import { KeyBindingsHelpOverlay } from "../KeyBindingsHelpOverlay";
+import { DefaultKeyboardShortcutsService } from "../api/keyboardShortcuts";
+import { ChannelType, OperatingSystem } from "@kogito-tooling/core-api";
+import { fireEvent, render } from "@testing-library/react";
+
+describe("KeyBindingsHelpOverlay", () => {
```
jomarko, line 23: I tried to increase the coverage as https://github.com/tiagobento/kogito-tooling/pull/13 but getting fail, would you help me  to understand me the fail?

jomarko (commented):
Code review, reviewed 17/31

tiagobento:
@jomarko Thanks for the review. See remarks below. Code review will be replied inline.

**Manual review - 01**
Wil fix.

**Manual review - 02**
Those shortcuts were left out of this first iteration. An effort [1] was made to enable them to be added in the future, but they'll be addressed separately.

[1] https://github.com/kiegroup/kie-wb-common/pull/3221/commits/02a95287680037c4b07902821f9cb089f43fb586

**Manual review - 03**
Undo/redo shortcuts are channel-specific. We have to think better about that, so we probably should create a new JIRA.

**Manual review - 04**
We don't want to block saving because the Keyboard Shortcuts overlay is displayed. The editor is still open, so you should be able to save your work normally.

**Manual review - 05**
Well... :)
This icon is added in the envelope, not in the editor, so it's not aware of UI changes in the editor. We can probably address that in the future, but I'd say not in this PR.

**Manual review - 06**
The icon should be there. If you built the Chrome Extension using `yarn build:prod`, the extension is pointing to the latest release, instead of your local changes. Please refer to README.md to more details on how to run the Chrome Extension with locally modified editors/envelope.

**Manual review - 07**
Adding a "search" option is a good idea. Since this panel is assembled dynamically, it will be "weird" to sort it IMHO. The shortcuts overlay respects the order in which they were registered.

CC @ederign

packages/microeditor-envelope/src/EditorEnvelopeView.tsx
tiagobento: Yes.

packages/microeditor-envelope/src/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
tiagobento: Yes, that's an issue on Stunner side. Some shortcuts are registered twice, but they're only executed once. I assumed it has to do with the Session, but I didn't dig deeper into it and added this workaround. We can create a JIRA for it.

packages/microeditor-envelope/src/__tests__/EditorEnvelopeView.test.tsx
tiagobento: We do. See `KeyBindingsHelpOverlay.test.tsx` The icon is rendered there.

packages/microeditor-envelope/src/__tests__/KeyBindingsHelpOverlay.test.tsx
tiagobento: Done.

packages/microeditor-envelope/src/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
tiagobento: It won't throw an exception because we're using Optional Chaining.

packages/microeditor-envelope/src/__tests__/KeyBindingsHelpOverlay.test.tsx
tiagobento: I think you should call a `cleanup` between tests. See this [1] for reference. Also, if that doesn't fix the issue, you should update the snapshots with `yarn run test -u` on `packages/microeditor-envelope` package.

[1] https://testing-library.com/docs/react-testing-library/api#cleanup

packages/chrome-extension-pack-kogito-kie-editors/src/envelope/index.ts
tiagobento: It's not common to return an Optional in Java as it is in TypeScript. I prefer to keep it that way, so that both APIs are idiomatic to their own languages.

packages/microeditor-envelope/src/__tests__/EditorEnvelopeController.test.tsx
tiagobento: This is used indirectly through `useContext`.

packages/microeditor-envelope/src/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
tiagobento: https://issues.redhat.com/browse/KOGITO-2266

tiagobento:
Created https://issues.redhat.com/browse/KOGITO-2268 for **Manual review - 03**

tiagobento:
Created https://issues.redhat.com/browse/KOGITO-2267 for **Manual review - 05**

tiagobento:
There was already https://issues.redhat.com/browse/KOGITO-1459 for **Manual review - 02**

packages/microeditor-envelope/src/__tests__/keyboardShortcuts/DefaultKeyboardShortcutsService.test.ts
```diff
@@ -0,0 +1,156 @@
+import { DefaultKeyboardShortcutsService, KeyboardShortcutsApi } from "../../api/keyboardShortcuts";
```
jomarko, line 17: We miss license

packages/microeditor-envelope/src/__tests__/keyboardShortcuts/DefaultKeyboardShortcutsService.test.ts
```diff
@@ -0,0 +1,156 @@
...
+import { fireEvent } from "@testing-library/react";
+
+describe("DefaultKeyboardShortcutsService", () => {
+  test("keyPress", async () => {
+    const keyboardShortcutsApi = new DefaultKeyboardShortcutsService({
+      operatingSystem: OperatingSystem.LINUX,
+      channel: ChannelType.ONLINE
+    });
+
+    const [wasFired] = resolveWhenKeyPressed("ctrl+a", keyboardShortcutsApi);
```
jomarko, line 28: Just question about naming. Is this intentionaly prefixed with `resolve...` as we call `Promise.resolve` inside? Just askung because to me it seems that the meaning is more `registerWhenKeyPressed`

packages/microeditor-envelope/src/__tests__/keyboardShortcuts/DefaultKeyboardShortcutsService.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    expect(keyboardShortcutsApi.registered().length).toStrictEqual(1);
+    fire("keydown", { metaKey: true, code: "KeyA" });
+    await wasFired;
+
+    expect(keyboardShortcutsApi.registered().length).toStrictEqual(1);
+  });
```
jomarko, line 48: I added same test for WIN but is failing, need to double check manually WIN

packages/microeditor-envelope/src/api/keyboardShortcuts/DefaultKeyboardShortcutsService.ts
```diff
@@ -0,0 +1,250 @@
...
+
+        if (e.type === "keydown") {
+          if (setsEqual(this.combinationKeySet(combination), this.pressedKeySet(e))) {
+            console.debug(`Fired (down) [${combination}]!`);
+            onKeyDown(e.target);
+          }
+        } else if (e.type === "keyup") {
+          if (
+            this.combinationKeySet(combination).has(MODIFIER_KEY_NAMES.get(e.code) ?? "") ||
+            this.combinationKeySet(combination).has(e.code)
```
jomarko, line 119: Could you please clarify why we can not again use 
```js
setsEqual(this.combinationKeySet(combination), this.pressedKeySet(e))
```

packages/microeditor-envelope/src/api/keyboardShortcuts/KeyboardShorcutsApi.ts
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { KeyBinding } from "./DefaultKeyboardShortcutsService";
+
+/**
+ * PUBLIC ENVELOPE API
+ */
+export interface KeyboardShortcutsApi {
```
jomarko, line 22: Could we improve docs of this interface?

packages/microeditor-envelope/src/__tests__/keyboardShortcuts/DefaultKeyboardShortcutsService.test.ts
jomarko: I amnually tested VSCODE and provided VSIX on WIN. Seems working.

packages/chrome-extension-pack-kogito-kie-editors/src/envelope/index.ts
jstastny-cz: @tiagobento not sure what you mean, Optional type is expected to be returned even by APIs to express that the method sometimes does not return a value (e.g. one should check what they got).

packages/microeditor-envelope/src/__tests__/keyboardShortcuts/DefaultKeyboardShortcutsService.test.ts
tiagobento: Done.

packages/microeditor-envelope/src/__tests__/keyboardShortcuts/DefaultKeyboardShortcutsService.test.ts
tiagobento: The goal of this method is to return a Promise that resolves once a key combination is pressed. We do so by registering a keyboard shortcut with given combination and calling `res()` on the keyboard shortcut action. Naming was a bit tricky on this structure but I think the way it is reflects best what's going on.

packages/microeditor-envelope/src/api/keyboardShortcuts/DefaultKeyboardShortcutsService.ts
tiagobento: Because when you lift your finger out of a key, `e.code` is actually the code of the key that you released, so we only have access to one key at the time. To identify `keyups` successfully, we have to observe if you released a modifier key or another key. Since the `keyup` event will always happen after one or more `keydown`s, we can simply check if one of the keys in the combination was present in the `keyup` event.

packages/microeditor-envelope/src/api/keyboardShortcuts/KeyboardShorcutsApi.ts
tiagobento: Done.

packages/chrome-extension-pack-kogito-kie-editors/src/envelope/index.ts
tiagobento: Sorry, I got confused. Optional types are not a good practice as parameters, not return types :) I'll change the API on appformer.

jomarko (approved):

PR #87, opened 2020-03-16
KOGITO-1320: Prevent editors from opening with an unsupported file

ljmotta:
## Task
https://issues.redhat.com/browse/KOGITO-1320

### Upload a file with the  `Choose a local file` button
- Only the supported files will be showed by default (.dmn, .bpmn and .bpmn2).
- If an unsupported file is selected, an error message will be displayed for 2 seconds and will be hidden in 0.3 seconds.

### Upload a file with the `Drag and Drop` zone
- Only the supported files will be accepted (.dmn, .bpmn and .bpmn2).
- If an unsupported file is dropped on the drag and drop zone, the color will be changed to red for 2 seconds and changed back to the original color in 0.3 seconds.

### Demo
ezgif com-video-to-gif

## Considerations
- It was observed that on the Chrome browser sometimes a bug occurs on the React.DragEvent when the `File` is undefined.


tiagobento (approved):
Thanks for the PR, @ljmotta! Looks great :)

Code looks great. Just one thing I thought is that maybe there are PatternFly variables with colours that we should use.. do you know that @srambach? Also, if you could give you opinion on how it looks, it would be very nice! 

Thanks! :))

srambach:
Thanks, @tiagobento - yes, the Patternfly color `--pf-global--danger-color--100` should be used here for the red. Patternfly doesn't recommend the pinkish background color, for one it reduces constrast below the guideline for a11y. I think the red border and message is enough, but if you disagree then I'd go with an Empty State component with the warning-triangle icon inside the upload box (the bullseye layout would center it).
I also see a padding in the new CSS; we should use either the utility classes for spacing https://www.patternfly.org/v4/documentation/core/utilities/spacing or the global spacer variables `--pf-global--spacer--xs` etc.
We also have a couple of global variables for animation - `--pf-global--TimingFunction` and `--pf-global--Transition` which might be useful here but I have not looked specifically at those to see.
Nice feature, @ljmotta thanks for adding it!

ljmotta:
Thanks for the reply! @srambach
I changed the colors and the padding as you asked but the animation is a little more complicated. It would be necessary to change a couple of things on the code and it would make it more complex. Hope that is ok :smile: 
Could you please review it again? @tiagobento @srambach

tiagobento (approved):
LGTM. I'll wait for @srambach to give the final review so we can merge it. Thanks @ljmotta :)

srambach:
@tiagobento @ljmotta  Thank you for switching to use PF variables. When I had a chance to look at this a little more, I am concerned about the way the A11y (doesn't) work for this method of reporting the error. There is no indication to a screen reader that a problem has occurred. In addition, the error message is transient, which can also be an a11y problem. The way it is done here looks nice, but I think that the right thing to do would be to use an Alert.
https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/alerts-and-notifications#main-content
If time is of the essence, I am fine with approving this so that there is error handling in place, but I believe there should be a Jira created to work with design to replace this with an appropriate alert.
@kywalker-rh

tiagobento:
@srambach Ok, that makes sense. So are we good to merge that as is and create a separate JIRA to improve a11y later?

domhanak:
I tested in Firefox and Chrome. Works well.

One small issue, in chrome, I wasn't able to select only supported files (*.dmn,*.bpmn,*bpmn2) like in firefox when I used 'Choose local file'

ljmotta:
@srambach I did't thought about the a11y at all, sorry about that, I'm going to pay more attention to that. Already create the JIRA (https://issues.redhat.com/browse/KOGITO-1517). Thanks for the review.

@domhanak That's a little odd, here's what I did (tested with Chrome 80.0.3987.132):
chrome


Please tell me if I'm missing something.

domhanak:
@ljmotta Yea I wrote in in a confusing way. Disregard. :) All works well.

domhanak (approved):

PR #91, opened 2020-03-24
KOGITO-1220: Business Modeler Hub Preview

tiagobento:


tiagobento:
~CI won't pass since there's a need to install `wine` to build the Electron app for Windows.~
Ci will now build on macOS, Linux and Windows. Each one will generate its own artifacts. On local machines, only the artifact for your platform will be generated.

domhanak:
@tiagobento Ping me when I am able to test :)

packages/hub/src/electron/Menu.ts
```diff
@@ -0,0 +1,120 @@
...
+  constructor(window: BrowserWindow, userData: HubUserData) {
+    this.window = window;
+    this.userData = userData;
+
+    this.macOSAppMenu = [
+      {
+        label: "Business Modeler Hub Preview",
+        submenu: [
+          {
+            label: "About Business Modeler Hub Preview",
```
domhanak, line 55: suggestion: Let's use separate file for constants - `LabelConstants.ts` and import them here and other places. It will be useful in testing aswell.

packages/hub/src/common/CommandExecutionResult.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { OperatingSystem } from "@kogito-tooling/core-api";
+
+export interface CommandExecutionResult {
```
domhanak, line 19: interface w/o doc

packages/hub/src/electron/UserData.ts
```diff
@@ -0,0 +1,60 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as electron from "electron";
+import * as path from "path";
+import * as fs from "fs";
+
+export class UserData {
```
domhanak, line 21: unit tests

domhanak (changes requested):
Left few comments on code.

I am exploratory testing the app locally (Linux), couple of issues to get started:
- Took my quite a while to figure out how to run the HUB, there was no README file in the ZIP and the `Business Modeler Preview` file is not executable by default.
So in order to run you need to :
1. `chmod -x` the file to be executable
2. Run it from CLI.

Clicking it from file explorer, produces following error dialog - 
Screenshot from 2020-03-30 09-35-01

We need to add installation instructions to ZIP's. I expected to download the HUB and install it by clicking one button or by running some startup script -  `./bmhub`.
Summary: We should make the hub file executable by default and there should be entry in README or docs explaining how to start the hub.

domhanak:
@tiagobento Is there a way to import file for file system in the HUB? I wasn't able to find it.

packages/hub/src/common/CommandExecutionResult.ts
tiagobento: This interface is not part of any public API. It's used only to share messages across electron and web environments inside the Hub module.

packages/hub/src/electron/Menu.ts
tiagobento: I don't think adding this indirection helps us in any sort of way at the moment. If when writing unit tests we identify that extracting constants will help us, we will do that. Let's not get ourselves caught up in premature optimisations.

tiagobento: If we do i18n at some point, for example, we can also revisit that and think of a more definitive solution.

packages/hub/src/electron/UserData.ts
tiagobento: Unit tests will be deferred for this PR.

tiagobento:
Hi Dominik! Thanks for your quick review.

1. As requested, I'll add instructions on how to build and package the Desktop and Hub modules (they work the same).

2. Having a script inside the Zip will have the same problem as the Hub's executable. You have to give permission to run it before you do. Did you try opening it from the UI *after* changing the executable's permissions? I'll talk to others about the possibility of adding a README file to the distributions Zip.

3. Ths Business Modeler Hub Preview is not intended to work as a "catch all" application. Its purpose is to bundle and help users access our tools. If you want to open a file, you should launch VSCode or Business Modeler Preview Desktop.

packages/hub/src/electron/Menu.ts
domhanak: Sure just wanted to mention this.

packages/hub/src/common/CommandExecutionResult.ts
domhanak: Okay, fine by me, but the docs are also useful for others within the team. For example, I would know that the this is needed `to share messages across electron and web environments inside the Hub module.` without asking.

domhanak:
Hi @tiagobento 
1.  Good, that helps alot.

2. Sure, I am not saying we have to have script, it seems as an overkill. I tried to add exectuable rights but I still wasn't able to run it without command line. (clicking from UI shows error about missing application for `Shared library files` and I didn't try to look for it as I am trying basic fucntions of the hub.

3. Cool, I wasn't sure. On the other hand shouldn't I be able to run VSCode from the hub?

domhanak:
I found one  issue so far:
1. Create new process OR dmn file
2. Model something (star -> end)
3. Input name without `.bpmn` suffix or `.dmn` suffix  and Save (save passes)
4. Use `Recent` to open it
5. There will be error about unsupported file.
WORKAROUND:
When saving, add corresponding file suffix - .bpmn or .dmn into file name.

I expected the HUB will add the suffix for me, for the reason this is tooling for BPMN and DMN.

domhanak:
I am unable to test the `Import via URL` for BPMN properly because of - https://issues.redhat.com/browse/KOGITO-513 Every file I try to open shows the error.

tiagobento:
@paulovmr See Dominik's issues above, please.

@domhanak These issues are related to the Desktop App and should be addressed in a separate PR.

domhanak:
@tiagobento @paulovmr Okay, do you want me to report it or do we already have such jira?
I ll check with master too.

tiagobento:
@domhanak I think we don't have one yet.

packages/hub/src/common/CommandExecutionResult.ts
tiagobento: That's a given since it is on the "common" folder.

tiagobento:
@domhanak Changes are pushed, please re-review it 😝. 

Also, could you please create a JIRA with the error you're getting when clicking the executable on the Linux file explorer UI? We'll track that and see what may be causing it. Thanks!

domhanak:
Ok so an update: this blocks me - https://issues.redhat.com/browse/KOGITO-1615 from verification of some scenarios. (updated from previous one, stunner team decided it is of different nature.)

Yes, I will add it to the jira (the issues with UI openning of HUB)

domhanak:
The changes look good to me. Thanks @tiagobento that is exactly what I wanted. :) 
Not sure about the builds, they seem to fail.

tiagobento:
@domhanak Yeah, GitHub Actions is having a problem with downloading the npm dependencies on macOS. A timeout is occurring for some reason.

domhanak (approved):
I tested the general behavior in common scenarios.
Open new BPMN/DMN file & save it.
Open sample BPMN/DMN files, modify & save,
General check on strings and typos in UI. @sterobin could maybe look at the *Learn * part if he didn't already. :) 
Open from URL seems to work I am able to open files from url's however I wasn't able to see BPMN models because of the error mentioned in this JIRA - https://issues.redhat.com/browse/KOGITO-1615
Issue about saving without suffixes: https://issues.redhat.com/browse/KOGITO-1630

Approving.

tiagobento:
@domhanak Thank you! I've created this JIRA [1] to track the error when trying to open the application from Linux file explorer UI.

[1] https://issues.redhat.com/browse/KOGITO-1633

PR #97, opened 2020-03-25
KOGITO-1547: Change online modeler to have a link to download the Hub

ljmotta:
## Task
https://issues.redhat.com/browse/KOGITO-1547

## Description
- Removed the links to GitHub and VS Code extensions.
- Added the link to the Get Business Modeler Hub Preview modal.
- Modal with description of all Kogito tools and the three operating systems options for download: Linux, macOS and Windows.
- Modal with a successful message and option to download using a link.

## Demo
kogito-1547

## Considerations
Regardless the successful modal, it wasn't made following the prototype (https://marvelapp.com/223hdgb6/screen/67389180). The download starts instantly, that way the spinner even wasn't going to show up.

PR #73, opened 2020-03-06
KOGITO-777: Include SCESIM editor

yesamer:
@danielezonca @jomarko @ederign @tiagobento can you please review it?

This PR introduces SCESIM editor inside kogito-tooling, and in particular to enable it in VSCode.

Thank you @tiagobento and @ederign for your support!

https://issues.redhat.com/browse/KOGITO-777

PR #102, opened 2020-03-31
KOGITO-1505: Stunner - CDATA block missing

hasys:
PR fixes CDATA missing block for following elements:
* bpmn2:completionCondition
* bpmn2:condition
* bpmn2:conditionExpression
* bpmn2:from
* bpmn2:to
* bpmn2:documentation
* drools:metaValue
* drools:script

Unfortunately **XSLTProcessor** doesn't support option `is CDATA present, just keep it`. The only possible solution is `put CDATA always for elements from the list`.

I checked big process with all elements and properties from [KOGITO-395](https://issues.redhat.com/browse/KOGITO-395) and didn't find any further elements with CDATA.

Most of the values with CDATA are present in our custom `drools:metaValue`

The only concern for me were `bpmn2:from` and `bpmn2:to`. It is attributes of Data Assignments. The concern was that `bpmn2:from` and `bpmn2:to` can contain not only content, but also an ID of another element. See example of data Input Association:
```xml
      <bpmn2:dataInputAssociation>
        <bpmn2:targetRef>_0DBFABE8-92B0-46E6-B52E-A9593AFA4371_GroupIdInputX</bpmn2:targetRef>
        <bpmn2:assignment>
          <bpmn2:from xsi:type="bpmn2:tFormalExpression"><![CDATA[managers]]></bpmn2:from>
          <bpmn2:to xsi:type="bpmn2:tFormalExpression"><![CDATA[_0DBFABE8-92B0-46E6-B52E-A9593AFA4371_GroupIdInputX]]></bpmn2:to>
        </bpmn2:assignment>
      </bpmn2:dataInputAssociation>
```
In Input and Output this value with ID is reversed so it is not possible to make one of it (for example `bpmn2:from`) with CDATA and the second without.

But I did the same test as I did in in this comment https://github.com/kiegroup/kie-wb-common/pull/3170#issuecomment-600958885 (the only change is that repo moved from **kogito-quickstarts** to **kogito-examples**) and it was successful, so I assume that it is OK to store IDs in CDATA.

### DIFF
--- a/packages/kie-bc-editors/src/DefaultXmlFormatter.ts
+++ b/packages/kie-bc-editors/src/DefaultXmlFormatter.ts
@@ -21,7 +21,7 @@ let cachedXsltProcessor: XSLTProcessor;
 function newXsltProcessor() {
   const xsltDoc = new DOMParser().parseFromString(
     [
-      '<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="3.0">',
+      '<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:drools="http://www.jboss.org/drools" xmlns:bpmn2="http://www.omg.org/spec/BPMN/20100524/MODEL" version="3.0">',
       '  <xsl:strip-space elements="*"/>',
       '  <xsl:template match="para[content-style][not(text())]">',
       '    <xsl:value-of select="normalize-space(.)"/>',
@@ -29,7 +29,9 @@ function newXsltProcessor() {
       '  <xsl:template match="node()|@*">',
       '    <xsl:copy><xsl:apply-templates select="node()|@*"/></xsl:copy>',
       "  </xsl:template>",
-      '  <xsl:output indent="yes"/>',
+      // indent="yes" prettifies output
+      // cdata-section-elements="list of nodes with cdata separated by space"
+      '  <xsl:output indent="yes" encoding="UTF-8" cdata-section-elements="bpmn2:completionCondition bpmn2:condition bpmn2:conditionExpression bpmn2:from bpmn2:to bpmn2:documentation drools:metaValue drools:script"/>',
       "</xsl:stylesheet>"
     ].join("\n"),
     "application/xml"


packages/kie-bc-editors/src/DefaultXmlFormatter.ts
```diff
@@ -21,15 +21,17 @@ let cachedXsltProcessor: XSLTProcessor;
...
+      '<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:drools="http://www.jboss.org/drools" xmlns:bpmn2="http://www.omg.org/spec/BPMN/20100524/MODEL" version="3.0">',
       '  <xsl:strip-space elements="*"/>',
       '  <xsl:template match="para[content-style][not(text())]">',
       '    <xsl:value-of select="normalize-space(.)"/>',
       "  </xsl:template>",
       '  <xsl:template match="node()|@*">',
       '    <xsl:copy><xsl:apply-templates select="node()|@*"/></xsl:copy>',
       "  </xsl:template>",
-      '  <xsl:output indent="yes"/>',
+      // indent="yes" prettifies output
```
romartin, line 32: I would remove comments

tiagobento:
Hi @hasys! Thanks for the PR. The changes look fine to me, although I can't comment about whether they are correct or not, since I'm not familiar with the XML structure of BPMN files, so I trust your solution.

One thing that I thought about was whether it works for DMN files as well. I'm concerned that this bit breaks something on DMN:  `xmlns:drools="http://www.jboss.org/drools" xmlns:bpmn2="http://www.omg.org/spec/BPMN/20100524/MODEL"`

romartin (commented):
Ok thanks @hasys , good job!

It doesn't result much friendly to be forced to specify all elements which may contain CDATA blocks, also because we're hardcoding domain specific stuff in this generic module, but yeah given the case, you did well by figuring out some other way for doing it. On the other hand, the BPMN2 spec is not going to change so often, so that should not be a big issue.

Also, did you tested DMN is also working as before? Are on DMN any CDATA blocks being removed as well, which should be included here? CC @manstis @tiagobento 

Thanks!

packages/kie-bc-editors/src/DefaultXmlFormatter.ts
hasys: I prefer to have it since it explain a bit what is going on here. But I can delete it, no problem, let's wait for others opinions :P

manstis (commented):
I tried adding some XML type text to some DMN elements I thought may have been problematic (`dmn:description`, `dmn:allowedAnswers` and `dmn:text`). They _appear_ to be encoded during marshalling (e.g. giving `<dmn:description>This contains some &lt;XML&gt; &amp;</dmn:description>`, where the text was `This contains some <XML> &`. The file could be re-opened without problem.

Did I miss something? How did the issue manifest itself in BPMN?

packages/kie-bc-editors/src/DefaultXmlFormatter.ts
```diff
@@ -21,15 +21,17 @@ let cachedXsltProcessor: XSLTProcessor;
 function newXsltProcessor() {
   const xsltDoc = new DOMParser().parseFromString(
     [
-      '<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="3.0">',
+      '<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:drools="http://www.jboss.org/drools" xmlns:bpmn2="http://www.omg.org/spec/BPMN/20100524/MODEL" version="3.0">',
```
manstis, line 24: Are you 100% sure, cast iron guarantee, that the namespace prefix for these are ALWAYS as you show?

I.e. Does the `http://www.omg.org/spec/BPMN/20100524/MODEL` namespace ALWAYS have the prefix `bpmn2` in everything your editor can handle? For example, if a file was opened that mapped `http://www.omg.org/spec/BPMN/20100524/MODEL` to `cheese` would saving it in your editor preserve the `cheese` prefix, in which case this PR breaks down.

hasys:
Hi @manstis, yes, you are right. This **XSLTProcessor** automatically encode your data. But there is several issues why we shouldn't relay on it and all user modified data are encoded/CDATAed on marshallers any way:

* We need to align processes with Business Central, so we can't relay that **XSLTProcessor** will process everything for us since we doesn't have it in Business Central.
* Google Chrome plugin doesn't have **XSLTProcessor** as well, so when you will save your file in Google Chrome extension your data won't be encoded (if you didn't encode it in your marshallers).
* Sometimes BPMN engine don't decode values, so we need it **as is** and that where CDATA blocks come in. See [example](https://issues.redhat.com/browse/KOGITO-1209).
* You anyway need to decode it back when you reading the file on your own, because this library used only on save but not on load.
* We can give up **XSLTProcessor** anytime for example we implement prettifying on our side for Chrome.

So by this PR we are pretty much duplicating in **XSLTProcessor** the same behavior we already have in our marshallers. So good point, need to discuss it better how to fix this duplication in the future, because now we are doing new marshallers, so it can be especially actual. Thanks for pointing out!

packages/kie-bc-editors/src/DefaultXmlFormatter.ts
hasys: Not sure what do you mean, we are using this namespace in both designers legacy and Stunner. If it is an issue, we have a pretty big issue than. Or am I missing something?

manstis:
@hasys Hello, I also tested DMN _outside_ of VSCode too; and whatever marshaller we're using escaped the XML text i.e. even without passing explicitly through `kogito-tooling` use of `XSLTProcessor` DMN's XML was escaped. We can however provide a list of DMN nodes that should be persisted as CDATA entries. @jomarko @karreiro WDYT?

hasys:
Hi @manstis it's interesting, because in BPMN Designer outside of VS Code it is never encoded until we call `encode` function on a property explicitly. How are you encoding/decoding attributes and element's values in DMN? Probably we can reuse it for new marshallers.

packages/kie-bc-editors/src/DefaultXmlFormatter.ts
manstis: I'm mean, what happens if you open a BPMN file from another vendor and they chose to represent the BPMN namespace (`http://www.omg.org/spec/BPMN/20100524/MODEL`) differently? For example they could have chosen to use it as their XML documents default namespace in which case there is no prefix required; or they could have used a different value to `bpmn2` as the prefix. For example:
```
<?xml version="1.0" encoding="UTF-8"?> 
<definitions id="Definition"
             targetNamespace="http://www.example.org/MinimalExample"
             typeLanguage="http://www.java.com/javaTypes"
             expressionLanguage="http://www.mvel.org/2.0"
             xmlns="http://www.omg.org/spec/BPMN/20100524/MODEL" <-- default namespace
             xmlns:xs="http://www.w3.org/2001/XMLSchema-instance"
             xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI"
             xmlns:dc="http://www.omg.org/spec/DD/20100524/DC"
             xmlns:di="http://www.omg.org/spec/DD/20100524/DI"
             xmlns:tns="http://www.jboss.org/drools">

<process processType="Private" isExecutable="true".....>
```
This snippet of a BPMN2 file (from Red Hat's [user guide](https://access.redhat.com/documentation/en-us/jboss_enterprise_brms_platform/5/html/brms_business_process_management_guide/an_example_bpmn_2.0_process) interestingly) has the BPMN namespace as the _default_ so there are no `bpmn2` prefixes on elements.

Here's another example:
```
<?xml version="1.0" encoding="ISO-8859-1" standalone="yes"?>
<semantic:definitions id="_1373649849716" 
                      name="A.1.0" 
                      targetNamespace="http://www.trisotech.com/definitions/_1373649849716"
                      xmlns:dc="http://www.omg.org/spec/DD/20100524/DC"
                      xmlns:bpsim="http://www.bpsim.org/schemas/1.0"
                      xmlns:di="http://www.omg.org/spec/DD/20100524/DI"
                      xmlns:semantic="http://www.omg.org/spec/BPMN/20100524/MODEL" <-- "semantic" prefix
                      xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI"
                      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <semantic:process isExecutable="false".....>
```
This example shows use of the `semantic` prefix for the BPMN2 namespace. If this file is opened by our BPMN2 editor is the BPMN2 namespace prefix `semantic` preserved in our conversion from XML->UI model->XML and hence the change to `DefaultXmlFormatter.ts` would not work as it's expecting the `bpmn2` prefix. If our conversion from XML->UI model overrides whatever was in the source XML file and forces it to use `bpmn2` as the prefix the change in this PR is OK.

manstis:
@hasys DMN uses a completely different XML library to BPMN for client-side marshalling! IIUC you use some port of the Eclipse EMF library whereas DMN uses a library called `jsonix` that generates `JAXB`-type JavaScript classes from the XSD.... it's possible the library DMN uses handles encoding (it appears that way).

packages/kie-bc-editors/src/DefaultXmlFormatter.ts
hasys: XML is generated in our editor always. If you will load file from other vendors, it still will be processed by our marshallers before it will reach **XSLTProcessor** so it always will be as we have it in our editor and in our editor we have **bpmn2** namespace. But good point, thank you!

hasys:
> @hasys DMN uses a completely different XML library to BPMN for client-side marshalling! IIUC you use some port of the Eclipse EMF library whereas DMN uses a library called `jsonix` that generates `JAXB`-type JavaScript classes from the XSD.... it's possible the library DMN uses handles encoding (it appears that way).

good point, thank you @manstis.

CC @romartin, @treblereel we will use the same one for new marshallers, right? Or how will we handle encoding/decoding in new marshallers?

manstis (approved):
@hasys says the namespace prefix usage is not an issue.. so all good for me.

hasys:
Great, thank you @manstis,

@mastis, @jomarko, @karreiro, don't forget to provide a list of DMN nodes that should be persisted as CDATA entries. Or we can to move it to separate jira issue if you want. Thank you!

manstis:
@hasys I don't think we need to provide any DMN elements that we need handled as `CDATA` sections as I was not able to replicate any issue with the (un)marshalling we have in place already (both inside of and outside of VSCode). If however you want to add some DMN ones then those I originally listed are fine (to begin with.. until/unless somebody finds issue somewhere). If you do add them to this PR I would want to re-test DMN to ensure their addition does not cause problems.

tiagobento:
@hasys Actually, `DefaultXmlFormatter` is used in all our channels -- VSCode, Chrome, Online and Desktop. See:
- `packages/desktop/src/envelope/index.ts`
- `packages/chrome-extension-pack-kogito-kie-editors/src/envelope/index.ts`
- `packages/vscode-extension-pack-kogito-kie-editors/src/webview/index.ts`
- `packages/online-editor/src/envelope/index.ts`

I think that's why @manstis was able to see the same behaviour on DMN outside VSCode too. Please tell me if that changes anything and if we should do another round of testing.

All in all, thank you very much for the discussion. I agree with @romartin when he says that it's not ideal to have BPMN/DMN specifics in a "XML formatter" class, but I'm okay with that for now until we figure out where this logic should go. We will probably have to revisit that when #73 (SceSim) comes in, so it should be soon. (cc @yesamer)

@hasys We currently have an issue with our CI that `actions/checkout@v1` can't be re-ran. There's a fix for it coming on #97. In the mean time, if you want to re-trigger the CI, do a pair of commits changing and un-changing something. CI works fine for new commits.

hasys:
Hi @manstis, if there is no issues, I suggest not to fix it 😄 

@romartin, @tiagobento, @cristianonicolai, @domhanak, @jomarko, did you have a chance to take a look at it?

I know it is not the best solution, but looks like the only one we can afford before the dev preview.

tiagobento:
@hasys Please rebase you PR with master, #97 was just merged.

ederign (changes requested):
Hi @hasys thanks for the PR!

Can you please add a test for the DefaultXmlFormatter that gives an input of an XML with C-DATA and make sure that we get it back as result?

ederign (changes requested):
Hi @hasys thanks for the PR!

Can you please add a test for the DefaultXmlFormatter that gives an input of an XML with C-DATA and make sure that we get it back as result?

romartin (approved):

hasys:
Hi @ederign,

good point. I tried to write some basic test, but I don't know how Node.js infrastructure works. I wrote the test like

```typescript
import {DefaultXmlFormatter} from "../DefaultXmlFormatter";

describe("DefaultXmlFormatter", () => {
    test("CDATA preserved", () => {
        const formatter = new DefaultXmlFormatter();
        expect(formatter.format("<bpmn2:documentation><![CDATA[some stuff]]></bpmn2:documentation>"))
            .toEqual("<bpmn2:documentation><![CDATA[some stuff]]></bpmn2:documentation>");
    });
});
```

But it is failing because can't find some symbols inside of `DefaultXmlFormatter` (see below). Also there was a test for formatter before XSLTProcessor was introduced and removed after migration to this library, probably for the same reason, don't know :) Any way, can I ask you to help me if know how to deal with this kind of issues, I am not an expert in TS/Node JS :( 

```
yarn run v1.19.1
$ jest
 PASS  src/__tests__/GwtAppFormerApi.test.ts
 PASS  src/__tests__/GwtEditorWrapper.test.ts
 PASS  src/__tests__/GwtEditorWrapperFactory.test.ts
 FAIL  src/__tests__/DefaultXmlFormatter.test.ts
  ● DefaultXmlFormatter › CDATA preserved

    ReferenceError: XSLTProcessor is not defined

      38 |   );
      39 |
    > 40 |   const xsltProcessor = new XSLTProcessor();
         |                         ^
      41 |   xsltProcessor.importStylesheet(xsltDoc);
      42 |   return xsltProcessor;
      43 | }

      at newXsltProcessor (src/DefaultXmlFormatter.ts:40:25)
      at DefaultXmlFormatter.Object.<anonymous>.DefaultXmlFormatter.format (src/DefaultXmlFormatter.ts:47:50)
      at Object.<anonymous> (src/__tests__/DefaultXmlFormatter.test.ts:22:26)

Test Suites: 1 failed, 3 passed, 4 total
Tests:       1 failed, 7 passed, 8 total
Snapshots:   0 total
Time:        2.576s, estimated 3s
Ran all test suites.
error Command failed with exit code 1.
```

tiagobento:
HI @hasys! The code of your test looks perfect. However, the reason why your test is failing is because it's running on a Node.js environment, where the `XSLTProcessor` class is not globally available. On the browser, however, it is. That's why it works.

To fix that issue, you'll have to look for a library that implements XSLTProcessing in Node.js. Idk if that exists, but it's worth the try. See this SO answer for references. https://stackoverflow.com/questions/56889281/xsltprocessor-is-not-defined-in-node-js

jomarko:
@manstis @hasys no nodes comes to my mind. I tried a business rule task, invoking a DMN model while whitespaces and &lt; or &gt; were in use , but save and reopen worked fine.

jomarko (approved):

domhanak:
@hasys imho if we can't test it just report it into JIRA for reference and let's proceed with this.

hasys:
To check that test work, just remove **bpmn2:documentation** from `cdata-section-elements` argument in **xsl:stylesheet tag** of stylesheet. You will see this:

image

hasys:
Hi @domhanak yes I agree with you. But anyway I added a test (I am not sure how good it is from Node ecosystem point of view but any way).

@ederign, @tiagobento, I tried several libraries most of them didn't work for me. For example: 
* [xslt](https://www.npmjs.com/package/xslt) doesn't support Stylesheets so it is not possible to use it. 
* [xslt-ts](https://www.npmjs.com/package/xslt-ts) not compiling due to recommended dependencies not compatible. I tried several versions of dependencies and this version, but non of them worked for me
* [enketo-xslt](https://www.npmjs.com/package/enketo-xslt) not sure what is it for, also looks like we will have some license issues.
* and so on

But I guess I found one working: [node10-libxslt](https://www.npmjs.com/package/node10-libxslt). It looks very heavy weight to me, but on the other side it is devDependency only, so maybe it is OK.

Anyway I don't know about Node infrastructure so much so please, review my last commit please :)

UPD: sorry wrote comments in different window this comment https://github.com/kiegroup/kogito-tooling/pull/102#issuecomment-610303241 should be below. It is about how to test the test :)

tiagobento:
@hasys That's a nice finding! TBH I tried thinking of a way to improve this test, but since it *looks* like `node10-libxslt` does *not* implement the same API as XSLTProcessor in the browser, sharing the XSLT stylesheet between them seems reasonable enough.

This test asserts that our stylesheet is valid for our intents, so any XSLT implementation should behave the same, and we should be covered. (At least that's what we expect 😝)

tiagobento (changes requested):
@hasys Can you please remove `packages/kie-bc-editors/package-lock.json `. I think this was created when you added `libxslt` using npm and it's breaking the CI build. You should use yarn on our project.

Also our `devDependencies` are all together on the root package.json, so that all modules can share the same version. Please move `libxslt` to root package.json.

hasys:
Thank you @tiagobento, fixed

hasys:
Hi @tiagobento,

looks like library I added is incompatible with our projects. According to the [documentation](https://www.npmjs.com/package/node10-libxslt#environment-compatibility):

```
Environment compatibility

For now 64bits linux and 32bits windows are confirmed. Other environments are probably ok, but not checked. Please report an issue if you encounter some difficulties.

Node-libxslt depends on node-gyp, you will need to meet its requirements. This can be a bit painful mostly for windows users. The node-gyp version bundled in your npm will have to be greater than 0.13.0, so you might have to follow these instructions to upgrade. There is no system dependancy otherwise, libxslt is bundled in the project.
```

And looks like build failing on this node-gyp library.

```
node-pre-gyp ERR! command "C:\\hostedtoolcache\\windows\\node\\12.13.0\\x64\\node.exe" "D:\\a\\kogito-tooling\\kogito-tooling\\node_modules\\node-pre-gyp\\bin\\node-pre-gyp" "install" "--fallback-to-build" "--loglevel" "http"
node-pre-gyp ERR! cwd D:\a\kogito-tooling\kogito-tooling\node_modules\node1-libxmljsmt
```

 Do you know how to fix it? Or we should to use another library for tests?

tiagobento:
@hasys 😅

Do we have other options for a XSLT node library? If so, I think we should give it a shot. If not, we can further discuss how to proceed. CC @ederign

ederign:
@hasys @tiagobento due to the limitations and challenges, if QE agrees (/cc @jomarko) we can create a JIRA for future research of this issue and, as an exception, merge this without test coverage.

@jomarko wdyt?

jomarko:
@ederign I am fine if @hasys will investigate xls libraries/testing in future. However I checked mostly from DMN point of view. I think still some QE ack for BPMN part is needed. 


@domhanak what is your opinion? Maybe I am wrong in this PR s good to go already. Conversation  in comments was quite long so maybe I overlooked some ack.

Should I have a look also on BPMN part?

hasys:
Hi @jomarko, I think @domhanak already answered here: https://github.com/kiegroup/kogito-tooling/pull/102#issuecomment-610249361

I already reverted state of this PR to pretest time and rebased to the master. I hope now it will be green and ready for merge. Thank you!

domhanak:
@jomarko I take a look on BPMN part.

domhanak (approved):

cristianonicolai (approved):

hasys:
All is done, @ederign, @tiagobento, your turn :)

ederign (approved):

tiagobento (approved):

romartin:
please @tiagobento @ederign feel free to merge this, this fix is important for the release. Thx!

ederign:
@hasys if you want to include this on the next release, all kogito tooling PRs should be cherry-picked for 0.3.x branch (https://github.com/kiegroup/kogito-tooling/tree/0.3.x).

Remember that we need the approval of EM + Architect to merge this on 0.3.x. (/cc @lazarotti @porcelli )

PR #104, opened 2020-04-01
KOGITO-1567: Export source code as GitHub gist

caponetto:
**Task**: [KOGITO-1567](https://issues.redhat.com/browse/KOGITO-1567)

**Demo**
gist-it

### DIFF
--- a/packages/online-editor/src/common/GithubService.ts
+++ b/packages/online-editor/src/common/GithubService.ts
@@ -32,15 +32,28 @@ export interface FileInfo {
   path: string;
 }
 
+export interface CreateGistArgs {
+  filename: string;
+  content: string;
+  description: string;
+  isPublic: boolean;
+}
+
 export class GithubService {
   private octokit: Octokit;
   private authenticated: boolean;
 
-  private init(): void {
+  constructor() {
+    this.init(false);
+  }
+
+  private init(resetToken: boolean): void {
     this.octokit = new Octokit();
     this.authenticated = false;
 
-    setCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME, EMPTY_TOKEN);
+    if (resetToken) {
+      setCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME, EMPTY_TOKEN);
+    }
   }
 
   private initAuthenticated(token: string): void {
@@ -63,26 +76,24 @@ export class GithubService {
   }
 
   public reset(): void {
-    this.init();
+    this.init(true);
   }
 
-  public authenticate(token: string = EMPTY_TOKEN): Promise<boolean> {
+  public async authenticate(token: string = EMPTY_TOKEN) {
     token = this.resolveToken(token);
 
-    return this.validateToken(token).then(isValid => {
-      if (isValid) {
-        this.initAuthenticated(token);
-      } else {
-        this.init();
-      }
+    if (!(await this.validateToken(token))) {
+      this.init(true);
+      return false;
+    }
 
-      return Promise.resolve(isValid);
-    });
+    this.initAuthenticated(token);
+    return true;
   }
 
   public resolveToken(token: string = EMPTY_TOKEN): string {
     if (!token) {
-      token = getCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME) || EMPTY_TOKEN;
+      return getCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME) ?? EMPTY_TOKEN;
     }
 
     return token;
@@ -136,24 +147,24 @@ export class GithubService {
       });
   }
 
-  public createGist(filename: string, content: string, description: string, publicGist: boolean): Promise<string> {
+  public createGist(args: CreateGistArgs): Promise<string> {
     if (!this.isAuthenticated()) {
       return Promise.reject("User not authenticated.");
     }
 
     const gistContent: any = {
-      description: description,
-      public: publicGist,
+      description: args.description,
+      public: args.isPublic,
       files: {
-        [filename]: {
-          content: content
+        [args.filename]: {
+          content: args.content
         }
       }
     };
 
     return this.octokit.gists
       .create(gistContent)
-      .then(response => Promise.resolve(response.data.files[filename].raw_url))
+      .then(response => response.data.files[args.filename].raw_url)
       .catch(e => Promise.reject("Not able to create gist on Github."));
   }
 
@@ -162,7 +173,7 @@ export class GithubService {
       .get({ gist_id: gistId })
       .then(response => {
         const filename = Object.keys(response.data.files)[0];
-        return Promise.resolve(response.data.files[filename].raw_url);
+        return response.data.files[filename].raw_url;
       })
       .catch(e => Promise.reject("Not able to get gist from Github."));
   }
--- a/packages/online-editor/src/common/GithubService.ts
+++ b/packages/online-editor/src/common/GithubService.ts
@@ -15,6 +15,15 @@
  */
 
 import * as Octokit from "@octokit/rest";
+import { getCookie, setCookie } from "./utils";
+
+export const GITHUB_OAUTH_TOKEN_SIZE = 40;
+export const GITHUB_TOKENS_URL = "https://github.com/settings/tokens";
+export const GITHUB_TOKENS_HOW_TO_URL =
+  "https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line";
+
+const GITHUB_AUTH_TOKEN_COOKIE_NAME = "github-oauth-token-kie-editors";
+const EMPTY_TOKEN = "";
 
 export interface FileInfo {
   gitRef: string;
@@ -22,15 +31,77 @@ export interface FileInfo {
   org: string;
   path: string;
 }
-export class GithubService {
-  private readonly octokit: Octokit;
 
-  constructor() {
+export class GithubService {
+  private octokit: Octokit;
+  private authenticated: boolean;
+
+  private init(): void {
     this.octokit = new Octokit();
+    this.authenticated = false;
+
+    setCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME, EMPTY_TOKEN);
+  }
+
+  private initAuthenticated(token: string): void {
+    this.octokit = new Octokit({ auth: token });
+    this.authenticated = true;
+
+    setCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME, token);
+  }
+
+  private validateToken(token: string): Promise<boolean> {
+    if (!token) {
+      return Promise.resolve(false);
+    }
+
+    const testOctokit = new Octokit({ auth: token });
+    return testOctokit.emojis
+      .get({})
+      .then(() => Promise.resolve(true))
+      .catch(() => Promise.resolve(false));
+  }
+
+  public reset(): void {
+    this.init();
+  }
+
+  public authenticate(token: string = EMPTY_TOKEN): Promise<boolean> {
+    token = this.resolveToken(token);
+
+    return this.validateToken(token).then(isValid => {
+      if (isValid) {
+        this.initAuthenticated(token);
+      } else {
+        this.init();
+      }
+
+      return Promise.resolve(isValid);
+    });
+  }
+
+  public resolveToken(token: string = EMPTY_TOKEN): string {
+    if (!token) {
+      token = getCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME) || EMPTY_TOKEN;
+    }
+
+    return token;
+  }
+
+  public isAuthenticated(): boolean {
+    return this.authenticated;
   }
 
   public isGithub(url: string): boolean {
     return /^(http:\/\/|https:\/\/)?(www\.)?github.com.*$/.test(url);
+  }
+
+  public isGist(url: string): boolean {
+    return /^(http:\/\/|https:\/\/)?(www\.)?gist.github.com.*$/.test(url);
+  }
+
+  public extractGistId(url: string): string {
+    return url.substr(url.lastIndexOf("/") + 1);
   }
 
   public retrieveFileInfo(fileUrl: string): FileInfo {
@@ -63,5 +134,36 @@ export class GithubService {
           `https://raw.githubusercontent.com/${fileInfo.org}/${fileInfo.repo}/${fileInfo.gitRef}/${fileInfo.path}`
         ).then(res => (res.ok ? res.text() : Promise.reject("Not able to retrieve file content from Github.")));
       });
+  }
+
+  public createGist(filename: string, content: string, description: string, publicGist: boolean): Promise<string> {
+    if (!this.isAuthenticated()) {
+      return Promise.reject("User not authenticated.");
+    }
+
+    const gistContent: any = {
+      description: description,
+      public: publicGist,
+      files: {
+        [filename]: {
+          content: content
+        }
+      }
+    };
+
+    return this.octokit.gists
+      .create(gistContent)
+      .then(response => Promise.resolve(response.data.files[filename].raw_url))
+      .catch(e => Promise.reject("Not able to create gist on Github."));
+  }
+
+  public getGistRawUrlFromId(gistId: string): Promise<string> {
+    return this.octokit.gists
+      .get({ gist_id: gistId })
+      .then(response => {
+        const filename = Object.keys(response.data.files)[0];
+        return Promise.resolve(response.data.files[filename].raw_url);
+      })
+      .catch(e => Promise.reject("Not able to get gist from Github."));
   }
 }
--- a/packages/online-editor/src/common/GithubTokenModal.tsx
+++ b/packages/online-editor/src/common/GithubTokenModal.tsx
@@ -45,7 +45,7 @@ interface Props {
 export function GithubTokenModal(props: Props) {
   const context = useContext(GlobalContext);
 
-  const [potentialToken, setPotentialToken] = useState("");
+  const [potentialToken, setPotentialToken] = useState(context.githubService.resolveToken());
   const [authenticated, setAuthenticated] = useState(context.githubService.isAuthenticated());
 
   const tokenToDisplay = useMemo(() => {
--- a/packages/online-editor/src/common/GithubTokenModal.tsx
+++ b/packages/online-editor/src/common/GithubTokenModal.tsx
@@ -72,12 +72,12 @@ export function GithubTokenModal(props: Props) {
       onClose={props.onClose}
       title=""
       header={
-        <React.Fragment>
+        <>
           <Title headingLevel={TitleLevel.h1} size={BaseSizes['2xl']}>
             GitHub OAuth Token
           </Title>
           <p className="pf-u-pt-sm">Authentication required for exporting to GitHub gist.</p>
-        </React.Fragment>
+        </>
       }
       footer={
         <div className="pf-u-w-100">
@@ -97,7 +97,6 @@ export function GithubTokenModal(props: Props) {
               isValid={!!authenticated}
               value={tokenToDisplay}
               onPaste={onPasteHandler}
-              onChange={() => { /**/ }}
               autoFocus={true}
             />
             {authenticated && (
--- a/packages/online-editor/src/common/GithubTokenModal.tsx
+++ b/packages/online-editor/src/common/GithubTokenModal.tsx
@@ -15,7 +15,7 @@
  */
 
 import * as React from "react";
-import { useCallback, useContext, useState } from "react";
+import { useCallback, useContext, useState, useMemo } from "react";
 import { GlobalContext } from "./GlobalContext";
 
 import {
@@ -37,6 +37,7 @@ import {
 } from './GithubService';
 
 interface Props {
+  isOpen: boolean;
   onClose: () => void;
   onContinue: () => void;
 }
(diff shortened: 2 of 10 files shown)


packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -112,12 +113,15 @@ export function EditorToolbar(props: Props) {
       </DropdownItem>,
       <DropdownItem key="downloadSVG" component="button" onClick={props.onPreview}>
         Download SVG
+      </DropdownItem>,
+      <DropdownItem key="exportGist" component="button" onClick={props.onExportGist}>
+        Gist it
```
tiagobento: We should probably change that to "Create GitHub Gist". cc @ederign @porcelli

packages/online-editor/src/editor/EditorPage.tsx
```diff
@@ -125,6 +135,23 @@ export function EditorPage(props: Props) {
...
+              const fileExtension = extractFileExtension(fileUrl.pathname);
+              // FIXME: KOGITO-1202
+              window.location.href = `?file=${gistUrl}#/editor/${fileExtension}`;
+            })
+            .catch(e => {
+              setGithubTokenModalVisible(true);
+            });
+        } else {
+          setGithubTokenModalVisible(true);
+        }
```
tiagobento: I would go with a guard-clause here, inverting the if and adding a `return` after making the GitHubTokenModal visible.

packages/online-editor/src/editor/EditorPage.tsx
```diff
@@ -125,6 +135,23 @@ export function EditorPage(props: Props) {
...
           setCopySuccessAlertVisible(true);
         }
+      } else if (action === ActionType.EXPORT_GIST) {
+        if (context.githubService.isAuthenticated()) {
+          context.githubService
+            .createGist(fileNameWithExtension, content.content, content.path || fileNameWithExtension, true)
+            .then(gistUrl => {
+              setGithubTokenModalVisible(false);
+              const fileUrl = new URL(gistUrl);
+              const fileExtension = extractFileExtension(fileUrl.pathname);
```
tiagobento: I think you could inline the `fileUrl` variable, since it's only a `temp` object.

packages/online-editor/src/editor/EditorPage.tsx
```diff
@@ -125,6 +135,23 @@ export function EditorPage(props: Props) {
         if (document.execCommand("copy")) {
           setCopySuccessAlertVisible(true);
         }
+      } else if (action === ActionType.EXPORT_GIST) {
+        if (context.githubService.isAuthenticated()) {
+          context.githubService
+            .createGist(fileNameWithExtension, content.content, content.path || fileNameWithExtension, true)
```
tiagobento: I don't understand why this expression is necessary `content.path || fileNameWithExtension`. Can `content.path` be undefined? If it can, why would it be different from `fileNameWithExtension`? If they're never different, could we only use fileNameWithExtension? If none of that is true, just replace `||` with `??` since it is a better "orElse" mechanism introduced recently on TypeScript.

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+        isSmall={true}
+        isOpen={open}
+        onClose={onCloseHandler}
+        header={header}
+        title=""
+        footer={footer}
+      >
+        {content}
+      </Modal>
+    </React.Fragment>
```
tiagobento: React.Fragment is meant for places where you want to return a list of components without wrapping them in a actual component. It can also be shortened with `<></>`. In this case, since we're only returning one component (<Modal>), we don't need it.

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+function obfuscate(token: string) {
+  if (token.length <= 8) {
+    return token;
+  }
+
+  const stars = new Array(token.length - 8).join("*");
+  const pieceToObfuscate = token.substring(4, token.length - 4);
+  return token.replace(pieceToObfuscate, stars);
+}
```
tiagobento, line 154: We should really think about having a commons module for that type of stuff. You don't have to change anything on your PR, I'm just sharing my concerns so that if you see an opportunity in future PRs, let's do it! :)

packages/online-editor/src/common/GithubTokenModal.tsx
tiagobento: Also, can you inline `header`, `content`, and `footer`? The way it's written right now requires a `useMemo` for each of them and also creates an unnecessary (IMHO) indirection. I think this will be more readable if inlined.

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+export function GithubTokenModal(props: Props) {
+  const context = useContext(GlobalContext);
+
+  const [potentialToken, setPotentialToken] = useState("");
+  const [open, setOpen] = useState(true);
+  const [authenticated, setAuthenticated] = useState(context.githubService.isAuthenticated());
+
+  const tokenToDisplay = obfuscate(context.githubService.resolveToken() || potentialToken);
```
tiagobento: We should have a `useMemo` for this one with dependencies on `context.githubService` and `potentialToken`.

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const onPasteHandler = useCallback(e => {
+    const token = e.clipboardData.getData("text/plain").slice(0, GITHUB_OAUTH_TOKEN_SIZE);
+    setPotentialToken(token);
+    setTimeout(async () => {
+      context.githubService.authenticate(token)
+        .then(isAuthenticated => setAuthenticated(isAuthenticated))
+    }, 0);
```
tiagobento: Is there a special reason for this `setTimeout`?

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+  }, []);
+
+  const onContinueHandler = useCallback(() => {
+    close();
+    props.onContinue();
+  }, []);
+
+  const onCloseHandler = useCallback(() => {
+    close();
+  }, []);
```
tiagobento: You can use `close` directly.

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+interface Props {
+  onClose: () => void;
+  onContinue: () => void;
+}
+
+export function GithubTokenModal(props: Props) {
+  const context = useContext(GlobalContext);
+
+  const [potentialToken, setPotentialToken] = useState("");
+  const [open, setOpen] = useState(true);
```
tiagobento: I think it's a better practice to have the "open/close" control in the parent component. Expose a `prop` called `isOpen` and let the parent of `GithubTokenModal` decide whether it's open or not.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -78,6 +80,7 @@ export function HomePage(props: Props) {
   const uploadDndRef = useRef<HTMLDivElement>(null);
 
   const [inputFileUrl, setInputFileUrl] = useState("");
+  const [gistRawUrl, setGistRawUrl] = useState("");
```
tiagobento, line 84: `gistRawUrl` can be derived from `inputFileUrl`, which means it shouldn't be a state, but a memo. In this particular case, since you need to make a remote call using `getGistRawUrlFromId`, `useMemo` does not apply. For **this particular case, where an async computation is needed**, you can keep the state and perform a `useEffect` instead. See comment below (*).

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -246,18 +249,36 @@ export function HomePage(props: Props) {
...
+          const rawUrl = new URL(rawUrlStr);
+          const fileExtension = extractFileExtension(rawUrl.pathname);
+          if (fileExtension && context.router.getLanguageData(fileExtension)) {
+            setGistRawUrl(rawUrlStr);
+            setInputFileUrlState(InputFileUrlState.VALID);
+          } else {
+            setInputFileUrlState(InputFileUrlState.INVALID_GIST_EXTENSION);
+          }
+        })
+        .catch(() => setInputFileUrlState(InputFileUrlState.INVALID_GIST));
```
tiagobento: (*) This can be done inside a `useEffect`. The reasoning behind it is "every time the `inputFileUrl` changes, `gistRawUrl` changes too", that's the same as thinking that a change to `inputFileUrl` has a side effect that changes `gistRawUrl`, hence the name, `useEffect`.

Now that we're here.....

`inputFileUrlState` can be calculated inside a `useMemo`, since its dependencies are `inputFileUrl` and `gistRawUrl`. Please change that too :)

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -274,19 +295,24 @@ export function HomePage(props: Props) {
 
   const openFileFromUrl = useCallback(() => {
     if (validatedInputUrl && inputFileUrlState !== InputFileUrlState.INITIAL) {
-      const fileUrl = new URL(inputFileUrl);
+      const urlToLoad = context.githubService.isGist(inputFileUrl) ? gistRawUrl : inputFileUrl;
+      const fileUrl = new URL(urlToLoad);
       const fileExtension = extractFileExtension(fileUrl.pathname);
```
tiagobento: Can we inline `fileUrl`? Looks like it's only a temporary object.

packages/online-editor/src/home/HomePage.tsx
tiagobento: Note how this entire validateUrl method will be removed and `inputFileUrlState` state will become a memo.

tiagobento (changes requested):
Thanks for the PR @caponetto! Welcome :))

Left some improvement requests inline. Please tell me if you have questions!

packages/online-editor/src/editor/EditorPage.tsx
caponetto: Not sure the implementation usecase but `content.path` can be undefined whereas `fileNameWithExtension` cannot. So I'm gonna go with `??`.

packages/online-editor/src/common/GithubTokenModal.tsx
caponetto: There isn't, thanks!

packages/online-editor/src/editor/EditorToolbar.tsx
caponetto: I'm up for suggestions :)
Actually, it wouldn't be the 1st time this name changed.

porcelli: I prefer: "Gist it!" (with the exclamation point). it's kinda modern/cooler :)

packages/online-editor/src/common/GithubService.ts
```diff
@@ -64,4 +135,35 @@ export class GithubService {
         ).then(res => (res.ok ? res.text() : Promise.reject("Not able to retrieve file content from Github.")));
       });
   }
+
+  public createGist(filename: string, content: string, description: string, publicGist: boolean): Promise<string> {
```
tiagobento: This method has a lot of parameters, including one that is "boolean". For readers, having 3 strings and 1 boolean next to each other might not be that meaningful. To avoid that (and since TS doesn't have named parameters), we can create an object. This method will then accept one single parameter:  `args: { filename: string, content: string, description: string, publicGist: boolean }`. I think this will improve readability a lot. Also, please change "publicGist" to "isPublic".

packages/online-editor/src/common/GithubService.ts
```diff
@@ -64,4 +135,35 @@ export class GithubService {
...
+
+    const gistContent: any = {
+      description: description,
+      public: publicGist,
+      files: {
+        [filename]: {
+          content: content
+        }
+      }
+    };
```
tiagobento: Is `any` really the only option here? I think you can look at the significa of `octokit.gists.create` and check what's its type.

packages/online-editor/src/common/GithubService.ts
```diff
@@ -64,4 +135,35 @@ export class GithubService {
...
+
+    return this.octokit.gists
+      .create(gistContent)
+      .then(response => Promise.resolve(response.data.files[filename].raw_url))
+      .catch(e => Promise.reject("Not able to create gist on Github."));
+  }
```
tiagobento: You don't need to create a "resolved" promise inside a "then". You can simply return the value on line 156.

packages/online-editor/src/common/GithubService.ts
```diff
@@ -15,24 +15,95 @@
...
+    return this.validateToken(token).then(isValid => {
+      if (isValid) {
+        this.initAuthenticated(token);
+      } else {
+        this.init();
+      }
+
+      return Promise.resolve(isValid);
+    });
+  }
```
tiagobento: What I would do:

```
public async authenticate(token: string = EMPTY_TOKEN) {
    token = this.resolveToken(token);

    if (!await this.validateToken(token)) {
      this.init();
      return false;
    }

    this.initAuthenticated(token);
    return true;
}
```

Not a change request, just a suggestion! :)

packages/online-editor/src/common/GithubService.ts
```diff
@@ -15,24 +15,95 @@
...
+
+  public resolveToken(token: string = EMPTY_TOKEN): string {
+    if (!token) {
+      token = getCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME) || EMPTY_TOKEN;
+    }
```
tiagobento: I also think we could go with a guard-clause here.

packages/online-editor/src/common/GithubService.ts
```diff
@@ -64,4 +135,35 @@ export class GithubService {
...
+      .catch(e => Promise.reject("Not able to create gist on Github."));
+  }
+
+  public getGistRawUrlFromId(gistId: string): Promise<string> {
+    return this.octokit.gists
+      .get({ gist_id: gistId })
+      .then(response => {
+        const filename = Object.keys(response.data.files)[0];
+        return Promise.resolve(response.data.files[filename].raw_url);
+      })
```
tiagobento: You can simply return the value on line 165.

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+  }, []);
+
+  return (
+    <Modal
+      isSmall={true}
+      isOpen={props.isOpen}
+      onClose={props.onClose}
+      title=""
+      header={
+        <React.Fragment>
```
tiagobento: Can be `<>`

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+          <Title headingLevel={TitleLevel.h1} size={BaseSizes['2xl']}>
+            GitHub OAuth Token
+          </Title>
+          <p className="pf-u-pt-sm">Authentication required for exporting to GitHub gist.</p>
+        </React.Fragment>
```
tiagobento: Can be </>

packages/online-editor/src/common/GithubTokenModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+              id="token-input"
+              name="tokenInput"
+              aria-describedby="token-text-input-helper"
+              placeholder="Paste your token here"
+              maxLength={GITHUB_OAUTH_TOKEN_SIZE}
+              isDisabled={authenticated}
+              isValid={!!authenticated}
+              value={tokenToDisplay}
+              onPaste={onPasteHandler}
+              onChange={() => { /**/ }}
```
tiagobento: We should avoid creating lambdas here. Every time this component's function runs, a new lambda is created and React might think that the "TextInput" component has to be re-rendered. It's not only a cause of unnecessary computation, but it can also result in bugs!

packages/online-editor/src/editor/EditorPage.tsx
```diff
@@ -103,6 +111,13 @@ export function EditorPage(props: Props) {
 
   const closeCopySuccessAlert = useCallback(() => setCopySuccessAlertVisible(false), []);
 
+  const closeGithubTokenModal = useCallback(() => setGithubTokenModalVisible(false), []);
+
+  const continueExport = useCallback(() => {
+    closeGithubTokenModal();
+    requestExportGist();
+  }, []);
```
tiagobento: `closeGithubTokenModal` and `requestExportGist` are callbacks, so you should declare them as dependencies here.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -238,55 +240,80 @@ export function HomePage(props: Props) {
...
 
-  const validateUrl = useCallback((fileUrl: string) => {
+  useEffect(() => {
+    if (context.githubService.isGist(inputFileUrl)) {
+      const gistId = context.githubService.extractGistId(inputFileUrl);
+      context.githubService.getGistRawUrlFromId(gistId)
+        .then(rawUrlStr => setGistRawUrl(rawUrlStr));
+    }
+
+    return () => { /* Do nothing */ };
```
tiagobento: I think this line is not necessary? Can't you simply not have a "return" clause?

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -238,55 +240,80 @@ export function HomePage(props: Props) {
...
+      return InputFileUrlState.NO_FILE_URL;
     } else if (!context.router.getLanguageData(fileExtension)) {
-      setInputFileUrlState(InputFileUrlState.INVALID_EXTENSION);
+      return InputFileUrlState.INVALID_EXTENSION;
     } else {
-      setInputFileUrlState(InputFileUrlState.VALID);
+      return InputFileUrlState.VALID;
     }
-  }, []);
+  }, [inputFileUrl, gistRawUrl]);
```
tiagobento, line 285: 😍

packages/online-editor/src/index.tsx
```diff
@@ -25,17 +25,26 @@ import { Alert, AlertVariant, AlertActionLink } from "@patternfly/react-core";
...
+githubService.authenticate()
+  .then(() => {
+    if (urlParams.has("ext")) {
+      waitForEventWithFileData();
+    } else if (urlParams.has("file")) {
+      openFileByUrl();
+    } else {
+      openDefaultOnlineEditor();
+    }
+  });
```
tiagobento: I think a comment of mine got lost. Can we not have this githubService.authenticate() call here? It's making previously sync code be async now and I feel like there's no need to authenticate on GitHub at the very start of the app. We should put it in a `useEffect` or something like that. Is that possible?

tiagobento (changes requested):
Thx @caponetto, looking much better now! Just some other minor comments inline. We're almost there.

packages/online-editor/src/common/GithubService.ts
caponetto: Apparently, there is [an issue](https://github.com/octokit/rest.js/issues/1046).
Can you think a better way around this?

tiagobento: That's fine then :)

porcelli (approved):

tiagobento (approved):

ederign (approved):

ederign:
@jomarko when you have a chance could you please take a look at this?

paulovmr (approved):

packages/online-editor/src/common/GithubService.ts
```diff
@@ -15,24 +15,106 @@
...
+    return this.authenticated;
   }
 
   public isGithub(url: string): boolean {
     return /^(http:\/\/|https:\/\/)?(www\.)?github.com.*$/.test(url);
   }
 
+  public isGist(url: string): boolean {
+    return /^(http:\/\/|https:\/\/)?(www\.)?gist.github.com.*$/.test(url);
+  }
```
jomarko, line 112: do we have tests for this method?

jomarko (commented):
Questions
- If user click *Gist it!* two times in a row, two same files with exactly same name are created. Is that intentional/appropriate?
- If token on github is deleted, user is correctly prompted for new token, however the **Reset** button doesn't clear the old one, is that expected? 
Screenshot from 2020-04-07 11-51-59
- files gisted from firefox are single line source, form chrome it works fine

packages/online-editor/src/common/GithubService.ts
caponetto: Yes. They are on `GithubService.test.ts` file.

caponetto:
Thanks for the review, @jomarko!

Here are the answers for your questions from my point of view:
1) This operation is like an export, so a new gist will be generated every time you click on the button, even if you do not change anything. In other words, this feature is not meant to update an existing gist but always create a new one regardless of their content.
2) It seems like a bug. I'll take a look on that.
3) It seems like an issue with the infrastructure that retrieves the file content. The same happens if you download the file. I'll take a look.

tiagobento:
@caponetto @jomarko Thanks Jozef for the review! As a suggestion, we could simply disable the "Gist it!" button once it's clicked. WDYT?

CC @srambach is "disabling a button after clicking" a thing? Does it have UX/a11y implications we should be aware of?

caponetto:
@tiagobento IMO disabling the button would bring more confusion than benefits. 
The download operation is similar if you think about it.

caponetto:
@jomarko
About your 3rd question: Is your Firefox version < 74? Apparently, the formatter only works on Firefox version >= 74. If you think this is an issue, I suggest opening a new Jira because it affects the other operations as well, such as download and copy.

jomarko:
@caponetto my firefox is 70

@caponetto @tiagobento about the point 1 - name of gist file, I think disabling button is not proper option. For me is also fine we will not override existing gist but creating new.

However for me was issue to distinguish which gist is latest vs. oldest, maybe just adding timestamp would help? But not sure if that is ok from UX point of view @srambach ?

ederign:
My 2 cents on this.

@jomarko if the user created it twice, he is exporting twice, so we expect that this will generate two gist files (with same name but different ids). I would say that adding a timestamp is not ideal because it will change the expected name from gist file due to a corner case and it will make the file name (on the URL) weird. 

About the formater issue on firefox < 74, @caponetto could you please create a JIRA?

jomarko (approved):
The firefox issue reported as https://issues.redhat.com/browse/KOGITO-1760

Thank you @caponetto

PR #118, opened 2020-04-14
KOGITO-859: Open file URL using Chrome Extension when it has CORS disabled

ljmotta:
## Task
https://issues.redhat.com/browse/KOGITO-859

## Description
URL's that doesn't exists and/or without the CORS headers were making the editor throw message errors user.

This PR adds validations to both cases and enable the "Open from source" button only if a valid URL is placed on the input.

## Demo
kogito-859

Tested URLs:
 - https://github.com/ljmotta/KieDora/blob/master/test.bpmn
 - https://www.dropbox.com/s/8frxhgb8jgv3xex/sample.bpmn
 - https://gist.github.com/ljmotta/37ae6a15b8a5daf38ffa527edb81a2dd

Obs: This demo was simulating a slow internet connection to show all components.

## Considerations
Since opening any URL using the Chrome Extension header-editing capabilities would result in a security flaw, this task changed to a validation on the input URL


tiagobento (approved):

packages/online-editor/src/__tests__/common/AnimatedTripleDotLabel.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+describe("AnimatedTripleDotLabel", () => {
+  test("should be valid", async () => {
+    const label = "label";
+    const validLabels = [
+      ["label", ""],
+      ["label", "."],
+      ["label", ".."],
+      ["label", "..."],
+      ["label", ""],
```
jomarko, line 29: Is this last item needed? Ii is same as first one? Or are we checking something like after three dots again zero dots is shown?

jomarko (commented):
Just one question to code, manual test is pending

jomarko (approved):
Checked on Chrome and Firefox, works pretty well. Good job!

packages/online-editor/src/__tests__/common/AnimatedTripleDotLabel.test.tsx
ljmotta: Yes (for the last question). The idea was to ensure that has a cycle on the animation. After three dots, the next step is zero dots. Maybe it's necessary to add more tests after the zero dots?

PR #140, opened 2020-04-24
KOGITO-1788: Desktop sort recently opened files

paulovmr:


tiagobento (approved):

jomarko (commented):
Thank you @paulovmr . Works fine, just two point before approval:
- Could we please increase coverage? https://github.com/paulovmr/kogito-tooling/pull/5
- Imagine recently opened files are (a.dmn, b.dmn, b.bpmn, c.bpmn). They will be ordered as (1. b.bpmn, 2. c.bpmn, 3. a.dmn, 4. b.dmn). In other words bpmn files are always before dmn files, is that ok? Just asking, I see it as valid from some point of view, but without any additional UI elements is not so much intuitive - I mean some visual indicator of bpmn section vs dmn section.

paulovmr:
Hi @jomarko , thanks for the review. 

- Thanks for the PR. Only one comment, actually when you click twice on the button there won't be a "reverse order", the files will return to the original default order, which is by access time (latest opened first). Nevertheless, I think the PR is still valid, so I merged it. I only modifed the parts where you wrote "reverse order" to "access time order".

- I couldn't reproduce this order, for me it was: (1. a.dmn, 2. b.bpmn, 3. b.dmn, 4. c.bpmn), as I would expect. I wrote another test with this exact case and it resulted in the same order inside the snapshot.

jomarko (approved):

PR #143, opened 2020-04-29
KOGITO-1815: Chrome Extension Release

ljmotta:
## Task
https://issues.redhat.com/browse/KOGITO-1858
https://issues.redhat.com/browse/KOGITO-1859

## Description
 - Remove unnecessary permissions of the Chrome Extension over the browser
 - Add a GitHub Actions workflow to upload and publish the released Chrome Extension on Chrome Store.

## Before accept this PR
It's necessary to set up these secret variables on the repository:
   - GOOGLE_DEVELOPER_CONSOLE_CLIENT_ID
   - GOOGLE_DEVELOPER_CONSOLE_CLIENT_SECRET
   - GOOGLE_DEVELOPER_CONSOLE_REFRESH_TOKEN
   - CHROME_EXTENSION_ID

## Video Overview
https://www.youtube.com/watch?v=TXuWErf6-Bo

### DIFF
--- a/.github/workflows/publish_chrome_ext.yml
+++ b/.github/workflows/publish_chrome_ext.yml
@@ -32,8 +32,8 @@ jobs:
       - name: Upload Chrome Extension to the Chrome Store
         id: update_extension
         run: |
-          access_token=$(curl -X POST -fsS "https://oauth2.googleapis.com/token" -d "client_id=${{ secrets.CLIENT_ID }}&client_secret=${{ secrets.CLIENT_SECRET }}&refresh_token=${{ secrets.REFRESH_TOKEN }}&grant_type=refresh_token" | jq -r '.access_token')
-          uploadResponse=$(curl -X PUT -sS "https://www.googleapis.com/upload/chromewebstore/v1.1/items/${{ secrets.EXTENSION_ID }}" -H "Authorization:Bearer ${access_token}" -H "x-goog-api-version:2" -T chrome_extension_kogito_kie_editors_${{ steps.release-tag.outputs.tag }}.zip)
+          access_token=$(curl -X POST -fsS "https://oauth2.googleapis.com/token" -d "client_id=${{ secrets.GOOGLE_DEVELOPER_CONSOLE_CLIENT_ID }}&client_secret=${{ secrets.GOOGLE_DEVELOPER_CONSOLE_CLIENT_SECRET }}&refresh_token=${{ secrets.GOOGLE_DEVELOPER_CONSOLE_REFRESH_TOKEN }}&grant_type=refresh_token" | jq -r '.access_token')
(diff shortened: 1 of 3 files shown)


.github/workflows/publish_chrome_ext.yml
```diff
@@ -0,0 +1,59 @@
...
+        id: publish_extension
+        run: |
+          access_token=$(curl -X POST -fsS "https://oauth2.googleapis.com/token" -d "client_id=${{ secrets.CLIENT_ID }}&client_secret=${{ secrets.CLIENT_SECRET }}&refresh_token=${{ secrets.REFRESH_TOKEN }}&grant_type=refresh_token" | jq -r '.access_token')
+          publishResponse=$(curl -X POST -sS "https://www.googleapis.com/chromewebstore/v1.1/items/${{ secrets.EXTENSION_ID }}/publish" -H "Authorization:Bearer ${access_token}" -H "x-goog-api-version:2" -H "Content-Length:0")
+          echo "$publishResponse"
+          echo ::set-output name=publish_status::$(echo "$publishResponse" | jq -r '.status | .[0]')
+
+      - name: Check Publish
+        run: |
+          [[ "${{ steps.update_extension.outputs.publish_status }}" == 'OK' ] || [ "${{ steps.update_extension.outputs.publish_status }}" == 'PUBLISHED_WITH_FRICTION_WARNING' ]]
```
tiagobento: We only need a separate job because if we want to re-run the "publish" part separately we can, is that correct?

tiagobento: `PUBLISHED_WITH_FRICTION_WARNING` really is the expected response? Seems like an odd success code 😝

packages/chrome-extension-pack-kogito-kie-editors/static/manifest.json
```diff
@@ -1,13 +1,13 @@
 {
-  "name": "BPMN and DMN Editors :: GitHub Extension",
+  "name": "Kie Group GitHub Chrome Extension",
```
tiagobento: Maybe this should be "KIE Group GitHub Chrome Extension"? cc @ederign

tiagobento (changes requested):
@ljmotta Thanks for the PR! Nice job creating the release jobs :)

I left some minor comments inline and also I'd like to ask if it's possible to rename the secrets so that we know what they're needed for (e.g. `CLIENT_ID` is a pretty generic name). We should use `CHROME_STORE_CLIENT_ID` or something like that.

.github/workflows/publish_chrome_ext.yml
ljmotta: Yes, it's a odd success code. `PUBLISHED_WITH_FRICTION_WARNING` is the code when the Chrome Extension needs extra time to be reviewed by Google before being published due to it current permissions. So it'll be published after the review.

ljmotta: No, the re-run operation restart all the workflow, it's not possible to re-run just one job. It's separated because they are two different operations (uploading an extension doesn't mean it will be published). I've found this way easier to track a possible error on the workflow but it could be just one job.

tiagobento:
Secrets were already created. Approving.

tiagobento (approved):

ederign (approved):

PR #150, opened 2020-05-08
KOGITO-1392: CSS tweaks

ljmotta:
## Task
https://issues.redhat.com/browse/KOGITO-1392

## Description
Added the features 1, 2 and 5 from https://marvelapp.com/8a8h595/screen/67998289

### DIFF
--- a/packages/online-editor/src/editor/EditorToolbar.tsx
+++ b/packages/online-editor/src/editor/EditorToolbar.tsx
@@ -184,6 +184,7 @@ export function EditorToolbar(props: Props) {
             toggle={
               <DropdownToggle
                 id={"toggle-id-lg"}
+                className={"kogito--editor__toolbar-toggle-icon-button"}
                 onToggle={isOpen => setMenuOpen(isOpen)}
                 iconComponent={CaretDownIcon}
               >
@@ -203,6 +204,7 @@ export function EditorToolbar(props: Props) {
             onSelect={() => setKebabOpen(false)}
             toggle={
               <DropdownToggle
+                className={"kogito--editor__toolbar-toggle-icon-button"}
                 id={"toggle-id-sm"}
                 onToggle={isOpen => setKebabOpen(isOpen)}
                 iconComponent={EllipsisVIcon}
@@ -216,14 +218,24 @@ export function EditorToolbar(props: Props) {
         </ToolbarItem>
 
         <ToolbarItem className={"pf-u-display-none pf-u-display-flex-on-lg"}>
-          <Button variant={"plain"} onClick={props.onFullScreen} aria-label={"Full screen"}>
-            <ExpandIcon className={"kogito--editor__toolbar-icon-button"}/>
+          <Button
+            className={"kogito--editor__toolbar-icon-button"}
+            variant={"plain"}
+            onClick={props.onFullScreen}
+            aria-label={"Full screen"}
+          >
+            <ExpandIcon />
           </Button>
         </ToolbarItem>
         {!context.external && (
           <ToolbarItem>
-            <Button variant={"plain"} onClick={props.onClose} aria-label={"Close"}>
-              <CloseIcon className={"kogito--editor__toolbar-icon-button"}/>
+            <Button
+              className={"kogito--editor__toolbar-icon-button"}
+              variant={"plain"}
+              onClick={props.onClose}
+              aria-label={"Close"}
+            >
+              <CloseIcon />
             </Button>
           </ToolbarItem>
         )}
--- a/packages/online-editor/src/editor/EditorToolbar.tsx
+++ b/packages/online-editor/src/editor/EditorToolbar.tsx
@@ -19,7 +19,6 @@ import { useCallback, useContext, useMemo, useState } from "react";
 import { GlobalContext } from "../common/GlobalContext";
 import {
   Button,
-  PageSection,
   TextInput,
   Title,
   Toolbar,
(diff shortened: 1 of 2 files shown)


manstis:
I'd have liked the changes for this to be made after https://github.com/kiegroup/kogito-tooling/pull/142 was merged (rather selfishly really, as it'll make my life easier!) If that's not possible can you please make sure there is enough test coverage on this PR to be sure my changes do not break anything (my PR moves desktop, online and github channels all to use a single reusable "embedded" editor and it touches some of the classes you  change here).

ljmotta:
Hi @manstis ! I've already removed some features of this PR because was decided to break it on another task. I believe it's not going to have any conflict now.

manstis:
Thanks @ljmotta it looks plain sailing now.

tiagobento (approved):

packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -205,13 +217,13 @@ export function EditorToolbar(props: Props) {
...
           <Button variant={"plain"} onClick={props.onFullScreen} aria-label={"Full screen"}>
-            <ExpandIcon />
+            <ExpandIcon height={16} width={16} />
           </Button>
         </ToolbarItem>
         {!context.external && (
           <ToolbarItem>
             <Button variant={"plain"} onClick={props.onClose} aria-label={"Close"}>
-              <CloseIcon />
+              <CloseIcon height={16} width={16} />
```
srambach: Instead of specifying a height and width on the icon, let's add a class to the button to size them based on our predefined icon sizes. Something like: 
```
.kogito--editor__toolbar-icon-button {
  --pf-c-button--FontSize: var(--pf-global--icon--FontSize--md);
}
```

packages/online-editor/static/resources/style.css
```diff
@@ -196,3 +195,14 @@
...
 }
+
+.kogito--editor__toolbar-title {
+  display: inline;
+  border: var(--pf-global--BorderWidth--md) solid var(--pf-c-page__header-nav--BackgroundColor);
+}
+
+.kogito--editor__toolbar-title:hover {
+  border: var(--pf-global--BorderWidth--md) solid var(--pf-global--BorderColor--200);
+  padding: 0.17em 0.3em;
```
srambach: I'm always a little wary of magical numbers. Are these trial and error or should they correspond to our spacer system?
We already had a little jump if the title had wrapped down in the header, but this creates another one on hover as well. 
headertitleshift

ljmotta: Hi Sarah! I've copied the values from the title box to maintain consistent but looks like it was a mistake :disappointed_relieved:. Do you have a suggestion?

packages/online-editor/src/editor/EditorToolbar.tsx
ljmotta: Indeed, looks better this way. About the variable, should I use this one or change to one that fit the 16px mentioned on the marvel? I just found a font size not destined for icons (--pf-global--FontSize--md). I've updated the PR description with the marvel that I've used to do this task.

srambach: Sorry, I should have mentioned that I talked with @kywalker-rh and he okayed `--pf-global--icon--FontSize--md` That comes out at 18px I think.

packages/online-editor/static/resources/style.css
srambach: I was trying to figure out how you are making the box editable but not quite following :) It's adding an editable text box over? Or swapping it out? Anyway, if the padding is not on hover, but on it all the time then that will get rid of that shift. I need to understand it better to get rid of the shift when the box becomes editable.

ljmotta: AH! I've added the border without the hover but I forgot to put the same padding!
About the title box, wasn't made by me, but I saw that is Title and on click is changed to a TextInput component over a hidden Title. Probably the shift will disappear when I adjust the Title padding.

ljmotta:
@srambach I've removed the magic numbers and fixed the title shift! I hope it's OK the new value I've chosen for the padding.

kogito-1392

packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -205,13 +217,13 @@ export function EditorToolbar(props: Props) {
...
           <Button variant={"plain"} onClick={props.onFullScreen} aria-label={"Full screen"}>
-            <ExpandIcon />
+            <ExpandIcon className={"kogito--editor__toolbar-icon-button"}/>
           </Button>
         </ToolbarItem>
         {!context.external && (
           <ToolbarItem>
             <Button variant={"plain"} onClick={props.onClose} aria-label={"Close"}>
-              <CloseIcon />
+              <CloseIcon className={"kogito--editor__toolbar-icon-button"}/>
```
srambach: Let's move these classes up to the `Button`. The svg is making its own height and width 1em, so we need to set the font size of the container, which is the Button. Once you do that, the SVG should be showing up correctly at 18px

srambach:
The padding on the title element looks good now 👍 and spacer variables FTW 😆

packages/online-editor/src/editor/EditorToolbar.tsx
ljmotta: My bad! I think it's correct now. The toggle icons size (CaretDownIcon and EllipsisVIcon) were with 16px too so I've changed their sizes as well, or should I revert and mantain with 16px?

srambach: I think as long as it's a button with just an icon, let's add that class. I've brought it up with the PF team whether maybe an icon-only button should do that automatically but for now, let's do it!

srambach (approved):
LGTM 👍

jomarko:
Check on browser:
- [x] chrome
- [x] firefox

jomarko (approved):
Works fine. Spotted new issue https://issues.redhat.com/browse/KOGITO-2198

Thank you @ljmotta

ederign (approved):

PR #155, opened 2020-05-15
KOGITO-1918: File doesn't save properly after tried to open an invalid one

paulovmr:


ljmotta (approved):
LGTM, bug fixed :rocket:

tiagobento (approved):

packages/desktop/src/__tests__/webview/App.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    expect(component.asFragment()).toMatchSnapshot();
+
+    act(() => {
+      electron.ipcRenderer.send("openFile", {
+        file: { filePath: "/a/a.dmn", fileType: "dmn", fileContent: "" }
+      });
+    });
+
+    expect(component.asFragment()).toMatchSnapshot();
+  });
```
tiagobento, line 55: Actually, I don't think this is correct. AFAIK, you can't have two different snapshots for the same test? Can you double check that please?

tiagobento (changes requested):

packages/desktop/src/__tests__/webview/App.test.tsx
paulovmr: More than one snapshot works, first one will have the "invalid file type alert alert is closed immediately after leaving the home page 1" identifier and the second will have "invalid file type alert alert is closed immediately after leaving the home page 2".

tiagobento (approved):

domhanak:
Hi, I wasn't able to reproduce with these changes. The file is properly saved and title is updated after working with invalid file.

domhanak (approved):

ederign (approved):

PR #156, opened 2020-05-18
KOGITO-2204: Initial integration tests for desktop channel

domhanak:
@paulovmr Hi Paulo,
here is initial "sneak-peak" as it is missing typescript and general refactor of some code to helper test classes and dependencies clean-up.

In order to run, go to `packages/desktop` and run `yarn run integration:test` or with `:clean` if you want to rebuild whole app again. 
Please ignore `yarn.lock` I think I messed it up, will fix in next commit.

I added some `data-ouia-component-id` attributes to relevant elements to help with testing and make them stable.
Let me know, feedback is welcome, spectron is new area for me.


jomarko:
@domhanak could you add instruction of running integration test into readme files?

domhanak:
@jomarko Once it is finished yes, I would very much like it to be part of `yarn run build`
Something like:
    `"build": "yarn run lint && yarn test && yarn run build:fast && yarn run integration:test"`

Wdyt? @paulovmr

domhanak:
TODO: 

- Typescript - so far it was not feasible, I always run into issues - let me know if we are ok with JS here or not
- Moving dependencies to parent.
- Make it part of regular build:prod - this is open for discussion, I would like for it to be valuable and not hinder development.
- Add test for saving a files and opening file from source

packages/desktop/src/integration-tests/global-setup.js
```diff
@@ -0,0 +1,53 @@
...
+    options.startTimeout = 30000
+  }
+
+  var app = new Application(options)
+  return app.start().then(function () {
+    assert.strictEqual(app.isRunning(), true)
+    chaiAsPromised.transferPromiseness = app.transferPromiseness
+    return app
+  })
+}
```
jomarko: May I ask you for clarification of seeting timeout in case of `process.env.CI`? My point is that in real test we call both so just trying to double check if we are not doing the same twice.

packages/desktop/src/integration-tests/launch.test.js
```diff
@@ -0,0 +1,87 @@
...
+var describe = global.describe
+var it = global.it
+var beforeEach = global.beforeEach
+var afterEach = global.afterEach
+
+const KOGITO_EDITOR_LOCATOR = "//iframe[@class = \'kogito--editor\']";
+const UNSAVED_FILE_H3_LOCATOR = "//h3[text() = 'unsaved file']"
+const BMP_TITLE= "Business Modeler Preview"
+
+describe('Application Startup', function () {
```
jomarko: I briefly checked  `chai` API, Noticed there are basically two options how to invoke an assertion - `expect` and `should`. I see you used `should`. To me from the documentation was not clear if one of them is significantly better/recommended. Could you please share some thoughts if should is a better option? Asking because if one of them is a better choice, we should recommend/require it in some `How to Contribute`

packages/desktop/src/integration-tests/launch.test.js
```diff
@@ -0,0 +1,87 @@
...
+var path = require('path')
+var assert = require('assert')
+
+var describe = global.describe
+var it = global.it
+var beforeEach = global.beforeEach
+var afterEach = global.afterEach
+
+const KOGITO_EDITOR_LOCATOR = "//iframe[@class = \'kogito--editor\']";
+const UNSAVED_FILE_H3_LOCATOR = "//h3[text() = 'unsaved file']"
```
jomarko: Just double checking, this locator is the title editor in the top , over editor itself?

packages/desktop/src/webview/home/FilesPage.tsx
```diff
@@ -316,6 +316,7 @@ export function FilesPage(props: Props) {
             isHoverable={false}
             isCompact={true}
             onClick={() => electron.ipcRenderer.send("createNewFile", { type: "bpmn" })}
+            data-ouia-component-id= "create-new-bpmn"
```
jomarko: Shouldn't we add also `data-ouia-component-type` for `Card`s that open sample and for `Card`s that create new file?

I am new to ouia so trying to find answer for a question of attributes of ouia component are optional or mandatory.

jomarko (commented):
@domhanak this seems very useful contribution. Thank you

ederign:
@paulovmr can you please take a look at this?

packages/desktop/src/integration-tests/global-setup.js
domhanak: The idea behind this, is that if it runs on CI we give a bit more timeout for stability. But it could be I set it twice, gonna check and fix thanks!

packages/desktop/src/integration-tests/launch.test.js
domhanak: Agreed, imho we could go with expect as it is browser compatible, allows adding assertion messages and we might use chai in other packages so we have unified approach

packages/desktop/src/integration-tests/launch.test.js
domhanak: Yes, will add more test to check it changes after renaming etc.

domhanak:
@paulovmr  I ll ping you when ready and finished, but feel free to do initial feedback so I can incorporate.

packages/desktop/src/webview/home/FilesPage.tsx
domhanak: It is up to us, I ll add them when I actually add test for those parts. But yes, we should add it to every element we want to test.

jomarko:
@domhanak thank you for answers

domhanak:
@paulovmr @jomarko Ready for review, tests are run using `yarn run ui-test` there is a `yarn run ui-test:clean` if you want to rebuild the application with changes and run them.

I made them part of prod build, let's see if actions are ok with that.

Typescript is in progress I found out there are indeed types for spectron and webdriverio so I ll add it when possible.

Test verifies desktop app has proper title and that it loads the editor when creating new dmn/bpmn or using samples. When editor loads we use close button and check Homepage is shown.

I wasn't able to test save yet as it requires manipulation with native dialogs and I didn't yet find a way or framework for that.

ederign:
@domhanak sorry for the delay on this review, we are focused on the release and some people are on PTO. We will get back on this as soon as possible.

packages/desktop/src/ui-test/global-setup.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+  return electronPath
+}
+
+exports.setupTimeout = function (test) {
+  if (process.env.CI) {
+    test.timeout(30000)
+  } else {
+    test.timeout(30000)
+  }
```
jomarko: can't we simplify this to the below?
```js
test.timenout(30000)
```

packages/desktop/src/ui-test/launch.test.js
```diff
@@ -0,0 +1,144 @@
+var helpers = require('./global-setup')
+var path = require('path')
+var assert = require('assert')
```
jomarko: we imported this already in 'global-setup' and do not use it in this file

packages/desktop/src/ui-test/launch.test.js
```diff
@@ -0,0 +1,144 @@
...
+var it = global.it
+var beforeEach = global.beforeEach
+var afterEach = global.afterEach
+
+const UNSAVED_FILE = 'unsaved file';
+const KOGITO_EDITOR_LOCATOR = '//iframe[@id = \'kogito-iframe\']';
+const UNSAVED_FILE_H3_LOCATOR = '//h3[text() = \'unsaved file\']';
+const GO_TO_HOMEPAGE_BUTTON_LOCATOR = '//button[@aria-label=\'Go to homepage\']';
+const LOADING_SCREEN_LOCATOR = '#loading-screen';
+const BMP_TITLE = "Business Modeler Preview";
```
jomarko: maybe we could organize somehow the constants, locators vs. non-locators or just sort them alphabetically? what is your opinion?

packages/desktop/src/ui-test/launch.test.js
```diff
@@ -0,0 +1,144 @@
...
+      .browserWindow.getBounds().should.eventually.have.property('width').and.be.above(0)
+      .browserWindow.getBounds().should.eventually.have.property('height').and.be.above(0)
+      .getTitle().should.eventually.equal(BMP_TITLE)
+  })
+
+  describe('Editors can be opened', function() {
+
+    it('opens BPMN editor - new file', async () => {
+      await app.client.waitUntilWindowLoaded();
+      await app.client.click('//article[@data-ouia-component-id = \'create-new-bpmn\']')
```
jomarko: As we construct similar locator multiple times, maybe we could ahve a utility function
```js
var articleComponent = function(componentId) {
  return `//article[@data-ouia-component-id = \'${componentId}\']`
}
```
to avoid repeating the locator prefix
```js
await app.client.click(articleComponent('create-new-bpmn'))
```

wdyt? if you decide to applythis feel free to choose any name for the utility function

packages/desktop/src/ui-test/launch.test.js
```diff
@@ -0,0 +1,144 @@
...
+    it('opens DMN editor - new file', async () => {
+      await app.client.waitUntilWindowLoaded();
+      await app.client.click('//article[@data-ouia-component-id = \'create-new-dmn\']')
+      await app.client.waitUntilWindowLoaded().getWindowCount().should.eventually.equal(1)
+      await app.client.isVisible(LOADING_SCREEN_LOCATOR).should.eventually.be.false
+      
+      await app.client.getText(UNSAVED_FILE_H3_LOCATOR).should.eventually.be.equal(UNSAVED_FILE)
+      await app.client.isVisible(KOGITO_EDITOR_LOCATOR).should.eventually.be.true
+      await app.client.browserWindow.focus().waitUntilWindowLoaded();
+      await app.client.browserWindow.focus().click(GO_TO_HOMEPAGE_BUTTON_LOCATOR).waitUntilWindowLoaded()
```
jomarko: in the same test for BPMN, we do not have `waitUntilWindowLoaded()`, is that intentional?

I see same difference when I compare opening sample BPMN vs sample DMN

jomarko (commented):
@domhanak thank you for updates, looks good to me and works on my machine, few comments inline

packages/desktop/src/ui-test/launch.test.js
domhanak: Or we can create a file that exports them? Wdyt? but yes, we can do alphabetical

packages/desktop/src/ui-test/launch.test.js
domhanak: I think it is a mistake I ll fix

packages/desktop/src/ui-test/launch.test.js
jomarko: hmm I like idea of exporting them, and keep single file just for constants, if in future someone will need a constant/locator.... will easily double check if it doesn't exist already

packages/desktop/src/ui-test/global-setup.js
```diff
@@ -0,0 +1,54 @@
+var Application = require('spectron').Application
```
ljmotta: Has any motive to use `var`? It would be better to use `let` if it's possible, `var` can lead to some unexpected behavior.

packages/desktop/src/ui-test/global-setup.js
```diff
@@ -0,0 +1,54 @@
...
+var assert = require('assert')
+var chai = require('chai')
+var chaiAsPromised = require('chai-as-promised')
+var chaiRoughly = require('chai-roughly')
+
+var path = require('path')
+
+global.before(function () {
```
ljmotta: All global functions can be used without accessing the global object, just `before(function () {})` works!

packages/desktop/package.json
```diff
@@ -45,5 +47,13 @@
...
-  ]
+  ],
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
+    "chai-roughly": "^1.0.0",
+    "electron-chromedriver": "^8.0.0",
+    "mocha": "^7.1.2",
+    "spectron": "^9.0.0"
+  }
```
tiagobento: Please declare devDependencies on root `package.json`.

packages/desktop/src/ui-test/launch.test.js
tiagobento: A good idea is to create 'Page Objects'. The idea is to create a class that represents a page or a section of a page. This way, we can avoid having an explosion of constants. Here's an example:

```
export class HomePage {
  public openBpmnDiagramButtonSelector() {
    return '#my-button-id';
  }
}
```

You can then create an instance of HomePage on your tests and call `homePage.openBpmnDiagramButtonSelector()`.

I think it makes code more reusable and you can even embed behaviour in this class later if you identify that many tests are doing similar operations.

tiagobento (changes requested):
Maybe I'm missing something, but that's the story behind `ouia`? 😝 

Left some comments inline. Also seen that you already commented about adding types to the tests. I think this will be a great step forward.

packages/desktop/src/ui-test/global-setup.js
domhanak: Probably a left-over from example of Spectron, Thanks I ll fix it

domhanak:
@tiagobento the story behind OUIA is that it brings in a `standard` in naming components in UI so that we can have test that are stable when locating elements and require less maintenance in future changes to the UI componetns.

Current status - figuring out how to make it work without `require` statements and use `imports`

packages/desktop/package.json
```diff
@@ -22,8 +22,10 @@
     "rebuild": "electron-rebuild -f -w sqlite3",
     "lint": "tslint -c ../../tslint.json 'src/**/*.{ts,tsx,js,jsx}'",
     "test": "jest",
+    "ui-test": "tsc && mocha ./out/src/ui-test",
+    "ui-test:clean": " yarn run build:fast && tsc && mocha ./src/ui-test",
```
tiagobento: I'd rather keep `jest` as our test execution framework if possible. Also, you're triggering `tsc` manually, which is compiling the whole project again (after webpack). Can we remove that and have it work integrated with webpack?

packages/desktop/src/webview/home/FilesPage.tsx
```diff
@@ -291,7 +291,8 @@ export function FilesPage(props: Props) {
             component={"article"}
             isHoverable={false}
             isCompact={true}
-            onClick={() => electron.ipcRenderer.send("createNewFile", { type: EditorType.BPMN })}
+            onClick={() => electron.ipcRenderer.send("createNewFile", { type: "bpmn" })}
+            data-ouia-component-id= "create-new-bpmn"
```
tiagobento: Please rename `data-ouia-component-id` to `data-testid` to comply with RTL standards while not giving up the OUIA spec guidelines.

tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "strictNullChecks": true,
     "experimentalDecorators": true,
     "downlevelIteration": true,
-    "jsx": "react"
+    "jsx": "react",
+    "skipLibCheck": true
```
tiagobento: I'm guessing these changes will not be necessary once the tests run integrated with webpack. Can you please check that?

tiagobento (changes requested):
Few comments inline, mostly about things we already talked about. Please reach out if you need something! Nice work @domhanak. Looking forward to have those running on our CI 😍

domhanak:
@tiagobento so it works on CI, but I am still not able to fix the windows. Looking into that. About the usage of `tsc`, we should schedule a small call and discuss a bit.

domhanak:
@tiagobento @ljmotta @jomarko So, current status of this PR:

1. Jest refactor does not seem to work, no matter what I do I can't get it to start the application. So I would like to ask your opinion on having `mocha` & `chai` as a devDeps for some time until I figure it out. Benefit is running UI tests for the desktop app in the meantime on CI.

2. Windows is having issues on CI, on my VM it seems to work I am fiddling with timeouts and trying to fix that.

packages/desktop/src/ui-test/utils/DesktopTestHelper.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/**
+ * Sets timout for test suite. 
+ * 
+ * @param test Suite of which to set timout
+ */
+export const setupTimeout = (test: Suite): void => {
+  test.timeout(30000);
+}
```
jomarko, line 38: It is 30 seconds, minutes? could we document it?

packages/desktop/src/ui-test/utils/EditorPage.ts
```diff
@@ -0,0 +1,13 @@
+export class EditorPage {
```
jomarko, line 1: I think we should distinguish this util/helper from the actual source `EditorPage.tsx`.

Currently the difference can be spotted just from package and file extension. I think I we rename this could facilitate browsing code.
 My proposals:
- EditorPageTestHelper
- EditorPageLocatorHelper
- EditorPageUtils
- EditorPageTestUtils

What do you think?

packages/desktop/src/ui-test/utils/HomePage.ts
```diff
@@ -0,0 +1,21 @@
+export class HomePage {
```
jomarko, line 1: similar question as above

packages/desktop/src/ui-test/smoke.test.ts
```diff
@@ -0,0 +1,122 @@
...
+const waitForLoadingSpinner = async (appClient: SpectronClient) => {
+  await appClient.waitUntil(
+    async () => await appClient.isVisible(new EditorPage().diagramLoadingScreenLocator()) === false)
+}
+
+const checkBasicDimensions = async (app: Application): Promise<void> => {
+    await app.client.waitUntilWindowLoaded()
+    app.browserWindow.focus();
+    await app.client.getWindowCount().should.eventually.equal(1);
+    await app.browserWindow.isMinimized().should.eventually.be.false;
```
jomarko, line 25: does `isMinimized()` return a promise? can't we use just `await app.browserWindow.isMinimized().should.be.false;` ?

packages/desktop/src/ui-test/smoke.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    await app.browserWindow.isVisible().should.eventually.be.true;
```
jomarko, line 26: does `isVisible()` return a promise? can't we use just `await app.browserWindow.isVisible().should.be.false;` ?

packages/desktop/src/ui-test/smoke.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    await app.browserWindow.isFocused().should.eventually.be.true;
+    await app.browserWindow.getBounds().should.eventually.have.property('width').and.be.above(0)
+    await app.browserWindow.getBounds().should.eventually.have.property('height').and.be.above(0)
+    await app.client.getTitle().should.eventually.equal(BMP_TITLE)
```
jomarko, line 30: Similar question also here, tsdoc doesn't document them as returning `Promise`

jomarko: Probably all test methods need to be revisited if we will drop `eventually`

jomarko (commented):
Few questions. However in overall looks good. 

Regarding question about using `chai` and `mocha`. I do not mind using them as seems be documented quite well.

I agree if they would be replaceable by `jest`, we would eliminate number of technologies we use, however I am not an expert in them, so can not do such decision.

packages/desktop/src/ui-test/utils/EditorPage.ts
domhanak: Sure sounds good!

packages/desktop/src/ui-test/smoke.test.ts
domhanak: I managed to make the Jest work, so I will update this to be better readable and we will use `expect`

domhanak:
Closing this in favor of new PR that will contain these changes and will use jest.

PR #165, opened 2020-06-02
FAI-173: Embed kogito-tooling editors in trusty-ai. Part 2 (trusty-ai changes)

manstis:
See https://issues.redhat.com/browse/FAI-173

@ederign This adds a new channel type for `EMBEDDED`.

It seams reasonable to distinguish from `ONLINE` WDYT? 

I'm happy to make it more explicit if you prefer; i.e. add `TRUSTY-AI` instead. AFAIK the flag is only used (by DMN for example) to determine some setup, like whether to add a dock to the `EAST` or `WEST` or to enable a feature etc. I checked DMN and it only [checks](changes)) for `VSCODE` and `GITHUB` otherwise everything is the same. 

Please let me know your thoughts.

### DIFF
--- a/packages/core-api/src/core/ChannelType.ts
+++ b/packages/core-api/src/core/ChannelType.ts
@@ -15,8 +15,9 @@
  */
 
 export enum ChannelType {
-    VSCODE = "VSCODE",
-    ONLINE = "ONLINE",
-    GITHUB = "GITHUB",
-    DESKTOP = "DESKTOP"
+  VSCODE = "VSCODE",
+  ONLINE = "ONLINE",
+  GITHUB = "GITHUB",
+  DESKTOP = "DESKTOP",
+  EMBEDDED = "EMBEDDED"
 }


jomarko:
@manstis hi, just question to understand stuff.

My understanding (when I was looking into original  #142 ) was embedded editor is not a separate channel, rather a way to unify injecting editors into channels. So it is not as that at all?

manstis:
@jomarko Correct. 

#142 was originally made to support embedding `kogito-tooling` editors in trusty-ai. However during its review @tiagobento also took the opportunity to have `kogito-tooling` refactored to use the new _embedded_ approach.

This PR adds another channel for general _embedded_ use (to distinguish from its current use for VSCode, Github and the _online_ editor etc). I'm leaving the decision as to whether it's needed - or whether we should use a more explicit channel for trusty-ai with @ederign.

Channel is used, for example by DMN, to determine what environment its running in and therefore adjust where docks are shown and what features are enabled. It's plausible that the editor, when running in trusty-ai may need to configure itself differently.

This [PR](https://github.com/kiegroup/trusty-ai-sandbox/pull/75) in `trusty-ai-sandbox` is the _real_ change that will make trusty-ai use the embedded editors; however I cannot complete it until there has been a release of `kogito-tooling` to `npm` (JavaScript/TypeScript's equivalent of Maven Central).

manstis:
@jomarko Oh, to answer your explicit question:

> ..embedded editor is not a separate channel..

Embedded is not a channel, but _trusty-ai_ is.... LOL, perhaps the channel name should therefore be `trusty-ai`. I'll let @ederign comment (as IDK if we want to add a different channel for every future use.. especially if third parties start using embedded - there has been some enquiry already).

ederign:
@manstis sorry for the delay on this review, I'll take a look on this tomorrow morning.

ederign:
@manstis for me makes total sense at least for now. The embedded channel makes it clear that it runs on someone else web site as (guess what) embedded! ;)

ederign (approved):

PR #166, opened 2020-06-02
Fetch git tags automatically

jomarko:
Fetch git tags automatically to simplify 'yarn run build:prod' build


ederign:
Let's re run the ci.

jomarko:
@ederign my motivation for this was I spend more than day with trying 'yarn run build:prod' overlooking a note about tag fetching in README.

However I wouldn't like to slowdown the build rapidly because just due to my inattention. If the count of tags is too big and takes too much time to download them, then I understand if we are not going to merge this.

ederign:
@jomarko this makes sense. let's merge this.
My comment was to let you know that I retrigger the ci build :)

ederign (approved):

PR #168, opened 2020-06-04
KOGITO-204: Initial cypress test of online kogito channel

jomarko:
This is first version of the cypress test for kogito online channel

For more details see: https://issues.redhat.com/browse/KOGITO-204


ederign:
@jomarko  sorry for the delay in this review, we are focused on the release and some people are on PTO. We will get back on this as soon as possible.

packages/online-editor/cypress/integration/kogito_comunity_release_checklist/create_sample_model.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    cy.getDiagramEditorBody().find('[data-title="Explore Diagram"]').click();
+    cy.getDiagramEditorBody()
+      .find('[data-field="explorerPanelBody"]')
+      .contains("Process travelers")
+  })
+
+  it('Try DMN sample', () => {
+    cy.trySampleModel('dmn')
+
+    cy.getDiagramEditorBody().find('.fa-chevron-right').click();
```
jstastny-cz: you can use some context helpers like:
```
cy.getDiagramEditorBody().within(()=>{
   cy.find('').click()
   cy.find('').should('', '' ,'')
})
```

packages/online-editor/cypress/support/commands.js
```diff
@@ -0,0 +1,52 @@
...
+// -- This will overwrite an existing command --
+// Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+/**
+ * Clicks the given 'Try Sample' according to the passed 'assetType'
+ * Doesn't wait for the result of the click
+ * @param assetType possible values: 'bpmn', 'dmn'
+ */
+Cypress.Commands.add("trySampleModel", (assetType) => {
+    cy.get(`[data-ouia-component-id='try-sample-model-${assetType}']`)
```
jstastny-cz, line 28: One concern here - does the click result in a page redirect with URL change? If so, then it's discouraged and should be divided in two tests:
a) check the root `/` context for the links and their href attribute
b) for each link write a separate test spec that uses cy.visit('/custom-context-the-link-leads-to')

This will separate the concerns and mainly - link navigation using correct href is a well-tested basic functionality of the browsers.

Plus it would remove the need for this command.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -547,7 +547,8 @@ export function HomePage(props: Props) {
       </PageSection>
       <PageSection className="pf-u-px-2xl-on-lg">
         <Gallery gutter="lg" className="kogito--editor-landing__gallery">
-          <Card>
+          <Card data-ouia-component-type="try-sample-model"
```
jstastny-cz, line 550: this is not ideal, definitely this doesn't lead to component-level compliance - reason is that the page is actually not divided into components, thus there's no distinction of what components are there. Might be a good chance to start the discussion about refactoring - possibly based on the ouia attributes you're setting here.

jstastny-cz (changes requested):

packages/online-editor/cypress/integration/kogito_comunity_release_checklist/create_sample_model.spec.js
jomarko: this is great tip, now I understand better usage of `within`, thank you

packages/online-editor/cypress/support/commands.js
jomarko: I think agree just partially. As this was we are sure link is clickable and cause a redirect action. However your proposal is better balance of what we can take as precondition and what focus to test. I will try to incorporate it. 

One more question, was this encouraged by cypress (good vs. bad practices) or this is general rule for UI testing?

jomarko: Ok, found this obstacle, in the online editor for each asset type, you can open editor by clicking:
- Try sample
- Create new

Clicking on both links  will result to same url. However in first case non-empty diagram will be shown.

If the url is typed manually to the browser, always an empty diagram is shown.

Due to the above, I think makes sense to keeps the tests written as it is, WDYT? 

Just guessing here. Probably I could capture somehow data (diagram) passed when clicking 'Try Sample' and manually pass it via cypress test, however that will not reflect changes in real sample used by online editor (if theoretically sample will be updated in future).

packages/online-editor/src/home/HomePage.tsx
jomarko: Hmm, not sure I understand. 

Is your point that the parts I marked with same data-ouia-component-type should have single source like this?

https://github.com/kiegroup/kogito-tooling/pull/168/commits/a1f498247a2f94253d7ad874cdfdfef1bf6ee82d

packages/online-editor/cypress/support/commands.js
jstastny-cz: It has been encouraged by the Cypress author.

We could separate the navigation testing from the rest:
* using visit('/context') to check the expected page contents
* using link clicking and checking:
  * correct redirect
  * safe check the rendering (for example the presence of a top-level wrapper or so)

And we can differentiate the type of test by a folder or suffix or sth like that.

Does this make sense? The navigation tests can be less stable though - just guessing.

packages/online-editor/src/home/HomePage.tsx
jstastny-cz: Yes something of that sort - currently the Page contains everything, IMO the best practice is for the Page to use components. And the components themselves should IMO be more fine-grained. I don't think this is something for this set of PRs though, let's add the attributes we need for testing now and think about some sort of refactoring in future.

jomarko: @jstastny-cz ok, I reverted it from this PR and keeping it separately for later discussion https://github.com/jomarko/kogito-tooling/tree/HomePageRefactoring

packages/online-editor/cypress/support/commands.js
jomarko: As agreed on private channel, this will be not addressed as part of this PR.

packages/online-editor/cypress/integration/kogito_comunity_release_checklist/create_sample_model.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// <reference path="../../support/index.d.ts" />
+
+context('Check sample models successfully created', () => {
```
caponetto, line 19: Please use prettier to format all the files

packages/online-editor/package.json
```diff
@@ -22,15 +22,22 @@
...
+    "build:prod": "yarn run build --mode production --devtool none && yarn run cypress-ui-test",
     "start": "webpack-dev-server -d --host 0.0.0.0"
   },
   "babel": {
     "presets": [
       "@babel/env",
       "@babel/react"
     ]
+  },
+  "devDependencies": {
```
caponetto, line 39: Can we move these dependencies to the [root](https://github.com/kiegroup/kogito-tooling/blob/master/package.json) `package.json`?

packages/online-editor/cypress/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
```
caponetto, line 1: Can we extend the [root](https://github.com/kiegroup/kogito-tooling/blob/master/tsconfig.json) `tsconfig.json`? [example](https://github.com/kiegroup/kogito-tooling/blob/master/packages/workspace/tsconfig.json)

packages/online-editor/cypress/support/index.d.ts
```diff
@@ -0,0 +1,21 @@
+// in cypress/support/index.d.ts
+// load type definitions that come with Cypress module
+/// <reference types="cypress" />
+
+declare namespace Cypress {
```
caponetto, line 5: So far, we haven't been using namespaces on the code. Personally, I don't see a problem using them.
Is it something required by Cypress?
IMO it'd be nice to follow a certain pattern, i.e., starting using namespaces or don't use them at all.
^ @tiagobento

packages/online-editor/cypress/support/index.d.ts
```diff
@@ -0,0 +1,21 @@
...
+         * Clicks the given 'Try Sample' according to the passed 'assetType'
+         * Doesn't wait for the result of the click
+         * @param assetType possible values: 'bpmn', 'dmn'
+         * @example cy.trySampleModel('dmn')
+         */
+       trySampleModel(assetType: string): Chainable<Element>
+
+       /**
+         * Returns body html element of the given (bpmn, dmn) diagram editor
+         * Acessig Diagram Editor via this method is preferred way as the Diagram Editor is inside an iFrame
```
caponetto, line 17:
```suggestion
         * Accessing Diagram Editor via this method is preferred way as the Diagram Editor is inside an iFrame
```

packages/online-editor/cypress/plugins/index.js
```diff
@@ -0,0 +1,21 @@
...
+//
+// You can change the location of this file or turn off loading
+// the plugins file with the 'pluginsFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/plugins-guide
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
```
caponetto, line 13: Can we replace this comment by our copyright header? Applicable to the other files as well.

packages/online-editor/cypress/integration/kogito_comunity_release_checklist/create_sample_model.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    cy.visit('/')
+  })
+
+  it('Try BPMN sample', () => {
+    cy.trySampleModel('bpmn')
+
+    cy.getDiagramEditorBody().within(($diagramEditor) => {
+      cy.get('[data-title="Explore Diagram"]').click();
+      cy.get('[data-field="explorerPanelBody"]')
+        .contains("Process travelers")
```
caponetto, line 32: Is it enough to be sure that the example was correctly loaded?

packages/online-editor/cypress/integration/kogito_comunity_release_checklist/create_sample_model.spec.ts
```diff
@@ -0,0 +1,45 @@
+/*
```
caponetto, line 1: Is `*.spec.ts` the recommended way to name the test files?
Notice that, for all the tests we have so far, we are using `*.test.ts`.

caponetto (commented):
Hi @jomarko, thanks for the PR!
I've left some comments, please check if they make sense.
Also, there are some conflicts that need to be resolved before merging this PR.

tiagobento:
Closing in favor of #327. This PR can be reopen if necessary.

PR #174, opened 2020-06-10
FAI-178: Score Cards: Move PMML marshaller to new module

manstis:
See https://issues.redhat.com/browse/FAI-178

This PR adds the basics for PMML marshalling to the PMML package.

Further work is covered by https://issues.redhat.com/browse/FAI-179.

### DIFF
--- a/packages/pmml-editor/pom.xml
+++ b/packages/pmml-editor/pom.xml
@@ -40,7 +40,7 @@
         <version.com.fasterxml.jackson.jaxrs.jackson-jaxrs-json-provider>2.11.0</version.com.fasterxml.jackson.jaxrs.jackson-jaxrs-json-provider>
         <version.org.eclipse.persistence.org.eclipse.persistence.core>2.7.7</version.org.eclipse.persistence.org.eclipse.persistence.core>
         <version.org.glassfish.jaxb.jaxb-runtime>2.3.3</version.org.glassfish.jaxb.jaxb-runtime>
-        <version.cz.habarta.typescript-generator.typescript-generator-maven-plugin>2.23-SNAPSHOT</version.cz.habarta.typescript-generator.typescript-generator-maven-plugin>
+        <version.cz.habarta.typescript-generator.typescript-generator-maven-plugin>2.24.612</version.cz.habarta.typescript-generator.typescript-generator-maven-plugin>
         <version.org.apache.maven.plugins.maven-resources-plugin>3.1.0</version.org.apache.maven.plugins.maven-resources-plugin>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.target>1.8</maven.compiler.target>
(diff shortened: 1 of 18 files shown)


manstis:
The PMML marshalling approach is this:

**_Java/static code generation_**
- Use `jpmml` Java PMML model (that is also used by @gitgabrio in the PMML runtime)
- Generate Typescript classes from the Java model using `typescript-generator-maven-plugin`

**_Typescript/browser/runtime_**
- Convert the XML to/from _full JSON_ using `xml-js`
- Convert the _full JSON_ to/from the Typescript model generated above using `jsonata`

I have covered PMML's `Header`, `DataDictionary`, `DataField`, `Interval` and `Value`.

There are many more but they will be added "on demand" so the immediate successors are:
- `ScoreCard` (https://issues.redhat.com/browse/FAI-179)
- `LinearRegressionModel` (https://issues.redhat.com/browse/FAI-181)

packages/pmml-editor/pom.xml
```diff
@@ -0,0 +1,282 @@
...
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>pmml-generation</artifactId>
+    <groupId>org.kie.kogito</groupId>
+    <packaging>jar</packaging>
+    <version>0.0.0</version>
```
manstis, line 24: This Maven module does not get packaged or deployed anywhere.

It's sole purpose for existing is to run the plugin to generate Typescript classes from Java source files (in `jpmml`).

manstis:
If you want to test `yarn run generate` you'll need to build the plugin from source as [this](https://github.com/vojtechhabarta/typescript-generator/pull/490) is needed.

packages/pmml-editor/tsconfig.json
```diff
@@ -3,5 +3,6 @@
   "compilerOptions": {
     "declaration": false
   },
-  "include": ["src"]
+  "include": ["src"],
+  "exclude": ["src/main"]
```
tiagobento, line 7: Do you have to copy `jpmml-model` resources to this exact path? I think it might cause confusion since it's inside the `src` directory, where the TypeScript sources are.

tiagobento (commented):
Is it necessary to commit the generated files or could we include the `generate` step during the build? I'm okay with either approach, just asking so I understand it better :)

Also, I'd like to know more about what it is your building here, I've been away on PTO for a while and looks like I missed the train a bit, do you mind scheduling a call so you can explain me things? I think it will help me provide better reviews.

tiagobento:
@manstis Did you create a new module? I see that's the commit message for 
f81f461 but I don't see any other modules than `packages/pmml-editor`?

manstis:
Replaced by #189

PR #179, opened 2020-06-11
Finally removing enable proposed api parameters

ederign:


ederign:
@domhanak @tiagobento one thing that I wondered and maybe worth to investigate next week. In development, what happens when we have the BPMN extension installed and run our extension on vscode? Don't we need to launch vscode with --disableExtensions to avoid overlap?

(/cc @caponetto @jomarko @tomasdavidorg )

tiagobento (approved):

PR #180, opened 2020-06-15
KOGITO-2447: Open the right online editor for new files

caponetto:
**Task**: [KOGITO-2447](https://issues.redhat.com/browse/KOGITO-2447)

### DIFF
--- a/packages/online-editor/src/__tests__/common/utils.test.ts
+++ b/packages/online-editor/src/__tests__/common/utils.test.ts
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


manstis (commented):
Fixes the issue, thank-you @caponetto.

However should we need some form of test before merging?

ederign:
+1 @caponetto could you please add some tests for it ?

packages/online-editor/src/__tests__/common/utils.test.ts
```diff
@@ -0,0 +1,64 @@
...
+import { EditorType } from "@kogito-tooling/embedded-editor";
+import { extractEditorTypeFromUrl } from "../../common/utils";
+import { join } from "path";
+
+describe("utils::extractEditorTypeFromUrl", () => {
+  const baseUrl = "https://kiegroup.github.io/kogito-online";
+
+  function setWindowLocationHref(url: string) {
+    window = Object.create(window);
+    delete window.location;
```
manstis, line 25: Is that _dangerous_ and possibly affect other tests? 

Will it still persist _after_ these tests have ran; or since it's in a closure is it temporary?

caponetto: Not sure, we already have some tests that do this though.
But anyway I can restore the window object on `afterAll` or `afterEach` methods.

manstis (approved):

ederign (approved):

Issue #173, opened 2020-06-10
React should be a peer dependencies


riccardo-forina:
https://github.com/kiegroup/kogito-tooling/blob/master/packages/core-api/package.json#L15-L18

One think I noticed working on [the AtlasMap POC](https://github.com/riccardo-forina/atlasmap-kogito-tooling) is that I was getting the [Invalid Hook Call Warning
](https://reactjs.org/warnings/invalid-hook-call-warning.html) error in the console. Ultimately, it was due to `@kogito/core-api` declaring React as direct dependency, which causes for multiple versions of React being embedded in the bundle. Since AtlasMap is written solely using hooks, this issue come up immediately.

A [quick workaround](https://github.com/riccardo-forina/atlasmap-kogito-tooling/blob/master/packages/simple-react-editors/webpack.config.js#L32-L45) was to instruct WebPack to consider React an external dependency, but it would be better to make React and React DOM peer dependencies in the `core-api` package.

tiagobento:
Hi @riccardo-forina!

Thanks for reporting that. I'm glad to see you're using our packages :)

Initially, we thought that we should be the ones in control of React/ReactDOM versions, since some things that we offer as a library/framework is dependent on the React API. You can simply remove your declaration of React and ReactDOM on your modules so that no conflicting versions exist.

However, I also think that we should give users the flexibility of choosing the exact React version to use. Since peerDependencies are not downloaded by default, we would be making it mandatory to declare your version of React/ReactDOM. More than that, it would be possible to choose an incompatible version of React/ReactDOM, resulting in build failures.

Maybe I'm missing something here, but I'm really not sure yet what's I think is the best approach, so if you could provide more information about why having your own declaring of React/ReactDOM is important, I guess it would help us make a decision :)

riccardo-forina:
React requires a single instance of React DOM for it to properly work. React and React DOM should be a dependency of the final consumer. Taking the template as an example, React and React DOM should be declared as a dependency of the `chrome-extension-pack-simple-react` and `vscode-extension-pack-simple-react` packages that are the "final" outputs. `simple-react-editor` should also declare React as a peer dependency, since that package acts as a library and not a final output.

[...]

PR #182, opened 2020-06-17
KOGITO-764: Create a Guided Tour for first time users on DMN editor

karreiro:
See https://issues.redhat.com/browse/KOGITO-764

Screen Shot 2020-06-17 at 10 09 17

---

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/988
- https://github.com/kiegroup/kie-wb-common/pull/3336
- https://github.com/kiegroup/kogito-tooling/pull/182

---

Assumptions:
- This PR does not include i18n for any tour. It will be separately implemented by https://issues.redhat.com/browse/KOGITO-2471;
- This PR does not include the final version of the tour, and the steps are a mix of the preliminary input provided by Liz and the Learn DMN in 15 minutes (as accordingly with @ederign).


yesamer:
@karreiro Impressive! Congratulations for your excellent job!

jomarko:
### 01
When full screen is active, the guided tour is not so close tho given element
Screenshot from 2020-06-18 19-19-31

jomarko:
### 02
Once I opened and exited full screen mode form comment 01 I checked also console and got error about memory leak. Could you please double check it is not caused by this PR?
Screenshot from 2020-06-18 19-21-09

jomarko:
### 03
When I was asked to name node 'Can drive?' was redirected to screen where help panel was shown unexpectedly. I guess because shortcut for its displaying contains '?'

I am not sure we have issue when that shortcut is active or we could just guide user to name node without '?' to be sure help panel will not unexpectedly appear. WDYT?

Screenshot from 2020-06-18 19-28-53

jomarko:
### 04
We should update the text of this step to be 'safe' if eager user as me already expanded properties panel somehow.
Screenshot from 2020-06-18 19-32-19

jomarko:
### 05
Could we please clarify which pencil icon, I was clicking the one representing properties panel as that I interacted with in previous step
Screenshot from 2020-06-18 19-36-58

jomarko:
### 06
We guide user to click on 'Select expression'. Should we mention also pressing enter as an alternative for clicking? WDYT?

jomarko:
### 07
I have selected an expression type but the guide seems not detected it.
Screenshot from 2020-06-18 19-39-06

jomarko:
### 08
We should update the text of 'Back to' step, as it is dependent on model name, not file name

Screenshot from 2020-06-18 19-41-56

jomarko:
### 09
Links on congratulations page for learn and doc websites should open new tabs/windows.

jomarko:
### 10
We have [different behavior](https://issues.redhat.com/secure/attachment/12475628/move-active.webm) chrome (first in video) vs. firefox (second in video)

jomarko:
### 11
When user doesn't satisfy next step, in this case I renamed to `Ag` instead of `Age`. He is asked if want to dismiss or continue. I decided to continue however the tour content is not centered and close buttons is missing.

Screenshot from 2020-06-18 20-02-37

jomarko:
### 12
When user is asked to open expression editor of 'Can drive?' node, it is assumed as done if user open an expression editor of any decision. Is it expected?

packages/guided-tour/.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "printWidth": 120
+}
```
tiagobento: This config is already present on `prettier.config.js`. Is this file necessary?

packages/guided-tour/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "@types/react": "16.8.8",
+    "@types/react-dom": "16.8.2",
+    "enzyme": "3.10.0",
+    "enzyme-adapter-react-16": "1.15.1",
+    "enzyme-to-json": "3.4.3",
+    "identity-obj-proxy": "^3.0.0",
+    "prettier": "1.19.1",
+    "react": "16.12.0",
+    "react-dom": "16.12.0"
+  },
```
tiagobento: Can you move these into the root package.json? We're trying to keep devDependencies there so that we don't have many versions of the same package.

Also, react and react-dom are currently declared on core-api. We'll probably make them `peerDependencies` as you did here, though. See #173.

packages/guided-tour/src/components/HighlightLayer.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+    return "";
+  }
+
+  return (
+    <svg style={{ opacity: highlightEnabled ? 1 : 0 }} className="kgt-svg-layer">
+      <path d={internalRectPath()} style={{ fill: "rgba(0, 0, 0, .5)" }} />
+    </svg>
+  );
+};
```
tiagobento: You should be able to memoize values using `useMemo`. This prevents all the calculations to be made every time this components function runs. I know there's a lot here and many inter-dependent values, but it looks to me that the only value you use on your component is the return of `internalRectPath`, so you should be able to wrap everything inside a `useMemo`.

packages/guided-tour/src/components/NavigationControls.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const numberOfSteps = getSteps(currentTutorial).length;
+  const currentStepNumber = (currentStep || 0) + 1;
+
+  function prev() {
+    setCurrentStep(currentStep - 1);
+  }
+
+  function next() {
+    setCurrentStep(currentStep + 1);
+  }
```
tiagobento: These should  be `useCallbacks`.

packages/guided-tour/src/components/NavigationControls.tsx
```diff
@@ -0,0 +1,81 @@
...
+        data-kgt-next="true"
+        onClick={next}
+        isDisabled={currentStep === numberOfSteps - 1}
+        variant="link"
+      >
+        <AngleRightIcon />
+      </Button>
+    </div>
+  );
+};
```
tiagobento: Basically, as a general rule, you want every statement on a function component to be wrapped with some hook. The simple fact that `next` and `prev` are functions declared here will already cause this component to re-render unnecessarily. Every time the `NavigationControls` function executes, a new reference is attributed to the `prev` and `next` names. These are passed as props to the `Button` components, making React think that one of their props has changes, therefore causing the Buttons to try and re-render themselves too.

packages/guided-tour/src/components/utils/GuidedTourBusListeners.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const usePositionListener = (onPositionReceived: (rect: Rect) => void) => {
+  const type = "GuidedTour.newPosition";
+  addListener(type, event => onPositionReceived(event.detail));
+};
+
+function addListener(eventLabel: EventLabel, consumer: (customEvent: CustomEvent) => void) {
+  const effect = createEffect(eventLabel, consumer);
+  useLayoutEffect(effect, []);
+}
```
tiagobento: It's a convention that custom hooks should *always* be named with the `use` prefix. This is how you know you're calling a React Hook. Also, in this case, I think inlining the `createEffect` function would make the code easier to read.

packages/guided-tour/src/components/utils/GuidedTourBusListeners.ts
```diff
@@ -0,0 +1,52 @@
...
+
+function createEffect(eventLabel: EventLabel, consumer: (customEvent: CustomEvent) => void) {
+  function listener(e: any) {
+    consumer(e as CustomEvent);
+  }
```
tiagobento: Notice how you're creating this function outside the hook. This is effectively the same thing as placing this function directly into the component's function body. This should be avoided.

packages/guided-tour/src/components/utils/GuidedTourBusListeners.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export const usePositionListener = (onPositionReceived: (rect: Rect) => void) => {
+  const type = "GuidedTour.newPosition";
+  addListener(type, event => onPositionReceived(event.detail));
+};
+
+function addListener(eventLabel: EventLabel, consumer: (customEvent: CustomEvent) => void) {
+  const effect = createEffect(eventLabel, consumer);
+  useLayoutEffect(effect, []);
```
tiagobento: Did using `useEffect` cause any issues? useLayoutEffect is generally slower than `useEffect`. It might be necessary sometimes to avoid flickers on the screen though. I'm just curious 😛

packages/guided-tour/src/components/utils/GuidedTourContent.tsx
```diff
@@ -0,0 +1,145 @@
...
+              Click on <b>Dismiss</b> to stop it or <b>Continue</b> to resume your tour :-)
+            </Text>
+          </EmptyStateBody>
+          <Button data-kgt-continue="true" onClick={showSuggestion}>
+            Continue
+          </Button>
+        </>
+      );
+    }
+  }
```
tiagobento: Change that to `useMemo` and add `message` as a dependency.

packages/guided-tour/src/components/utils/GuidedTourContent.tsx
```diff
@@ -0,0 +1,145 @@
...
+  return () => (
+    <>
+      <ModalBoxCloseButton data-kgt-close="true" onClose={onCloseAction} />
+      {renderContent(content, onCloseAction, nextStep, prevStep)}
+      <NavigationControls />
+    </>
+  );
+};
+
+export const renderNegativeReinforcementDialog = (suggestion: string, onCloseAction: () => void) => {
```
tiagobento: This is a component, therefore should have a Capitalized function name. `NegativeReinforcementDialog` probably? :)

packages/guided-tour/src/components/utils/GuidedTourContent.tsx
```diff
@@ -0,0 +1,145 @@
...
+  return content;
+}
+
+export const renderStepDialog = (
+  content: React.ReactNode | ((props: object) => React.ReactNode) | string,
+  onCloseAction: () => void
+) => {
+  const { currentStep, setCurrentStep } = useContext(CurrentTutorialContext);
+  const nextStep = () => setCurrentStep(currentStep + 1);
+  const prevStep = () => setCurrentStep(currentStep - 1);
```
tiagobento: Use `useCallback`.

packages/guided-tour/src/components/utils/GuidedTourSelectorHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const element = querySelector(selector);
+    const isValidQuerySelector = !!element;
+    if (isValidQuerySelector) {
+      setCurrentRefElementPosition(element!.getBoundingClientRect());
+      return;
+    }
+
+    setCurrentRefElementPosition(DEFAULT_RECT);
+  }
```
tiagobento: Do not declare functions inside a component's function body.

packages/guided-tour/src/components/utils/GuidedTourSelectorHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+import { KogitoGuidedTour } from "../..";
+import { DEFAULT_RECT } from "../../api";
+import { CurrentTutorialContext } from "../../contexts";
+
+export const useSelectorHandler = () => {
+  const { currentStep, currentTutorial, latestUserInteraction, setCurrentRefElementPosition } = useContext(
+    CurrentTutorialContext
+  );
+  const dialogStep = getCurrentStep(currentStep, currentTutorial);
+  const selector = dialogStep?.selector || "";
```
tiagobento: Use `useMemo`

packages/guided-tour/src/components/utils/GuidedTourStepsHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+import { KogitoGuidedTour, Step, Tutorial, SubTutorialMode, DemoMode } from "../..";
+
+const concat = <T>(array1: T[], array2: T[]) => array1.concat(array2);
+const flatMap = <T>(f: (array: T) => T[], array: T[]) => array.map(f).reduce(concat, []);
+
+export const getCurrentStep = (currentStep: number, currentTutorial: Tutorial | undefined): Step | undefined => {
+  return getSteps(currentTutorial)[currentStep];
+};
+
+export const getSteps = (currentTutorial: Tutorial | undefined) => {
```
tiagobento: `currentTutorial?: Tutorial` will make this parameter optional.

packages/guided-tour/src/components/utils/GuidedTourStepsHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (!("label" in stepMode)) {
+      return [step];
+    }
+
+    const stepTutorialLabel = (stepMode as SubTutorialMode).label;
+    const stepTutorial = registeredTutorials.find(tutorial => {
+      return tutorial.label === stepTutorialLabel;
+    });
+
+    return stepTutorial?.steps || [];
```
tiagobento: Prefer using the `??` operator instead of `||`.

packages/guided-tour/src/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+  Step,
+  SubTutorialMode,
+  Tutorial,
+  UserInteraction,
+  DEFAULT_RECT
+} from "./api";
+import { GuidedTour } from "./components";
+import { GuidedTourDomUtils, GuidedTourEventBus, GuidedTourCookie } from "./core";
+
+class KogitoGuidedTour {
```
tiagobento, line 35: As we discussed, I'd rather have this class as a singleton instead of static.

packages/microeditor-envelope-protocol/src/EnvelopeBusOuterMessageHandler.ts
```diff
@@ -41,6 +42,9 @@ export interface EnvelopeBusOuterMessageHandlerImpl {
   notify_editorUndo(): void;
   notify_editorRedo(): void;
   receive_newEdit(edit: KogitoEdit): void;
+  receive_guidedTourUserInteraction(userInteraction: UserInteraction): void;
+  receive_guidedTourRegisterTutorial(tutorial: Tutorial): void;
+  receive_guidedTourElementPositionRequest(rect: Rect): void;
```
tiagobento: This should be named `receive_guidedTourElementPositionResponse`

packages/microeditor-envelope/src/api/tour/GuidedTourService.ts
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { UserInteraction, Tutorial } from "@kogito-tooling/guided-tour";
+
+/**
+ * PUBLIC ENVELOPE API
+ */
+export interface GuidedTourService {
```
tiagobento: Can we rename that to GuidedTourApi?

packages/online-editor/src/App.tsx
```diff
@@ -72,12 +74,16 @@ export function App(props: Props) {
     [file]
   );
 
+  useDMNTour(isEditorReady, file);
```
tiagobento: We're trying to keep acronyms as camel-case in this project. So this method should be called `useDmnTour`

packages/online-editor/src/common/GlobalContext.ts
```diff
@@ -25,6 +25,8 @@ export interface GlobalContextType {
   router: EmbeddedEditorRouter;
   readonly: boolean;
   external: boolean;
+  isEditorReady: boolean;
+  setIsEditorReady: (isEditorReady: boolean) => void;
```
tiagobento: If possible, I'd like to not have editor-specifics on the global context. The online-editor has pages that do not relate to the editor, so having something like that on the global context makes it feel like it's misplaced.

packages/online-editor/src/tour/GuidedTourInitializer.tsx
```diff
@@ -0,0 +1,102 @@
...
+export function useDMNTour(isEditorReady: boolean, file: File) {
+  useEffect(() => {
+    KogitoGuidedTour.setup();
+    return () => KogitoGuidedTour.teardown();
+  }, []);
+
+  useEffect(() => {
+    if (isEditorReady && file.editorType === EditorType.DMN) {
+      KogitoGuidedTour.registerTutorial(getOnlineEditorTutorial());
+      KogitoGuidedTour.start("DMN Online Editor Tutorial");
```
tiagobento: Should this string be the same as the tutorial label? I feel you could do `tutorial.label` here 😛

tiagobento (changes requested):
Hi @karreiro. Thanks for the PR. The feature looks great and I'm sure it will really increase the value of our Online Editor.

I left some comments inline, but as a general comment I'd like to point out that this was a very smooth review. I didn't comment everywhere I saw it, but declaring inner-functions is generally not a good idea when creating function components with React Hooks. Also, it's always good to memoize your values/functions (either with `useMemo` or `useCallback`), so that we're sure that no unnecessary renders are happening.

Please reach out if you have any questions :)

packages/guided-tour/.prettierrc
karreiro: Removed! :-)

packages/guided-tour/package.json
karreiro: Fixed! Thanks!

packages/guided-tour/src/components/HighlightLayer.tsx
karreiro: Thanks for the comment. I agree with the idea of using `useMemo` to avoid unnecessary calculations, but the `currentRefElementPosition` will almost every time hold a different value, so the amount of memoized values will be quit large. Thus, I believe we should keep `HighlightLayer` as is, but please let me know wdyt.

packages/guided-tour/src/components/NavigationControls.tsx
karreiro: Fixed! :-)

packages/guided-tour/src/components/NavigationControls.tsx
karreiro: Makes sense, thanks for the explanation, @tiagobento!

packages/guided-tour/src/components/utils/GuidedTourBusListeners.ts
karreiro: Renamed! Thanks!

packages/guided-tour/src/components/utils/GuidedTourBusListeners.ts
karreiro: Fixed!

packages/guided-tour/src/components/utils/GuidedTourBusListeners.ts
karreiro: Yeah, I agree, but this kind of listener must have its setup before the browser paints the change, see:

img1

_Source: Learning React, 2nd Edition (by Alex Banks, Eve Porcello, release Date: June 2020)_

packages/guided-tour/src/components/utils/GuidedTourContent.tsx
karreiro: Thanks! I applied the fix on `Dialog.tsx`.

packages/guided-tour/src/components/utils/GuidedTourContent.tsx
karreiro: +1 for that! Thanks! :-)

packages/guided-tour/src/components/utils/GuidedTourContent.tsx
karreiro: Fixed!

packages/guided-tour/src/components/utils/GuidedTourSelectorHandler.ts
karreiro: Fixed!

packages/guided-tour/src/components/utils/GuidedTourSelectorHandler.ts
karreiro: Fixed!

packages/guided-tour/src/components/utils/GuidedTourStepsHelper.ts
karreiro: Fixed! Thanks!

packages/guided-tour/src/components/utils/GuidedTourStepsHelper.ts
karreiro: Fixed! (I'm keeping it only on `HighlightLayer`, because there it has a logical operation (since I want to return the right-side when the left-side is `false`).

packages/guided-tour/src/index.tsx
karreiro: Yeah! It's a singleton now!

packages/microeditor-envelope-protocol/src/EnvelopeBusOuterMessageHandler.ts
karreiro: Renamed!

packages/microeditor-envelope/src/api/tour/GuidedTourService.ts
karreiro: Renamed!

packages/online-editor/src/App.tsx
karreiro: Renamed!

packages/online-editor/src/common/GlobalContext.ts
karreiro: Fixed! Good catch, thanks! :-)

packages/online-editor/src/tour/GuidedTourInitializer.tsx
karreiro: Well spotted 😜

karreiro:
@tiagobento I've answered your comments inline, thank you for the valuable tips.

----

@jomarko Thank you for your review and the reported issues. Regarding each item you've raised:

- 1 - I believe we can handle this fullscreen bug separately here: https://issues.redhat.com/browse/KOGITO-2513

- 2 - So... it will be handled by [KOGITO-2513](https://issues.redhat.com/browse/KOGITO-2513)

- 3 - Yeah, this is an issue, but it's being fixed by Tiago here: https://github.com/kiegroup/kogito-tooling/pull/185

- 4 - Makes sense... Fixed!

- 5 - Well spotted :-) Fixed!

- 6 - That's a good idea. Fixed!

- 7 - I've added some enhancements on the grid observer, and I couldn't reproduce this issue after that. Before, these changes, I could intermittently reproduce it. So, I believe it's fixed! :-)

- 8 - We would need extra infrastructure code to get this kind of information. Maybe we can have that kind of infra in the future, but for now I believe we can just point the link by using the dialog.

- 9 - Fixed!

- 10 - Well spotted. Fixed!

- 11 - Makes sense :-) Fixed!

- 12 - Well spotted, thanks. They are fixed now.

packages/guided-tour/src/components/HighlightLayer.tsx
tiagobento: When a new value comes to play, the old value is discarded and garbage-collected, so no need to worry about memory consumption.

karreiro: Thanks for the explanation :-) Fixed!

tiagobento (approved):

karreiro:
(conflicts fixed)

karreiro:
(conflicts fixed) 👀

ederign (approved):
Awesome work @karreiro

karreiro:
Hi @jomarko, I'm not sure if you could solve the problems to build this ensemble locally, but you can test this PR by following these steps:

- Unzip [this zip file](https://www.dropbox.com/s/b5c4d4osyhl45yv/dmn.zip?dl=0) into the `packages/kie-bc-editors-unpacked/dmn/` directory;
- Run `yarn run start` on `packages/online-editor`.

I hope this is helpful :-)

jomarko:
### Re-review 01
New issue reported https://issues.redhat.com/browse/KOGITO-2602

jomarko:
### Re-review 02
After user select some expression type the tour step provides link for 'learn-in-15-minutes'. This should be also opened in new tab. see [video](https://issues.redhat.com/secure/attachment/12476527/learn-in-15-minutes.webm).

jomarko:
### Re-review 03
New issue reported https://issues.redhat.com/browse/KOGITO-2603

jomarko:
### Re-review 04
New issue reported https://issues.redhat.com/browse/KOGITO-2605

packages/guided-tour/src/__tests__/components/NavigationControls.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+      },
+      currentTutorial: { steps: [{}, {}, {}, {}] }
+    };
+
+    act(() => {
+      useContextMock(ctx);
+      render(<NavigationControls />);
+
+      triggerClick("[data-kgt-next]");
+    });
```
jomarko, line 42: I am not able to manually switch between steps. Is this  test for tour that will be launched in other mode?

karreiro: Yeah, currently none of the current steps is enabling button-based navigation. This is something we already support, but this feature is not being used on the current steps.

karreiro:
Thank you for the comments, @jomarko. May I handle the “Re-review 2” in a separate JIRA as well? I can open another PR by the EOD. But I would like to get this ensemble in :-)

Thanks again for the comments. Well spotted issues.

jomarko (approved):

PR #189, opened 2020-06-22
FAI-178: FAI-179: PMML Scorecard marshaller

manstis:
This supersedes #174 

I replaced the earlier PR as my local folder structure (in `pmml-editor`) had been changed quite considerably and included other changes for FAI-179. It was therefore easier to restructure into two `pmml-editor-xxx` modules on my FAI-179 branch than the original FAI-178 branch.  

This PR covers:
- https://issues.redhat.com/browse/FAI-178
- https://issues.redhat.com/browse/FAI-179


.github/workflows/publish_artifacts.yml
```diff
@@ -81,4 +81,4 @@ jobs:
           NPM_TOKEN: ${{ secrets.KIEGROUP_NPM_TOKEN }}
         run: |
           echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
-          cd kogito-tooling && npx lerna exec --scope @kogito-tooling/* --ignore @kogito-tooling/kie-bc-editors-unpacked --ignore @kogito-tooling/desktop --ignore @kogito-tooling/hub --ignore @kogito-tooling/pmml-editor -- npm publish --access public && cd -
+          cd kogito-tooling && npx lerna exec --scope @kogito-tooling/* --ignore @kogito-tooling/kie-bc-editors-unpacked --ignore @kogito-tooling/desktop --ignore @kogito-tooling/hub --ignore @kogito-tooling/pmml-editor --ignore @kogito-tooling/pmml-editor-codegen -- npm publish --access public && cd -
```
manstis: Exclude `pmml-editor-codegen` from publication (`pmml-editor` is already excluded).

packages/pmml-editor-codegen/src/generated/pmml-4-4.xsd
```diff
@@ -0,0 +1,4426 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns="http://www.dmg.org/PMML-4_4" elementFormDefault="unqualified" targetNamespace="http://www.dmg.org/PMML-4_4"><xs:element name="DefineFunction">
```
manstis: This is the PMML 4.4 XSD from the internet.. Nothing to review here... Move on.

packages/pmml-editor-codegen/src/generated/pmml.ts
```diff
@@ -0,0 +1,5562 @@
+/* tslint:disable */
```
manstis: This file is generated.. Nothing to review here.

manstis:
@tiagobento @jomarko This PR is ready for review.

@tiagobento It now contains split packages. @jomarko It now has everything necessary for the Scorecard Editor to begin... @kelvah FYI. Did you want to review too? I can do a quick call for you and @jomarko to provide an overview as I know @tiagobento found it invaluable to understand what's going on.

kelvah:
> I can do a quick call for you and @jomarko to provide an overview as I know @tiagobento found it invaluable to understand what's going on.

hi @manstis, sure, let's do the call. Thank you.

.github/workflows/publish_artifacts.yml
tiagobento: @manstis Sorry for not commenting that earlier, but don't you want `pmml-editor` to be published? How are you going to use it on trusty-AI if it's not? Or do you not want to publish it until you have a more "stable" version?

tiagobento (approved):
@manstis I didn't notice it then, but since you added new libraries to `pmml-editor`, I thought I give a look to the "externals" entry on `pmml-editor/webpack.config.js` and it turns out you don't have any externals declared.

The impact of that is that all libraries will be included in this package's `index.js`, resulting in a big module, since it depends on things like React, etc. Since `pmml-editor` is a library module, I think you could go ahead and add everything as external for this module like so `externals: [nodeExternals({ modulesDir: "../../node_modules" })]`. Does that make sense?

manstis:
@tiagobento We will publish `pmml-editor` when it exists (at the moment it's massively incomplete and only exists as a marshaller... UI, integration with channels etc all to follow... then, and only then, should we publish it IMO).

tiagobento:
@manstis Ok, perfect! Thank you. My review is done :))

.github/workflows/publish_artifacts.yml
manstis: Just to complete the circle, as discussed on Zulip, both `pmml-editor` and `pmml-editor-codegen` are not to be published yet as the editor is massively incomplete.. When it reaches maturity we will move to publishing it along with the other packages.

kelvah (commented):
Everything is fine for me.
I opened [a ticket](https://issues.redhat.com/browse/FAI-212) to investigate if `pmml-editor-codegen` could be a ts only package, without the js transpilation.

kelvah (approved):

manstis:
Hi @jomarko this is ready for your review.

packages/pmml-editor-codegen/pom.xml
```diff
@@ -0,0 +1,282 @@
...
+        <version.org.jpmml.pmml-model>1.5.1</version.org.jpmml.pmml-model>
+        <version.org.jpmml.pmml-agent>1.5.1</version.org.jpmml.pmml-agent>
+        <version.com.fasterxml.jackson.jaxrs.jackson-jaxrs-json-provider>2.11.0</version.com.fasterxml.jackson.jaxrs.jackson-jaxrs-json-provider>
+        <version.org.eclipse.persistence.org.eclipse.persistence.core>2.7.7</version.org.eclipse.persistence.org.eclipse.persistence.core>
+        <version.org.glassfish.jaxb.jaxb-runtime>2.3.3</version.org.glassfish.jaxb.jaxb-runtime>
+        <version.cz.habarta.typescript-generator.typescript-generator-maven-plugin>2.24.612</version.cz.habarta.typescript-generator.typescript-generator-maven-plugin>
+        <version.org.apache.maven.plugins.maven-resources-plugin>3.1.0</version.org.apache.maven.plugins.maven-resources-plugin>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>1.8</maven.compiler.source>
```
jomarko: In `kogito-runtimes` we already use `<version.jdk>11</version.jdk>`, do we still should keep `8` in tooling? Probably question for @ederign

packages/pmml-editor-codegen/src/generated/pmml-4-4.xsd
jomarko: Maybe in some comment we could mentioned where it was downloaded. Could facilitate future tasks like upgrade PMML model version. However not sure if comment in `xsd` will not cause some troubles.

packages/pmml-editor/src/__tests__/marshaller/DataDictionary.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      </PMML>
+    `);
+
+    expect(pmml).not.toBeNull();
+
+    const dataDictionary: DataDictionary = pmml.DataDictionary;
+    const dataField: DataField = new DataField({
+      name: "field1" as FieldName,
+      dataType: "string",
+      optype: "categorical"
```
jomarko, line 52: Looking on `pmml.ts` also `dataType` and `optype` has some non buil in type. Why we do not need to use `as` operator, as we do for `name`

packages/pmml-editor/src/__tests__/marshaller/Interval.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const dataDictionary: DataDictionary = pmml.DataDictionary;
+    expect(dataDictionary.DataField.length).toBe(1);
+    expect(dataDictionary.DataField[0].Interval).not.toBeUndefined();
+    expect(dataDictionary.DataField[0].Interval?.length).toBe(0);
+
+    const xml: string = PMML2XML(pmml);
+
+    expect(xml).toContain(`<DataField name="field1" optype="categorical" dataType="string"/>`);
+  });
```
jomarko: I would like to double check this test is correct. I am looking on `pmml.ts`. There is class with mandatory `closure` field, however I do not see it in `pmml` or `xml` variable.

I think I am confused from 
```
    expect(dataDictionary.DataField[0].Interval).not.toBeUndefined();
    expect(dataDictionary.DataField[0].Interval?.length).toBe(0);
```
We are saying some interval to be defined - first line
But then we are again saying maybe is not there - question mark on second line

Or it is just my wrong understanding of typescript?

packages/pmml-editor/src/__tests__/marshaller/Interval.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    const interval: Interval = intervals[0];
+    expect(interval).not.toBeUndefined();
+
+    intervals.splice(0, 1);
+
+    const xml: string = PMML2XML(pmml);
+
+    expect(xml).toContain(`<DataField name="field1" optype="categorical" dataType="string"/>`);
+  });
+});
```
jomarko, line 163: Should we add also test where is just `leftMargin` without `rightMargin` and vice versa? not PMML expert, maybe does not make sense at all.

packages/pmml-editor/src/__tests__/marshaller/Scorecard.test.ts
```diff
@@ -0,0 +1,551 @@
...
+} from "@kogito-tooling/pmml-editor-codegen";
+import * as XMLJS from "xml-js";
+import { PMML2XML, XML2PMML } from "../../marshaller";
+import {
+  SCORE_CARD_BASIC_COMPLEX_PARTIAL_SCORE,
+  SCORE_CARD_COMPOUND_PREDICATE,
+  SCORE_CARD_NESTED_COMPLEX_PARTIAL_SCORE,
+  SCORE_CARD_NESTED_COMPOUND_PREDICATE,
+  SCORE_CARD_SIMPLE_PREDICATE
+} from "./TestData_ScoreCards";
```
jomarko, line 40: Shouldn't we use 4.4 model version instead of 4.2 inside of these constants?

packages/pmml-editor/src/__tests__/marshaller/Scorecard.test.ts
```diff
@@ -0,0 +1,551 @@
...
+    //Characteristic 1, Attributes
+    expect(characteristic1Attributes.length).toBe(2);
+    expect(characteristic1Attributes[0].predicate).toBeInstanceOf(SimplePredicate);
+    expect(characteristic1Attributes[1].predicate).toBeInstanceOf(True);
+
+    //Characteristic 1, Attributes' ComplexPartialScores
+    expect(characteristic1Attributes[0].ComplexPartialScore).not.toBeUndefined();
+    expect(characteristic1Attributes[1].ComplexPartialScore).toBeUndefined();
+    const c1a0cps: string = XMLJS.json2xml(JSON.stringify(characteristic1Attributes[0].ComplexPartialScore));
+    expect(c1a0cps).toBe(
```
jomarko, line 564: using such variables is quite risk. as it is not human readable, quite easily we can overlook that using `c1a0cps` instead of `c0a0cps` or similar. I am not saying this is case here. 
However more preferable would be probably have smaller functions for verification of same things. 

At the same time this is test for generated class, that will probably not change much. I think leaving the decision up to you.

jomarko (commented):
I am done with review of `kogito-tooling/packages/pmml-editor/src/__tests__/marshaller`

If I am correct there are tests of `pmml.ts`. Just for record, that file exports ~300 classes and we have tests for <10 of them. I am no PMML expert and can not assess if coverage is enough. 

Furthermore all covered classes have optional `extension` field, however we do not cover it. Again I can not asses if that is issue. For DMN and BPMN we used extension for non standard requirements as coloring, positioning of elements. Not sure if sth. like that needed for PMML.

Reviewed 21/38

packages/pmml-editor-codegen/src/generated/pmml-4-4.xsd
manstis: Comment added.

packages/pmml-editor-codegen/pom.xml
manstis: JDK11 is needed by `kogito-runtimes` to run in Quarkus.

JDK11 is not supported for GWT and hence the majority of UI related work in `kiegroup` uses JDK1.8.

I only need the JDK1.7+ to compile Java source classes to then generate the TypeScript classes. 

JDK1.7+ is needed as the Java source files contain the _diamond_ `<>` operator.

There is no runtime element that actually uses Java. 

Switching between JDK11 and JDK1.8 is a PITA as the latter does not contain Javax annotations by default. Consequentially many of our UI related modules in `appformer`, `kie-wb-common` or `drools-wb` compile with JDK11 without extensive changes to the `pom`'s. There is work underway to move everything to JDK11 however it needs a new version of GWT, Errai and changes to a lot of our poms.

packages/pmml-editor/src/marshaller/jsonata/json2ui/Header.ts
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+export const HEADER: string = `
+"Header": elements.elements[(name = "Header")] ~> $map(function($v, $i) {
+  {
+    "copyright": $v.attributes.copyright,
+    "description": $v.attributes.description,
+    "modelVersion": $v.attributes.modelVersion,
+    "Application": $v.elements[(name = "Application")].attributes,
```
jomarko, line 22: is this correct? in `pmml.ts` just annotation seems to be array. `attributes` makes me feel we return array here.

jomarko (commented):
Finished also `jsonata` part. However I do not know details of all those transformations `$map` `$numer` ... As we store those logic into string constants, I was focusing more on `typo` mistakes. Didn't found such.

packages/pmml-editor/src/__tests__/marshaller/DataDictionary.test.ts
manstis: I'll try to give you a really detailed Typescript explanation :-)

`name` is of type `FieldName` that is a class.
`optype` is of type `OpType` that is an alias for a list of allowed `string` values.
`dataType` is also of type `DateType` that again is an alias for a list of allowed `string` values.

Therefore I do not need to _cast_ the `string` values for either `optype` or `datatType`. But they must be a value defined in the list. However `name` needs to be an object of type `FieldName`. I can either write object syntax `name: { value: "hello" }` or cast the value to the object with `as` i.e. `name: "hello" as FieldName`. I'm happy to do either... the object form is easier to understand, the cast form is shorter.

packages/pmml-editor-codegen/pom.xml
jomarko: ok, thank you

packages/pmml-editor/src/__tests__/marshaller/DataDictionary.test.ts
jomarko: thank you, keep it as it is then

packages/pmml-editor/src/__tests__/marshaller/Interval.test.ts
manstis: This test asserts that the collection of `Interval` objects exists (not `undefined`) but is empty. As opposed to being `undefined` in which cases I would not be able to do `dataField.Interval.push(new Interval(...)` without first checking if `dataField.Interval === undefined` and then setting it to an empty array `dataField.Interval = []`. 

I have updated the name of the test.

packages/pmml-editor/src/__tests__/marshaller/Scorecard.test.ts
manstis: Fair point. Let me refactor a bit.

packages/pmml-editor/src/marshaller/jsonata/json2ui/Header.ts
manstis: This is black magic, a dark art... It is a JSONata mapping that is effectively the JSON equivalent of a XSLT in the XML world. `$map`, `$number`, `$append`, `$merge` used in this PR are JSONata functions. 

L#22 map the following:
```
"elements": [
  {
    "type": "element",
    "name": "Application",
    "attributes": {
      "name": "name",
      "version": "version"
    }
  }
]
```
To
```
"Application": {
  "name": "name",
  "version": "version"
}
```

packages/pmml-editor-codegen/pom.xml
```diff
@@ -0,0 +1,282 @@
...
+            <url>https://repository.jboss.org/nexus/content/groups/public</url>
+            <snapshots>
+                <enabled>true</enabled>
+                <updatePolicy>always</updatePolicy>
+            </snapshots>
+        </repository>
+    </repositories>
+
+    <properties>
+        <version.org.jpmml.pmml-model>1.5.1</version.org.jpmml.pmml-model>
```
gitgabrio: hi @manstis 
Those two versions are already defined somewhere in our bom: why is it necessary to declare again theme here ? (just to understand)

packages/pmml-editor-codegen/pom.xml
```diff
@@ -0,0 +1,282 @@
...
+                <configuration>
+                    <!-- The jpmml-model classes contain both JAXB and Jackson2 annotations. However the -->
+                    <!-- JAXB annotations contain better concrete information on optional properties so use -->
+                    <!-- those annotations to generate the TypeScript -->
+                    <jsonLibrary>jaxb</jsonLibrary>
+                    <classPatterns>
+                        <pattern>org.dmg.**</pattern>
+                    </classPatterns>
+                    <customTypeNaming>
+                        <naming>org.dmg.pmml.AnovaRow$Type:AnovaRowType</naming>
```
gitgabrio: @manstis 
Did you manually generate all those entries?

packages/pmml-editor-codegen/src/generated/pmml-4-4.xsd
gitgabrio: @manstis 
IMHO there should be no file inside "generated" folders - i.e those folders are usually meant for files that are created during build - generally those folder should be excluded from VCS

packages/pmml-editor-codegen/src/generated/pmml.ts
gitgabrio: @manstis 
Same comment as above - if this file is generated at every build, I think it should not put under VCS. If, on the other side, it is not, I think  it should not be in "generated" folder.

gitgabrio (changes requested):
Hi @manstis 
Beside the position of a couple of files, I would like to ask you to add some test (if not already present) where the model name is not defined.
Thx

packages/pmml-editor/src/__tests__/marshaller/Scorecard.test.ts
manstis: Done.

packages/pmml-editor/src/__tests__/marshaller/Scorecard.test.ts
manstis: Updated.. much nicer. Thank-you for pulling me up on it!

packages/pmml-editor/src/__tests__/marshaller/Interval.test.ts
manstis: The properties are optional so I have added a couple of tests covering missing margins.

packages/pmml-editor-codegen/pom.xml
manstis: No. It was a lovely morning spent eliminating issues generating the TS files...

manstis:
Hi @gitgabrio thank-you for looking at this PR.

To answer your questions:-
- I've moved the PMML XSD to a folder excluded from the build. The XSD is really only there for my convenience. It is not used by the Maven plugins however it will be used when I add XML validation - to ensure we're only attempting to open a valid PMML file. The `XML-JS` library is not a _validating_ XML library; so I need to find something to valid the file against the XSD before loading..... 

- The `pmml.ts` file **is** generated however not on every build as it's a once in a (PMML version) lifetime event. Therefore I prefer to (1) keep it in the `generated` folder, (2) commit it to VCS - it's generation is a relatively expensive operation and I don't want to slow builds by ~30s just to rebuild the same file with a different timestamp in the comment in the source code.

- Why do I include the `jpmml` version here... well.. You have to go and open that can of worms. The kogito-tooling repository is not handled by our build and release [mechanism](https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/script/release). Therefore even if I do set the `pom` in this PR to have it's parent as `kiegroup`'s `BOM` the parent version I reference will not be updated as successive `BOM` releases are made. Therefore if you were to change the version of `jpmml` in the `BOM` this module would always depend on an older version. Manually changing to a newer `BOM` version is no different to manually changing to a newer `jpmml` version (work/effort-wise). That said it should be something investigated further. See https://issues.redhat.com/browse/KOGITO-2637

packages/pmml-editor-codegen/pom.xml
manstis: See my [comment](https://github.com/kiegroup/kogito-tooling/pull/189#issuecomment-653030728).

gitgabrio:
> Hi @gitgabrio thank-you for looking at this PR.
> 
> To answer your questions:-
> 
> * I've moved the PMML XSD to a folder excluded from the build. The XSD is really only there for my convenience. It is not used by the Maven plugins however it will be used when I add XML validation - to ensure we're only attempting to open a valid PMML file. The `XML-JS` library is not a _validating_ XML library; so I need to find something to valid the file against the XSD before loading.....
+1

> * The `pmml.ts` file **is** generated however not on every build as it's a once in a (PMML version) lifetime event. Therefore I prefer to (1) keep it in the `generated` folder, (2) commit it to VCS - it's generation is a relatively expensive operation and I don't want to slow builds by ~30s just to rebuild the same file with a different timestamp in the comment in the source code.
Gotcha... I always thought to "generated" folder as "always generated by build", anyway it is a matter of personal taste


> * Why do I include the `jpmml` version here... well.. You have to go and open that can of worms. The kogito-tooling 
repository is not handled by our build and release [mechanism](https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/script/release). Therefore even if I do set the `pom` in this PR to have it's parent as `kiegroup`'s `BOM` the parent version I reference will not be updated as successive `BOM` releases are made. Therefore if you were to change the version of `jpmml` in the `BOM` this module would always depend on an older version. Manually changing to a newer `BOM` version is no different to manually changing to a newer `jpmml` version (work/effort-wise). That said it should be something investigated further. See https://issues.redhat.com/browse/KOGITO-2637

I sort of feared that.

gitgabrio (approved):

ederign (changes requested):
I would like to discuss this further. @manstis I'll drop you an email.

jomarko (approved):

manstis:
Hi @ederign any news on the email, so I can address any issues?

ederign (approved):

ederign:
LGTM! @tiagobento can you please merge this when green?

PR #193, opened 2020-06-24
KOGITO-2542: Update Hub to install VS Code extension from Marketplace

tiagobento:
See https://issues.redhat.com/browse/KOGITO-2542

### DIFF
--- a/packages/hub/src/webview/App.tsx
+++ b/packages/hub/src/webview/App.tsx
@@ -114,7 +114,7 @@ export function App() {
 
   const vscode_install = useCallback((e: React.MouseEvent) => {
     e.preventDefault();
-    electron.shell.openExternal("vscode:extension/kie-group.vscode-extension-kogito-bundle");
+    electron.shell.openExternal(`vscode:extension/${Constants.VSCODE_EXTENSION_PACKAGE_NAME}`);
   }, []);
 
   const vscode_requestUninstall = useCallback(() => {
--- a/packages/hub/src/webview/App.tsx
+++ b/packages/hub/src/webview/App.tsx
@@ -27,7 +27,6 @@ import {
   AlertProps,
   Brand,
   Button,
-  ButtonVariant,
   Card,
   CardBody,
   CardFooter,
@@ -36,7 +35,6 @@ import {
   Dropdown,
   DropdownItem,
   Gallery,
-  InputGroup,
   KebabToggle,
   Modal,
   Page,
@@ -44,7 +42,6 @@ import {
   PageSection,
   Text,
   TextContent,
-  TextInput,
   TextList,
   TextListItem,
   TextListVariants,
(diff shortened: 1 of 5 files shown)


ljmotta (approved):
LGTM!

ederign (approved):

packages/hub/src/webview/App.tsx
```diff
@@ -106,70 +101,26 @@ export function App() {
...
   }, [vscode_kebabOpen]);
 
-  const vscode_requestInstall = useCallback(() => {
-    electron.ipcRenderer.send("vscode__install_extension", { location: vscode_location });
-    setVscode_installModalOpen(false);
-    setVscode_status(ExtensionStatus.INSTALLING);
-  }, [vscode_location]);
+  const vscode_install = useCallback((e: React.MouseEvent) => {
+    e.preventDefault();
+    electron.shell.openExternal("vscode:extension/kie-group.vscode-extension-kogito-bundle");
```
jomarko: Should we rather use constant VSCODE_EXTENSION_PACKAGE_NAME ?

jomarko (commented):
- I am getting continuously [1] in HUB log, is this expected?
- I am able to install `Kogito Bundle`, however files are not opened with our editors see [video](https://issues.redhat.com/secure/attachment/12476098/kogito-bundle-installed.webm)
- I am able to launch VS code via HUB just once. When I closed VS Code and clicked launch second time, just [1] was appearing in HUB log but no VS code started

[1]
```
Executing command ' '/usr/share/code/bin/code' --list-extensions ' on LINUX
Success: true
JosefBiehler.cypress-fixture-intellisense
k--kato.intellij-idea-keybindings
PatternFly.patternfly-snippets
redhat.java
VisualStudioExptTeam.vscodeintellicode
vscjava.vscode-java-debug
vscjava.vscode-java-dependency
vscjava.vscode-java-pack
vscjava.vscode-java-test
vscjava.vscode-maven
```

packages/hub/src/webview/App.tsx
tiagobento: Sure! Good catch.

tiagobento:
@jomarko 

1. Yes, I added a 2s polling mechanism to check whether you have the extension installed or not.
2. That's odd. This extension is supposed to download the DMN Editor and BPMN Editor extensions. What's the version of VS Code that you're using? It must be 1.46.0 or above.
3. Hmm.. I've seen VS Code leave some processes open, which can confuse the Hub thinking it's open. Do you mind trying a `ps aux | grep [/usr/share/code]` command on the terminal for us to see the output?

tiagobento:
@jomarko Actually we spotted a bug on the Kogito Bundle extension, but we fixed that already. Can you please give it another try? Issue `2.` should be gone :)

jomarko:
@tiagobento thank you for updates.
- 2s polling - I understand now we can update the HUB UI (Install vs. Launch) thanks to this polling, however once the VSCode extension is detected as installed by HUB, I think we should stop polling. My understanding of HUB was an installer to speed up getting started steps. Or is the polling OK because most users close HUB once they select demanded kogito channel? 
- Opening files after installing `Kogito Bundle` works fine now.
- Re-launching the vscode extension works with some delay. Not possible to relaunch immediately after closing but need to wait few seconds. It is shown in the [video](https://issues.redhat.com/secure/attachment/12476215/hub-launch.webm) and [log](https://issues.redhat.com/secure/attachment/12476216/hub.log). That is not merge blocking, however we should ensure it is not a symptom of other problem.

Before merge I would like to ask if running command `ps aux` isn't a security risk. If I read the help correctly it lists processes of all machine users.

tiagobento:
@jomarko 

1. We can't stop polling after the extension is installed because you can uninstall it 😛 
2. Nice! Thanks for re-testing
3. Yes.. It happens on macOS too, I think VS Code takes a while to clean-up all of its processes after we tell it to close.

About the `ps aux`, we only run commands that the user can run and never ask for admin permission. Also we never run a `ps aux` in isolation, we always do that with `grep`, so our application doesn't "see" all the processes on the machine. More than that, we never store nor share this data. It all happens locally.

jomarko (approved):

PR #198, opened 2020-06-30
KOGITO-2580, KOGITO-2581: Cleanups

caponetto:
Cleanups:
- [KOGITO-2580](https://issues.redhat.com/browse/KOGITO-2580): Adding noImplicitAny + missing types
- [KOGITO-2581](https://issues.redhat.com/browse/KOGITO-2581): Delete isDirty method on core-api

### DIFF
--- a/packages/embedded-editor/src/__tests__/common/EditorTypes.test.ts
+++ b/packages/embedded-editor/src/__tests__/common/EditorTypes.test.ts
@@ -16,7 +16,7 @@
 import { EditorType } from "../../common/EditorTypes";
 
 function extractValue(key: string): string {
-  return EditorType[key];
+  return (EditorType as any)[key];
 }
 
 describe("EditorTypes::definitions", () => {
--- a/packages/keyboard-shortcuts/src/__tests__/api/DefaultKeyboardShortcutsService.test.tsx
+++ b/packages/keyboard-shortcuts/src/__tests__/api/DefaultKeyboardShortcutsService.test.tsx
@@ -123,7 +123,7 @@ describe("DefaultKeyboardShortcutsService", () => {
   });
 });
 
-function fire(type: "keydown" | "keyup", opts: any) {
+function fire(type: "keydown" | "keyup", opts: KeyboardEventInit) {
   fireEvent(window, new KeyboardEvent(type, opts));
 }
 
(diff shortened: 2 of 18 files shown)


packages/keyboard-shortcuts/src/__tests__/api/DefaultKeyboardShortcutsService.test.tsx
```diff
@@ -123,7 +123,7 @@ describe("DefaultKeyboardShortcutsService", () => {
   });
 });
 
-function fire(type: "keydown" | "keyup", opts) {
+function fire(type: "keydown" | "keyup", opts: any) {
```
tiagobento: I think this should be `KeyboardEventInit`?

packages/embedded-editor/src/__tests__/common/EditorTypes.test.ts
```diff
@@ -16,7 +16,7 @@
 import { EditorType } from "../../common/EditorTypes";
 
 function extractValue(key: string): string {
-  return EditorType[key];
+  return (EditorType as any)[key];
```
tiagobento, line 19: Really? 😅 

I figured TS would not ask for us to cast the enum type to `any` before indexing it, but ok :)

packages/online-editor/src/common/GithubService.ts
```diff
@@ -192,7 +192,7 @@ export class GithubService {
       .get({ gist_id: gistId })
       .then(response => {
         const filename = Object.keys(response.data.files)[0];
-        return response.data.files[filename].raw_url;
+        return (response.data as any).files[filename].raw_url;
```
tiagobento, line 195: Doesn't Octokit offer types for response data? I see that `octokit.gists.create` returns a `Promise<Octokit.Response<Octokit.GistsCreateResponse>>`, meaning that `response.data` has type `Octokit.GistsCreateResponse`, which has a `files` attribute. Do we need the cast to `any` here anyways? Same for line 186.

tiagobento (changes requested):
Thanks for the PR @caponetto! IMHO, things like that really make our codebase grow stronger. Few minor change requests inline.

packages/online-editor/src/common/GithubService.ts
caponetto: > Doesn't Octokit offer types for response data? I see that `octokit.gists.create` returns a `Promise<Octokit.Response<Octokit.GistsCreateResponse>>`, meaning that `response.data` has type `Octokit.GistsCreateResponse`, which has a `files` attribute. Do we need the cast to `any` here anyways? Same for line 186.

Octokit does offer types but they don't seem to work on these situations.
Same reason we had to use `any` [here](https://github.com/kiegroup/kogito-tooling/blob/master/packages/online-editor/src/common/GithubService.ts#L174), instead of the corresponding type.

tiagobento (approved):

ederign (approved):

PR #203, opened 2020-07-03
KOGITO-1883: Improving pattern management for ResourceContentAPI to support Windows enviroments.

yesamer:
ResourceContentApi provides a `list(pattern : string )` method which returns a list of files matching the given GLOB `pattern`.
In case the given GLOB patterns contains a path separator, the method doesn't work correctly in Windows environment (eg. scesim uses `src/**/*.dmn`), because the pattern is matched against windows path (eg. `C:\test\src\Test.dmn`)
To correctly support both POSIX and Windows environments, VSCode API provides the following hints:

```
     * Note: a backslash (`\`) is not valid within a glob pattern. If you have an existing file
     * path to match against, consider to use the [relative pattern](#RelativePattern) support
     * that takes care of converting any backslash into slash. Otherwise, make sure to convert
     * any backslash to slash when creating the glob pattern.
```
Therefore, I changed the parameter given to `findFiles()` method to a `RelativePattern` object;

```
         * The *difference* to the [`path`](#Uri.path)-property is the use of the platform specific
         * path separator and the handling of UNC paths. The sample below outlines the difference:
```
Then, I replaced `uri.path` with `uri.fsPath`

Linux
Screenshot from 2020-07-03 15-07-51

Windows
Cattura

@tiagobento @ederign Can you please review it OR assign to other reviewers?


packages/vscode-extension/src/VsCodeResourceContentService.ts
```diff
@@ -39,9 +39,10 @@ export class VsCodeResourceContentService implements ResourceContentService {
   }
 
   public async list(pattern: string, opts?: ResourceListOptions): Promise<ResourcesList> {
-    const expr = opts?.type === SearchType.ASSET_FOLDER ? this.currentAssetFolder + pattern : pattern;
-
-    const files = await vscode.workspace.findFiles(expr);
+    const workspaceFolder = vscode.workspace!.workspaceFolders![0].uri.fsPath + nodePath.sep;
```
tiagobento: Can you rename that to `workspaceFolderPath`?
Also, what happens if `vscode.workspace` is undefined?

packages/vscode-extension/src/VsCodeResourceContentService.ts
```diff
@@ -39,9 +39,10 @@ export class VsCodeResourceContentService implements ResourceContentService {
...
+    const basePath = opts?.type === SearchType.ASSET_FOLDER ? workspaceFolder + this.currentAssetFolder : workspaceFolder;
+    const includingRelativePattern = new RelativePattern(basePath, pattern);
```
tiagobento: Why is this called `includingRelativePattern` and not `relativePattern`?

tiagobento (changes requested):
Thanks for the PR @yesamer. Minor questions inline! Also, can you please configure Prettier on your IDE? I noticed that there are formatting issues.

packages/vscode-extension/src/VsCodeResourceContentService.ts
yesamer: @tiagobento I chose that name because of `findFile()` method parameters: 
`export function findFiles(include: GlobPattern, exclude?: GlobPattern ...) `
It supports both inclusion and exclusion GLOB pattern. My scope was to underline that an inclusion GLOB is currently passed.
Anyway, considering is the only parameter passed, I can rename it as suggested.

packages/vscode-extension/src/VsCodeResourceContentService.ts
yesamer: @tiagobento 
1. Sure!
2. In my opinion both `workspace` and `workspaceFolder` can't be `undefined`, because the code flow is called from a file which is opened inside a VSCode workspace. Consider that here https://github.com/kiegroup/kogito-tooling/pull/203/files/4839c7af7bafe41686a366dd494f684a8b900de9#diff-d79a48ded68ac338330c5545e6f185d5R68 there is the same assumption.

tiagobento: 2. So, it turns out `vscode.workspace` is a namespace, meaning that it cannot indeed be `undefined`. I asked because there's a `!` there, and that's usually means that you're forcing the TS compiler to ignore that this property _can_ be `undefined`. Can you please remote the `!` there then? :D

tiagobento: Now looking again at this code, I see that you're forcing the compiler to assume that `workspaceFolders` is not `undefined` and also you're assuming that it _has_ at least one element. The link you sent is just opening this PR, can you send me another one?
I'm afraid that when you open a a file without opening a folder, this code will break because `vscode.workspace.workspaceFolders` will be `undefined`, can you please check that?

tiagobento: image

yesamer: @tiagobento Good point, I'll check this case as well. Thank you!

yesamer: @tiagobento I double checked. Basically, it's correct to assume `vscode.workspace.workspaceFolders` is not undefined and with at least one element inside `VsCodeResourceContentService` class. The reason is inside `KogitoEditorFactory`, here https://github.com/kiegroup/kogito-tooling/pull/203/files/4839c7af7bafe41686a366dd494f684a8b900de9#diff-b162a87c3b2489e6e759295a8b2daacdL89. 
As you can see, the logic here determines which Service to instantiate: `VsCodeResourceContentService` or `VsCodeResourceContentService` or `VsCodeNodeResourceContentService`. The first one is instantiated iff the given `path` is inside a workspaceFolder (please refer to `isAssetInWorkspace(path: string)`) method. Therefore, if `VsCodeResourceContentService` is instantiated, `vscode.workspace.workspaceFolders` must be NOT undefinied.

tiagobento: Thanks @yesamer, that makes sense.

tiagobento (approved):

tiagobento:
@jstastny-cz Can you please take a look at this PR when you have time?

tiagobento:
@jstastny-cz @jomarko Hi guys, is some of you available to look at this PR? Should I tag someone else?

ederign (approved):

PR #213, opened 2020-07-16
KOGITO-2659: Update PatternFly to the latest version

ljmotta:
## Task
https://issues.redhat.com/browse/KOGITO-2659
https://issues.redhat.com/browse/KOGITO-2212

## Demos
Some changes were needed to fit the new PF version:
- Altered the unsaved warning alert to use the new PF layout
image

- Option to collapse the desktop sidebar (and automatically collapse it when resizes)
image
image


packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -90,57 +90,65 @@ export function EditorToolbar(props: Props) {
...
+        props.onSave,
+        props.onDownload,
+        props.onCopyContentToClipboard,
+        props.onExportGist
+      ]
+    );
 
   const filenameInput = (
-    <>
+    <React.Fragment>
```
tiagobento: This can be replaced by `<>`

packages/desktop/src/webview/home/HomePage.tsx
```diff
@@ -31,17 +31,77 @@ enum NavItems {
...
+        break;
+      case NavState.MANUAL_OPEN:
+        setNavState(NavState.RESIZED_CLOSE);
+        break;
+    }
+  }, [navState]);
+
+  const navIsOpen = useMemo(() => {
+    return navState === NavState.OPEN || navState === NavState.RESIZED_OPEN || navState === NavState.MANUAL_OPEN;
+  }, [navState]);
```
tiagobento: Can you please provide more details about this logic? Also, I think `isNavOpen` is more common for naming boolean variables.

tiagobento (changes requested):
Thanks for the PR @ljmotta! Looks really good. I left some comments inline.

packages/desktop/src/webview/home/HomePage.tsx
ljmotta: I changed the variable name and I've made the logic a little bit simpler.
When the application is first opened, the `Page` component will trigger the `onPageResize` callback and depending on the size of the screen, the `navState` will be changed to `RESIZED_OPEN` or `RESIZED_CLOSE`.
The `MANUAL_OPEN`/`MANUAL_CLOSE` prevent a behavior when the `nav` is manually opened/closed. To illustrate this behavior, suppose the following scenario:
1. Open our application and the nav is open.
2. Close manually the nav.
3. Resize to fullscreen. (or any size that would keep the `mobileView` equals to `false`)
4. The nav stay closed due to `MANUAL_OPEN`/`MANUAL_CLOSE`

When the user set the nav state to the spected state (a small window = nav closed or a big window = nav opened), the control of the nav goes back to `onPageResize`.

packages/online-editor/src/editor/EditorToolbar.tsx
ljmotta: Done.

packages/desktop/src/webview/home/HomePage.tsx
tiagobento: Ok, I get it know. Thanks @ljmotta!

tiagobento (approved):

packages/desktop/src/webview/App.tsx
```diff
@@ -16,14 +16,14 @@
...
-import "@patternfly/patternfly/patternfly-variables.css";
-import "@patternfly/patternfly/patternfly.css";
 import { Alert, AlertActionCloseButton, AlertVariant } from "@patternfly/react-core";
 import * as electron from "electron";
 import * as React from "react";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import "../../static/resources/style.css";
+import "@patternfly/patternfly/base/patternfly-variables.css";
+import "@patternfly/patternfly/patternfly-addons.scss";
+import "@patternfly/patternfly/patternfly.scss";
```
srambach: Does this mean that the application styles are loaded before the PatternFly styles? If so, this is incorrect - local application styles should always be after PatternFly.

ljmotta: Yes, my bad! I carelessly changed the import order. I double checked and I missed a PatternFly import on the EditorPage.tsx which was overriding the application styles as well.

packages/desktop/src/webview/editor/EditorToolbar.tsx
```diff
@@ -54,20 +62,20 @@ export function EditorToolbar(props: Props) {
 
   const headerToolbar = (
     // TODO: The toolbar should be switched out for DataToolbar and possibly the Overflow menu
```
srambach: This comment can be removed

packages/desktop/src/webview/editor/EditorToolbar.tsx
```diff
@@ -54,20 +62,20 @@ export function EditorToolbar(props: Props) {
...
-    <Toolbar>
-      <ToolbarGroup className={"kogito--right"}>
-        <ToolbarItem>
+    <PageHeaderTools>
+      <PageHeaderToolsGroup className={"kogito--right"}>
```
srambach: I don't see `.kogito--right` being passed through and it doesn't look like it's needed any more.

packages/desktop/src/webview/editor/EditorToolbar.tsx
ljmotta: Done

packages/desktop/src/webview/editor/EditorToolbar.tsx
ljmotta: Done

srambach:
Online editor, editor page - when narrow, the pageheader doesn't leave enough space around the logo. 
image

I think the best fix for this is to adjust the MinHeight of the page header - but this only needs to happen at the narrow width. Then we can use a media query to set it back to what it was. Unfortunately, we can't use the Patternfly breakpoint variables unless we start using scss, since you cannot use custom variables in a media query.

```
.kogito--editor__toolbar {
  --pf-c-page__header--MinHeight: 6.75rem;
}
@media (min-width: 1200px) {
  .kogito--editor__toolbar {
    --pf-c-page__header--MinHeight: unset;
  }
```
I have not exhaustively tested this...😁

ljmotta:
> Online editor, editor page - when narrow, the pageheader doesn't leave enough space around the logo.
> image
> 
> I think the best fix for this is to adjust the MinHeight of the page header - but this only needs to happen at the narrow width. Then we can use a media query to set it back to what it was. Unfortunately, we can't use the Patternfly breakpoint variables unless we start using scss, since you cannot use custom variables in a media query.
> 
> ```
> .kogito--editor__toolbar {
>   --pf-c-page__header--MinHeight: 6.75rem;
> }
> @media (min-width: 1200px) {
>   .kogito--editor__toolbar {
>     --pf-c-page__header--MinHeight: unset;
>   }
> ```
> 
> I have not exhaustively tested this...

It works really good. The desktop has the same problem so I added this same fix on it.

srambach:
This is looking really good :) 
I notice one more thing on the desktop. On the editor screen, the file name is not centered. It appears to be because the container `pf-c-page__header-nav` doesn't go the width of the window. You can see very very faintly that it doesn't because it's also a very slightly different color. (I think the color is also a mistake.) The nav container has a 14em margin-right on it, which maybe was trying to center the filename when next to the header tools, but shouldn't be there when the filename wraps to the next line. We might want to come up with another solution for the 14em also, since that's not really a foolproof way to get it centered :) 
image

ljmotta:
@srambach I don't think the color change is a mistake because I found a PatternFly example that has the same behavior, so I adjusted to the entire div to have the same color instead of just a piece. (https://www.patternfly.org/v4/documentation/react/components/page/horizontal-nav). For now I fixed the title centralization putting the `14rem` inside the `@media (min-width: 1200px)`, but we can think of another solution.

It's looking like this now:
image

srambach (approved):
LGTM

ederign (approved):

ederign:
@ljmotta please ping me when you fix the conflicts!

PR #218, opened 2020-07-20
KOGITO-2235: Fixed Scesim artifact generation

yesamer:
@paulovmr A follow up of your previous PR just to fix the Scesim directory as I reported in my mail this morning.

### DIFF
--- a/.github/workflows/pull_request_full_downstream_ci.yml
+++ b/.github/workflows/pull_request_full_downstream_ci.yml
@@ -121,7 +121,7 @@ jobs:
           fi
           mvn clean install -DskipTests -Denforcer.skip -pl drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-runtime -am -T 2
 
-      - name: Upload SceSim editor
+      - name: Upload Test Scenario editor
         uses: actions/upload-artifact@v1
         with:
           name: scesim-editor-fdb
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request_full_downstream_ci.yml
```diff
@@ -125,7 +125,7 @@ jobs:
         uses: actions/upload-artifact@v1
```
yesamer, line 125: @paulovmr A question regarding line 122: Usually changes on editor are not inside `drools-wb-scenario-simulation-editor-kogito-runtime`, which is only the module which generates the war file. Most of the time changes are inside `drools-wb-scenario-simulation-editor-kogito-client` or `drools-wb-scenario-simulation-editor-client` (but there are other modules to consider). Should we add them as well, in your opinion? Thanks!

paulovmr: Hi @yesamer , by using the `-am` parameter with this command we include all `drools-wb-scenario-simulation-editor-kogito-runtime` dependencies in the build, so if the runtime depends on those other modules, they will be built as well.

paulovmr (approved):

.github/workflows/pull_request_full_downstream_ci.yml
tiagobento: Hi @yesamer! The `-am` flag means "also make", which builds every depending module. From `mvn --help`:

```
 -am,--also-make                        If project list is specified, also
                                        build projects required by the
                                        list
```

yesamer: @paulovmr @tiagobento Thank you both for your synchronized answer! :D

PR #235, opened 2020-08-03
KOGITO-2877: Improve EmbeddedEditor and EmbeddedView usage experience

tiagobento:
On this PR:
- Removed `Router` and `Routes` abstractions. Created `EditorEnvelopeLocator` instead. The URL prefix to load resources like JS and CSS files are now passed in `EditorInitArgs` during Envelope initialization.
- Made `EmbeddedEditorRef` implement the `EditorApi` interface. The idea of `EmbeddedEditor` is to abstract the existence of the iframe and the Envelope application, so it makes sense that it exposes the same API as an `Editor`.
- Made `EmbeddedEditor` props implement `KogitoEditorChannelApi` partially, since some features are implemented by default on `EmbeddedEditor`.
- Removed `editorType` on `File`, since EmbeddedEditor is supposed to represent *any* Editor, and not only those which `kogito-tooling` provides.
- Made KogitoEditor on `vscode-extension` implement `EditorApi` too, for the same reasons of `EmbeddeEditorRef`.
- Created improved abstractions of `Envelope`s and `Channel`s. These will be further improved on a next PR.
- Improved `vscode-extension-pack-kogito-kie-editors` a little bit. This code will also be further improved on a next PR.

PR #236, opened 2020-08-03
FAI-235: PMML: Enhance marshaller for DMN's use case

manstis:
See https://issues.redhat.com/browse/FAI-235

This adds basic marshalling of ALL PMML model types for use by the DMN editor.


manstis:
@yesamer FYI... this will add support for all PMML model types to the client side marshaller.

At present only Scorecards and Regression Models are supported.

manstis:
@yesamer This should be ready to roll!

@tiagobento @jiripetrlik @kelvah (@ederign) @yesamer needs this to re-use the PMML client-side marshaller in DMN.

jiripetrlik (commented):
Were these files somehow automatically generated from XSD? Whch version have you used? Do you plan to somehow maintain or check it with next versions of PMML? As far as I know there is no Kie PMML implementation which supports these models.

jiripetrlik:
@manstis 
One more question. There is screenshot from Business Central in the JIRA. Is there something like that in Kogito? Where does it come from?

yesamer:
> @manstis
> One more question. There is screenshot from Business Central in the JIRA. Is there something like that in Kogito? Where does it come from?

@jiripetrlik I can answer to this question. That is the place where PMML are consumed in DMNEditor. I'm currently working for enabling the same feature in Kogito, for this reason I need to use the PMML marshaller @manstis wrote in this repo.

jiripetrlik:
> > @manstis
> > One more question. There is screenshot from Business Central in the JIRA. Is there something like that in Kogito? Where does it come from?
> 
> @jiripetrlik I can answer to this question. That is the place where PMML are consumed in DMNEditor. I'm currently working for enabling the same feature in Kogito, for this reason I need to use the PMML marshaller @manstis wrote in this repo.

@yesamer 
So, it is screenshot from 7.x stream right?

yesamer:
@jiripetrlik Yes. Once I'll complete my task I expect the same screenshot will be valid for Kogito too.

manstis:
Hi @jiripetrlik the main JIRA is https://issues.redhat.com/browse/FAI-174

The `pmml-editor-marshaller` is to ultimately be used by the (work in progress) editor.

Some files are generated (from PMML4.4 XSD) and some are not.

- This [file](https://github.com/kiegroup/kogito-tooling/tree/master/packages/pmml-editor-marshaller/src/marshaller/model) **is** generated. 
- These [files](https://github.com/kiegroup/kogito-tooling/tree/master/packages/pmml-editor-marshaller/src/marshaller/jsonata/json2ui) and [these](https://github.com/kiegroup/kogito-tooling/tree/master/packages/pmml-editor-marshaller/src/marshaller/jsonata/ui2json) are hand-written.
- We will support other [versions](https://issues.redhat.com/browse/FAI-220).

The XML->UI model marshalling works like this:-
- We have an object [model](https://github.com/kiegroup/kogito-tooling/blob/master/packages/pmml-editor-marshaller/src/marshaller/model/pmml4_4.ts) generated from the XSD (well, it's generated from Java classes generated from the XSD)
- We use a TypeScript [library](https://github.com/kiegroup/kogito-tooling/blob/master/packages/pmml-editor-marshaller/package.json#L17) to convert the XML to JAXB-like[1] JSON.
- We use another TypeScript [library](https://github.com/kiegroup/kogito-tooling/blob/master/packages/pmml-editor-marshaller/package.json#L16) to convert the JAXB-like[1] JSON to JSON matching the generated classes[2].

The longer term plan is to replace the use of `xml-js` and `jsonata` with an in-house XML marshaller that follows Java's Jackson but generates Java code that can be transpiled by the GWT compiler/J2CL into JavaScript and hence usable from TypeScript. It is a development/research work in progress at the moment.
  
Notes:
[1] JAXB-like JSON is like this:
```
{
  "elements": [
    {
      "type": "element",
      "name": "Header"
    },
    {
      "type": "element",
      "name": "DataDictionary",
      "attributes": {
        "numberOfFields": "3"
      }
    }
  ]
}
```
[2] JSON matching the object model is like this:
```
{
  "Header": {},
  "DataDictionary": {
    "numberOfFields": 3
  }
}
```

tiagobento (approved):
@manstis Approved. I just got a little confuse on how the SceSim Editor will consume this marshaller library, can you share your thoughts about that?

manstis:
@tiagobento Hello, well the _plan_ is to re-use the resulting TS/JS library the same as we do for others in Business Central. AFAIK this entails unpacking the NPM, copying the applicable JS file to somewhere within the consuming application; then having GWT inject the said JS into it's host HTML page; writing some Java/GWT classes annotation with `@JsInterop` (and related) annotations to expose the JS classes to GWT and then use like any other class inside GWT... [Here's](https://github.com/kiegroup/appformer/blob/master/uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml#L348) the starting point for where we use the Monaco editor from GWT too..... feel free to follow the code down the rabbit hole.

Of course we first need a new release of `kogito-tooling` to publish `pmml-editor-marshaller`.. (and then to fix anything I've probably forgotten/did not know to do in order for it to be consumed the same as Monaco).

@karreiro @yesamer Could you please confirm, correct or elaborate on what I've said above?

tiagobento:
@manstis Ok, I understand. My concern is that the editors come first on the "build pipeline", since `kogito-tooling` depends on them. If we make the SceSim/DMN editors depend on a package released by `kogito-tooling`, we'd be making it always the case that this package will always be one version behind everything.

To address that, I'd recommend adding the `pmml-marshaller` as a library on [`GwtMapping`](https://github.com/kiegroup/kogito-tooling/blob/406f3db4cd2d33e4492c47ca8a54d1f3cc258a28/packages/kie-bc-editors/src/GwtEditorMapping.ts), (See #235. Formerly `GwtEditorRoutes`). This way, we make sure that the DMN/SceSim Editors define a contract with JsInterop and that the `pmml-marshaller` implements it, like we do with other things we provide (e.g. Resource Content API). WDYT?

paulovmr:
Build: manstis/FAI-235

manstis:
@tiagobento Ah, I understand your concern completely.. and TBH I had not thought about that.

> To address that, I'd recommend adding the `pmml-marshaller` as a library on [`GwtEditorMapping`](https://github.com/kiegroup/kogito-tooling/blob/406f3db4cd2d33e4492c47ca8a54d1f3cc258a28/packages/kie-bc-editors/src/GwtEditorMapping.ts), (See #235. Formerly `GwtEditorRoutes`). This way, we make sure that the DMN/SceSim Editors define a contract with JsInterop and that the `pmml-marshaller` implements it, like we do with other things we provide (e.g. Resource Content API). WDYT?

IIUC are you suggesting to add `pmml-editor-marshaller` as a `(dev)-dependency` to `kie-bc-editors`?

Should it be a `dev` dependency or a full/normal one?

tiagobento:
@manstis Since we'll not use the PMML Marshaller directly on `kie-bc-editors`, as they only provide the glue code between Kogito Tooling and the Editors themselves, I believe we can add `pmml-marshaller` as a dependency of our apps (e.g. VS Code and Chrome Extensions, Desktop and Online). I think placing the library in a `pmml` folder on the same level as the other ones is a good idea. When we have the `pmml` editor itself, we can use the same folder to store its JS bundles etc, WDYT?

karreiro:
@tiagobento IIUC, the current guideline for Kogito Tooling libraries re-usability on editors is to expose APIs via envelope for now.

With that said, I believe that @yesamer will re-use the marshaller that @manstis made by exposing an envelope API that will be consumed in the editor (just like Guided Tour, Resource Content API, State Control, etc). This is a bit different from the approach we use for importing Monaco.

Am I missing something? :-)

tiagobento:
@karreiro It can be done both ways. We can either load the library like we do with `Jsonix` or expose it on `window.envelope`. Both ways will work fine, but I think for this case loading the library and consuming it directly makes more sense to me. Note that `window.envelope` is provided to all GWT-based editors at the moment, so BPMN would have PMML-Marshalling capabilities, although I think it won't need that at all. Making it this way would only make the EditorEnvelope bundle heavier for the BPMN Editor.

`window.envelope` is supposed to expose things that are provided by the EditorEnvelope, and I don't think that a PMML Marshaller is something that the EditorEnvelope provides, but rather a library that's packaged with the PMML Editor and can be loaded by other editors that need it.

A third option would be binding it to a global object that's not `window.envelope`, but it would have the same problem of making the EditorEnvelope bundle bigger for the BPMN Editor.

manstis:
I'm not sure I'm following @tiagobento @karreiro.

`online-editor` (for example) would have a dependency on `pmml-editor-marshaller`. How would that JS then get exposed to the editor itself? Firstly it'd be bundled into the single JS for the whole of `online-editor` (unless there's some webpack magic we can do), secondly it'd live "outside" the envelope (in kogito land, not editor land). `Jsonix` is bundled with the editor in editor land so is accessible via a [path](https://github.com/kiegroup/kogito-tooling/blob/master/packages/kie-bc-editors/src/GwtEditorMapping.ts#L69). We can't do that with the `pmml-editor-marshaller` w/o it being a version behind.

I'm on PTO as from next week (2 weeks). Does this PR need something major before it's mergable? Does it make sense to merge "as is" so others can contribute anything that may be needed to assist @yesamer while I'm on PTO?

tiagobento:
@manstis We can merge the PR. How we are going to make it usable in other Editors can be addressed separately.

The idea is to make the `pmml-editor-marshaller` be available through a path as well. We can copy the whole `dist/index.js` bundle produced by the `pmml-editor-marshaller` module and copy it to a `pmml` folder that's beside the `bpmn`, `dmn` and `scesim` folders. On the DMN Editor, we add a new path on the GwtMapping file, saying that we'll need to load ``${editorInitArgs.resourcesPathPrefix}/../pmml/marshaller.js`` too. This file would then create a PMMLMarshaller global namespace, where everything it provides would go under. In GWT code, we would create the respective JsInterop clases to map it. Does that make sense?

yesamer:
@manstis @tiagobento @karreiro For your information, I'm following the solution @karreiro described, it's almost completed.
My concern to evaluate a different solution is a mixture of lack of time and experience, in my case. Consider I'm a novice in TS/JS world and it's mandatory to have this feature available next week. Therefore, I'm available to consider a different solution, but I'll need a support in that case.

tiagobento:
@yesamer Don't worry. In practice, the difference would be that the methods you'd call would be in `window.PMMLMarshaller` instead of `window.envelope`. Let's wait for @manstis' and @karreiro's input so we can agree on a solution.

karreiro:
Considering the [work we have already done](https://github.com/kiegroup/kogito-tooling/compare/master...yesamer:KOGITO-2895?expand=1), I would just use the envelope for now. Also, this approach gives us the flexibility of doing some pre-processing before the GWT edit properly consumes the entities provided by the marshaller. However, let's stay vigilant for what we're sharing in the envelope on future development. Wdyt @tiagobento?

tiagobento:
@karreiro Sure. That doesn't seem like a big overhead for the moment. We should keep an eye on that and act accordingly in the future. +1 to keep binding the PMML Marshaller to the `window.envelope` object.

manstis:
@tiagobento OK, I completely understand your proposal.

> The idea is to make the `pmml-editor-marshaller` be available through a path as well. We can copy the whole `dist/index.js` bundle produced by the `pmml-editor-marshaller` module and copy it to a `pmml` folder that's beside the `bpmn`, `dmn` and `scesim` folders. On the DMN Editor, we add a new path on the `GwtEditorMapping` file, saying that we'll need to load `${editorInitArgs.resourcesPathPrefix}/../pmml/marshaller.js` too. This file would then create a `PMMLMarshaller` global namespace, where everything it provides would go under. In GWT code, we would create the respective JsInterop clases to map it.

@karreiro @yesamer OK, I completely understand your solution.

> Considering the work we have already done, I would just use the envelope for now. Also, this approach gives us the flexibility of doing some pre-processing before the GWT edit properly consumes the entities provided by the marshaller. However, let's stay vigilant for what we're sharing in the envelope on future development.

However adding the PMML marshaller to the envelope does expose it to every editor; even if not needed. That said (it's fortunate) DMN, PMML, BPMN (and possibly, one day, scesim) all need the PMML marshaller! Furthermore exposing it via the envelope does mean we get a single bundle with it; whereas adding separately to each editor would lead to the PMML bundle being loaded into each `IFRAME`.... pros and cons for both.

Since @tiagobento et al is the _architect_ of `kogito-tooling` I'd be inclined to incorporate his wishes longer term.

manstis:
@tiagobento @jiripetrlik I'm on PTO for two weeks as from the end of today (UK time).

I think it should be merged "as is" and it allows @yesamer to progress DMN's usage..  after @jiripetrlik has approved.

tiagobento:
@manstis Thanks for the reply, I'm glad we're able to reason about these trade-offs together. We'll go with adding the PMML Marshaller to the Envelope right now, but ideally we should migrate to not having editor-specific libraries bundled with the GWT Editors Envelope. Enjoy your PTO 😉

tiagobento:
Waiting for @jiripetrlik's and @kelvah's review then we're good to merge.

tiagobento:
@kelvah @jiripetrlik Can you please review that?

packages/pmml-editor-marshaller/src/__tests__/marshaller/TestData_AnomalyDetectionModel.ts
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+export const ANOMALY_DETECTION_MODEL_1: string = `
+<PMML xmlns="http://www.dmg.org/PMML-4_4" version="4.4"> 
+  <Header copyright="DMG.org"/>
+  <DataDictionary numberOfFields="4">
+    <DataField name="field1" optype="continuous" dataType="double"/>
+  </DataDictionary>
```
jomarko: same question for all models, shouldn't we have `numberOfFields="1"` instead of `numberOfFields="4"` ? Asking because I see everywhere just single `field1` `DataField`.

jomarko: furthermore, shouldn't we cover `numberOfFields` value in unit tests?

kelvah (approved):
Besides the latest observations by @jomarko, it looks fine for me.
Interesting discussion about how the marshaller should be consumed by other packages, but I agree that's a whole other topic.

packages/pmml-editor-marshaller/src/__tests__/marshaller/TestData_AnomalyDetectionModel.ts
manstis: Done in new commit. Thank-you @jomarko.

manstis:
Hi @jomarko @jiripetrlik this PR is ready for your re-review :-)

manstis:
CI issue (downloading dependencies).. retesting..

tiagobento:
@jiripetrlik @jomarko CI is green. Can you guys please take a look at this PR?

jomarko (approved):
Approving, however would be good to have tests also for some real models, that is just (not blocking) comment.  Maybe I will improve that when I will test effect of this PR on DMN Editor.

ederign (approved):

PR #250, opened 2020-08-12
KOGITO-3042: Replace MessageBusClient with proxified version of ApiToConsume

tiagobento:
Consuming the API provided by your counterpart as an Envelope or a Channel is not very complex, but making it be type-safe makes things way more complicated. This PR makes it possible to consume your counterpart APIs by using an object with the same interface that it implements. This was achieved by using a `Proxy` object. See [EnvelopeBusMessageManager.ts](https://github.com/kiegroup/kogito-tooling/compare/master...tiagobento:envelope-client-proxy-to-PR?expand=1#diff-f1ce77da7a0aaca4664b4e53167fd3ef).

I believe this is a much better way to communicate with the Envelope/Channel APIs than we had before, although just as type-safe.

I also provided a convenient `messageBusClientApiMock` function to be used by tests.


caponetto (approved):
Top!

tiagobento:
@caponetto Decided to remove the KogitoEditorEnvelopeServer too, since the way to call the Envelope API is now much better. 

Had to make the Proxy objects for `requests` and `notifications` a little bit more complex, since we need to spy on those methods on some of our tests (`EmbeddedEditor.test.tsx`). I needed to memoize the values returned by the `get` handlers and add a `set` handler, since that's what `jest.spyOn` needs internally. 

This has made the "proxied" objects behave much more like a normal object, too :)

Take a look at e4ee88c8ca34c909fa974fb07cf4c2cf9c69687d

EDIT: Oops, had to add a new commit to fix tests.

ljmotta (approved):
LGTM! :rocket:

tomasdavidorg (approved):
Looks good to me.

ederign (approved):

PR #256, opened 2020-08-13
KOGITO-1808: Editor's Backend Support

caponetto:
**Proposal for the backend support infrastructure**

This PR includes the following new packages:
- **backend** and its sub-modules:
  - api: Core API for supporting generic backend services. This module must not depend on any other package.
  - channel-api: Backend services API for channels. This module should include all API definitions that channels can support.
   - http-bridge: Bridge implementations between channels and http services. Required for channels supporting such a communication.
  - node: Common implementations related to backend services that target node.
- **backend-quarkus-runner-unpacked**: Retrieve the quarkus runner jar from maven (similar to what's done on kie-bc-editors-unpacked).
- **vscode-extension-backend**: The VS Code extension that should be installed along the editor extensions to support backend services.

Most changes are additions, i.e., the current code won't have any initial impact.

**All related PRs**:
- kogito-tooling: https://github.com/kiegroup/kogito-tooling/pull/256
- appformer: https://github.com/kiegroup/appformer/pull/1028
- kogito-tooling-java: https://github.com/kiegroup/kogito-tooling-java/pull/1

A complete documentation on the usage of the backend services infrastructure will be provided afterwards.
I've removed the sample code on the following commits, which can give an idea on how this works:
- kogito-tooling: https://github.com/kiegroup/kogito-tooling/commit/038601ca32c4803673114038307007f6fc2b3db8
- appformer: https://github.com/kiegroup/appformer/commit/e53551b0035f0ba1dea46bc982092c5507c45b6f
- kogito-tooling-java: https://github.com/kiegroup/kogito-tooling-java/commit/f337a0b34b3eb9b13f48c14eca25244b97a7f17c

**Additional notes**:
- The backend extension should be included into our bundle. This will be done once we have our first real backend service.


packages/backend-api/src/core/BackendManagerService.ts
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * @param bridge Service bridge. Required to start up HTTP services.
+   * @param localHttpServer Local HTTP server. Required to start up local HTTP services.
+   * @param initialServices Services that should be started up along with the backend manager.
+   * @param lazyServices Services that should be started upon the first usage.
+   */
+  public constructor(
+    private readonly bridge?: HttpBridge,
+    private readonly localHttpServer?: LocalHttpServer,
+    private readonly initialServices?: Service[],
```
paulovmr:
```suggestion
    private readonly bootstrapServices?: Service[],
```

packages/backend-node/src/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+      }
+
+      const [, , , major, minor, patch] = regexMatch;
+      const actualVersion: Version = { major: +major, minor: minor ? +minor : 0, patch: patch ? +patch : 0 };
+      resolve(isRequiredVersionSatisfied(requiredVersion, actualVersion));
+    });
+  });
+}
+
+function isRequiredVersionSatisfied(required: Version, actual: Version): boolean {
```
paulovmr: Shouldn't the following if's have else's before the next one? (i.e. required 9.1 and actual 10.0 should return true)

packages/vscode-extension/src/index.ts
```diff
@@ -31,13 +34,19 @@ import { generateSvg } from "./generateSvg";
...
+export async function startExtension(args: {
   extensionName: string;
   context: vscode.ExtensionContext;
   viewType: string;
   getPreviewCommandId: string;
   editorEnvelopeLocator: EditorEnvelopeLocator;
 }) {
+  backendProxy = new VsCodeBackendProxy();
+  if (!(await backendProxy.tryLoadBackendExtension())) {
+    // TODO: Show somehow to the user the possibility of using the backend extension.
```
paulovmr: Probably a native VS Code warning notification in the bottom right of the screen would be nice here.

paulovmr (changes requested):
Really nice! Just a few minor comments. :)

packages/backend-node/src/utils.ts
caponetto: You're right, thanks. I also spotted this issue while writing unit tests. The code is gonna be improved.

packages/vscode-extension/src/index.ts
caponetto: Agree. I guess a notification with two buttons would be nice: (1) "Install the extension"; (2) "Do not show again".

packages/vscode-extension/src/VsCodeBackendProxy.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import * as vscode from "vscode";
+import { BackendProxy, Service, Capability, CapabilityResponse } from "@kogito-tooling/backend-api";
+import { BackendExtensionApi } from "@kogito-tooling/backend-channel-api";
+
+/**
+ * Augmented {@link BackendProxy} for the VS Code channel.
+ */
+export class VsCodeBackendProxy extends BackendProxy {
+  private readonly BACKEND_EXTENSION_ID = "kie-group.backend-vscode-extension";
```
tiagobento: We shouldn't have this name hard coded here. Our extensions on the store will probably have a different name, so this should be configurable on the init method of our `vscode-extension` package.

packages/vscode-extension/src/index.ts
caponetto: Screenshot from 2020-08-18 13-55-29

packages/backend-vscode-extension/README.md
```diff
@@ -0,0 +1 @@
+Kogito Tooling Backend VSCode extension
```
tiagobento: Replace 'VSCode' by 'VS Code'

packages/backend-vscode-extension/src/backend/VsCodeBackendManagerService.ts
```diff
@@ -0,0 +1,31 @@
...
+export class VsCodeBackendManagerService extends BackendManagerService {
+  public constructor(context: vscode.ExtensionContext) {
+    super(
+      new DefaultHttpBridge(),
+      new QuarkusLocalServer(context.asAbsolutePath(path.join("dist", "server", "quarkus-runner.jar"))),
+      [],
+      []
+    );
+  }
+}
```
tiagobento: Can we make that with composition instead of inheritance?

packages/vscode-extension/src/KogitoEditor.ts
```diff
@@ -41,7 +41,13 @@ export class KogitoEditor implements EditorApi {
     private readonly messageBroadcaster: EnvelopeBusMessageBroadcaster,
     private readonly envelopeServer = new EnvelopeServer<KogitoEditorChannelApi, KogitoEditorEnvelopeApi>(
       {
-        postMessage: msg => this.panel.webview.postMessage(msg)
+        postMessage: msg => {
+          try {
+            this.panel.webview.postMessage(msg);
+          } catch (e) {
+            /* The webview has been disposed, thus cannot receive messages anymore. */
```
tiagobento, line 48: Is this really the only case in that this method can fail? If it isn't we will be masking error messages. Is there a way to not even call this method when after the webview is disposed? Looks like we're not "killing" everything properly before the Webview closes.

packages/vscode-extension/src/configUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+export function maybeShowInstallBackendExtensionHint(backendExtensionId: string): void {
+  const config = vscode.workspace.getConfiguration(CONFIG_SECTION);
+
+  if (!config.get<boolean>(BACKEND_HINT_CONFIG)) {
+    return;
+  }
+
+  const message = "Consider installing the backend extension to augment the capabilities of the editors.";
+  const installOption = "Install extension";
+  const dontShowAgainOption = "Don't show again";
```
tiagobento: This PR ir probably going to be merged after #257, so this will have to be updated.

packages/vscode-extension/src/index.ts
```diff
@@ -66,3 +79,10 @@ export function startExtension(args: {
     vscode.commands.registerCommand(args.getPreviewCommandId, () => generateSvg(editorStore, workspaceApi))
   );
 }
+
+/**
+ * Stops the Kogito extension.
+ */
+export function stopExtension() {
+  backendProxy?.stopServices();
+}
```
tiagobento: This package has become too coupled with the `Editor` abstraction. Not that we're adding a Backend that's dedicated to Editors, it's even more clear. I'm not saying we should do anything on this PR, I'm just commenting here so we can think together of how can we make this package more generic so that other extensions can be implemented using a different Envelope abstraction.

packages/vscode-extension-pack-kogito-kie-editors/package.json
```diff
@@ -90,7 +90,19 @@
...
+      "title": "Kogito",
+      "properties": {
+        "kogito.installBackendExtensionHint": {
+          "type": "boolean",
+          "description": "Specifies if a hint should be shown in case the backend extension is not installed.",
+          "default": true,
+          "scope": "application"
+        }
+      }
+    }
```
tiagobento: What happens if we have two extensions defining the same property? This will happen on the BPMN/DMN extensions bundles, since we install them both at the same time. Should we also parameterise this on `vscode-extension/src/index.ts`? We need BPMN and DMN extensions to work well together. Also, as a side note, do you know how to internationalize the "description" string?

packages/backend-vscode-extension/src/extension/extension.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  backendManager = new VsCodeBackendManagerService(context);
+  await backendManager.start();
+
+  return { backendManager: backendManager };
+}
+
+export function deactivate() {
+  backendManager?.stop();
+}
```
tiagobento: Since the activation elements are the languages we support in our editors, does that mean that when we close all editors, this extension will de-activate?

packages/backend-vscode-extension/webpack.config.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+const CopyWebpackPlugin = require("copy-webpack-plugin");
+const { merge } = require("webpack-merge");
+const common = require("../../webpack.common.config");
+
+module.exports = [
+  merge(common, {
+    output: {
+      library: "AppFormer.VsCodePack",
```
tiagobento: We should probably change that name.

tiagobento (changes requested):
@caponetto Thanks for the PR! I left some minor comments inline, but on a more general note, here are some thoughts:

1. It would be nice to have the `backend` packages be merged into a single package containing multiple modules, like we did on `@kogito-tooling/editor`. WDYT?
2. The `backendExtensionId` on `vscode-extension/src/index.ts` is not optional. Should we change that? We don't wanna enforce the creation of a backend by every one creating a VS Code extesion.
3. When we start having backend capabilities like the Java LS, how will we separate them from non-Java related editors? Let's imagine I'm creating an editor for a `.svg` file, since it will also know the KogitoEditorChannelApi, it will be able to call it too, when available. Is that our intent?
4. I think it's a *great* idea to have the Editors communicate with the Channel and not with the backends directly. It gives us nice encapsulation of the Editor and makes us much more flexible for changes on the backend communication implementation 😍
5. What on this PR wil lmove to `kogito-tooling-java`?
6. Something is not clear to me. How does the BackendProxy communicates with whatever is running on the `backend-vscode-extension`?

packages/vscode-extension/src/KogitoEditor.ts
caponetto: The only scenario that I noticed this fail was closing the editor while waiting for a promise to be resolved. It's not specific to backend though. Afaik, it's just an annoying warning log on the console but the user wouldn't notice anything.
Maybe we can think of a promise cancelation mechanism.

I couldn't find a proper way to check if messages can be posted before posting it.
Please let me know if you have any suggestions.

packages/backend-vscode-extension/src/extension/extension.ts
caponetto: No. From what I have observed on my tests and also reading the docs, the extension will be deactivated on the following scenarios: (1) VS Code is closed; (2) the extension is disabled; (3) the extension is uninstalled.

packages/backend-vscode-extension/webpack.config.js
caponetto: Yes, I'm open for suggestions :smiley:

packages/backend-vscode-extension/src/backend/VsCodeBackendManagerService.ts
caponetto: Actually, I'm thinking of getting rid of this class. 
It helped isolate things when I was developing but I think that we can do it directly [here](https://github.com/kiegroup/kogito-tooling/pull/256/files#diff-2c5a3b64337676b4b5d169e0b8727d06R26). Wdyt?

packages/vscode-extension-pack-kogito-kie-editors/package.json
caponetto: That's a very good question!
I couldn't find any doc about it but I've observed on my tests that two extensions can define/use the same property.
In the end, it is going to be a key/value pair on the user's settings.json.
Do you see the need of having one for each editor?

ATM I don't know how to do i18n with strings on package.json. Let me take a look and I'll get back to you.
Since it is just a boolean value, I'm 100% open for another approach that avoids the need of a configuration property.

packages/vscode-extension/src/KogitoEditor.ts
tiagobento: Yes, good point. I'll create a JIRA for us to cancel whatever pending requests are registered on the EnvelopeMessageBusManager.

packages/backend-vscode-extension/src/extension/extension.ts
tiagobento: Ok. Good. Thanks for the info :)

packages/backend-vscode-extension/webpack.config.js
tiagobento: VsCodeBackendExtension 😛

packages/vscode-extension-pack-kogito-kie-editors/package.json
tiagobento: Well, I don't think it's possible to internationalise this file, since it is the "manifest" of a VS Code extension and we can only have one such file. However, VS Code should provide a mechanism for us to apply i18n to that, given that it's a text string written in English. If we didn't need a config property, that would be great, but I'm afraid I can't point to a solution since I don't know VS Code's storing capabilities. Do you mind checking if there's such thing? Like a "per-extension database" or something like that. WDYT?

caponetto: > ExtensionContext.globalState: A global storage where you can write key/value pairs. VS Code manages the storage and will restore it for each extension activation.

:smiley:

caponetto: EDIT: Btw, about i18n: https://github.com/microsoft/vscode-extension-samples/tree/master/i18n-sample#considerations

caponetto:
Hi @tiagobento, thanks for the review! Follow the answers for the items you've raised:
1. Great suggestion, this is done.
2. It makes sense. `backendExtensionId` is optional now.
3. That's a good point. This part is simplified right now but we can create more layers while we accommodate the services.
4. :smile: :tada:
5. Nothing else on this PR will be moved to kogito-tooling-java repo.
6. The proxy redirects to the Backend manager, which resolves the required service. There is an example [here](https://github.com/kiegroup/kogito-tooling/commit/038601ca32c4803673114038307007f6fc2b3db8#diff-96133d009e7dae1bca9f98bd33d92349L93).

I've done all inline change requests, including the ones from @paulovmr.
The test scenario runner service is also included in this PR.
You can give it a try - the `vscode-extension-backend` artifact is available in the CI workflow.

packages/vscode-extension/src/runTestScenario.ts
```diff
@@ -0,0 +1,68 @@
...
+import * as vscode from "vscode";
+import { VsCodeI18n } from "./i18n";
+import { VsCodeBackendProxy } from "./VsCodeBackendProxy";
+
+export async function runTestScenario(
+  backendProxy: VsCodeBackendProxy,
+  workspaceApi: WorkspaceApi,
+  vsCodeI18n: I18n<VsCodeI18n>
+) {
+  const i18n = vsCodeI18n.getCurrent();
```
tiagobento:
```suggestion
  const i18n = vscodeI18n.getCurrent();
```

🙊

packages/vscode-extension/src/VsCodeBackendProxy.ts
```diff
@@ -0,0 +1,142 @@
...
+import { I18n } from "@kogito-tooling/i18n/dist/core/I18n";
+import * as vscode from "vscode";
+import { VsCodeI18n } from "./i18n";
+
+const SUGGEST_BACKEND_KEY = "SUGGEST_BACKEND";
+
+/**
+ * Augmented {@link BackendProxy} for the VS Code channel.
+ */
+export class VsCodeBackendProxy extends BackendProxy {
```
tiagobento: Can't we change that to composition instead of inheritance?

packages/vscode-extension/src/index.ts
```diff
@@ -69,4 +80,15 @@ export function startExtension(args: {
   args.context.subscriptions.push(
     vscode.commands.registerCommand(args.getPreviewCommandId, () => generateSvg(editorStore, workspaceApi, vsCodeI18n))
   );
+
+  args.context.subscriptions.push(
+    vscode.commands.registerCommand("extension.kogito.runTest", () => runTestScenario(backendProxy, workspaceApi, vsCodeI18n))
+  );
```
tiagobento: Please move that to `vscode-extension-pack-kogito-kie-editors`.

tiagobento (changes requested):
We're almost there! :)

Left some minor comments inline, but also, as a general observation:
1. Can we try to avoid inheritance and prefer composition? I think this will make our code more future-proof.
2. Can we rename `test-runner` to `test-scenario-runner`? I really want the names to connect so that people instantly know that a Test Scenario Runner runs tests from the Test Scenario Editor.

domhanak:
Hi @caponetto great work, went through the code, looks good to me. 
I would like to give it a try >  `You can give it a try - the vscode-extension-backend artifact is available in the CI workflow.`
Could you please share more details on how to do a this? Is that possible with this PR, I understand it is just infra and there shouldn't be any visible changes but I would like you to confirm before I approve. Thanks.


Food for thought - what do you think about using `graphql` to  communicate across the extension? Channels could define a schema and orchestrate the communication between editors and backend or similar. Could that be considered? While I was reviewing it seemed it could work, and it that could prove beneficial in a longer run (flexibility). Wanted to hear your opinion. :)

caponetto:
Hi @domhanak, thanks for the review!

You can try it out with our _development_ extension by doing the following:

Go to the last build associated with this PR ([this one](https://github.com/kiegroup/kogito-tooling/pull/256/checks?check_run_id=1061444445) atm).
You will find two `vsix` artifacts: `vscode-extension` and `vscode-extension-backend`.

With these artifacts, here are some things you can do:

(1) Only `vscode-extension` installed.
- Popup is shown when a diagram is opened.
Open a `bpmn` or `dmn` file. You should see a popup informing the user to install the backend extension.
Note: The "_Install_" button does not do anything because the backend extension hasn't been published yet.
Note: This popup should be shown only once per session. It should not be shown anymore if the user chooses "_Don't Show Again_".

- Popup is shown when a backend service is called when the backend extension is not installed.
Open a `scesim` file. You should see a "_Run_" button on top right.
If you click on it, a popup will be shown because the backend extension is not installed.
Note: This popup should be shown every time you click on the button.

(2) Both `vscode-extension` and `vscode-extension-backend` installed. Make sure to reopen VS Code after installation.
- Verify our Quarkus app.
Open a `bpmn` or `dmn`. Access http://localhost:8082/. This is our Quarkus app that will contain java services.
Note: The extension tries to find the first available port starting from 8082.
Note: Since we don't have any java service yet, there is no endpoint registered to be accessed.
Note: Our Quarkus app will be started up only if you have Java 11+.
Note: http://localhost:8082/ should only be killed when the VS Code is closed.

- Test runner service sample.
Open [dmn-quarkus-example](https://github.com/kiegroup/kogito-examples/tree/stable/dmn-quarkus-example) on VS Code and then `TrafficViolationTest.scesim`. Click on the "_Run_" button and wait for the test runner to complete. 
Note: It works only if you have Java 11+ and Maven 3.6.2+.
Note: It should take a while on the first run because Maven needs to download everything.

Notice that you won't see anything above on our next _release_ extension.
Please let me know if you have any questions. :)

------
I must say that I haven't thought about `graphql` at all but that's a very good point.
Since we don't have any java service yet, we assumed having a basic rest API for now.
I see this infrastructure evolving while we accommodate future services and I personally think `graphql` can be considered in the next steps. Thanks for bringing this up. :)

packages/vscode-extension/src/runTestScenario.ts
caponetto: Changed to `backendI18n`.

packages/vscode-extension/src/VsCodeBackendProxy.ts
caponetto: I understand the concern but inheritance is more suited for this case since the goal is to extend the functionalities of the `BackendProxy` for the VS Code channel.

packages/vscode-extension/src/index.ts
caponetto: Done in https://github.com/kiegroup/kogito-tooling/pull/256/commits/a9ba46950624bd7d48949d8a2025853e1ed73bb4.

caponetto:
@tiagobento Thanks again for the review. You will see on https://github.com/kiegroup/kogito-tooling/pull/256/commits/a9ba46950624bd7d48949d8a2025853e1ed73bb4:
- All _testRunner_ references have been renamed to _testScenarioRunner_.
- Some minor improvements for the test scenario runner service.
- The `backend` package has its own `i18n`.
- Created the `workspace` package with `api` and `vscode` submodules.
- Created the `vscode` submodule for the `backend` package.
- The test runner command is now registered on `vscode-extension-pack-kogito-kie-editors`.
- The responsibility to create the `BackendProxy` and stop the services were moved to `vscode-extension-pack-kogito-kie-editors` - now our `vscode-extension/index` is almost untouched.

tiagobento (approved):
Nice! Approving. Nice work @caponetto :)

domhanak:
Hi @caponetto I was able to successfully perform the scenarios from [your comment](https://github.com/kiegroup/kogito-tooling/pull/256#issuecomment-685723424)

There is only one issue, or maybe not, might be that I misunderstood the sentence: 
> Note: This popup should be shown only once per session. It should not be shown anymore if the user chooses "Don't Show Again".

Clicking "Don't Show Again" works as expected, but when I didn't click "Don't Show Again" but I clicked the `X` and next time I opened a file, the popup didn't show up. Very minor issue, so I am approving anyway.

domhanak (approved):

caponetto:
Hi @domhanak, thanks again for the review!
That's the expected behavior. I didn't want to annoy the user with the popup every time they open a diagram.
Instead, it is shown only the first time they use our extension as a reminder.
So the popup will be shown again only if VS Code is reopen.

paulovmr (approved):

PR #261, opened 2020-08-27
KOGITO-3240: Fix convenience classes for use in `kogito-tooling-examples`

tiagobento:
**EmbeddedEnvelope**: This change is necessary due to the generic nature of this component. By creating a HoC component for it, we're able to keep the generics constraints without loosing functionality.

**EmbeddedEditor**: Added new effect to notify the Enveloped Editor that the content changed when the file prop changes.


paulovmr (approved):

tiagobento:
@paulovmr Thx for the review, but this PR will change still. I should've put that in Draft mode. Sorry about that.

ljmotta:
**LoadingScreen**: Simplified the component and removed inline CSS to solve a render bug.
**EditorEmbeddedView**: Refactor into a RefForwardingComponent.

packages/editor/src/envelope/KogitoEditorEnvelope.tsx
```diff
@@ -71,15 +71,20 @@ export class KogitoEditorEnvelope {
...
       </KogitoEditorEnvelopeContext.Provider>
     );
 
-    return new Promise<EditorEnvelopeView>(res => {
+    return new Promise<EditorEnvelopeViewApi>((res, rej) => {
       setTimeout(() => {
-        ReactDOM.render(app, container, () => res(view!));
+        ReactDOM.render(app, container, () => {
+          if (editorEnvelopeViewRef.current) {
+            res(editorEnvelopeViewRef.current);
```
caponetto, line 84: Don't we need a `return` after this line? Otherwise line 86 will run even though the `resolve` is called.

packages/editor/src/envelope/LoadingScreen/LoadingScreen.tsx
```diff
@@ -15,48 +15,54 @@
...
+    e.stopPropagation();
+    setMustRender(false);
+  }, []);
+
+  const loadingScreenClassName = useMemo(() => {
+    if (props.loading) {
+      return "";
+    } else {
+      return "loading-finished";
+    }
```
caponetto:
```suggestion
   return props.loading ? "" : "loading-finished";
```
A suggestion to make it shorter.

packages/editor/src/envelope/KogitoEditorEnvelope.tsx
ljmotta: Yes we need! :+1:  . The 86 is not going to run because it is the reject, but if we had any other code, it would be a problem.

packages/editor/src/envelope/LoadingScreen/LoadingScreen.tsx
ljmotta: It's shorter, but I think it can be trickier to read. I left like this to be easier, but I should've removed the else clause.  :grimacing:

caponetto (approved):

paulovmr (approved):

ljmotta (approved):

ederign (approved):

PR #272, opened 2020-09-11
FAI-253: Score Cards: Research feasibility of reducers for undo/redo

manstis:
See https://issues.redhat.com/browse/FAI-253

This is my work on support for undo/redo support for complex models (nesting depth > 1).

I invite you to `yarn run build:showcase` on `pmml-editor` and `yarn run start`.


packages/pmml-editor/src/editor/reducers/CompositeReducer.ts
```diff
@@ -0,0 +1,54 @@
...
+  return [
+    (state: PMML, action: AllActions) => {
+      const pmmlClone: PMML = cloneDeep(state);
+      for (const reducer of reducers) {
+        const newStateForReducer: PMML = reducer(pmmlClone, action);
+        if (!isEqual(pmmlClone, newStateForReducer)) {
+          return newStateForReducer;
+        }
+      }
+      return state;
```
kelvah: hi @manstis, is it necessary to clone and deep compare states? It's an expensive operation and I assume when changes are applied, `newStateForReducer` will always be a new tree generated using `produce`, so an identity comparison should be enough. Do you agree?

manstis: Hi @kelvah I've remove the use of `deepClone` as `produce` does indeed return a cloned updated sub-tree.

However I'm less cetain `isEquals(..)` can be so easily removed. Each `Reducer` returns the full PMML and I will not know what, if anything, has changed without a deep comparison. I'm not certain I can simply merge the existing state and new state and return the new state as it may lead to some entries not being deleted!?! I'd need to write some tests to check.

Give me a bit more time and I'll commend on here again..

packages/pmml-editor/src/editor/history/HistoryProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+  return state;
+};
+
+const apply = (state: PMML, path: string | null, patch: any) => {
+  if (path === null) {
+    return applyPatches(state, patch);
+  }
+  const branch: any = get(state, path);
+  const branchUndone: any = applyPatches(branch, patch);
+  const newState: PMML = cloneDeep(state);
```
manstis: @kelvah I cannot avoid a deep clone with the changes for reducers to accept the sub-tree. 

The old way had `immer` patches stored that were relative to the root `PMML` object. When undoing state changes and applying the "reverse" `immer` returned a new instance of the `PMML` object. Now the deltas are relative to whatever object was passed to the reducer and undoing them mutates the sub-tree not the top-level `PMML`. Consequentially React is not detecting a state change w/o the deep clone. I also tried both `Object.assign(..)` and spread operators but they only perform shallow clones so some state on other sub-trees was lost.

packages/pmml-editor/src/editor/history/HistoryProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { applyPatches, produce } from "immer";
+import { WritableDraft } from "immer/dist/types/types-external";
+import { PMML } from "@kogito-tooling/pmml-editor-marshaller/dist/marshaller/model";
+import { cloneDeep, get, set } from "lodash";
+
+interface Change {
+  path: string | null;
```
manstis, line 22: I now have to store the relative path to where the change was made. Originally all changes were performed on the `PMML` root object so the deltas were relative to that too. Now the changes made by the reducers are relative to whatever object was passed to them. Without storing the path it is impossible to know to which sub-tree the undo/redo needs to be applied.

packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
```diff
@@ -0,0 +1,71 @@
...
+  };
+  [Actions.DeleteDataField]: {
+    readonly dataFieldIndex: number;
+  };
+}
+
+export type DataDictionaryActions = ActionMap<DataDictionaryPayload>[keyof ActionMap<DataDictionaryPayload>];
+
+export const DataDictionaryReducer: PathAwareReducer<DataDictionary, DataDictionaryActions> = (
+  state: DataDictionary,
```
manstis: At least I can now pass the actual object to the reducer...

packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    case Actions.DeleteDataField:
+      newState = mutate(state, path, draft => {
+        draft.DataField.splice(action.payload.dataFieldIndex, 1);
+      });
+      break;
+  }
+
+  let dataFieldIndex: number | undefined;
+  if (hasDataFieldIndex(action.payload)) {
```
manstis: The parent reducer handles indexing of children elements...

In this instance if the action has a `dataFieldIndex` property only the applicable `DataField` is reduced. If the property does not exist on the action then all `DataField` objects are passed to the reducer. This is a choice we can agree on.. currently the reducer is free to decide what to do with actions; e.g. `SetDataFieldName` only applies to a single `DataField` but `CapitaliseAllDataFieldNames` would apply to all `DataField`s (assuming the payload lacks a `dataFieldIndex` property). Does this make sense? Or we decide if the index is missing we don't handle any of the `DataField` children.

packages/pmml-editor/src/editor/reducers/DataFieldReducer.ts
```diff
@@ -0,0 +1,44 @@
...
+export type DataFieldActions = ActionMap<DataFieldPayload>[keyof ActionMap<DataFieldPayload>];
+
+export const DataFieldReducer: PathAwareReducer<DataField, DataFieldActions> = (
+  state: DataField,
+  action: DataFieldActions,
+  path: string
+) => {
+  let newState = state;
+  switch (action.type) {
+    case Actions.SetDataFieldName:
```
manstis: There is no need to check we're handling the correct instance as it was filtered by the `DataDictionaryReducer`. I did have an alternative that handled the filtering here (i.e. I checked in `DataFieldReducer` that the index encoded in the `path` parameter matched the index in the action payload); but I thought it was not in the spirit of our discussion.

manstis:
@kelvah @tiagobento I've updated the PR to have only the applicable sub-tree passed to reducers.

We can still have actions that span the whole model by scoping the command to the applicable model context; i.e. a `CapitaliseNames` action could be configured as a `PMMLAction` and handled by the `PMMLReducer` that has access to the whole model. Or a `CapitaliseAllDataFieldNames` action would be a `DataDictionaryAction` and handled by the `DataDictionaryReducer` etc.

WDYT now?

manstis:
Hi @kelvah @tiagobento, following our discussions I've refactored this PR to use:-

- Pass the relevant sub-tree to reducers
- Use an existing library to handle reducer composition
- Use redux's store and hooks to support correct re-rendering of sub-tree UI components

The end result is that:-

- The reducers are much clearer
- state selection and action dispatch simplified
- The UI correctly re-renders following sub-tree changes 

All in all; I'm quite happy.

I'll write tests for the reducers (leaving the UI as it is entirely mocked to trial state management).

packages/pmml-editor/src/editor/reducers/CompositeReducer.ts
manstis: This component does not exist any more. Resolving conversation.

packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
manstis: Out of date. Resolving.

packages/pmml-editor/src/editor/reducers/DataFieldReducer.ts
manstis: Out of date. Resolving.

packages/pmml-editor/src/editor/PMMLEditor.tsx
```diff
@@ -16,6 +16,37 @@
...
+import mergeReducers from "combine-reducer";
+import { HeaderReducer } from "./reducers/HeaderReducer";
+import { DataDictionaryReducer } from "./reducers/DataDictionaryReducer";
+import { DataFieldReducer } from "./reducers/DataFieldReducer";
+
+enableAllPlugins();
+
+const reducer: Reducer<PMML, AllActions> = mergeReducers(PMMLReducer, {
+  Header: HeaderReducer,
+  DataDictionary: mergeReducers(DataDictionaryReducer, { DataField: DataFieldReducer })
```
kelvah: I know you split these two reducers from the beginning, but is it necessary? I was wondering if `DataDictionaryReducer` is enough to handle all the actions and if `DataFieldReducer` is adding complexity.

Maybe you have a more clear picture about how many actions will be there and so you decided to separate them. Or were you just testing how to handle nested reducers?

packages/pmml-editor/src/editor/history/HistoryProvider.ts
kelvah: yes, I think with nested reducers there is no other way to do the `apply` than deep cloning the state. Let's keep an eye on this for potential performance issues later on.

packages/pmml-editor/src/editor/history/HistoryProvider.ts
kelvah: I saw that when `immer` generates patches, they have a path inside them referring to the root level property involved.
Is it correct? I guess if we only had top level reducers this was just handled by `immer`.

Asking this because we could also manipulate the PMML before storing it, so that it could be split at top level only, and maybe let the store handle also other things not present inside the PMML itself. I recognize It's a quite different take. What do you think?

packages/pmml-editor/package.json
```diff
@@ -19,7 +19,14 @@
     "@types/react-router": "5.1.1",
     "@types/react-router-dom": "5.1.1",
     "react-router": "5.1.1",
-    "react-router-dom": "5.1.1"
+    "react-router-dom": "5.1.1",
```
kelvah: There's no need to add `react-router` as a dependency. `react-router-dom` is enough because it's the one we actually use, and it relies on `react-router` internally.

It's made clear in a note on the `react-router` [npm package](https://www.npmjs.com/package/react-router).

kelvah (approved):
Thank you @manstis for your accurate work on this. I left some observations/questions.

packages/pmml-editor/package.json
manstis: Done.. thank-you.

packages/pmml-editor/src/editor/PMMLEditor.tsx
manstis: I'll keep the `DataFieldReducer` separate; it contains many more fields than just `name` and so will become more complex (the reducers present in this PR are "just enough" to research undo/redo/state management but are all incomplete for the full editor):
```
export class DataField extends Field<DataField> {
    Extension?: Extension[];
    Interval?: Interval[];
    Value?: Value[];
    name: FieldName;
    displayName?: string;
    optype: OpType;
    dataType: DataType;
    taxonomy?: string;
    isCyclic?: Cyclic;
```

packages/pmml-editor/src/editor/history/HistoryProvider.ts
manstis: `immer` does indeed generate paths but they are based on the sub-tree passed to the `produce` method and not the root of the whole data model. For example here's a screenshot of the change recorded for an updated `DataField` name.

Screenshot from 2020-09-22 12-53-09

`immer` has the path as `[0, "name"]` that is relative to the object passed to the reducer; being the `DataField[]` array. However when this change is "undone" it needs to be relative to the root to be correctly applied; i.e. `DataDictionary.DataField` that is the path I store.

To answer your questions:

> I guess if we only had top level reducers this was just handled by immer

Correct.

> ...we could also manipulate the PMML before storing it...

I've had my fair share of writing editors that follow the same principle: (1) get the XML/text, (2) unmarshall it to an object model, (3) convert it to another object model and repeat in the opposite direction when saving. Usually step (3) has been undertaken when serialising the model across the wire (server to browser) however since all editors need to be 100% client/browser based I felt it an unnecessary burden and I was going to try to use the object model produced from the first marshalling operation in the UI itself. Furthermore the PMML object model has many levels of nesting and if we attempt to flatten them (to simplify reducer composition) we're going to need to add and maintain lots of primary/foreign key constraints between objects.

Consider the simple `DataField` hierarchy: `DataDictionary -> DataField[] -> Interval[]`.

We'd need to _key_ `Interval` instances to their applicable `DataField` instance.

I am however not [pig headed](https://dictionary.cambridge.org/dictionary/english/pigheaded) and remain open for discussion.

manstis:
Hi @jiripetrlik I've added you as the formal QE review. This PR adds infrastructure support for undo/redo/state management. The widgets presented are completely throw away and only exist to get some visualisation of state changes and undo/redo support. As the editor comes into being these widgets will be deleted. Hence no test coverage for them. If you feel you don't want to be involved at these early stages just say so - although I will continue to add you to PMML Editor PRs FYI and you can become more involved as you see appropriate.

packages/pmml-editor/src/editor/history/HistoryProvider.ts
kelvah: > Furthermore the PMML object model has many levels of nesting and if we attempt to flatten them (to simplify reducer composition) we're going to need to add and maintain lots of primary/foreign key constraints between objects.

That's a fair point. Things could get difficult to manage for our case.
Thank you for taking the time to elaborate your view.

packages/pmml-editor/src/editor/MockSummaryUI.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { TypedUseSelectorHook, useSelector } from "react-redux";
+import { PMML } from "@kogito-tooling/pmml-editor-marshaller";
+
+const style = {
+  padding: "5px 5px 5px 5px"
+};
+
+const MockSummaryUI = () => {
+  const typedUseSelector: TypedUseSelectorHook<PMML> = useSelector;
+  const pmml: PMML = typedUseSelector(state => state);
```
tiagobento: IIUC, we're adding a type out of the blue here to `useSelector`, is that correct? If so, instead of having these two lines, could we combine that into a `useTypedSelector<T>` or something like that?

packages/pmml-editor/src/editor/MockSummaryUI.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+import * as React from "react";
+import { Timestamp, Title } from "./PMMLEditor";
+import { TypedUseSelectorHook, useSelector } from "react-redux";
+import { PMML } from "@kogito-tooling/pmml-editor-marshaller";
+
+const style = {
+  padding: "5px 5px 5px 5px"
+};
```
tiagobento: This is not typed properly, since if you add a property like `foo: bar` there, the variable declaration won't break. You'll only see it breaking on like 30. To improve that, I'd type this object with `CSS.Properties`

packages/pmml-editor/src/editor/PMMLEditor.tsx
```diff
@@ -16,6 +16,37 @@
...
+import { PMMLReducer } from "./reducers/PMMLReducer";
+import { AllActions } from "./reducers/Actions";
+import { PMML } from "@kogito-tooling/pmml-editor-marshaller";
+import { Provider } from "react-redux";
+import mergeReducers from "combine-reducer";
+import { HeaderReducer } from "./reducers/HeaderReducer";
+import { DataDictionaryReducer } from "./reducers/DataDictionaryReducer";
+import { DataFieldReducer } from "./reducers/DataFieldReducer";
+
+enableAllPlugins();
```
tiagobento: I'm not a big fan of code floating around like that. Since this will be executed when this file's module is imported, I think it's kinda unpredictable. It's probably inoffensive at this point, but having that inside a `useEffect` call would make it better IMHO.

packages/pmml-editor/src/editor/history/HistoryProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+
+interface History {
+  changes: Change[];
+  index: number;
+}
+
+const history: History = {
+  changes: [],
+  index: 0
+};
```
tiagobento: Global variable.. not ideal. But ok to keep it until we need to refactor it. My fear is that it may be harder and harder to remove this after a while. I'll leave the decision to you, just wanted to point that out.

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { ActionMap, Actions, AllActions } from "./Actions";
+import { mutate, redo, undo } from "../history/HistoryProvider";
+import { PMML } from "@kogito-tooling/pmml-editor-marshaller";
+import { Reducer } from "react";
+
+export const ROOT = null;
```
tiagobento: What is this used for?

tiagobento (commented):
@manstis Left some comments inline. Please consider them.

As a general note, I really like the new concepts you introduced and I'm very glad you're thinking about the best approach to have good code while enabling features like undo/redo from the beginning. I think some things on this PR will end up changing as you dive deeper on all the PMML features, but I think we're off to a great start. Thanks!

packages/pmml-editor/src/editor/MockSummaryUI.tsx
manstis: Corrected; I was (blindly) following advice on the interweb; but following this comment found a [better approach](https://stackoverflow.com/questions/57472105/react-redux-useselector-typescript-type-for-state).

packages/pmml-editor/src/editor/MockSummaryUI.tsx
manstis: Great catch. Updated.

packages/pmml-editor/src/editor/PMMLEditor.tsx
manstis: Sure. I've moved to the `PMMLEditor` constructor. I can't move to a hook as the `PMMLEditor` itself needs to be a class (not Functional Component) due to, AFAIK, `kogito-tooling` limitations (or at least the inability to have FC's implement interfaces?). Is there any other alternative (to moving to the constructor)?

packages/pmml-editor/src/editor/history/HistoryProvider.ts
manstis: I've refactored to have a single instance provided by a `Context` and propagated to the reducers via the payload. It's included in a second commit. I don't love it... I don't like having to pass the History Service to the reducers via payload. Any better suggestions?

manstis:
@kelvah I welcome your comments on my last commit moving `History` from a global to a `Context` and how it is passed to reducers. I am begining to think it may not be too terrible as it fortuitously supports different histories (local and global contexts for modal dialogs if you remember) and even different implementations (other than `immer`). However as always I welcome other opinions.

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
manstis: Ah, yes, it's only used internally (now... there have been ongoing revisions).

It was referenced from some tests however I don't think it is good to export it just for use in tests.

I'll remove it.

packages/pmml-editor/src/editor/PMMLEditor.tsx
tiagobento: Of course, I understand. That's right, to implement the Editor interface right now you need a class :)

packages/pmml-editor/src/editor/MockHistoryUI.tsx
```diff
@@ -51,14 +56,23 @@ const StateButtons = () => {
 
 const HistoryLog = () => {
   const [state, setState] = useState(true);
+  const service: HistoryService = useContext(HistoryContext).service;
```
tiagobento:
```suggestion
  const { service } = useContext(HistoryContext);
```

I think that's more idiomatic :)

packages/pmml-editor/src/editor/history/HistoryProvider.ts
tiagobento: Well, I actually don't mind having that extra parameter on the payload, since it makes your dependencies more explicit (i.e. an Action requires a HistoryService and some other parameters to be performed). If anything, it makes your code "more functional" in the sense of functional programming. As you said in your comment below, you now have the flexibility to use a different HistoryService implementation on an Action, which you didn't have before, so it's making your code more flexible too.

I understand that it looks like a boilerplate, but that's because your Actions and Reducers are simple now and it looks like this HistoryService is not supposed to be there. In the future, I'm sure that other Services will be required to perform other Actions.

If that's not convincing still, I believe you could create an instance of HistoryService before creating your Reducers and passing it as a parameter. That's somewhat better than having a "global variable", since your dependency tree will be explicit now and you won't have that much boilerplate. You'd then *not* export this HistoryService instance, meaning that it would only be available as a parameter to a "Factory" function of your Reducers.

Hope that's insightful! 😅

packages/pmml-editor/src/editor/MockHistoryUI.tsx
manstis: Changed throughout (I still have yet to learn some of this syntax sugar).

packages/pmml-editor/src/editor/PMMLEditor.tsx
kelvah: I'm late but I have a different take on this. `enableAllPlugins()` feels more related to the `import` phase. It seemed therefore reasonable to place it _floating around_ but together with other imports inside the application entry point. Anyway, just personal preferences at this point.

packages/pmml-editor/src/editor/history/HistoryProvider.ts
kelvah: I share @manstis perplexity on the HistoryService passed along inside the payload. There's definitely room for improvement I think, maybe wrapping reducers with a history reducer taking care of that part? Anyway, I guess it could be addressed later to avoid blocking this PR.

manstis: Thank-you @kelvah for the objection and proposal. 

I've refactored again. The results of which I am happier with.

tiagobento:
@manstis Michael, I just ran the PMML showcase you provided and something crossed my mind. I figured I'd share that here.
When editing text inputs, although we have this capability, I don't think we need to store each key stroke as a "step" on the History. I think it makes for a much better UX if we "aggregate" some actions into one. I say that now because the Editor will probably evolve to be more complex and refactoring it might be harder in the future. I believe more scenarios like that will show up (e.g. Moving elements around with the arrow keys). Have you thought about that? What do you think if this subject?

manstis:
Hi @tiagobento bear in mind the UI presented here is mock, disposable and only exists to facilitate testing state management.

That said what you raise has been considered in part already and there are a myriad of different approaches that can be considered. IDK if there is a "one size fits all" solution. For example, we can handle action flooding at the widget level:

`TextBox` and `TextAreas` elements could only dispatch actions `onBlur`, `Checkbox` and `Listbox` `onChange`. Drag and drop operations `onMouseUp` etc. Or we could accumulate changes for certain DOM event types (e.g. `onKeyDown` in a `Textbox`) and only dispatch actions periodically (e.g. every 500ms and flush any caching `onBlur`). 

Or, we can handle action flooding at the `HistoryService` level:

We could record every action and have a periodic "reaper" timer task that aggregates actions in the history stack (e.g. it would combine 100 `SetText` actions into a single action and flush the other 99 from the stack) or we could pipe all actions to an OpenShift Persistence Storage cluster and wave the OCP flag ;-)

My preference is to limit the number of actions raised as opposed to performing house keeping.

Whilst this is no doubt an important topic it is outside the scope of this PR.

tiagobento (approved):
@manstis Thanks! Just wanted to point that out. Approving :)

manstis:
Hey @tiagobento @ederign do you have anyone in QE you normally use?

I'm getting no replies from @jiripetrlik so assume he's on PTO... it'd be nice to get this merged.

I'm wondering if there's an alternative QE representative that could give this the once over?

tiagobento:
@manstis Hi! @tomasdavidorg is our QE representative. I'm adding him as reviewer on this PR :)

jiripetrlik:
Hello @manstis I'm back from my PTO. :-) Will look at it today. However, my knowledge of frontend technologies like React and TypeScript is still rather low. So, I will appreciate also any feedback from @tomasdavidorg if you have a time.

manstis:
Hi @jiripetrlik welcome back. 

@jiripetrlik @tomasdavidorg The guts of this PR is the undo/redo mechanism. 

ALL UI components are throw-away as they're there to simply "help" test/check undo/redo was working in the UI.

To save you scrolling throw the review:-

- Fetch this PR locally
- `yarn install`
- `yarn run build:prod`
- `cd ./packages/pmml-editor`
- `yarn run build:showcase`
- `yarn run start`

Unit tests are provided for the undo/redo stuff; but not the UI (since it is to be discarded).

Any questions please ask :-)

jiripetrlik (approved):
Looks good to me. I was able to run it according to @manstis info and undo/redo buttons seems to be working.

tomasdavidorg (approved):

tomasdavidorg:
Just quickly checked the showcase and the code a bit. Looks good to me.

PR #283, opened 2020-10-05
FAI-183: Score Cards: Opening a PMML file story

manstis:
See https://issues.redhat.com/browse/FAI-183

This represents the Landing Page when opening (or for now) creating a new PMML file.

When first ran an empty _wrapper_ is shown. It is not part of the editor. It is akin to VSCode without a file opened/created.

editor-empty

When you open a file the list of models will be shown.

editor-opening-file

You can filter the model list by name or supported/unsupported (only Scorecards are supported).

The undo/redo/PMML buttons are not part of the editor. The buttons are provided by the container ("Undo"/"Redo" should be self explanatory, "PMML" gets the XML content of the editor and is used, for example, by VSCode's "Save"). You can delete models and undo/redo the deletion however be aware of https://issues.redhat.com/browse/FAI-270


manstis:
@tiagobento IDK if you have any interest in this for now (it's really specific PMML stuff).

Feel free to add yourself as a reviewer if you want to be involved.

manstis:
@kelvah I've implemented _atomic design_, in principle, although I've cut it the other way (top level folders are feature, with sub-folders the atomic elements). See what you think.. I can realign with `kogito-apps/ui-packages` if you'd prefer to maintain some form of consistency. @tiagobento the leads over in `kogtio-apps` are imposing atomic design.. Roberto has had to follow their mandate over in `kogito-apps` to get a PR approved and merged. We're not lovers of it. Opinion welcome. We've raised the concern.

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <PageSection variant={PageSectionVariants.light}>
+          <Split>
+            <SplitItem>
+              <TextContent>
+                <Title size="3xl" headingLevel="h2">
+                  {props.path}
+                </Title>
+              </TextContent>
+            </SplitItem>
+            <SplitItem isFilled={true}>&nbsp;</SplitItem>
```
kelvah: Any reason to add this `SplitItem`? You could set `isFilled` to the the first one.

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const models: Model[] | undefined = useSelector<PMML, Model[] | undefined>((state: PMML) => state.models);
+
+  const [filter, setFilter] = useState("");
+  const [showUnsupportedModels, setShowUnsupportedModels] = useState(true);
+
+  const filterModels = useCallback((): Model[] => {
+    const _lowerCaseFilter: string = filter.toLowerCase();
+    const _filteredModels: Model[] | undefined = models?.filter((_model: Model) => {
+      const _modelName: string | undefined = getModelName(_model);
+      const _isSupportedModelType: boolean = isSupportedModelType(_model);
```
kelvah: I think all these type definitions could be inferred by typescript without having to write them down. It could be personal taste but in this case I find them too verbose while not adding anything. What's your opinion?

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
```diff
@@ -0,0 +1,121 @@
...
+      const _supportMatch: boolean = showUnsupportedModels || _isSupportedModelType;
+      return _nameMatch && _supportMatch;
+    });
+    return _filteredModels ? _filteredModels : [];
+  }, [filter, showUnsupportedModels, models]);
+
+  const filteredModels: Model[] = useMemo(() => filterModels(), [filter, showUnsupportedModels, models]);
+
+  return (
+    <>
```
kelvah: I'm sorry for being extremely pedant, but `<>` is necessary only when returning a list of elements. We should definitely add an eslint rule for unnecessary fragments when we'll get rid of tslint.

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
```diff
@@ -0,0 +1,121 @@
...
+            setFilter={setFilter}
+            showUnsupportedModels={showUnsupportedModels}
+            setShowUnsupportedModels={setShowUnsupportedModels}
+          />
+        </PageSection>
+
+        <PageSection isFilled={true}>
+          <section>
+            {filteredModels.length > 0 && (
+              <Gallery hasGutter={true}>{filteredModels.map(model => renderCard(model))}</Gallery>
```
kelvah: I think the `GalleryItem` code is simple enough to be just put inline. I don't see any particular reason to move it to a function.

packages/pmml-editor/src/editor/index.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import "@patternfly/patternfly/patternfly.css";
+import "@patternfly/react-core/dist/styles/base.css";
```
kelvah: I think importing just the `react-core` styles is enough.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ActionSelector.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <Dropdown
+      data-testid="action-selector"
+      onSelect={onSelect}
+      toggle={
+        <DropdownToggle
+          id="action-selector-id"
+          data-testid="action-selector__toggle"
+          onToggle={onToggle}
+          toggleIndicator={CaretDownIcon}
```
kelvah: This is the default value for `toggleIndicator`, so no need to set it and import the icon.

packages/pmml-editor/src/editor/components/LandingPage/organisms/LandingPageToolbar.tsx
```diff
@@ -0,0 +1,90 @@
...
+        </ToolbarItem>
+        <ToolbarItem>
+          <Switch
+            id="only-supported-models-switch"
+            data-testid="landing-page-toolbar__supported-models"
+            label="Show unsupported models"
+            labelOff="Hide unsupported models"
+            isChecked={showUnsupportedModels}
+            onChange={setShowUnsupportedModels}
+          />
```
kelvah: Some comments not strictly related to the code.

What if the switch is displayed only when there are unsupported model types? Or maybe it's just disabled when there are none?

I also have some concerns with switches with different labels according to their state. I know it's a UX matter and PF examples include this behavior, but it always confuses me. In my opinion there should be a fixed label (like "Show unsupported  models" and, optionally, a state description (like "On"/"Off"). I could find some literature supporting this but probably it needs to be discussed  also with Liz.

packages/pmml-editor/src/editor/components/LandingPage/organisms/EmptyStateNoModels.tsx
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as React from "react";
+import { Button, EmptyState, EmptyStateBody, EmptyStateIcon, EmptyStateVariant, Title } from "@patternfly/react-core";
+import BoxesIcon from "@patternfly/react-icons/dist/js/icons/boxes-icon";
```
kelvah: Maybe it was your IDE auto importing it like that, but this could be
```suggestion
import { BoxesIcon } from "@patternfly/react-icons";
```
I've noticed it in other components too.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const ModelCard = (props: ModelCardProps) => {
+  const { model } = props;
+  const modelType: string = coalesce(getModelType(model), "<Unknown>");
+
+  return (
+    <Card data-testid="model-card" isHoverable={true} className="model-card">
+      <CardHeader>
+        <CardHeaderMain>
+          <ModelCardIcon model={model} />
```
kelvah: I would try to pass to components only the necessary information they need to work. I see what you did here encapsulating the logic to retrieve the icon inside the `ModelCardIcon`, but passing down the whole model seems overkill. Could the `modelType` that you already retrieved be enough?

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+        </CardHeaderMain>
+      </CardHeader>
+      <ModelCardTitle model={model} />
```
kelvah: The card title hardly deserves a separate component. It's good to separate things when they could be reused or to break down complexity, but I don't see the value here.

packages/pmml-editor/src/editor/utils/PropertyUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const coalesce = (value: string | undefined | null, defaultValue: string) => {
+  if (value === null || value === undefined) {
+    return defaultValue;
+  }
+  return value;
+};
```
kelvah: This could be done with ts nullish coaleshing, the `??` thing. Right?

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.scss
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+.model-card {
+  height: 280px;
+
+  &__title {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
```
kelvah, line 24: This is repeated inside `ModelCardTitle.scss`

kelvah (commented):
Hi @manstis, I left some comments. 

I see your are often declaring variables with a leading `_`. Is that just to silence the name shadowing warning? Are there any other reasons?

I didn't add considerations on file structure because we are discussing it elsewhere, by now you already know my opinion :)

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
manstis: Changed as suggested.

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
manstis: You can't teach an old Java programmer new tricks ;-)

I've removed those you call out here and will endeavour to be a better Type Script citizen!

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
manstis: Removed.

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
manstis: OK.. Fair enough. Changed.

packages/pmml-editor/src/editor/index.ts
manstis: Seems OK. Removed.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ActionSelector.tsx
manstis: Done.

packages/pmml-editor/src/editor/components/LandingPage/organisms/EmptyStateNoModels.tsx
manstis: Fixed all occurrences.

packages/pmml-editor/src/editor/utils/PropertyUtils.ts
manstis: Yes, you are right. I've removed the function and all use. Thank-you.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.scss
manstis: I've removed the `ModelCardTitle` component and added it back to `ModelCard` as suggested above.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.tsx
manstis: Removed. I went to the sub-atomic particle level with my use of atomic design ;-)

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.tsx
manstis: I've changed this occurrence (and related code to support `getModelIconByType(type)`).

However I would usually pass a higher level object, when available, in case other properties are needed.

Java Script passes objects by value (but the value is the reference to the object, there is no deep cloning of objects to pass by value, just the pointer is duplicated). Therefore there is no overhead passing a `Model` vs `string` (in fact there is possibly less overhead passing an object reference by value vs a primitive that is cloned).

packages/pmml-editor/src/editor/components/LandingPage/organisms/LandingPageToolbar.tsx
manstis: Removed different labels (yay) and only render switch when there are unsupported models.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.tsx
kelvah: Thank you Michael. I think the issue is more about having a more clear component interface. Passing down individual properties instead of a whole object helps quickly understanding what are the inputs of a component and what's its purpose. It simplifies collaboration when looking at a components tree and eventually it improves debugging.

manstis: Fair point. Did you want more changes on this PR or can it be approved (by you)?

manstis:
Hi @kelvah,to answer your first question some variable names (usually those in closures) start with an underscore for the sole purpose of keeping the lint checks happy. Thank-you.

kelvah (approved):

manstis:
@tiagobento @jiripetrlik This is ready for your reviews. Thank-you.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ActionSelector.tsx
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  const dropdownItems = [
+    <DropdownItem key="createModel" isDisabled={true}>
+      Create model
+    </DropdownItem>,
+    <DropdownItem key="viewDataDictionary" isDisabled={true}>
+      View Data Dictionary
+    </DropdownItem>
+  ];
```
tiagobento: I know there's some debate about whether this is necessary or not, but I'm trying to give people the advice of memoizing in-component functions and values using `useCallback` and `useMemo` respectively. This will prevent us from having unnecessary re-renders. In the case of `dropdownItems`, since it's all static, it can even leave outside the component as a constant.

packages/pmml-editor/src/editor/components/LandingPage/organisms/LandingPageToolbar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const onFilterSubmit = (): void => {
+    if (filterField && filterField.current) {
+      setFilter(filterField.current.value);
+    }
+  };
+  const onFilterEnter = (event: React.KeyboardEvent): void => {
+    if (filterField && filterField.current && event.key === "Enter") {
+      setFilter(filterField.current.value);
+    }
+  };
```
tiagobento: I'm not an expert, but I think if you wrap this input inside a `form`, you're able to have the `Enter` key working out-of-the-box, removing the need for a `ref` and making you write less code. Beware that preventing the default behaviour of the `onSubmit` event is necessary.

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const filterModels = useCallback((): Model[] => {
+    const _lowerCaseFilter = filter.toLowerCase();
+    const _filteredModels = models?.filter((_model: Model) => {
+      const _modelName = getModelName(_model);
+      const _isSupportedModelType = isSupportedModelType(_model);
+      const _nameMatch = _modelName === undefined || _modelName.toLowerCase().includes(_lowerCaseFilter);
+      const _supportMatch = showUnsupportedModels || _isSupportedModelType;
+      return _nameMatch && _supportMatch;
+    });
+    return _filteredModels ? _filteredModels : [];
```
tiagobento: You can go for a `??` operator here.

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
```diff
@@ -0,0 +1,119 @@
...
+                <GalleryItem key={uuid()} data-testid="landing-page__model-card">
+                  <ModelCard
+                    model={model}
+                    onDelete={(_model: Model) => {
+                      dispatch({
+                        type: Actions.DeleteModel,
+                        payload: {
+                          model: _model
+                        }
+                      });
```
tiagobento: Un-memoized function. Might result in unnecessary re-renders.

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
```diff
@@ -44,6 +47,18 @@ export const PMMLReducer: HistoryAwareReducer<PMML, AllActions> = (
           draft.version = action.payload.version;
         });
 
+      case Actions.DeleteModel:
+        return service.mutate(state, null, draft => {
+          const model: Model = action.payload.model;
```
tiagobento: Unnecessary explicit type.

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
```diff
@@ -44,6 +47,18 @@ export const PMMLReducer: HistoryAwareReducer<PMML, AllActions> = (
...
+          if (draft.models !== undefined) {
+            const models: Model[] = draft.models;
```
tiagobento: Same here.

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
```diff
@@ -44,6 +47,18 @@ export const PMMLReducer: HistoryAwareReducer<PMML, AllActions> = (
           draft.version = action.payload.version;
         });
 
+      case Actions.DeleteModel:
+        return service.mutate(state, null, draft => {
+          const model: Model = action.payload.model;
+          if (draft.models !== undefined) {
```
tiagobento:
```suggestion
          if (draft.models) {
```

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
```diff
@@ -44,6 +47,18 @@ export const PMMLReducer: HistoryAwareReducer<PMML, AllActions> = (
...
+            const models: Model[] = draft.models;
+            const index: number = models.indexOf(model);
```
tiagobento: Same here 🙊

tiagobento (commented):
@manstis Thanks for the PR! Left some minor comments inline.

As general comments:
- I'm not a big fan of "utils" or "common" packages, since they tend to aggregate un-related features that are used across the application. This might lead to bad unnecessary coupling and, well.. you know the consequences. If you can, I think it would be better to reorganize things to avoid it.
- I saw that there's some inline texts on the components. I know this PR is not definitive and things will change in the future, but trying to start using the I18n module might be a good idea. Me or @ljmotta can help with that if you need.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ActionSelector.tsx
manstis: Done :-)

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
manstis: Not the first time I've forgotten about the nullish operator... and probably won't be the last.

Fixed.

packages/pmml-editor/src/editor/components/LandingPage/templates/LandingPage.tsx
manstis: Fixed.

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
manstis: I must remember this is not Java.... Fixed (and elsewhere).

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
manstis: Fixed.

manstis:
Thank-you for the review @tiagobento.

In response to your general comments:-

> utility classes

I've moved and renamed them to "better" places.

> i18n

We're already tracking the need to use your i18n [here](https://issues.redhat.com/browse/FAI-269).

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
manstis: Done.

packages/pmml-editor/src/editor/reducers/PMMLReducer.ts
manstis: Done.

packages/pmml-editor/src/editor/components/LandingPage/organisms/LandingPageToolbar.tsx
manstis: Updated.

tiagobento (approved):

manstis:
Hi @jiripetrlik just waiting for your review please.

manstis:
Hi @tomasdavidorg IDK if you want to review this instead of/as well as @jiripetrlik.

TBH IDK who from QE has an interest in PMML (editor) UI work.

tomasdavidorg (approved):

PR #289, opened 2020-10-13
KOGITO-3449: Add an option to update the current opened Gist on the Online Editor

ljmotta:
## Task
https://issues.redhat.com/browse/KOGITO-3347
https://issues.redhat.com/browse/KOGITO-3449

## Demo
ezgif com-video-to-gif


packages/online-editor/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "scripts": {
     "lint": "tslint -c ../../tslint.json 'src/**/*.{ts,tsx,js,jsx}'",
-    "test": "jest --silent --verbose",
+    "test": "jest --verbose",
```
tiagobento: Can we have that back? I really don't like the logs of testing showing up on the CI. If you need to see the logs you can simply use `jest` instead of `yarn test`.

packages/online-editor/src/editor/EditorPage.tsx
```diff
@@ -43,10 +42,12 @@ export function EditorPage(props: Props) {
   const [isEditorReady, setIsEditorReady] = useState(false);
   const [fullscreen, setFullscreen] = useState(false);
   const [copySuccessAlertVisible, setCopySuccessAlertVisible] = useState(false);
+  const [updateGistSuccessAlertVisible, setUpdateGistSuccessAlertVisible] = useState(false);
   const [githubTokenModalVisible, setGithubTokenModalVisible] = useState(false);
   const [showUnsavedAlert, setShowUnsavedAlert] = useState(false);
   const isDirty = useDirtyState(editor);
   const { locale, i18n } = useOnlineI18n();
+  const fileUrl = useMemo(() => getFileUrl(), [window.location]);
```
tiagobento: I think this has an unclear dependency here. If you don't know the implementation of `getFileUrl`, you'd never guess you needed to add it here. I think making a custom hook for this is the ideal. This way, you can even "listen" to changes on window.location.

packages/online-editor/src/editor/EditorPage.tsx
```diff
@@ -233,6 +263,15 @@ export function EditorPage(props: Props) {
             />
           </div>
         )}
+        {!fullscreen && updateGistSuccessAlertVisible && (
+          <div className={"kogito--alert-container"}>
+            <Alert
+              variant="success"
+              title={"Your gist was updated."}
```
tiagobento: i18n? :D

packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -85,6 +89,21 @@ export function EditorToolbar(props: Props) {
     [saveNewName, cancelNewName]
   );
 
+  const fileUrl = useMemo(() => getFileUrl(), [window.location]);
+
+  const [updateGist, setUpdateGist] = useState(false);
```
tiagobento: Can we rename that `userCanUpdateGist`? `updateGist` sounds like an imperative function.

packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -85,6 +89,21 @@ export function EditorToolbar(props: Props) {
...
+
+  useEffect(() => {
+    if (fileUrl) {
+      const userLogin = context.githubService.extractUserLoginFromGistRawUrl(fileUrl);
+      if (userLogin === context.githubService.getLogin()) {
+        setUpdateGist(true);
+      } else {
+        setUpdateGist(false);
+      }
```
tiagobento:
```suggestion
      setUserCanUpdateGist(userLogin === context.githubService.getLogin());
```

😜

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -190,7 +190,12 @@ export function HomePage(props: Props) {
         urlToOpen: undefined
       });
 
-      const gistId = context.githubService.extractGistId(inputFileUrl);
+      let gistId: string;
+      if (context.githubService.isGistDefault(inputFileUrl)) {
+        gistId = context.githubService.extractGistId(inputFileUrl)
+      } else {
+        gistId = context.githubService.extractGistIdFromRawUrl(inputFileUrl);
+      }
```
tiagobento: I'd go for a ternary here so that we don't have a mutable local variable.

packages/online-editor/src/__tests__/editor/EditorToolbar.test.tsx
```diff
@@ -15,10 +15,15 @@
...
 import * as React from "react";
-import { render } from "@testing-library/react";
+import { render, fireEvent } from "@testing-library/react";
 import { EditorToolbar } from "../../editor/EditorToolbar";
 import { StateControl } from "@kogito-tooling/editor/dist/embedded";
 import { usingTestingGlobalContext, usingTestingOnlineI18nContext } from "../testing_utils";
+import { GithubService } from "../../common/GithubService";
+import {EditorPage} from "../../editor/EditorPage";
+
+jest.clearAllMocks();
```
tiagobento: Shouldn't that be inside an `afterAll` too?

packages/online-editor/src/common/GithubService.ts
```diff
@@ -63,15 +65,15 @@ export class GithubService {
     setCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME, token);
   }
 
-  private validateToken(token: string): Promise<boolean> {
+  private validateToken(token: string): Promise<string | false> {
```
tiagobento:
```suggestion
  private validateToken(token: string): Promise<string | undefined> {
```

packages/online-editor/src/common/GithubService.ts
```diff
@@ -103,18 +107,38 @@ export class GithubService {
...
   public extractGistId(url: string): string {
     return url.substr(url.lastIndexOf("/") + 1);
   }
 
+  public extractGistIdFromRawUrl(url: string): string {
+    return url.split("gist.githubusercontent.com/")[1].split("/")[1];
+  }
+
+  public extractUserLoginFromGistRawUrl(url: string): string {
+    return url.split("gist.githubusercontent.com/")[1].split("/")[0];
```
tiagobento: Can we extract a constant for this domain -> `gist.githubusercontent.com`?

packages/online-editor/src/common/GithubService.ts
```diff
@@ -187,9 +211,26 @@ export class GithubService {
...
+
+    return this.octokit.gists.update(gistContent);
+  }
+
   public getGistRawUrlFromId(gistId: string): Promise<string> {
     return this.octokit.gists
-      .get({ gist_id: gistId })
+      .get({
+        gist_id: gistId
+      })
```
tiagobento: Prettier did that? :D

packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -125,21 +144,47 @@ export function EditorToolbar(props: Props) {
...
         context.external,
         context.readonly,
         props.onSave,
         props.onDownload,
         props.onCopyContentToClipboard,
-        props.onExportGist
+        props.onExportGist,
+        props.onUpdateGist,
+        window.location,
+        updateGist
```
tiagobento: This is a long list of dependencies. Sure we wouldn't benefit if this was inlined?

tiagobento (changes requested):
Thanks for the PR @ljmotta! Left some comments inline.

packages/online-editor/src/editor/EditorPage.tsx
ljmotta: Oops! Looks like I've forgot the i18n there. :disappointed:

packages/online-editor/src/common/GithubService.ts
ljmotta: It's not more natural to return a `false` than an `undefined` on a `validate` method?

packages/online-editor/src/common/GithubService.ts
ljmotta: Sure

packages/online-editor/src/common/GithubService.ts
ljmotta: Yes, probably because I've previously added more parameters in this object. I'm going to change it back to how it was. :grimacing:

packages/online-editor/src/common/GithubService.ts
tiagobento: Well... I guess if you change the name of this method to "getAuthenticatedUser", it's better then?

packages/online-editor/src/__tests__/editor/EditorToolbar.test.tsx
ljmotta: Yes

packages/online-editor/src/editor/EditorToolbar.tsx
ljmotta: If we inline this method we'll have duplicated code.

tiagobento:
@tomasdavidorg Can you please take a look at this?

tomasdavidorg:
Hello @ljmotta ,
I found two issues:
1. When a gist is updated, modal dialog `Your gist was updated.` appears and it covers some functionality. It cannot be switched between Properties and Explore Diagram panels or panel cannot be hide by the `fa-chevron-right` button.
Screenshot from 2020-10-21 11-41-44
2. When file is renamed in the online editor and gist is updated, it will create a new file in the gist but it won't rename the previous one. 

Gist can contain multiple files and it is not clear which will be opened in the editor.

packages/online-editor/src/common/i18n/locales/en.ts
```diff
@@ -57,13 +58,16 @@ export const en: OnlineI18n = {
     sendChangesToGitHub: `Send changes to ${en_common.names.github}`,
     copySource: "Copy Source",
     downloadSVG: `${en_common.terms.download} ${en_common.names.svg}`,
+    setGitHubToken: `Set up you ${en_common.names.github} token`,
```
tomasdavidorg:
```suggestion
    setGitHubToken: `Set up your ${en_common.names.github} token`,
```

tomasdavidorg (changes requested):

ljmotta:
@tomasdavidorg Hi Tomas. I'll work on fix both issues. Thanks for your review!

ljmotta:
@tomasdavidorg I've fixed the CSS error by removing the pointer event from the div. This same problem was on the other alerts as well.

The Update Gist functionality now renames the currently opened file. In case the user renames the Gist, it'll re-render the Editor page. This can't be avoided because it's necessary to update the current URL.

When the user opens a Gist from the Home page, it'll open the first Gist file. I've created a [JIRA](https://issues.redhat.com/browse/KOGITO-3680) so we can improve this behavior.

Additionally, the user can specify which file from the Gist should be opened, for example:
https://gist.github.com/ljmotta/db1c79d9919ffa3eb40ad8b7cada5b7f

This Gist has two files, and the user can use these specif URL's to open the chosen file:
- https://gist.github.com/ljmotta/db1c79d9919ffa3eb40ad8b7cada5b7f#file-test100-bpmn
- https://gist.github.com/ljmotta/db1c79d9919ffa3eb40ad8b7cada5b7f#file-test4-bpmn

tomasdavidorg:
Hello @ljmotta ,
thanks for the CSS fix. It works fine.

The opening does not work to me. Here is my Gist https://gist.github.com/tomasdavidorg/c7125375cff4a88499ad842b5864fc8e

- If you open the main link to the Gist - https://gist.github.com/tomasdavidorg/c7125375cff4a88499ad842b5864fc8e
  - The aaa.bpmn is opened - :heavy_check_mark: 
- If you open link to aaa - https://gist.github.com/tomasdavidorg/c7125375cff4a88499ad842b5864fc8e#file-aaa-bpmn
  - The aaa.bpmn is opened - :heavy_check_mark
- If you open link to bbb - https://gist.github.com/tomasdavidorg/c7125375cff4a88499ad842b5864fc8e#file-bbb-bpmn
  - The aaa.bpmn is opened - :x: 
- If you open raw link to aaa - https://gist.githubusercontent.com/tomasdavidorg/c7125375cff4a88499ad842b5864fc8e/raw/64c9183fa8ebebf0877fa135359c111e83fb6060/aaa.bpmn
  - The aaa.bpmn is opened - :heavy_check_mark: 
- If you open raw link to bbb - https://gist.githubusercontent.com/tomasdavidorg/c7125375cff4a88499ad842b5864fc8e/raw/64c9183fa8ebebf0877fa135359c111e83fb6060/bbb.bpmn
  - The aaa.bpmn is opened - :x: 

Because the first file is always opened, I checked only these test scenarios for renaming:
- If you rename first file:
  - The file is renamed in the Gist. :heavy_check_mark: 
- If you rename first file and use name which is alphabetically after the second file
  - It will display the second file content :x: 
- If you rename the first file and use the name of the second file
  - Second file content is replaced :x: 
- If you rename first file and use name which is alphabetically after the second file and then rename again.
  - A new file will be created (There will be 3 files instead of 2) :x:

tomasdavidorg (approved):
Thanks @ljmotta . Looks good to me. :slightly_smiling_face:

packages/online-editor/src/common/GithubService.ts
```diff
@@ -63,16 +78,16 @@ export class GithubService {
     setCookie(GITHUB_AUTH_TOKEN_COOKIE_NAME, token);
   }
 
-  private validateToken(token: string): Promise<boolean> {
+  private getAuthenticateUser(token: string): Promise<string | undefined> {
```
tiagobento, line 85:
```suggestion
  private getAuthenticatedUserLogin(token: string): Promise<string | undefined> {
```

packages/online-editor/src/common/GithubService.ts
```diff
@@ -183,16 +253,47 @@ export class GithubService {
...
+            content: args.content,
+            filename: args.filename
+          }
+        }
+      });
+      return this.removeCommitHashFromGistRawUrl((updateResponse.data as any).files[args.filename].raw_url);
+    } catch (err) {
+      throw Alerts.ERROR;
+    }
+  }
```
tiagobento: `throw` should be used to exceptional cases, where the application cannot recover and is forced to show an error message saying that something went wrong. Alerts are not an exceptional case, since they are expected and part of the application flow. `try/catch` statements are not `if/else` statements, nor are `then/catch` methods on Promises. Please change that so that the errors are returned by the function normally, and not thrown as exceptions. You can create an object like `GistUpdate` with the information about how did the Gist updated happened.

tiagobento: NOTE: Octokit rejecting their Promises is okay, since it might be that an unexpected behaviour happened. We *could* propagate this Promise rejection in our application too, but since we are not Octokit and for our domain it makes sense that a 3rd-party library (Octokit) request fails, it's not an exceptional flow too.

tiagobento (changes requested):
Some more minor comments inline. We're almost there!

packages/online-editor/src/common/GithubService.ts
ljmotta: Thanks for the exaplanation @tiagobento . I'll make the corrections.

tiagobento (approved):

PR #308, opened 2020-11-06
KOGITO-3450: Export the Online Editor content using the Standalone Editor

ljmotta:
## Task
https://issues.redhat.com/browse/KOGITO-3450


tiagobento (approved):

tiagobento:
@domhanak This is ready for review.

ederign (approved):

domhanak:
@ljmotta  Hi in JIRA ticket there is this line `The user should choose if the embedded editor is read-only or not using a checkbox.` I do not see such option available in this PR. Could you please clarify?

ljmotta:
@domhanak  Hi. I've forgotten to update the JIRA. It was decided to remove the option to tweak between read-only/editable based on UX feedback, and now it's always read-only. I'm going to update the JIRA, I'm sorry about that!

domhanak:
@ljmotta No worries, thanks for confirmation.
So far I was able to verify following:
**Chrome**
- Create DMN model and use `Embed` option in dropbox to export `iframe` of modeled diagram to different web application. I created simple static HTML page for this.
- Create BPMN model and use `Embed` option in dropbox to export `iframe` of modeled diagram to different web application. I created simple static HTML page for this.

Possible issue: Concerns the Height of the embedded editor. How could I control the height of it? See screenshot, where there are 2 `iframes` above each other. I expected them to stretch and each take half of the screen, but they are both squeezed. Same thing happens when only one editor is embedded.
Screenshot from 2020-12-07 15-23-06

**Firefox**
I tried to embedd the modelled diagrams but both DMN and BPMN do not load and I keep seeing infinite loading.

Screenshot from 2020-12-07 15-27-13

domhanak:
@ljmotta `and now it's always read-only.` When it comes to BPMN editor. I am able to work with the editor, move nodes around, change properties. I believe, this is because BPMN editor does not have read-only mode yet.

When it comes to DMN the read-only mode is working as expected.

So we need to clarify firefox, the height and it should be good to go.

ljmotta:
@domhanak Thanks for your review. Here are my answeres for the your topics:
**Chorme**
The `iframe` element will fill the size of the container where it was rendered on, so you need to adjust this container by adding a CSS property, in this case the `height` property.

**Firefox**
Which version of the Firefox are you using? Apparently this issue was happening on older versions. (I'm utilizing the 82.0, and this issue doesn't happen)

**BPMN**
Yes, the BPMN editor doens't support a read-only option yet.

domhanak (approved):
Turns out I used older version of Firefox ESR release when testing. (74.0.1) after upgrading to latest  ESR release (78.5.0) the issue is gone and embedded editor is loaded properly.

PR #314, opened 2020-11-12
Productization job script

Ginxo:
preparing the script to just execute `./build-publish.sh REGISTRY_URL`

### DIFF
--- a/build-publish.sh
+++ b/build-publish.sh
@@ -1,3 +1,5 @@
+#!/bin/sh
+# input parameter NPM Registry URL instead of default ones (npm, yarn)
 npm run install-locktt
 npm run locktt --registry=$1
 yarn install
--- a/build-publish.sh
+++ b/build-publish.sh
@@ -1,0 +1,5 @@
+npm run install-locktt
+npm run locktt --registry=$1
+yarn install
+yarn build
+yarn publish
--- a/build-publish.sh
+++ b/build-publish.sh
@@ -2,6 +2,6 @@
 # input parameter NPM Registry URL instead of default ones (npm, yarn)
 npm run install-locktt
 npm run locktt --registry=$1
-yarn install
-yarn build
-yarn publish
+yarn run init
+yarn build:prod
+yarn publish:all
--- a/build-publish.sh
+++ b/build-publish.sh
@@ -1,7 +1,9 @@
 #!/bin/sh
 # input parameter NPM Registry URL instead of default ones (npm, yarn)
+npm install -g yarn
 npm run install-locktt
-npm run locktt --registry=$1
+npm run locktt -- --registry=$1
+yarn install
 yarn run init
-yarn build:prod
-yarn publish:all
+yarn build:productization
+yarn run publish --yes --cd-version patch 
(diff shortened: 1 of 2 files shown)


mareknovotny:
looks generally ok, few comments:
please add directly there a shell script header and some comment for the argument directly into script.

@paulovmr would be helpful to have a script also for non-linux shells ?

build-publish.sh
```diff
@@ -0,0 +1,5 @@
+npm run install-locktt
```
mareknovotny, line 4:
```suggestion
#!/bin/sh
# input parameter NPM Registry URL instead of default ones (npm, yarn)
npm run install-locktt
```

mareknovotny: just possible suggestion ^^ to what I could see there.

mareknovotny (approved):

Ginxo:
@tiagobento @paulovmr discussing with @mareknovotny we notice probably `yarn publish` probably is not the right command to execute but we are not 100% sure. 
Probably we should replace it by `yarn publish-all` and add a new script to https://github.com/kiegroup/kogito-tooling/blob/c7bcb0c7e2acf4e6c12a02e9337039d86baa83bb/package.json#L9 with `"publish-all": "npx lerna run publish"`
WDYT guys?

tiagobento:
@Ginxo Yes, you're right. Since this repo is composed by many different packages, we should use Lerna tooling to execute the publish commands on each package. See my comments inline.

build-publish.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# input parameter NPM Registry URL instead of default ones (npm, yarn)
+npm run install-locktt
+npm run locktt --registry=$1
+yarn install
```
tiagobento, line 6: This should be `yarn run init`, since it executes a script called `init` that's configured on the root `package.json`. This will run the `lerna bootstrap` command, which will download the dependencies and link the inter-dependant packages on the repo.

build-publish.sh
```diff
@@ -0,0 +1,7 @@
...
+yarn build
```
tiagobento: This should be `yarn build:prod`

build-publish.sh
```diff
@@ -0,0 +1,7 @@
...
+yarn publish
```
tiagobento: I believe we can run lerna here to execute the `publish` Yarn goal on each package, as you mentioned.

tiagobento:
@mareknovotny Since this script is only going to be used by our CI/Prod builds, I believe we don't need a Windows version, unless you run these processes in a Windows instance too.

mareknovotny (approved):
lgtm

tiagobento:
@Ginxo Please take a look at https://github.com/kiegroup/kogito-tooling/blob/b71e7810b115ab4d1da20dae236323badb6df2a8/.github/workflows/publish_artifacts.yml#L82-L87. This is how we publish our packages to NPM. Of course the filtering doesn't need to be applied in this case, but I believe the commands are still slightly different. What's the purpose of this `publish:all` script?

tiagobento (approved):

tiagobento:
Closing as I think this subject is resolved now. cc @Ginxo

PR #351, opened 2020-12-10
Update README with more local development options

tiagobento:
cc @manstis

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,17 @@ Develop
 --------------------
 
 > **NOTE:** We recommend using *LTS* version of Node.js when developing this project. Our CI uses Node `12.16.3` and Yarn `1.19.1`.
+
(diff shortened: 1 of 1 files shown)


paulovmr (approved):

README.md
```diff
@@ -78,6 +78,17 @@ Develop
 --------------------
 
 > **NOTE:** We recommend using *LTS* version of Node.js when developing this project. Our CI uses Node `12.16.3` and Yarn `1.19.1`.
+
+> **NOTE:** This repo now consumes NPM packages containing the exploded WARs of the BPMN, DMN, and SceSim Editors. For this reason, if you want to use a local build of one of them during development, you can use some environment variables to point to the exploded WAR directories.
```
ederign, line 82: I would make explicit that this only affects GWT based editor. I think more and more this will be our corner case than our default path.

PR #355, opened 2020-12-11
FAI-336: Fix undoing new fields (DD, Outputs, Characteristics) not exiting editing mode

manstis:
See https://issues.redhat.com/browse/FAI-336

Since this the editor remains a PoC there are still no Unit Tests...

### DIFF
--- a/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsTable.tsx
+++ b/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsTable.tsx
@@ -74,12 +74,13 @@ export const CharacteristicsTable = (props: CharacteristicsTableProps) => {
     }
   }, [activeOperation]);
 
+  //Exit "edit mode" when the User adds a new entry and then immediately undoes it.
   useEffect(() => {
     if (selectedCharacteristicIndex === characteristics.length) {
       setSelectedCharacteristicIndex(undefined);
       setActiveOperation(Operation.NONE);
     }
-  }, [characteristics]);
+  }, [characteristics, selectedCharacteristicIndex]);
 
   const onEdit = (index: number | undefined) => {
     setSelectedCharacteristicIndex(index);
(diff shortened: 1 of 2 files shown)


packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsTable.tsx
```diff
@@ -74,6 +74,13 @@ export const CharacteristicsTable = (props: CharacteristicsTableProps) => {
     }
   }, [activeOperation]);
 
+  useEffect(() => {
+    if (selectedCharacteristicIndex === characteristics.length) {
+      setSelectedCharacteristicIndex(undefined);
+      setActiveOperation(Operation.NONE);
+    }
+  }, [characteristics]);
```
tiagobento: I believe you need `selectedCharacteristicIndex` as a dependency of this effect as well. Ideally, `setSelectedCharacteristicIndex` too 🙊 . Same for below.

manstis: @tiagobento Yes and no (but mainly no)... it is impossible to select a row in the table whose index is equal to the length of the collection. Indexes are zero-based. There is however one such scenario where a User can add a new row and immediately undoes the addition. The selected index in this scenario points to the now missing row. So the hook is really to catch this specific scenario. That said, it does no harm to include the selected index; so I have do so in a new commit.

kelvah (approved):

sjamboro (approved):

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsTable.tsx
tiagobento: Yeah, I understand your point.. and I'm actually not sure how to proceed on such scenarios. We, when writing the code, know the intricacies of the possible execution branches and therefore can reason about the dependencies on hooks being necessary or not, however, my concern is about the mid/long-term future. I'm not sure we can remember those scenarios when changing something after a while.. Therefore, leaving dependencies out might cause headaches. But again, that's my reasoning. I'm not sure what's the "standard" or "recommended" way to handle that. I'm fine with whatever you decide :)

manstis: Hi @tiagobento I added a new commit including the dependency and a comment explaining the hook's purpose.

PR approval beckons....

tiagobento (approved):

PR #356, opened 2020-12-13
Enable multi-package live-reload to improve Online Editor development

tiagobento:
This PR adds a new CLI tool (`@kogito-tooling/monorepo-reload`) that enables multiple packages to be rebuilt after a change on one of the dependents of a running application (`@kogito-tooling/online-editor`). This is particularly convenient for development purposes. Since we have many packages that compose the webapp that is the Online Editor, it's nice not to have to manually rebuild the packages every time a change is made.

By using Lerna's built-in dependency graph and [`chokidar`](https://github.com/kimmobrunfeldt/chokidar-cli) as the directory watcher, we can watch for changes on all the dependency packages of the Online Editor and automatically trigger a rebuild. See the new `dev:online-editor` script.

To make this possible, I had to remove the `envelope.js` and `envelope.html` files from `@kogito-tooling/kie-bc-editors` and replicate them to the Desktop and Online Editor channels, since that's the places these files were being used. This is aligned with the recommended way of consuming Envelopes -- each Channel should be responsible for creating its own Envelopes. So no harm done.

Please feel free to try it and give feedback! I created this new tool in a way that other projects can benefit from it too, so on our next release, this will be publicly available on NPM.

#### Known issues:
- After a change, the Online Editor webapp keeps trying to recompile for each package rebuilt. This causes some pollution on the Terminal log. I wasn't able to address that yet, but once all the packages are done rebuilding, the webapp is correctly recompiled and refreshed on the browser.
- Only tested on macOS. Probably works on Linux, but probably doesn't on Windows.


packages/kie-bc-editors/src/__tests__/envelope/envelope.test.ts
```diff
@@ -1,69 +0,0 @@
...
-        }
-      }
-    };
-  })
-});
-
-//Lazy load module as it executes once loaded and, if import'ed, it's before the mocks are setup.
-import module = require("../../envelope/envelope");
-
-describe("EditorEnvelope.init", () => {
```
jstastny-cz, line 39: Not needed anymore or got lost during refactoring?

tiagobento: Not needed anymore.

caponetto:
Hi @tiagobento,
As we talked, I needed to force a rebuild of the `online-editor` after changing something on, for example, the package `i18n-common-dictionary`. It'd be nice to avoid that.

Very useful package for development, by the way. :)

tiagobento:
@caponetto Thanks for looking into this.

Yeah... looks like this needs more work. Webpack DevServer watchers are getting lost with the many changes that rebuilding other packages trigger.

tiagobento:
Updated. This is not perfect yet, but I guess can be very convenient in some development cases. See https://github.com/webpack/webpack-dev-server/issues/2930.

tiagobento:
After discussing this with @caponetto, I decided to close this PR. Although I think it would be a nice improvement to our development workflows, it didn't work as expected and `webpack-dev-server` does not seem to be built with this use-case in mind. I'll continue researching the subject and hopefully come up with a better alternative. Thanks for all who looked at it.

PR #357, opened 2020-12-15
Fix FDB

tiagobento:
On this PR:
- Fixed FDB script
- Moved all checkouts to the beginning of the Workflow to avoid 
- Prevented WEB-INF from being copied from the Editors directories
- Improved build of `kie-wb-common` by only building the dependnecies of DMN and BPMN Editors
- Improved directory organization with multiple repositories by using the `path` property on `actions/checkout@v2`

### DIFF
--- a/.github/workflows/pull_request_full_downstream_ci.yml
+++ b/.github/workflows/pull_request_full_downstream_ci.yml
@@ -25,7 +25,7 @@ jobs:
         run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
 
       - name: Fetch user membership to kiegroup
-        if: github.repository == 'kiegroup/kogito-tooling' && (github.event_name != 'issue_comment' || startsWith(github.event.comment.body, 'Build:'))
+        if: github.repository == 'kiegroup/kogito-tooling'
         uses: octokit/request-action@v2.x
         id: checkUserMembership
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/pull_request_full_downstream_ci.yml
```diff
@@ -293,44 +282,64 @@ jobs:
...
+        uses: actions/upload-artifact@v1
+        with:
+          name: desktop-app-linux-fdb
+          path: ${{ github.workspace }}/kogito-tooling/packages/desktop/out/Business Modeler Preview-linux-x64.tar.gz
+
+      - name: Upload Hub App
+        uses: actions/upload-artifact@v1
+        with:
+          name: hub-app-linux-fdb
+          path: ${{ github.workspace }}/kogito-tooling/packages/hub/out/Business Modeler Hub Preview-linux-x64.tar.gz
```
tomasdavidorg: @tiagobento I think these steps should be before the `Rebuild Chrome Extension to use servers running on localhost and run integration tests` because we rebuild Chrome extensions there.

tiagobento: Good point! I'll change it.

tiagobento:
@domhanak Can you please take a look at the error that happened on the VS Code integration tests? It broke `CI :: Default` on `macOS`.

tomasdavidorg (approved):

paulovmr (approved):

PR #360, opened 2020-12-16
Revert removal of integrity checks in yarn.lock

jstastny-cz:


tiagobento:
@jstastny-cz Thanks for the PR. It would be great for us to understand what caused that. Since `yarn.lock` is a big file and we have good-enough checks on our CI, we many times skip looking at it more carefully on reviews. It's clear to me that there's no reason for us to remove those integrity checks, so understanding the commands ran during development that made the integrity checks be removed will be valuable to us.

domhanak:
@tiagobento it is usage of `locktt --registry=<URL>` without `--skipIntegrity`

tiagobento:
@domhanak Right. Thanks for reply!

@Ginxo Hi! So, I'm not that familiar with the everyday use of `locktt`, I only know it's used inside our Jenkins/Prod environments, but given what happened here, maybe we could make `--skipIntegrity` be turned on by default? Or is it that the standard usage of `locktt` always aims at removing the integrity checks from `yarn.lock`/`package-lock.json` files?

domhanak (approved):
LGTM, some dependenciess resolved to newer version I will update them in separate PR. Apologies for causing this and thanks putting the integrity checksums back.

tiagobento:
@domhanak What do you mean "some dependencies resolved to newer version"? `yarn.lock` is basically a recipe for reproducible builds, so everything that's written there is what's used to download the dependencies. Also, `yarn.lock` has to be aligned with what's declared on our `package.json` files. If it's not, then after a `yarn run init`, we'll see changes on `yarn.lock`. That's why we have the `Check yarn.lock` steps on our CI builds. If everything is green, we shouldn't have to update anything. Am I missing something?

domhanak:
@tiagobento you are correct, all I mean that since we use `^` yarn resolved newer version than we have declared(expected).

tiagobento (approved):

tiagobento (approved):

Ginxo:
+1 to this PR. `locktt` it's basically used by `locktt` in production build in order to replace registry (prod build system only access to our internal npm registry and it fails due to `.lock` files).
It seems to be solved by yarn 2+ so we will remove locktt from this project once we test it and we migrate prod system to this new version.

tiagobento:
@Ginxo Thx for chiming in! Upgrading to Yarn 2 sounds like a good idea :) Thanks!

PR #358, opened 2020-12-15
FAI-335: Plugging data dictionary to state management (PMML editor PoC)

kelvah:
See https://issues.redhat.com/browse/FAI-335

@manstis I plugged data dictionary fields add, batch add, edit, delete, order and partial range constraints handling (only one interval at the moment).
Following PR's will cover full range and enums constraints and validation.

PR #361, opened 2020-12-18
FAI-288: Score Cards: Reducer operation

manstis:
See https://issues.redhat.com/browse/FAI-288

- 2-Phase commit.
- Access to model sub-trees in all Reducers.

### DIFF
--- a/packages/pmml-editor/src/editor/history/BatchDispatchHook.ts
+++ b/packages/pmml-editor/src/editor/history/BatchDispatchHook.ts
@@ -14,6 +14,25 @@
  * limitations under the License.
  */
 
-export * from "./BatchDispatchHook";
-export * from "./HistoryAwareReducer";
-export * from "./HistoryProvider";
+import { useDispatch } from "react-redux";
+import { HistoryService } from "./HistoryProvider";
+import { Dispatch } from "redux";
+import { PMML } from "@kogito-tooling/pmml-editor-marshaller";
+import { Actions } from "../reducers";
+
+export const useBatchDispatch = (service: HistoryService, getCurrentState: () => PMML | undefined): Dispatch<any> => {
+  const dispatch = useDispatch();
+
+  return (action: any): any => {
+    const result = dispatch(action);
+
+    dispatch({
+      type: Actions.Refresh,
+      payload: {
+        pmml: service.commit(getCurrentState())
+      }
+    });
+
+    return result;
+  };
+};
--- a/packages/pmml-editor/src/editor/reducers/DelegatingModelReducer.ts
+++ b/packages/pmml-editor/src/editor/reducers/DelegatingModelReducer.ts
@@ -16,38 +16,63 @@
 import { HistoryAwareModelReducer, HistoryService, ModelReducerBinding } from "../history";
 import { Model } from "@kogito-tooling/pmml-editor-marshaller";
 import { Reducer } from "react";
-import { AllScorecardActions } from "./ScorecardReducer";
-import { getModelType, ModelType } from "..";
+import { getModelType, ModelType } from "../PMMLModelHelper";
+import { AllActions } from "./Actions";
+import get = Reflect.get;
 
-export const DelegatingModelReducer: HistoryAwareModelReducer<AllScorecardActions> = (
+const reduce = (model: Model, action: AllActions, reducers: Map<ModelType, ModelReducerBinding<any, any>>) => {
+  const modelType = getModelType(model);
+  const reducer = reducers.get(modelType);
+  if (reducer !== undefined) {
+    return reducer.factory(reducer.reducer(model, action));
+  }
+  return model;
+};
+
+export const DelegatingModelReducer: HistoryAwareModelReducer<AllActions> = (
   service: HistoryService,
   reducers: Map<ModelType, ModelReducerBinding<any, any>>
-): Reducer<Model[], AllScorecardActions> => {
-  return (state: Model[], action: AllScorecardActions) => {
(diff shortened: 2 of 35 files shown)


manstis:
@tiagobento IDK if you're interested in this..

It's about batching changes made by different Reducers into a single commit applied to the model with immer.

It's easier to explain face to face than in words.. something for the New Year!

packages/pmml-editor/src/editor/history/BatchDispatchHook.ts
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+import { useDispatch } from "react-redux";
+import { HistoryService } from "./HistoryProvider";
+import { Dispatch } from "redux";
+import { PMML } from "@kogito-tooling/pmml-editor-marshaller";
+import { Actions } from "../reducers";
+
+export const useBatchDispatch = (service: HistoryService, getCurrentState: () => PMML | undefined): Dispatch<any> => {
```
manstis, line 23: Here's the guts of it:-

- `dispatch( action )` gets reducers to record a change to be applied
- `service.commit(..)` applies the recorded changes to the model in a single change
- `dispatch( Refresh )` signals the state change to React.

manstis:
Depends on https://github.com/kiegroup/kogito-tooling/pull/358

#358 Contains multiple changes to `DataDictionary` support that affects this PR.

I have a working update on top op #358 locally. I'll update this PR when #358 is merged.

packages/pmml-editor/src/editor/reducers/DelegatingModelReducer.ts
```diff
@@ -16,38 +16,63 @@
...
+    //to be correctly detected.
+    let changed = false;
+    const newState: Model[] = [];
+    state.forEach(m => newState.push(m));
+    const modelIndex: number = get(action.payload, "modelIndex");
+
+    //Delegate Model agnostic actions to all Model reducers
+    if (modelIndex === undefined) {
+      state.every((model, index) => {
+        const modelAction = Object.assign({}, action, { payload: { ...action.payload, modelIndex: index } });
```
manstis, line 53: @kelvah Whilst `Actions.UpdateDataDictionaryField` does not need a `modelIndex` I clone the payload and add the `modelIndex` when delegating to model specific Reducers as `modelIndex` becomes part of the "path" stored in the history log so the correct model segment is reverted when the User undoes an operation. The payload for `Actions.UpdateDataDictionaryField` [now has](https://github.com/kiegroup/kogito-tooling/pull/361/files#diff-5cc11150206c93edb897e5ad2efbf2dce519dc8a19628a6156559cae66855d94R26) `modelIndex` as an optional member although I suspect that can be removed as the use of `Object.assign(..)` does not care whether it's declared or not. Please advise your preference. That said I'd need to change [this](https://github.com/kiegroup/kogito-tooling/pull/361/files#diff-e43c5ad43eba3495f0cf28ddbe194bf127c410279e12965dbfd2ac38af0fa920R57) to use reflection if the `modelIndex` is removed from the definition.

manstis:
@kelvah @tiagobento This is about as good as it's going to get...

Happy to arrange a meeting to discuss the approach used herein... just ask.

packages/pmml-editor/src/editor/reducers/DelegatingModelReducer.ts
kelvah: It's difficult to choose between the two options. I would probably keep `modelIndex` in the payload interface. It could help to understand how the updated `MiningSchemaFieldReducer` is operating.

packages/pmml-editor/src/editor/reducers/DelegatingModelReducer.ts
```diff
@@ -16,38 +16,63 @@
...
+    //this reducer needs to return a *new* Model[] instance in order for a state change
+    //to be correctly detected.
+    let changed = false;
+    const newState: Model[] = [];
+    state.forEach(m => newState.push(m));
+    const modelIndex: number = get(action.payload, "modelIndex");
+
+    //Delegate Model agnostic actions to all Model reducers
+    if (modelIndex === undefined) {
+      state.every((model, index) => {
```
kelvah: any particular reason to use `every` here?

tiagobento (approved):
Thank for taking the time to explain that in more depth! Really nice addition with powerful capabilities. Nice work! Approving.

tiagobento:
@manstis Abou the FDB failures, I believe the VS Code Extension tests are too unstable right now. I'll disable them temporarily until we have a fix. cc @domhanak

kelvah (approved):
Hi @manstis, approved and left a minor comment.

packages/pmml-editor/src/editor/reducers/DelegatingModelReducer.ts
manstis: Nope.. just where I forget to select `forEach` instead and `every` pops into the code!

I'll fix.

sjamboro (approved):
Approved. Can you consider to add some test which will check that MiningSchema react on action UpdateDataDictionaryField and REDO/UNDO? Not fully coverage, just simple use case to show interaction between "Editors".

manstis:
@sjamboro I've added some basic test coverage.

sjamboro (approved):

manstis:
Closing and re-opening.. in an attempt to drop the FDB (that has been disabled).

manstis:
@tiagobento Can this be merged? Only the FDB is failing.

PR #384, opened 2021-01-20
FAI-366: Score Cards: Validation: MiningField.importance valid range

manstis:
See https://issues.redhat.com/browse/FAI-366

The JIRA contains details of how to test.


manstis:
I'm going to submit a single PR for FAI-366, FAI-367 and FAI-368 as I've discussed some changes with @kelvah.

PR #391, opened 2021-01-21
KOGITO-3718: Importing DMN as Included Model in VS Code on Windows

tiagobento:
After analysing this issue more carefully, the issue was that, on Windows, the extension was not correctly identifying the current directory of the open file, and it was always falling back to the root of the Workspace.

A workaround for this issue is to always work on the root folder of the Workspace. Keep in mind that even after this fix, only DMNs on the same folder can be used as Included Models.

### DIFF
--- a/packages/vscode-extension/src/KogitoEditableDocument.ts
+++ b/packages/vscode-extension/src/KogitoEditableDocument.ts
@@ -28,6 +28,7 @@ import { KogitoEdit } from "@kogito-tooling/channel-common-api";
 import { KogitoEditor } from "./KogitoEditor";
 import { VsCodeI18n } from "./i18n";
 import { I18n } from "@kogito-tooling/i18n/dist/core";
+import * as nodePath from "path";
 
 export class KogitoEditableDocument implements CustomDocument {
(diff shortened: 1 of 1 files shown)


ederign (approved):

ederign:
@danielzhe can you please help us test this?

caponetto (approved):

danielzhe (approved):
It's fixed. Thank you!

jomarko (commented):
We should contribute also unit test to this

tiagobento:
@jomarko Thanks for the review! VS Code does not provide an easy way for us to unit-test it, so we would have to rely on heavy mocking to write them at the moment. We're researching what's the best way to approach unit tests for our VS Code Extension yet, but so far we don't have anything in place. This is something very important that will be addressed as soon as we have the resources to do so.

jomarko (approved):
Found new issue during review
- https://issues.redhat.com/browse/KOGITO-4261

Next I decided to prepare an automated test
- https://github.com/kiegroup/kogito-tooling/pull/395

I listed PRs that block from merging it. Please Prioritize them.

PR #394, opened 2021-01-25
FAI-366, FAI-367, FAI-368, FAI-369: MiningSchema validation

manstis:
See the following JIRA

https://issues.redhat.com/browse/FAI-366
https://issues.redhat.com/browse/FAI-367
https://issues.redhat.com/browse/FAI-368
https://issues.redhat.com/browse/FAI-369

The different JIRA contain details of the change and steps to test.

### DIFF
--- a/packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaContainer/MiningSchemaContainer.tsx
+++ b/packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaContainer/MiningSchemaContainer.tsx
@@ -1,8 +1,8 @@
 import * as React from "react";
-import { useEffect, useState } from "react";
-import { Bullseye, Stack, StackItem, Title, TitleSizes } from "@patternfly/react-core";
+import { useEffect, useMemo, useState } from "react";
+import { Alert, Bullseye, Stack, StackItem, Title, TitleSizes } from "@patternfly/react-core";
 import { CSSTransition, SwitchTransition } from "react-transition-group";
-import { pickBy, isEqual } from "lodash";
+import { isEqual, pickBy } from "lodash";
 import EmptyMiningSchema from "../EmptyMiningSchema/EmptyMiningSchema";
 import MiningSchemaFields from "../MiningSchemaFields/MiningSchemaFields";
 import MiningSchemaAddFields from "../MiningSchemaAddFields/MiningSchemaAddFields";
@@ -11,8 +11,10 @@ import "./MiningSchemaContainer.scss";
 
 import { DataDictionary, MiningField, MiningSchema } from "@kogito-tooling/pmml-editor-marshaller";
 import NoMiningSchemaFieldsOptions from "../NoMiningSchemaFieldsOptions/NoMiningSchemaFieldsOptions";
+import { useValidationService } from "../../../validation";
 
 interface MiningSchemaContainerProps {
+  modelIndex: number;
   dataDictionary?: DataDictionary;
   miningSchema?: MiningSchema;
   onAddField: (name: string[]) => void;
@@ -21,7 +23,7 @@ interface MiningSchemaContainerProps {
 }
 
 const MiningSchemaContainer = (props: MiningSchemaContainerProps) => {
-  const { dataDictionary, miningSchema, onAddField, onDeleteField, onUpdateField } = props;
+  const { modelIndex, dataDictionary, miningSchema, onAddField, onDeleteField, onUpdateField } = props;
 
   const [fields, setFields] = useState<MiningSchemaOption[]>(prepareFieldOptions(dataDictionary, miningSchema));
   const [viewSection, setViewSection] = useState<MiningSchemaSection>("overview");
@@ -80,6 +82,9 @@ const MiningSchemaContainer = (props: MiningSchemaContainerProps) => {
     setFields(prepareFieldOptions(dataDictionary, miningSchema));
   }, [dataDictionary, miningSchema]);
 
+  const validationService = useValidationService().service;
+  const validations = useMemo(() => validationService.get(`models[${modelIndex}].MiningSchema`), [miningSchema]);
+
   return (
     <section className="mining-schema">
       <MiningSchemaContext.Provider value={editingField}>
@@ -122,14 +127,22 @@ const MiningSchemaContainer = (props: MiningSchemaContainerProps) => {
                               </Bullseye>
                             ))}
                           {miningSchema && miningSchema.MiningField.length > 0 && (
-                            <MiningSchemaFields
-                              fields={miningSchema?.MiningField}
-                              onAddProperties={goToProperties}
-                              onDelete={handleDeleteField}
-                              onPropertyDelete={handlePropertyDelete}
-                              onEdit={handleEditField}
-                              onCancel={handleCancelEditing}
-                            />
+                            <>
+                              {validations.length > 0 && (
+                                <section className="mining-schema__validation-alert">
+                                  <Alert variant="warning" title="Some items are invalid and need attention." />
+                                </section>
+                              )}
+                              <MiningSchemaFields
+                                modelIndex={modelIndex}
+                                fields={miningSchema?.MiningField}
+                                onAddProperties={goToProperties}
+                                onDelete={handleDeleteField}
+                                onPropertyDelete={handlePropertyDelete}
+                                onEdit={handleEditField}
+                                onCancel={handleCancelEditing}
+                              />
+                            </>
                           )}
                         </>
                       )}
@@ -139,6 +152,8 @@ const MiningSchemaContainer = (props: MiningSchemaContainerProps) => {
               )}
               {viewSection === "properties" && (
                 <MiningSchemaPropertiesEdit
+                  modelIndex={modelIndex}
+                  miningFieldIndex={editingField}
                   field={miningSchema!.MiningField[editingField]}
                   onSave={handlePropertiesSave}
                   onClose={handlePropertiesClose}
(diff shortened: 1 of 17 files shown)


packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.scss
```diff
@@ -1,5 +1,6 @@
 .data-dictionary {
   height: 60vh;
+
```
kelvah: Do you have any idea why scss are being reformatted?

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaContainer/MiningSchemaContainer.tsx
```diff
@@ -122,14 +127,22 @@ const MiningSchemaContainer = (props: MiningSchemaContainerProps) => {
...
-                              onAddProperties={goToProperties}
-                              onDelete={handleDeleteField}
-                              onPropertyDelete={handlePropertyDelete}
-                              onEdit={handleEditField}
-                              onCancel={handleCancelEditing}
-                            />
+                            <>
+                              {validations.length > 0 && (
+                                <section className="mining-schema__validation-alert">
+                                  <Alert variant="warning" title="Some items are invalid and need attention." />
```
kelvah: This alert could be placed outside `mining-schema__fields` so it would stay visible even when scrolling a long list of fields.
I would also suggest to use an inline Alert, the drop shadow doesn't seem to fit with the overall style of the list of fields.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaFields/MiningSchemaFields.tsx
```diff
@@ -99,11 +115,26 @@ const MiningSchemaItem = (props: MiningSchemaFieldProps) => {
     >
       <section className={"editable-item__inner"}>
         <Split hasGutter={true}>
+          {index !== editing && validations.length > 0 && (
```
kelvah, line 118: Do you think the filed level validation warning should be displayed only when not editing? Asking because it's always visible inside Data Dictionary. I understand it could be seen as a duplication if a more specific warning is there next to a label. I would probably keep them both to avoid a difference in the validation style between editing and viewing items. What's your opinion?

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaPropertiesEdit/MiningSchemaPropertiesEdit.tsx
```diff
@@ -111,114 +172,185 @@ const MiningSchemaPropertiesEdit = ({ field, onSave, onClose }: MiningSchemaProp
...
-                    id="highValue"
-                    name="highValue"
-                    aria-describedby="High Value"
-                    value={highValue}
-                    onChange={value => setHighValue(Number.parseFloat(value))}
-                    onBlur={handleSave}
-                  />
-                </FormGroup>
+              <SplitItem style={{ width: "50%" }}>
+                <Card isCompact={true} style={{ margin: "1em 0" }}>
```
kelvah: The card for outliers related fields seems a bit overkill. They seems "special" while their behavior as conditionally required fields is similar to missing/invalid values fields. I would probably put them in a single row.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaPropertiesEdit/MiningSchemaPropertiesEdit.tsx
```diff
@@ -111,114 +172,185 @@ const MiningSchemaPropertiesEdit = ({ field, onSave, onClose }: MiningSchemaProp
...
+                            selection={outliers}
+                          />
+                        </FormGroup>
+                      </StackItem>
+                      <StackItem>
+                        <FormGroup
+                          label="Low Value"
+                          fieldId="lowValue"
+                          className="mining-schema__properties__field"
+                          helperText={validationsLowValue.length === 0 ? "" : validationsLowValue[0].message}
```
kelvah, line 280: I am not discussing a change here but I have some doubts about using the general validation messages everywhere. They include the field name, which seems unnecessary here. I don't have a solution that doesn't imply duplicating the messages for local validations. Maybe we could revise the topic when we'll finish adding validations everywhere and we'll have a bigger picture.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaPropertiesEdit/MiningSchemaPropertiesEdit.tsx
```diff
@@ -111,114 +172,185 @@ const MiningSchemaPropertiesEdit = ({ field, onSave, onClose }: MiningSchemaProp
...
+                          validated={validationsLowValue.length === 0 ? "default" : "warning"}
+                          labelIcon={
+                            <Tooltip
+                              content={`Low Value is required when Outliers is "asExtremeValues" or "asMissingValues"`}
+                            >
+                              <button
+                                aria-label="More information for Low Value field"
+                                onClick={e => e.preventDefault()}
+                                aria-describedby="simple-form-name"
```
kelvah: This aria attribute should be probably changed. Also in other places where this kind of tooltip is present. (Sorry, the other day I linked you an example of this from data dictionary that had the same problem, I later realized it and fixed it.)

kelvah (commented):
Hi @manstis, I left some comments. Thank you.

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.scss
manstis: I reformatted it at some point; but I have reverted to the formatting you have.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaFields/MiningSchemaFields.tsx
manstis: As discussed, changed to be consistent with your `DataDictionary` PR.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaPropertiesEdit/MiningSchemaPropertiesEdit.tsx
manstis: Removed the `aria-describedby` attribute to be consistent with `DataDictionaryPropertiesEdit`.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaPropertiesEdit/MiningSchemaPropertiesEdit.tsx
manstis: Changed as discussed. There is no doubt room for more improvement however it should suffice for now.

Screenshot from 2021-01-27 11-51-17

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaContainer/MiningSchemaContainer.tsx
manstis: Corrected.

Screenshot from 2021-01-27 12-17-37

manstis:
@kelvah I've updated the PR following your feedback and our discussion.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaFields/MiningSchemaFields.tsx
kelvah: We agreed to display it always I think, right? I don't see changes about it :thinking:

manstis: TBH I thought this comment related to the label warnings when not in edit mode. That said, the {{DataDictionary}} does now show the row level warning when editing (the following screen shots are taken with your latest commits):-

**_Read-only_**
DF-readonly

**_Editing_**
DF-editing

So I am consistent with you! But I am happy to change if you are too.

kelvah: You're right, I got confused. I think we can leave it as is for now. The fact that DD has a form while editing an item while MS does not have one makes finding a common approach a bit tricky. Don't know what's the best for all cases at this point. Let's revisit the topic later on. Thank you.

kelvah (approved):

manstis:
@tiagobento @sjamboro <nudge> please review when you're able.. the PR is 100% green. A sight to behold and hard to obtain.

tiagobento:
@manstis Expect to see a lot more all-green builds from now on :) I want to make red builds to be respected!!

tiagobento (approved):

sjamboro:
@manstis it seems that FAI-369 works weird. If I rewrite invalid value to another invalid value than value is changed to "not needed" but error is still presented.
image

manstis:
@sjamboro Thank-you for finding the issue. I have pushed a new commit fixing it.

You will probably have to delete your local branch and fetch again entirely (as I have also sync'ed this branch with `master`).

sjamboro (approved):
@manstis now it works fine, thank you for quick fix

PR #396, opened 2021-01-26
KOGITO-4273: Enable Read-only mode it-tests

jstastny-cz:
Enabling the test - should be merged only when all builds are green.

### DIFF
--- a/packages/kie-editors-standalone/it-tests/cypress/integration/dmn-readonly.spec.ts
+++ b/packages/kie-editors-standalone/it-tests/cypress/integration/dmn-readonly.spec.ts
@@ -14,13 +14,13 @@
  * limitations under the License.
  */
 
-describe.skip("Dmn Read Only.", () => {
+describe("Dmn Read Only.", () => {
   before("Visit page", () => {
     cy.visit("localhost:9001/dmn-read-only");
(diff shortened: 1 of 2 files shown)


jomarko (approved):

jstastny-cz:
@tiagobento could you please advise on which builds are not relevant anymore? This PR was opened before the changes to CI, so I think we won't get all checks green here, right?

tiagobento:
@jstastny-cz If you don't have changes on the Editors, everything except FDB should be green. If you have changes on the Editors, you need a new snapshot version based on `master` and then everything, including FDB, should be green. Since these checks are already old, please push a new empty commit to re-trigger them. Keep in mind that we'll remove FDB once the checks on the v7 repos are in place, so the default will always be everything green.

jstastny-cz:
@tiagobento all checks except of FDB are green. FDB failed in kie-wb-common with 
```
Failed to execute goal org.codehaus.mojo:gwt-maven-plugin:2.9.0:compile (gwt-compile) on project kie-wb-common-stunner-bpmn-kogito-runtime: Command failed with status 137 -> [Help 1]
```

tiagobento (approved):

PR #405, opened 2021-02-01
FAI-380 [Scorecard Editor] Output validation

kelvah:
JIRA ticket: https://issues.redhat.com/browse/FAI-380

This PR adds validation for the outputs section.

### DIFF
--- a/packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabels.tsx
+++ b/packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabels.tsx
@@ -16,10 +16,13 @@
 import * as React from "react";
 import { FieldName, OpType, RankOrder, ResultFeature } from "@kogito-tooling/pmml-editor-marshaller";
 import { OutputFieldLabel } from "./OutputFieldLabel";
+import { ValidationEntry } from "../../../validation";
+import { ValidationIndicatorLabel } from "../../EditorCore/atoms";
 
 interface OutputLabelsProps {
   optype: OpType | undefined;
   targetField: FieldName | undefined;
+  targetFieldValidation: ValidationEntry[];
   feature: ResultFeature | undefined;
   value: any | undefined;
   rank: number | undefined;
@@ -29,12 +32,30 @@ interface OutputLabelsProps {
 }
 
 export const OutputLabels = (props: OutputLabelsProps) => {
-  const { optype, targetField, feature, value, rank, rankOrder, segmentId, isFinalResult } = props;
+  const {
+    optype,
+    targetField,
+    targetFieldValidation,
+    feature,
+    value,
+    rank,
+    rankOrder,
+    segmentId,
+    isFinalResult
+  } = props;
 
   return (
     <>
       {optype && OutputFieldLabel("OpType", optype)}
-      {targetField && OutputFieldLabel("TargetField", targetField)}
+      {targetFieldValidation.length > 0 ? (
+        <ValidationIndicatorLabel validations={targetFieldValidation} cssClass="output-fields-list__item__label">
+          <strong>TargetField:</strong>&nbsp;
+          <em>Missing</em>
+        </ValidationIndicatorLabel>
+      ) : (
+        targetField && OutputFieldLabel("TargetField", targetField)
+      )}
+      {/*{targetField && OutputFieldLabel("TargetField", targetField)}*/}
       {feature && OutputFieldLabel("Feature", feature)}
       {value && OutputFieldLabel("Value", value)}
       {rank !== undefined && OutputFieldLabel("Rank", rank)}
--- a/packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabels.tsx
+++ b/packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabels.tsx
@@ -55,7 +55,6 @@ export const OutputLabels = (props: OutputLabelsProps) => {
       ) : (
         targetField && OutputFieldLabel("TargetField", targetField)
       )}
-      {/*{targetField && OutputFieldLabel("TargetField", targetField)}*/}
       {feature && OutputFieldLabel("Feature", feature)}
       {value && OutputFieldLabel("Value", value)}
       {rank !== undefined && OutputFieldLabel("Rank", rank)}
--- a/packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabelsEditMode.tsx
+++ b/packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabelsEditMode.tsx
@@ -145,7 +145,7 @@ export const OutputLabelsEditMode = (props: OutputLabelsEditModeProps) => {
         <Label
           style={PADDING}
           variant="outline"
-          color="orange"
+          color="cyan"
           href="#outline"
           icon={<ArrowAltCircleRightIcon />}
           onClick={e => {
@@ -153,7 +153,7 @@ export const OutputLabelsEditMode = (props: OutputLabelsEditModeProps) => {
             viewExtendedProperties();
           }}
         >
-          Edit properties...
+          Edit properties
         </Label>
       )}
     </>
--- a/packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabelsEditMode.tsx
+++ b/packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabelsEditMode.tsx
@@ -108,7 +108,7 @@ export const OutputLabelsEditMode = (props: OutputLabelsEditModeProps) => {
             value: undefined
           });
         })}
-      {rank &&
+      {rank !== undefined &&
         OutputFieldLabel("Rank", rank, () => {
           setRank(undefined);
           commit({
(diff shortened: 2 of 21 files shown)


packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabels.tsx
```diff
@@ -29,12 +32,29 @@ interface OutputLabelsProps {
...
+    rankOrder,
+    segmentId,
+    isFinalResult
+  } = props;
 
   return (
     <>
       {optype && OutputFieldLabel("OpType", optype)}
-      {targetField && OutputFieldLabel("TargetField", targetField)}
+      {targetFieldValidation.length > 0 ? (
```
manstis, line 50: What happens if it's not missing but either (a) references a non-existent `MininingField`, (b) does not need to be populated as there is only one `MiningField` with usage of `target` (but the `OutputField.targetField` is set in the PMML)?

packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabelsEditMode.tsx
```diff
@@ -76,13 +80,20 @@ export const OutputLabelsEditMode = (props: OutputLabelsEditModeProps) => {
             optype: undefined
           });
         })}
-      {targetField &&
+      {targetFieldValidation.length > 0 ? (
+        <ValidationIndicatorLabel validations={targetFieldValidation} cssClass="output-fields-list__item__label">
+          <strong>TargetField:</strong>&nbsp;
+          <em>Missing</em>
```
manstis, line 86: Same comment as above.

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
```diff
@@ -70,6 +73,117 @@ export const ScorecardReducer: HistoryAwareReducer<Scorecard, ScorecardActions>
...
+            action.payload.modelIndex,
+            state.Output?.OutputField,
+            state.MiningSchema.MiningField,
+            validation
+          );
+        }
+        if (isOutputsTargetFieldRequired(state.MiningSchema.MiningField)) {
+          const startIndex = state.Output?.OutputField.length ?? 0;
+          action.payload.outputFields.forEach((name, index) => {
+            validation.set(
```
manstis, line 176: We traditionally (by accident really!) have not added validation messages to the registry in reducers but delegated to the "validation" components. Did you want to retain the approach for consistency? Although I suspect at least some of this type of code (validation on actions) could change when we move to validating after commit. I'll leave the decision up to you (just let me know what you decide so I don't pick it up as a review topic again!).

manstis (commented):
I've put a couple of comments on the code.

I also observed the following unexpected behaviours:

- Disable Rank Order when Value entered? (Rank Textbox is already disabled)
- Delete entered Value from about. Rank Textbox does not re-enable. It should.

- Set more than one MF field as "target". Outputs light up. Good.
- Correct MFs to pick a target field. Warnings disappear. Good.
- Go back to MF and change field back to "active" (from "target").
- Outputs now reference a MF field that is not type "target". Would expect this to be an error?

- Output Label shows a Target field although the field has been changed back to "active". Should Reducer clear OF.targetField if a MF is changed to not be "target"? 
- The Outputs Properties Edit screen does not populate the Target Field dropdown. Correct. I suspect this is by chance as we filter MF's based on usage (of "target"). 
- The PMML still shows it (as the field changed back to "active").
- These errors could be fixed if the Reducer cleared OF.targetField when MF.usage changes.

I've attached some files to the JIRA that may help you reproduce these.

packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabels.tsx
kelvah: fixed now. if the reference is non-existent it's automatically deleted.
if the `targetField` is populated but not required, it's still kept. it could be useful if it becomes required again.

packages/pmml-editor/src/editor/components/Outputs/atoms/OutputLabelsEditMode.tsx
kelvah: Fixed as above. (The fix is in the scorecard reducer though)

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
kelvah: I agree with you. Validation should stay inside "validation" methods. I did it here directly in the action because it seemed the easiest way to do it without having to manually update the state with the new output fields.
As you pointed, this part will become useless and removed when moving the validation after commits, because we will validate the updated state.
If it's OK for you, I would leave it like that for now.

kelvah:
> I've put a couple of comments on the code.
> 
> [...]

Thank you @manstis. 

I fixed the issue with the disabled rank field (and now I am disabling rank order too).

I also added the logic that automatically cleans the `targetField` property if it contains the name of a mining schema field that got deleted or that is no more of type `target`.

manstis (approved):
LGTM. Thank-you @kelvah.

sjamboro (approved):

manstis:
Hi @tiagobento can we get this one in please?

packages/pmml-editor/src/editor/components/Outputs/organisms/OutputFieldExtendedProperties.tsx
```diff
@@ -224,9 +305,20 @@ export const OutputFieldExtendedProperties = (props: OutputFieldExtendedProperti
...
+
+const GenericSelectorEditor = (
+  id: string,
+  items: Array<string | GenericSelectorOption>,
+  selection: string,
+  onSelect: (_selection: string) => void,
+  isDisabled?: boolean
+) => {
+  return <GenericSelector id={id} items={items} selection={selection} onSelect={onSelect} isDisabled={isDisabled} />;
+};
```
tiagobento, line 324: Does that add something to `GenericSelector`? Why not use `GenericSelector` directly?

packages/pmml-editor/src/editor/components/Outputs/molecules/OutputFieldEditRow.tsx
```diff
@@ -126,7 +138,11 @@ const OutputFieldEditRow = (props: OutputFieldEditRowProps) => {
       eventTypes: ["click"]
     }
   );
-
+  const { service } = useValidationService();
```
tiagobento, line 141: Since `useValidationService` has a `service` property, maybe this Custom Hook should become `useValidation` and you could change this to

```suggestion
  const validationService = useValidation().service;
```

WDYT?

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
```diff
@@ -70,6 +73,160 @@ export const ScorecardReducer: HistoryAwareReducer<Scorecard, ScorecardActions>
...
+              });
+            }
+          });
+          validateOutputs(
+            action.payload.modelIndex,
+            state.Output?.OutputField,
+            state.MiningSchema.MiningField.filter((field, index) => index !== action.payload.miningSchemaIndex),
+            validation
+          );
+        }
```
tiagobento, line 106: Lots of `!` here. Is there an alternative? What happens if `draft` is undefined?

tiagobento (commented):
Left some minor comments inline. Please consider them!

packages/pmml-editor/src/editor/components/Outputs/molecules/OutputFieldEditRow.tsx
manstis: This has all changed in my PR #406.. 

It is now `ValidationRegistry` and the `string` literal paths changed to the builder I added.

I propose this PR be merged (once you're OK with it) and my #406 will then be updated to include the work in here.

I pasted a dependency chain of PMML PRs on `tooling` on Zulip and updated PR labels in GitHub.

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
manstis: This too is set to be refactored once all the existing "validation" PRs are merged.

Then I'll be doing https://issues.redhat.com/browse/FAI-384 to remove much of the filering etc shown above.

packages/pmml-editor/src/editor/components/Outputs/organisms/OutputFieldExtendedProperties.tsx
kelvah: `GenericSelectorEditor` precedes my work here. I just moved it to the bottom of the file to have `OutputFieldExtendedProperties` interface and declaration on top.

I agree we could get rid of it but I didn't want to refactor the whole component in this PR. We have a [JIRA](https://issues.redhat.com/browse/FAI-385) for replacing all `Select` usages with `FormSelect`, I think we could take the chance to do some cleanup also.

packages/pmml-editor/src/editor/components/Outputs/molecules/OutputFieldEditRow.tsx
kelvah: :+1:

tiagobento (approved):

PR #406, opened 2021-02-01
FAI-370: Score Cards: Validation: Removal of MiningField needs to update Attributes' predicates

manstis:
See https://issues.redhat.com/browse/FAI-370

Steps to test are covered in the JIRA.


packages/pmml-editor/src/editor/components/EditorScorecard/molecules/CharacteristicsTableRow.tsx
```diff
@@ -76,18 +91,35 @@ export const CharacteristicsTableRow = (props: CharacteristicsTableRowProps) =>
...
   return (
     <ul>
       {characteristic.Attribute.map((item, index) => (
         <li key={index}>
+          {validations(index).length > 0 && (
+            <span className="characteristic-list__item__label">
+              <ValidationIndicator validations={validations(index)} />
+            </span>
+          )}
           {CharacteristicPredicateLabel(toText(item.predicate, dataFields))}
```
kelvah: Are you avoiding `ValidationIndicatorLabel` because the predicate label could already have a tooltip in case of truncated text?
Maybe that's also the reason why you added `ValidationIndicator` at attribute level while not having one at characteristic level.

packages/pmml-editor/src/editor/reducers/MiningSchemaReducer.ts
```diff
@@ -55,7 +56,6 @@ export const MiningSchemaReducer: HistoryAwareValidatingReducer<MiningSchema, Al
           if (miningSchemaIndex >= 0 && miningSchemaIndex < draft.MiningField.length) {
             draft.MiningField.splice(miningSchemaIndex, 1);
           }
-          validation.clear(`models[${action.payload.modelIndex}].MiningSchema`);
```
kelvah, line 59: Why you are not clearing out the validation before re-validating?
Let's say the first mining field in the list has invalid properties and you delete it. The second mining field in the list becomes the first and it _"inherits"_ the validation info of the deleted one.

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
```diff
@@ -48,11 +62,73 @@ export type ScorecardActions = ActionMap<ScorecardPayload>[keyof ActionMap<Score
...
+        service.batch(state, `models[${action.payload.modelIndex}].Characteristics`, draft => {
+          validation.clear(`models[${action.payload.modelIndex}].Characteristics`);
+          validateCharacteristics(
+            action.payload.modelIndex,
+            state.Characteristics.Characteristic,
+            state.MiningSchema.MiningField.filter(mf => mf.name !== action.payload.name),
+            validation
+          );
+        });
+        break;
```
kelvah: I suppose you have to validate characteristics also when adding a new mining field.
If I delete a MF, the attributes referring to it get invalid. If I manually add it again the attributes are still invalid.

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
```diff
@@ -70,8 +146,66 @@ export const ScorecardReducer: HistoryAwareReducer<Scorecard, ScorecardActions>
...
+  service: HistoryService
+) => {
+  if (predicate === undefined) {
+    return;
+  } else if (predicate instanceof True) {
+    return;
+  } else if (predicate instanceof False) {
+    return;
+  } else if (predicate instanceof SimpleSetPredicate) {
+    const ssp: SimpleSetPredicate = predicate as SimpleSetPredicate;
```
kelvah: just a minor detail, you are already checking the predicate type in the if condition, so the following type assertions are redundant. of course they do no harm, so if you prefer to have them I see no issue.
it's something that could be covered by eslint rules eventually.

kelvah (commented):
Hi @manstis, I left you some comments.

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
manstis: Great. Thank-you. Fixed elsewhere too. Java would need the cast... I have an excuse!

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
manstis: Fixed thank-you.

packages/pmml-editor/src/editor/reducers/MiningSchemaReducer.ts
manstis: I am now clearing the validation registry for `MiningSchema` in the `ScorecardReducer` where I also handle the `AddMiningSchemaFields` and `DeleteMiningSchemaField` actions. I potentially add validation entries in the `ScorecardReducer` if the field used in a predicate does not exist in the `MiningSchema`. Clearing the same path here would remove those entries. The `ScoecardReducer` will always be called before the `MiningSchemaReducer` and hence I do not need to clear here too.

manstis: Hmm.. but on testing something is not quite right. I will investigate further.

manstis: LOL. OK, they clear different paths. Doh. I've added this back.

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/CharacteristicsTableRow.tsx
manstis: Updated. Thank-you.

manstis: Screenshot from 2021-02-02 15-59-42
Screenshot from 2021-02-02 15-59-44

manstis:
@kelvah PR updated. Thank-you.

kelvah (approved):
:ok_hand:

sjamboro (approved):

manstis:
Hi @tiagobento. Could you please try to take a look soon.

We have a number of other PRs and JIRAs that depend on this.

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryHandler/DataDictionaryHandler.tsx
```diff
@@ -56,12 +56,14 @@ const DataDictionaryHandler = () => {
   };
 
   const deleteField = (index: number) => {
-    dispatch({
-      type: Actions.DeleteDataDictionaryField,
-      payload: {
-        index
-      }
-    });
+    if (window.confirm(`Delete Output "${dictionary[index].name}"?`)) {
```
tiagobento, line 59: So this is a bit dangerous... I know `window.confirm` provides a nice and synchronous way to display a "modal" to users, but often those become more complex and `window.confirm` can't handle the complexity of these "modals". I would advise to not use those, since they tend to be used in places where the presentation logic is not present. Like it seems like this is the case. I'm not sure how you could overcome using `window.confirm` on this scenario, but if it's possible to not use that, I think it would be beneficial to the project on the mid/long-run.

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributesTableRow.tsx
```diff
@@ -14,27 +14,52 @@
...
+    miningFields,
+    onEdit,
+    onDelete
+  } = props;
+
+  const { service } = useValidationService();
+  const validations = useMemo(
+    () =>
+      service.get(
+        `models[${modelIndex}].Characteristics.Characteristic[${characteristicIndex}].Attribute[${attributeIndex}].Predicate`
```
tiagobento: Are these strings typed in any way? What happens if the model changes? Will these break? Or simply return an empty validation list? I think throwing an exception if this string is not 'correct' would be the best approach.

packages/pmml-editor/src/editor/reducers/AttributesReducer.ts
```diff
@@ -46,16 +46,17 @@ interface AttributesPayload {
...
   };
 }
 
 export type AttributesActions = ActionMap<AttributesPayload>[keyof ActionMap<AttributesPayload>];
 
-export const AttributesReducer: HistoryAwareReducer<Attribute[], AllActions> = (
-  service: HistoryService
+export const AttributesReducer: HistoryAwareValidatingReducer<Attribute[], AllActions> = (
+  service: HistoryService,
+  validation: ValidationService
```
tiagobento: Names are a bit generic here. Could these be changed to `history` and `validation` or even `historyService` and `validationService`?

packages/pmml-editor/src/editor/reducers/CharacteristicReducer.ts
```diff
@@ -35,10 +36,11 @@ interface CharacteristicPayload {
 
 export type CharacteristicActions = ActionMap<CharacteristicPayload>[keyof ActionMap<CharacteristicPayload>];
 
-export const CharacteristicReducer: HistoryAwareReducer<Characteristic[], AllActions> = (
-  service: HistoryService
+export const CharacteristicReducer: HistoryAwareValidatingReducer<Characteristic[], AllActions> = (
+  service: HistoryService,
+  validation: ValidationService
```
tiagobento: Same here.

packages/pmml-editor/src/editor/reducers/MiningSchemaReducer.ts
```diff
@@ -28,6 +28,7 @@ interface MiningSchemaPayload {
   [Actions.DeleteMiningSchemaField]: {
     readonly modelIndex: number;
     readonly miningSchemaIndex: number;
+    readonly name: FieldName | undefined;
```
tiagobento:
```suggestion
    readonly name?: FieldName;
```

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
```diff
@@ -48,11 +63,85 @@ export type ScorecardActions = ActionMap<ScorecardPayload>[keyof ActionMap<Score
...
+                  Attribute: attributes
+                };
+              }
+              return characteristic;
+            }),
+            state.MiningSchema.MiningField,
+            validation
+          );
+        });
+        break;
```
tiagobento: Don't wanna be picky here, but this code looks like too much for me 😝 

I'm not experienced in using reducers and all, but this method is very difficult to read at first. Maybe extract the mappings to a variable? Not sure what to do, just saying what my impression was.

tiagobento (commented):
Left some comments inline. Please consider them!

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryHandler/DataDictionaryHandler.tsx
manstis: Hello.. this is biting us already (see my VSCode integration PR, where `window.confirm(..)` is blocked by VSCode. I will refactor all use of `window.confirm(..)` and `window.alert(..)` (if any) to PatternFly's `Modal` as part of the integration work. Is that OK with you?

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributesTableRow.tsx
manstis: Hello, these strings are effectively `xpath`'s into the model.

The model is an OMG standard and hence, for the version of the standard we support, not going to change.

If the path is wrong then we'll get zero validation entries (or worst case validation entries that don't relate to the path). We only store validation records where part of the model is invalid and we store them effectively in a shallow clone of the model itself; i.e. the structure is identical but only holds validation records. 

I could possibly validate the path against the _real_ model and throw an exception if it does not exist. We could also move them from repeated literals to a library of functions that build the string based on parameters. WDYT? Can this also be done as part of a new JIRA?

packages/pmml-editor/src/editor/reducers/AttributesReducer.ts
manstis: Done throughout the codebase.

packages/pmml-editor/src/editor/reducers/CharacteristicReducer.ts
manstis: Done throughout the codebase.

packages/pmml-editor/src/editor/reducers/MiningSchemaReducer.ts
manstis: Done.

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
manstis: What!?!? It's _lovely_code!

Nah, it's all a bit contrived. Due to an unforeseen problem with our V&V design we need to "pre-apply" changes to the model when validating different sub-trees. We have https://issues.redhat.com/browse/FAI-384 for me to refactor all of this after the current PRs adding the final bits of V&V are merged (@kelvah and myself decided to hit it all in one refactor once all validation was "in" as we'd gone 75% of the way; rather than try to refactor in the middle of it all).

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryHandler/DataDictionaryHandler.tsx
tiagobento: Yes, absolutely! :)

manstis:
@tiagobento Updates made and comments added.

packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts
tiagobento: Got it. Thanks for the explanation. Glad to know you're aware of that and already have plans to refactor it :)

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributesTableRow.tsx
tiagobento: For sure this can be done separately. I'm not sure what is the best approach, though. Static typing is also a very important matter for me, but I understand that there are limitations to what compilers can type-check :)

I'm not a big fan of removing legibility too, and these look very simple. Maybe this runtime-checking of this string actually hits and existing part of the model is enough and we combine it with unit-tests at _some point in the future_?

manstis: I had a quick look around @tiagobento and passing the _real_ model to the validation service (just) to check the paths are valid seems an overkill. The service is used a lot and everywhere it is invoked would need to also provide the _real_ model. It doesn't smell right. I'd rather prefer to move the magic strings to a suite of builder functions that take explicit parameters for the various indexed items and return a type representing the path. WDYT?

@kelvah WDYT too?

manstis: I'm working on replacing the `string` literals with this type of thing:
```
Builder()
  .forModel(modelIndex)
  .forMiningSchema()
  .forMiningField(miningFieldIndex)
  .forImportance()
  .build()),
```
Where `Builder` is erm, a builder pattern, that returns a _type_ used for the paths.

Provided the builders are correct the risk of incorrect paths is minimised.

@kelvah I can also reuse this for the paths used in the `HistoryService`...

kelvah: Hi @manstis, I like your idea for the builder. Honestly I didn't have a big issue with the string literal approach but I guess the builder could help. I would handle it in a new JIRA like you said, I see it like an improvement.

tiagobento: @manstis Understand. Don't know if that makes sense but to me it's a bit odd that you're validating something that you is not "at hand" at the moment of validation. But again, I'm not familiar with the code, so can't argue about that and trust your feeling that it's too much of an overkill to use the actual object to check the validation request.

About the builder pattern: it can be helpful, but it can also be a pain to maintain, if it is hand-written. Not sure how to approach that actually. Honestly, I'm beginning to think that using the plain strings is not a big point of concern.

On the other hand, though, there's one thing I *know* is possible is to do. We could add type-checking with the property names directly from the model interface. Such "builder" would look like as below.

For the XPath:
`models[${modelIndex}].Characteristics.Characteristic[${characteristicIndex}].Attribute[${attributeIndex}].Predicate`

We would have:
```
new XPathBuilder<MyModel>()
    .inList("models", modelIndex)
    .inList("Characteristics", characteristicIndex)
    .at("Characteristic")
    .inList("Attribute", attributeIndex)
    .at("Predicate")
    .build();
```

`inList` and `at` would all be generic methods with the highest level of TypeScript wizardry. We have something similar with the Envelope/Channel communication APIs.

The advantage would be that this is generic enough to handle any type/interface and it's also type-checked. Disadvantage is that I don't know if something like `MyModel` exists at the codebase and also it might be hard to get right.

I know this is starting to sound like too much, and maybe it is, but let me know what you think! :)

manstis:
@tiagobento @kelvah There are two new commits on this PR:-

1. Change to the builder I proposed for paths used for validation.
2. Rename and refactor some classes used for validation to better reflect their use.

The old `ValidationService` did not perform any validation; it is just a registry of validation entries stored in a sparse copy of the real model (i.e same structure, but content is just the validation entries). The old `ValidationService` has been renamed `ValidationRegistry` and the old `ValidationRegistry` has been renamed `ValidationStore`.

@tiagobento I was 90% of the way through the hand-written builders. It might be nice to move to the form you proposed but I'll just log as a JIRA for future consideration (we don't have validation on every array and property in the whole model; indeed we even have some "keys" that don't exist in the model.. which could be problematic for your proposal).

@kelvah I'll also make a JIRA to move the paths used by the reducers/history to use the builders (rename and move to something more generic and add whatever is missing for the reducers). Then we should have zero magic strings scattered throughout the code, which I think will make it a nicer place to be.

Please consider the last two commits and (hopefully) approve.

manstis:
@tiagobento @kelvah This PR has been updated following merge of @kelvah's #405.

Other than the move to using a "path builder" and rename of the validation related classes nothing else should have changed.

@kelvah The `ScorecardReducer` was a mess; so I'd appreciate it if you could double check your `Output` work is still OK.
  
Could you please re-review.

kelvah (approved):
Hi @manstis, all the validation work on `Output` is working as expected.
Everything else looks good to me. Thank you.

manstis:
@tiagobento Are you OK with this being (approved) and merged w/o the green FDB?

tiagobento:
@manstis Yes

tiagobento (approved):

PR #415, opened 2021-02-09
FAI-397: Score Cards: Support deletion of Attribute labels

manstis:
See https://issues.redhat.com/browse/FAI-397

### DIFF
--- a/packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributeLabel.tsx
+++ b/packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributeLabel.tsx
@@ -18,12 +18,31 @@ import { Label } from "@patternfly/react-core";
 
 import "./AttributeLabel.scss";
 
-export const AttributeLabel = (name: string, value: any) => {
+export const AttributeLabel = (name: string, value: any, onClose?: () => void) => {
   return (
-    <Label color="cyan" className="attribute-list__item__label">
-      <strong>{name}:</strong>
-      &nbsp;
-      <span>{value}</span>
-    </Label>
+    <>
+      {!onClose && (
+        <Label color="cyan" className="attribute-list__item__label">
+          <strong>{name}:</strong>
+          &nbsp;
+          <span>{value}</span>
+        </Label>
+      )}
+      {onClose && (
+        <Label
+          color="cyan"
+          className="attribute-list__item__label"
+          onClose={e => {
+            e.nativeEvent.stopImmediatePropagation();
+            e.stopPropagation();
+            onClose();
+          }}
+        >
+          <strong>{name}:</strong>
+          &nbsp;
+          <span>{value}</span>
+        </Label>
+      )}
+    </>
   );
 };
--- a/packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributeLabel.tsx
+++ b/packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributeLabel.tsx
@@ -18,7 +18,15 @@ import { Label } from "@patternfly/react-core";
 
 import "./AttributeLabel.scss";
 
-export const AttributeLabel = (name: string, value: any, onClose?: () => void) => {
+interface AttributeLabelProps {
+  name: string;
+  value: any;
+  onClose?: () => void;
+}
+
+export const AttributeLabel = (props: AttributeLabelProps) => {
+  const { name, value, onClose } = props;
+
   return (
     <>
       {!onClose && (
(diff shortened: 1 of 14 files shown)


manstis:
@kelvah @tiagobento @sjamboro This PR has finally become available for review...

After this one I should be able to work on moving the _inlined_ functions to `useCallbacks` and the Validation refactor :-)

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributeEditor.tsx
```diff
@@ -145,12 +141,12 @@ export const AttributeEditor = (props: AttributeEditorProps) => {
     const _text = toText(attribute.predicate, dataFields);
     setText({
       value: _text,
-      valid: true
+      valid: validateText(_text)
```
kelvah: I would keep the [predicate validation related to this flag](https://github.com/kiegroup/kogito-tooling/blob/20eaba43153fc2d4a096dca156da87c0b0c716f8/packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributeEditor.tsx#L178) as `warning` to keep consistency. Until we will eventually switch everything to `error`.

packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributeLabels.tsx
```diff
@@ -26,8 +26,10 @@ interface AttributeLabelsProps {
   characteristicIndex: number;
   activeAttributeIndex: number;
   activeAttribute: Attribute;
+  attributes: Attribute[];
```
kelvah: Why did you add `attributes` to props? I see it's not used.

manstis: Yes, why did I?!?! It is now removed.

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributeEditor.tsx
manstis: I have made the change. Unfortunately PF does not show the icon for warnings so I added an `Alert` too.

Screenshot from 2021-02-24 11-56-49

Also, for consistency, I added support to delete the predicate attribute.

Screenshot from 2021-02-24 11-56-41

Finally the predicate validation is now performed the same as everywhere else; i.e. using `ValidationRegistry` (it _was_ being validated this way, but the UI component showed the validation issue the "old" way so I replaced it).

packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributeLabel.tsx
```diff
@@ -18,12 +18,31 @@ import { Label } from "@patternfly/react-core";
 
 import "./AttributeLabel.scss";
 
-export const AttributeLabel = (name: string, value: any) => {
+export const AttributeLabel = (name: string, value: any, onClose?: () => void) => {
```
kelvah: This should probably be a standard functional component instead of a function. We also have many specific labels around that could be replaced by a generic label component at some point. So It may not be worth to fix just this one if it's not going to last. I think we could just add the generic label to the improvements to-do list and take care of all of them later.

packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributeLabels.tsx
kelvah: :+1:

kelvah (approved):

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributeEditor.tsx
kelvah: > I have made the change. Unfortunately PF does not show the icon for warnings so I added an Alert too.

You are right. I forgot that. I am not sure if the Alert does its job here, maybe because it's not really clear that the item needing attention is the predicate editor. Given it's not a PF component it doesn't have apparent validation styles. I would probably add manually the validation message at the bottom. What you think?

manstis:
@kelvah FYI updates completed:

Screenshot from 2021-02-25 09-46-40

sjamboro (approved):

manstis:
Hi @tiagobento could you please review. Thank-you.

manstis:
Hi @tiagobento this is just needing your review please.

tiagobento (approved):

PR #417, opened 2021-02-09
KOGITO-4399: Validation errors does not affect save anymore

adrielparedes:
**JIRA**: https://issues.redhat.com/browse/KOGITO-4399

A try/catch was added to the validation part not to break save.

* https://github.com/kiegroup/appformer/pull/1101
* https://github.com/kiegroup/kogito-tooling/pull/417

### DIFF
--- a/packages/vscode-extension/src/KogitoEditableDocument.ts
+++ b/packages/vscode-extension/src/KogitoEditableDocument.ts
@@ -73,12 +73,16 @@ export class KogitoEditableDocument implements CustomDocument {
     try {
       const editor = this.editorStore.get(this.uri);
       if (!editor) {
-        console.error(`Cannot save because there's no open Editor for ${this.uri.fsPath}`);
+        this.vsCodeLogger.error(`Cannot save because there's no open Editor for ${this.uri.fsPath}`);
         return;
       }
 
-      const notifications = await editor.validate();
-      this.vsCodeNotifications.setNotifications(destination.fsPath, notifications);
+      try {
+        const notifications = await editor.validate();
+        this.vsCodeNotifications.setNotifications(destination.fsPath, notifications);
+      } catch (e) {
+        this.vsCodeLogger.warn(`File was not validated: ${e}`);
+      }
 
       const content = await editor.getContent();
       if (cancellation.isCancellationRequested) {
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ Develop
 > **Example:**
 >
 > $ export EXTERNAL_RESOURCE_PATH__bpmnEditor=/Users/tiago/redhat/kie-wb-common/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-kogito-runtime/target/kie-wb-common-stunner-bpmn-kogito-runtime
+> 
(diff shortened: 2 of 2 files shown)


tiagobento (approved):

ederign (approved):

jomarko (commented):
Can not download vsix from github actions - zero bytes size.

Built locally rebased with master however still can not save file chages.

packages/vscode-extension/src/KogitoEditableDocument.ts
```diff
@@ -73,12 +73,16 @@ export class KogitoEditableDocument implements CustomDocument {
...
       }
 
-      const notifications = await editor.validate();
-      this.vsCodeNotifications.setNotifications(destination.fsPath, notifications);
+      try {
+        const notifications = await editor.validate();
+        this.vsCodeNotifications.setNotifications(destination.fsPath, notifications);
+      } catch (e) {
+        this.vsCodeLogger.warn(`File was not validated: ${e}`);
+      }
```
jomarko, line 85: Could you please clarify when user can see effect of this change? My understanding was it is if user tries to save an invalid DMN file, then an message appears in problems view, but I had no luck with this approach.

karreiro (commented):
Thank you, @adrielparedes. 

I've tried this PR locally and I couldn't save the asset too:

2021-02-16 12 18 23 copy

Maybe I'm missed something during the artifact generation.. so, if you link a `.vsix` plugin, I can retest it.

Thanks!

adrielparedes:
@karreiro @jomarko I will check what is happening here, but just to be sure, did you also compiled appformer? Did you exported DMN Editor variable so kogito-tooling can find the compiled editor instead of nightly build?

karreiro:
@adrielparedes I built:
- your branch on **appformer**
- DMN and BPMN web apps on **kie-wb-common** (on 7.50.0-SNAPSHOT)
- and then **kogito-tooling** (exporting `EXTERNAL_RESOURCE_PATH__bpmnEditor` and `EXTERNAL_RESOURCE_PATH__dmnEditor`)

adrielparedes:
@karreiro that's right, I will check it again, thanks for the confirmation!

karreiro (approved):
Thank you, @adrielparedes. I raised a separate JIRA for the issue (https://issues.redhat.com/browse/KOGITO-4458) I mentioned, as it's already present on Marketplace:

jomarko (commented):
@adrielparedes thank you for vsix provided via email. And explanation there is one more separate issue for bpmn. I linked the jiras.

I can confirm the fix works. But there is still my non answered question from my previous review https://github.com/kiegroup/kogito-tooling/pull/417#discussion_r576440671

Could you please reply? From my point of view it is important to be sure we have all same understanding. As said in comment, right now I would expect some validation messages to appear somewhere in VS Code also for dmn.

packages/vscode-extension/src/KogitoEditableDocument.ts
adrielparedes: Sorry I didn't answer before. Yes, you are right, those messages appear when the user saves a Diagram, but right now you can't see them because those validation methods should be implemented in editor side. Since I'm not the expert I just copied what It used to be the BPMN validation, but for sure Roger will need to improve what I wrote. The tests I did and the video I have is for  mock notifications I used to check the validation method. I can share it with you if you want.

ederign: Just adding to that, DMN/BPMN validation will be planned/prioritized for the next sprints/cycles (some of them require backend, so not a quick thing to do)

PR #421, opened 2021-02-15
FAI-348: Score Cards: When characteristic filter matches no rows, filter text is cleared

manstis:
See https://issues.redhat.com/browse/FAI-348

@sjamboro This can be tested as follows:-
- Open a Scorecard with Characteristics
- Enter a value in the Characteristics filter `TextBox` and press enter/click magnifying glass icon
- The view should show the filtered Characteristics. The Filter text remains as entered.
- Change the filter so that no Characteristics match
- The view should show some text about nothing matching. The Filter text remains as entered.

The last step was previously the bug. The Filter text would become empty.

### DIFF
--- a/packages/pmml-editor/src/editor/components/EditorScorecard/molecules/CharacteristicsToolbar.tsx
+++ b/packages/pmml-editor/src/editor/components/EditorScorecard/molecules/CharacteristicsToolbar.tsx
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 import * as React from "react";
-import { useState } from "react";
 import {
   Button,
   ButtonVariant,
@@ -32,14 +31,14 @@ import { SearchIcon } from "@patternfly/react-icons";
 import "./CharacteristicsToolbar.scss";
 
 interface CharacteristicsToolbarProps {
-  onFilter: (filter: string) => void;
+  filter: string;
+  setFilter: (filter: string) => void;
+  onFilter: () => void;
   onAddCharacteristic: () => void;
 }
 
 export const CharacteristicsToolbar = (props: CharacteristicsToolbarProps) => {
-  const { onFilter, onAddCharacteristic } = props;
-
-  const [filter, setFilter] = useState("");
+  const { filter, setFilter, onFilter, onAddCharacteristic } = props;
 
   return (
     <Toolbar id="characteristics-toolbar" data-testid="characteristics-toolbar">
@@ -74,7 +73,7 @@ export const CharacteristicsToolbar = (props: CharacteristicsToolbarProps) => {
                       data-testid="characteristics-toolbar__submit"
                       variant={ButtonVariant.control}
                       aria-label="filter button for filter input"
-                      onClick={() => onFilter(filter)}
+                      onClick={() => onFilter()}
                     >
                       <SearchIcon />
                     </Button>
--- a/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
+++ b/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
@@ -43,7 +43,6 @@ interface CharacteristicsContainerProps {
   isBaselineScoreRequired: boolean;
   characteristics: Characteristic[];
   filteredCharacteristics: IndexedCharacteristic[];
-  filter: string;
   onFilter: (filter: string) => void;
   deleteCharacteristic: (index: number) => void;
   commit: (index: number | undefined, characteristic: Characteristic) => void;
@@ -58,7 +57,6 @@ export const CharacteristicsContainer = (props: CharacteristicsContainerProps) =
     isBaselineScoreRequired,
     characteristics,
     filteredCharacteristics,
-    filter,
     onFilter,
     deleteCharacteristic,
     commit
@@ -68,6 +66,7 @@ export const CharacteristicsContainer = (props: CharacteristicsContainerProps) =
   const { service, getCurrentState } = useHistoryService();
   const dispatch = useBatchDispatch(service, getCurrentState);
 
+  const [filter, setFilter] = useState("");
   const [selectedCharacteristicIndex, setSelectedCharacteristicIndex] = useState<number | undefined>(undefined);
   const [selectedAttributeIndex, setSelectedAttributeIndex] = useState<number | undefined>(undefined);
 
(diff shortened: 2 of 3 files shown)


manstis:
Green!

kelvah (approved):
Hi @manstis, thank you, approving and leaving you some observations.

- it could be probably nice to trim  `filter` string on change or on filter, to prevent searching an empty space inadvertently.
- do you think it's worth clearing `filter` after clicking on `add characteristic`? I reckon it should be another jira eventually
- I wasn't expecting to find the filtering logic inside `ScorecardEditorPage.tsx`. it felt more related to the characteristic container. Is there any reason why you placed it high in the hierarchy? I'm hair-splitting of course :monocle_face:

manstis:
@kelvah For the record I've moved Characteristic filtering to the correct place and performed the other tidy-up suggested.

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
```diff
@@ -113,16 +127,39 @@ export const CharacteristicsContainer = (props: CharacteristicsContainerProps) =
...
       setActiveOperation(Operation.UPDATE_CHARACTERISTIC);
+
+      //Clear filter to ensure new Characteristic is visible
+      doSetFilter("");
     }
   }, [characteristics]);
 
+  const deleteCharacteristic = useCallback(
+    (characteristicIndex: number) => {
+      if (window.confirm(`Delete Characteristic "${characteristics?.[characteristicIndex].name}"?`)) {
```
tiagobento, line 150: Won't this break on VS Code?

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
```diff
@@ -28,50 +28,64 @@ import {
...
+      return scorecard.Characteristics.Characteristic;
+    }
+    return [];
+  });
+
+  useEffect(() => onFilter(), [modelIndex, characteristics]);
+
+  const doSetFilter = (_filter: string) => {
+    setFilter(_filter.toLowerCase().trim());
+  };
```
tiagobento: I feel like `setFilter` and `doSetFilter` are the same thing. Probably renaming this to `setLowercaseTrimmedFilter` would be better?

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
```diff
@@ -28,50 +28,64 @@ import {
...
-    commit
-  } = props;
+  const { modelIndex, areReasonCodesUsed, isBaselineScoreRequired } = props;
 
   const { setActiveOperation } = useOperation();
   const { service, getCurrentState } = useHistoryService();
   const dispatch = useBatchDispatch(service, getCurrentState);
 
+  const [filter, setFilter] = useState("");
+  const [filteredCharacteristics, setFilteredCharacteristics] = useState<IndexedCharacteristic[]>([]);
```
tiagobento, line 57: From my understanding, `filteredCharacteristics` is directly derivable from `filter`. This usually means that `filteredCharacteristics` does not need to be a state, but rather a memo. Is that the case?

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/CharacteristicsToolbar.tsx
```diff
@@ -74,7 +73,7 @@ export const CharacteristicsToolbar = (props: CharacteristicsToolbarProps) => {
                       data-testid="characteristics-toolbar__submit"
                       variant={ButtonVariant.control}
                       aria-label="filter button for filter input"
-                      onClick={() => onFilter(filter)}
+                      onClick={() => onFilter()}
```
tiagobento, line 76: No need to create a new lambda expression here. You can use `onFilter` directly. The same applies for changes below.
```suggestion
                      onClick={onFilter}
```

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
manstis: Yes but I have only removed use of `window.confirm(..)` on the VSCode related PR.

I prefer to keep their removal there for now. This PR will cause a merge conflict on the VSCode one so I'll fix it then.

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
manstis: Renamed to `setLowercaseTrimmedFilter` as proposed. I also renamed `onFilter` to `applyFilter`.

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
manstis: Yes it is; however we only apply the filter when the User presses `<enter>` or clicks the filter icon. Therefore whilst `filter` holds the text for the filter criteria it is not applied until a subsequent action. I therefore cannot have `filteredCharacteristics` a memorised value dependent on the `filter` as it would be applied on each character entered in the filter text box. There's various nasty hacks I could do to make it work as you're proposing but I think it'll be less clear than it currently is.

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/CharacteristicsToolbar.tsx
manstis: Corrected (in both places).

manstis:
@tiagobento Questions answered and updates made in new commit.

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CharacteristicsContainer.tsx
tiagobento: Ok!

manstis:
Hi @sjamboro could you please try to review some of the PRs?

@tiagobento I assume you're OK with this PR now? Could you please approve it.

sjamboro (approved):

tiagobento (approved):

PR #423, opened 2021-02-17
FAI-414: Score Cards: Characteristics container should scroll

manstis:
See https://issues.redhat.com/browse/FAI-414

After much fiddling around this is the only solution I found to work. The height can be any value. Absolutely any.

I'd love to know the correct way.. so if you get both the time and urge to look properly it'd be appreciated.

PR #424, opened 2021-02-17
FAI-415: Score Cards: Max width for editor

manstis:
See https://issues.redhat.com/browse/FAI-415


manstis:
@kelvah I tried having the two different sections, as we discussed, (Model Title/_buttons_ and Model Setup/Characteristics) but the whole editor is positioned relatively and hence both sections are effectively floating over their absolute parent meaning they're both getting a grey background anyway. So this PR just has the single container (and sets the margins - when max width is met - to a nice dark grey that I'm sure you'll hate!)

kelvah:
@manstis it's not that bad after all :smile:. You're right, it cannot be done with two separate sections. If you want to give it a try though you'll have to wrap the content of each `PageSection` inside `ScorecardEditorPage.tsx` with an `.editor-page` container.

manstis:
Thank-you @kelvah for the suggestion.

I've included a new commit that allows the background colour for the margins to be set correctly.

Screenshot from 2021-02-18 12-02-49

The screenshot was taken with me setting a `max-width` to suit my monitor's resolution.

I've also re-tested with #423 and it continues to work fine (scrollbar and this `max-width`).

kelvah (approved):

packages/pmml-editor/src/editor/components/EditorScorecard/templates/ScorecardEditorPage.tsx
```diff
@@ -92,134 +92,142 @@ export const ScorecardEditorPage = (props: ScorecardEditorPageProps) => {
...
+                          payload: {
+                            modelIndex: modelIndex,
+                            characteristicIndex: _index,
+                            name: _characteristic.name,
+                            reasonCode: _characteristic.reasonCode,
+                            baselineScore: _characteristic.baselineScore
+                          }
+                        });
+                      }
+                    }}
```
tiagobento: I know this is not something you added on this PR, but I believe there is too much logic here. Extracting this prop to a useCallback or a plain variable should improve this component's readability. The DOM structure should ideally not be mixed with heavy logic.

manstis: Hi @tiagobento I plan on reviewing the whole codebase and moving the inlined callbacks under https://issues.redhat.com/browse/FAI-382.

tiagobento: Ok!

manstis:
Hi @sjamboro could you please try to review some of the PRs?

@tiagobento I assume you're OK with this PR now? Could you please approve it.

tiagobento (approved):

sjamboro (approved):

PR #432, opened 2021-02-25
FAI-406: Score Cards: Model Name: Improvements: Edit cues and validation

manstis:
See https://issues.redhat.com/browse/FAI-406


kelvah:
Hi @manstis, I'll give you my opinion before reviewing the code. I know you'll hate me a bit :speak_no_evil:

I think the box-style hover here doesn't fit very well. The model name is not a box, so surrounding it with one on hover feels strange to me. I would try something more subtle in this case. Like a light border that hints at an input field. 

Something like:
Screenshot from 2021-02-25 18-20-20

Not sure blue is the right choice here but it could be a start. Maybe a darker variant.
What you think? :thinking: 

I would also try to make the model name font while editing as similar as possible to the one while non-editing, same size, same weight. Not having a "jump" between the two sizes may help understand what's happening, that's still the same string.

Finally I would do something to the input field width. It feels too long. I understand the intention to cover also very long names but we could assume they are an exception for now. I would try to put a max-width on it of about 50%, that could be increased at lower resolutions with a media query.

Let me know what you think or if you want to discuss it more elsewhere.

sjamboro:
@manstis
I am confused as an user, now. It is not clear that this warning represents "Model Name". The warning "Model Name will be generated at runtime." can be interpreted as automatically generated model name (for example: based on filename). 
 
I looked at google docs and it uses gray title "Untitled document", after insert a valid name, the title turns black. Can you add a title `<not set>` or  `<no model name>` above warning? `<Undefined>` says nothing it can be any attribute. Missing link between the file name and PMML editor, missing function name, characteristic name, etc. 
image

I not sure, if Model Name should be represented separately or in Model Setup Box. Based on XML structure. Model name is an attribute of Scorecard element same as "isScorable", "functionName", etc.

This is my feels so be free to decide if you want implement something from this comment

sjamboro:
@manstis
Steps to reproduce:

- open New file
- Warning is present
       -> check PMML modelName is missing
- Click on the warning to edit it, cancel edit mode without change
       -> check PMML: `modelName="<Undefined>"`

If I do not change value, the PMML code should stay same. Isn't it?

manstis:
@kelvah @sjamboro Fixes and changes made.

Following discussion with @kelvah this is the expected behaviour:-
- Authoring a new file adds a default model name.
- Opening a file with no model name (attached to JIRA) shows the warning about runtime generation.
- During editing of the name, if it is invalid, attempts to commit it (i.e. click outside) rollback to the last valid name.
- Pressing `<Escape>` cancels the edit.
- Hover-over styling changed.
- Editing/viewing font sizes and positioning are consistent.

Hopefully this is "close enough" to resolve the JIRA topic.

Subsequent refinement can be the result of feedback and additional JIRA.

@tiagobento One for you too. Thank-you.

sjamboro:
@manstis
You wrote:
- Opening a file with no model name (attached to JIRA) shows the warning about runtime generation.

I can not reach this state. For example steps to reproduce:
1. delete current name "Untitled model" -> the warning is present
2. confirm it
There is the old name.

I just want to check that you expect this behaviour. Otherwise I can open new JIRA.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.tsx
```diff
@@ -60,16 +60,18 @@ export const ModelCard = (props: ModelCardProps) => {
     [index]
   );
 
+  const _modelName = useMemo(() => (modelName === "" ? "<Undefined>" : modelName), [modelName]);
```
sjamboro: Do you use `"<Undefined>"` or `undefined`

manstis:
> @manstis
> You wrote:
> 
> * Opening a file with no model name (attached to JIRA) shows the warning about runtime generation.
> 
> I can not reach this state. For example steps to reproduce:
> 
> 1. delete current name "Untitled model" -> the warning is present
> 2. confirm it
>    There is the old name.
> 
> I just want to check that you expect this behaviour. Otherwise I can open new JIRA.

This is by design.

Our editor prevents the entry of an empty model name. The previous valid state is restored (which is the same behaviour for Characteristics, Data Fields and Output Fields). `Untitled model` is the default name when you create a NEW scorecard. If you open the file attached to the JIRA it has no name defined and the editor will show the warning even when not editing.

packages/pmml-editor/src/editor/components/LandingPage/molecules/ModelCard.tsx
manstis: I want to use `<Undefined>` 

This is the string shown to Users when a model has no name.

That said it is inconsistent with what is shown in the Scorecard Editor so I have added a new commit making them equal.

manstis: I have added two files to the JIRA.

One with a Scorecard with no model name set (to test the Scorecard editor), another with multiple models (to test the "Landing Page" shown when the file contains multiple models and the subject of this conversation regarding `ModelCard`).

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.tsx
```diff
@@ -14,108 +14,167 @@
...
-            autoFocus={true}
-            value={modelName.value}
-            validated={modelName.valid ? "default" : "error"}
-            onChange={e => setModelName({ value: e, valid: e.length > 0 })}
-            onBlur={e => {
-              if (modelName.valid) {
-                onCommit();
-              }
+      <Stack hasGutter={true} className={"modelTitle--full-width"}>
+        <StackItem>
```
kelvah: This `Stack` has only one `StackItem` so it can probably be replaced by just a `div`

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.scss
```diff
@@ -14,13 +14,72 @@
...
+  border-top: 1px solid #ffffff;
+  border-bottom: 1px solid #ffffff;
+  transition: margin-bottom 0.15s var(--sizeEasing), border-left 0.12s ease-in, border-right 0.12s ease-in, border-top 0.12s ease-in, border-bottom 0.12s ease-in;
+
+  &:not(.modelTitle--editing):hover {
+    border-left: 1px solid var(--pf-global--BorderColor--100);
+    border-right: 1px solid var(--pf-global--BorderColor--100);
+    border-top: 1px solid var(--pf-global--BorderColor--100);
+    border-bottom: 1px solid var(--pf-global--BorderColor--100);
+  }
```
kelvah: Shortening things: 
```suggestion
  &:not(&--editing):hover {
    border: 1px solid var(--pf-global--BorderColor--100);
  }
```

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.scss
```diff
@@ -14,13 +14,72 @@
...
+  margin-top: 1px;
+  padding-left: 8px;
+  padding-right: 8px;
+
+  // Ensure the border transition does not flicker with default black borders
+  border-left: 1px solid #ffffff;
+  border-right: 1px solid #ffffff;
+  border-top: 1px solid #ffffff;
+  border-bottom: 1px solid #ffffff;
+  transition: margin-bottom 0.15s var(--sizeEasing), border-left 0.12s ease-in, border-right 0.12s ease-in, border-top 0.12s ease-in, border-bottom 0.12s ease-in;
```
kelvah: `border` shorthand property to the rescue:
```suggestion
  border: 1px solid #ffffff;
  transition: margin-bottom 0.15s var(--sizeEasing), border 0.12s ease-in;
```

manstis:
Hey @kelvah this is ready for another review.

Hi @sjamboro we decided on making Model Name entirely optional in the UI now.

It's absence is neither an error or warning but perfectly acceptable value (the PMML Specification states it as optional).

**_Model Name set_**
FAI-406-set


**_Model Name not set_**
FAI-406-not-set


**_Model Name tooltip_**
FAI-406-tooltip

tiagobento (approved):

sjamboro (approved):

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.tsx
```diff
@@ -14,108 +14,135 @@
...
+  const ref = useOnclickOutside(event => onCommitAndClose(), {
     disabled: activeOperation !== Operation.UPDATE_NAME,
     eventTypes: ["click"]
   });
 
   useEffect(() => {
-    setModelName({ value: props.modelName, valid: true });
-  }, [props]);
+    setModelName(props.modelName);
+  }, [props.modelName]);
```
kelvah: Here you access `props.modelName` while `commitModelName` has been destructured from the props object. I find things easier to understand when a single approach is followed.
(That's another thing that could be [linted](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md) by the way)

packages/pmml-editor/src/editor/components/EditorCore/molecules/EditorHeader.scss
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.editorHeader {
+
+  &--modelName {
```
kelvah: This should be a child, not a modifier.

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.scss
```diff
@@ -14,13 +14,69 @@
...
+
+  // Ensure the Textbox when editing does not affect content positioning below.
+  &--editing {
+    margin-bottom: 1px;
+  }
+
+  // Overrides for pf-c-form__group-label-help
+  &__icon {
+    padding: 0 0 2px 0;
+    margin: 0 0 0 0;
```
kelvah: This is the same as `margin: 0`

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.scss
```diff
@@ -14,13 +14,69 @@
...
  */
 
+// The title text itself has padding-left of 8px to align with the Textbox content when editing.
+// Therefore the default PF Split Gutter spacing is too large so reduce. Total remains 16px.
+.pf-l-split__item.modelTitle__icon {
+  margin-right: 8px !important;
+  display: flex;
+}
+
 .modelTitle {
```
kelvah, line 24: Could you please increase the font-size of the model name (both here and in the `--editing` class)? It seems to have lost some importance now. I did a test with `1.25em` and that's probably enough.

kelvah (commented):
Hi @manstis, thanks for the work on this PR, and for your patience. I left you some comments for your consideration.

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.scss
manstis: Applied.

packages/pmml-editor/src/editor/components/EditorCore/molecules/EditorHeader.scss
manstis: Applied.

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.scss
manstis: Applied.

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.scss
manstis: Applied.

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.tsx
manstis: Changed to use state for `title` and `setTitle`. `modelName` is now extracted from `props` (like `commitModelName`).

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.tsx
manstis: Changed to `div`. Good catch.

packages/pmml-editor/src/editor/components/EditorCore/atoms/ModelTitle.scss
manstis: I have made the text when displayed and being edited to use the same styling as our Modals' headers.

kelvah (approved):
thank you @manstis :+1:

PR #433, opened 2021-02-25
FAI-382: Score Cards: Refactor inline callbacks to functions

manstis:
See https://issues.redhat.com/browse/FAI-382

### DIFF
--- a/packages/pmml-editor/src/editor/components/EditorScorecard/atoms/CharacteristicLabel.tsx
+++ b/packages/pmml-editor/src/editor/components/EditorScorecard/atoms/CharacteristicLabel.tsx
@@ -23,7 +23,7 @@ import { DataField, Predicate } from "@kogito-tooling/pmml-editor-marshaller";
 
 interface CharacteristicLabelProps {
   name: string;
-  value: any;
+  value: string;
 }
 
 export const CharacteristicLabel = (props: CharacteristicLabelProps) => {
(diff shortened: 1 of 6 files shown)


manstis:
Just the last commit.. I'll rebase once the dependent PR has gone in..

There are still some small inline functions throughout the codebase but the "monsters" have been moved.

manstis:
@kelvah @sjamboro @tiagobento This is ready for review, thank-you.

packages/pmml-editor/src/editor/components/EditorScorecard/atoms/CharacteristicLabel.tsx
```diff
@@ -21,31 +21,20 @@ import { ValidationIndicatorLabel } from "../../EditorCore/atoms";
 import { toText } from "../organisms";
 import { DataField, Predicate } from "@kogito-tooling/pmml-editor-marshaller";
 
-export const CharacteristicLabel = (name: string, value: any, tooltip?: string) => {
+interface CharacteristicLabelProps {
+  name: string;
+  value: any;
```
tiagobento: Can we have this changed to `string`? Since it's something that's going to be on rendered, I don't think we should be able to put anything here. Also, this being `any` can lead to people passing the wrong arguments without being notified by a compilation error. 

Leaving the responsibility of converting some value to string to the caller is less error-prone, I believe. In the case where you need a number, it's easy enough to convert, but let's say that something that once was a string is not a complex object? If it's passed as `value` here, compiler won't complain and [Object object] will be shown on the screen, which is not ideal.

tiagobento (changes requested):

packages/pmml-editor/src/editor/components/EditorScorecard/atoms/CharacteristicLabel.tsx
manstis: Fixed in new commit.

sjamboro (approved):

kelvah (approved):

manstis:
Hi @tiagobento this just needs your approval please.

tiagobento (approved):

PR #436, opened 2021-03-05
FAI-384: Score Cards: Move validation to post-commit update

manstis:
See https://issues.redhat.com/browse/FAI-384

This is an exciting PR :1st_place_medal: 

It refactors the entire validation sub-system to perform validation after all of the batch commits have completed.

Previously we were validating sub-trees after the sub-tree had been updated. However this was problematic when one sub-tree depended on the state/update made to another since the dependent sub-tree only had access to the stale/prior state and had to re-apply the change to the sub-tree it could access to mimic the complete batch. The JIRA may explain it better (of course it may not).

Sadly, with regard to testing, every RESOLVED child JIRA under https://issues.redhat.com/browse/FAI-333 should be revisited.

### DIFF
--- a/packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaContainer/MiningSchemaContainer.tsx
+++ b/packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaContainer/MiningSchemaContainer.tsx
@@ -92,7 +92,7 @@ const MiningSchemaContainer = (props: MiningSchemaContainerProps) => {
           .forMiningSchema()
           .build()
       ),
-    [miningSchema]
+    [dataDictionary, miningSchema]
   );
 
   return (
@@ -142,6 +142,7 @@ const MiningSchemaContainer = (props: MiningSchemaContainerProps) => {
                             <>
                               <MiningSchemaFields
                                 modelIndex={modelIndex}
+                                dataDictionary={dataDictionary}
                                 fields={miningSchema?.MiningField}
                                 onAddProperties={goToProperties}
                                 onDelete={handleDeleteField}
--- a/packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
+++ b/packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
@@ -22,6 +22,7 @@ import { Builder } from "../paths";
 
 interface DataDictionaryPayload {
   [Actions.AddDataDictionaryField]: {
+    readonly modelIndex?: number;
     readonly name?: string;
     readonly type: DataType;
     readonly optype: OpType;
--- a/packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
+++ b/packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
@@ -32,6 +32,7 @@ interface DataDictionaryPayload {
     readonly index: number;
   };
   [Actions.AddBatchDataDictionaryFields]: {
+    readonly modelIndex?: number;
     readonly dataDictionaryFields: FieldName[];
   };
   [Actions.ReorderDataDictionaryFields]: {
--- a/packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
+++ b/packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
@@ -27,6 +27,7 @@ interface DataDictionaryPayload {
     readonly optype: OpType;
   };
   [Actions.DeleteDataDictionaryField]: {
+    readonly modelIndex?: number;
     readonly index: number;
   };
   [Actions.AddBatchDataDictionaryFields]: {
@@ -73,12 +74,14 @@ export const DataDictionaryReducer: HistoryAwareValidatingReducer<DataDictionary
             if (index >= 0 && index < draft.DataField.length) {
               draft.DataField.splice(index, 1);
             }
+          },
+          pmml => {
             validationRegistry.clear(
               Builder()
                 .forDataDictionary()
                 .build()
             );
-            validateDataFields(draft.DataField, validationRegistry);
+            validateDataFields(pmml.DataDictionary.DataField, validationRegistry);
           }
         );
         break;
@@ -92,12 +95,14 @@ export const DataDictionaryReducer: HistoryAwareValidatingReducer<DataDictionary
           draft => {
             const [removed] = draft.DataField.splice(action.payload.oldIndex, 1);
             draft.DataField.splice(action.payload.newIndex, 0, removed);
+          },
+          pmml => {
             validationRegistry.clear(
               Builder()
                 .forDataDictionary()
                 .build()
             );
-            validateDataFields(draft.DataField, validationRegistry);
+            validateDataFields(pmml.DataDictionary.DataField, validationRegistry);
           }
         );
         break;
--- a/packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
+++ b/packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
@@ -32,8 +32,10 @@ import {
   isInvalidValueReplacementRequired,
   isMissingValueReplacementRequired,
   validateMiningField,
-  validateMiningFields
+  validateMiningFields,
+  validateMiningFieldsDataFieldReference
 } from "../validation/MiningSchema";
+import { getDataDictionary, getMiningField, getMiningSchema } from "../PMMLModelHelper";
 
 interface MiningSchemaFieldPayload {
   [Actions.UpdateMiningSchemaField]: {
@@ -62,27 +64,6 @@ export const MiningSchemaFieldReducer: HistoryAwareValidatingReducer<MiningField
 ): Reducer<MiningField[], AllActions> => {
   return (state: MiningField[], action: AllActions) => {
     switch (action.type) {
-      case Actions.UpdateDataDictionaryField:
-        state.forEach((mf, index) => {
-          if (mf.name === action.payload.originalName) {
-            historyService.batch(
-              state,
-              Builder()
-                .forModel(action.payload.modelIndex)
-                .forMiningSchema()
-                .forMiningField()
-                .build(),
-              draft => {
-                draft[index] = {
-                  ...draft[index],
-                  name: action.payload.dataField.name
-                };
-              }
-            );
-          }
-        });
-        break;
-
       case Actions.UpdateMiningSchemaField:
         historyService.batch(
           state,
@@ -92,7 +73,6 @@ export const MiningSchemaFieldReducer: HistoryAwareValidatingReducer<MiningField
             .forMiningField()
             .build(),
           draft => {
-            const modelIndex = action.payload.modelIndex;
             const miningSchemaIndex = action.payload.miningSchemaIndex;
             const _areLowHighValuesRequired = areLowHighValuesRequired(action.payload.outliers);
             const _isMissingValueReplacementRequired = isMissingValueReplacementRequired(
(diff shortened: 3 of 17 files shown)


packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaContainer/MiningSchemaContainer.tsx
```diff
@@ -92,7 +92,7 @@ const MiningSchemaContainer = (props: MiningSchemaContainerProps) => {
           .forMiningSchema()
           .build()
       ),
-    [miningSchema]
+    [dataDictionary, miningSchema]
```
manstis, line 95: I've also added validation to invalid `MiningField`'s that referenced `DataField`'s when the `DataField` is deleted. This requires the validations for the "Mining Schema" button and "Mining Field Rows" to be updated when the `DataDictionary` changes.

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
```diff
@@ -62,27 +64,6 @@ export const MiningSchemaFieldReducer: HistoryAwareValidatingReducer<MiningField
 ): Reducer<MiningField[], AllActions> => {
   return (state: MiningField[], action: AllActions) => {
     switch (action.type) {
-      case Actions.UpdateDataDictionaryField:
```
manstis, line 65: I've just moved this block, keeping the order the actions are handled by the reducers more consistent.

packages/pmml-editor/src/editor/reducers/OutputFieldReducer.ts
```diff
@@ -59,6 +63,23 @@ export const OutputFieldReducer: HistoryAwareReducer<OutputField[], AllActions>
                 isFinalResult: action.payload.outputField.isFinalResult
               };
             }
+          },
```
manstis, line 66: I've moved some validations that were handled by the `ScorecardReducer` to reducers more appropriate for the validation being performed. For example `MiningSchema`\`MiningField` validation is now handled by the `MiningSchemaReducer`\`MiningSchemaFieldReducer`. The same can now be said for `Outputs` and `OutputField`.

packages/pmml-editor/src/editor/reducers/OutputReducer.ts
```diff
@@ -95,8 +148,116 @@ export const OutputReducer: HistoryAwareReducer<Output, AllActions> = (
...
+        break;
+
+      case Actions.DeleteMiningSchemaField:
+        if (state.OutputField.length > 0) {
+          const modelIndex = action.payload.modelIndex;
+          historyService.batch(
+            state,
+            Builder()
+              .forModel(modelIndex)
+              .build(),
```
kelvah: This is not working because of the missing `output` in the path

```suggestion
            Builder()
              .forModel(modelIndex)
              .forOutput()
              .build(),
```

packages/pmml-editor/src/editor/reducers/OutputReducer.ts
```diff
@@ -95,8 +148,116 @@ export const OutputReducer: HistoryAwareReducer<Output, AllActions> = (
...
+        break;
+
+      case Actions.UpdateMiningSchemaField:
+        if (state.OutputField.length > 0) {
+          const modelIndex = action.payload.modelIndex;
+          historyService.batch(
+            state,
+            Builder()
+              .forModel(modelIndex)
+              .build(),
```
kelvah: Missing `output` here too

packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
```diff
@@ -27,6 +27,7 @@ interface DataDictionaryPayload {
     readonly optype: OpType;
   };
   [Actions.DeleteDataDictionaryField]: {
+    readonly modelIndex?: number;
```
kelvah, line 31: Do we need `modelIndex` to delete a DD field?

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
```diff
@@ -144,20 +131,71 @@ export const MiningSchemaFieldReducer: HistoryAwareValidatingReducer<MiningField
...
+                    .forMiningSchema()
+                    .build()
+                );
+                validateMiningFields(modelIndex, miningSchema.MiningField, validationRegistry);
+                validateMiningFieldsDataFieldReference(
+                  modelIndex,
+                  dataDictionary.DataField,
+                  miningSchema.MiningField,
+                  validationRegistry
+                );
```
kelvah: I see you are marking as invalid the mining fields here. I am guessing what the user should do with them afterwords. Delete them? Create or rename a DD field? In that case mining fields need a re-validation on DD change. Also wondering if it is dangerous to replace a DD field pointed by a MiningField if there are attributes referring to the latter.

kelvah (changes requested):
Hi @manstis, thank you for the huge refactor. I've found two bugs inside `OutputReducer`. Plus left a consideration.

packages/pmml-editor/src/editor/reducers/DataDictionaryReducer.ts
kelvah: Yes, we do. For the new validation you just added for Mining Schema. And that I commented minutes ago. :man_facepalming:

packages/pmml-editor/src/editor/reducers/OutputReducer.ts
manstis: Corrected. Thank-you.

packages/pmml-editor/src/editor/reducers/OutputReducer.ts
manstis: Corrected. Thank-you.

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
manstis: At this point it is only the references I need to re-validate so perhaps I'll make the clean-up smarter and just clear `forDataFieldMissing()`. I can then remove invalidation of the whole `MiningField`. I should probably re-validate the `MiningField` references when a DD is added or updated too. Did I understand correctly? I did not understand your second point "if it is dangerous to replace a DD field...". Perhaps we can discuss?

kelvah: > At this point it is only the references I need to re-validate so perhaps I'll make the clean-up smarter and just clear `forDataFieldMissing()`. I can then remove invalidation of the whole `MiningField`. I should probably re-validate the `MiningField` references when a DD is added or updated too. Did I understand correctly?

Yes, that's correct. :+1:

> I did not understand your second point "if it is dangerous to replace a DD field...". Perhaps we can discuss?

Sure. I was referring to characteristics attributes, specifically to predicates. But on a second thought I don't think it's a problem strictly related to DD field deletion. I think it's more about what happens to predicates when doing changes to DD fields in general. So that's another story.

tiagobento (approved):

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
manstis: @kelvah I've added a new commit fixing "Issue 1 : At this point it is only the references I need to re-validate ...."

As discussed "Issue 2 : Complex validation scenarios" will be a new JIRA.

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
```diff
@@ -119,18 +120,22 @@ export const MiningSchemaFieldReducer: HistoryAwareValidatingReducer<MiningField
...
                   .forModel(modelIndex)
                   .forMiningSchema()
                   .forMiningField(miningSchemaIndex)
                   .build()
               );
-              validateMiningField(
-                action.payload.modelIndex,
-                action.payload.miningSchemaIndex,
+              validateMiningField(modelIndex, miningSchemaIndex, miningField, validationRegistry);
+              validateMiningFieldDataFieldReference(
```
manstis, line 135: When a `MiningField` is updated we clear the `ValidationRegistry` of all entries for the _whole_ `MiningField`.

We therefore have to re-validate the references. This issue would also exist in `master`.

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
```diff
@@ -139,6 +144,27 @@ export const MiningSchemaFieldReducer: HistoryAwareValidatingReducer<MiningField
...
+              draft => {
+                draft[index] = {
+                  ...draft[index],
+                  name: action.payload.dataField.name
+                };
+              }
+            );
+          }
+        });
+      //Note no break here, since we want the validation below to apply.
```
manstis, line 166: This, I think, is the lessor of two evils.

I either allow fall-through of the `case` to include validation or duplicate the code for validation.

~~Or move to a function that is re-used~~

Do you have a preference?

packages/pmml-editor/src/editor/validation/MiningSchema.ts
```diff
@@ -168,11 +168,28 @@ export const validateMiningFieldsDataFieldReference = (
...
+  dataFields: DataField[],
+  miningFieldIndex: number,
+  miningField: MiningField,
+  validationRegistry: ValidationRegistry
+): void => {
+  const dataFieldNames = dataFields.map(dataField => dataField.name);
+  _validateMiningFieldDataFieldReference(modelIndex, dataFieldNames, miningFieldIndex, miningField, validationRegistry);
+};
+
+const _validateMiningFieldDataFieldReference = (
```
manstis: Renamed, prefixing `_` (since it's an internal function).

packages/pmml-editor/src/editor/validation/MiningSchema.ts
```diff
@@ -168,11 +168,28 @@ export const validateMiningFieldsDataFieldReference = (
...
+      dataFieldNames,
+      miningFieldIndex,
+      miningField,
+      validationRegistry
+    )
   );
 };
 
-const validateMiningFieldDataFieldReference = (
+export const validateMiningFieldDataFieldReference = (
```
manstis, line 173: New function to validate the references of a single `MiningField`.

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
kelvah: No problems with the case fall-through for me.
The only issue I see is that data fields can be added also via `Actions.AddBatchDataDictionaryFields`.

packages/pmml-editor/src/editor/validation/MiningSchema.ts
kelvah: I don't get why you introduced the `_validateMiningFieldDataFieldReference` abstraction :thinking:

manstis: Changed in a new commit to only have the exposed functions `validateMiningFieldDataFieldReference(..)` and `validateMiningFieldsDataFieldReference(..)`.

packages/pmml-editor/src/editor/validation/MiningSchema.ts
manstis: Refactored.

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
manstis: Validation of references also added for `AddBatchDataDictionaryFields`.

packages/pmml-editor/src/editor/validation/MiningSchema.ts
manstis: Refactored.

kelvah (approved):
thanks @manstis for the changes :+1:

packages/pmml-editor/src/editor/reducers/MiningSchemaFieldReducer.ts
manstis: See https://issues.redhat.com/browse/FAI-434

sjamboro (approved):

PR #440, opened 2021-03-09
FAI-426: Score Cards: DataField validation

manstis:
See https://issues.redhat.com/browse/FAI-426

### DIFF
--- a/packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
+++ b/packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
@@ -127,6 +127,9 @@ const DataDictionaryContainer = (props: DataDictionaryContainerProps) => {
 
   const dataTypeNameValidation = (dataTypeName: string) => {
     let isValid = true;
+    if (dataTypeName.trim().length === 0) {
+      return false;
+    }
     const match = dataTypes.find((item, index) => item.name === dataTypeName.trim() && index !== editing);
     if (match !== undefined) {
       isValid = false;
--- a/packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
+++ b/packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
@@ -18,7 +18,7 @@ import {
   StackItem,
   TextInput
 } from "@patternfly/react-core";
-import { ArrowAltCircleRightIcon, TrashIcon } from "@patternfly/react-icons";
+import { ArrowAltCircleRightIcon, ExclamationCircleIcon, TrashIcon } from "@patternfly/react-icons";
 import { DDDataField } from "../DataDictionaryContainer/DataDictionaryContainer";
 import "./DataTypeItem.scss";
 import ConstraintsLabel from "../ConstraintsLabel/ConstraintsLabel";
@@ -193,7 +193,12 @@ const DataTypeItem = (props: DataTypeItemProps) => {
             }
           }}
         >
-          <Form onSubmit={handleSave}>
+          <Form
+            onSubmit={e => {
+              e.stopPropagation();
+              e.preventDefault();
+            }}
+          >
             <Split hasGutter={true}>
               <SplitItem>
                 <Stack hasGutter={true}>
@@ -204,6 +209,7 @@ const DataTypeItem = (props: DataTypeItemProps) => {
                           fieldId="name"
                           label="Name"
                           helperTextInvalid="Name is mandatory and must be unique"
+                          helperTextInvalidIcon={<ExclamationCircleIcon />}
                           validated={validation}
                           style={{ width: 280 }}
                           isRequired={true}
--- a/packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
+++ b/packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
@@ -115,7 +115,7 @@ const DataTypeItem = (props: DataTypeItemProps) => {
   };
 
   const handleNameSave = () => {
-    if (name.trim().length === 0 || validation === "error") {
+    if (validation === "error") {
       setName(dataType.name);
       setValidation("default");
     } else if (name !== dataType.name) {
@@ -203,9 +203,10 @@ const DataTypeItem = (props: DataTypeItemProps) => {
                         <FormGroup
                           fieldId="name"
                           label="Name"
-                          helperTextInvalid="Name already used by another Data Type"
+                          helperTextInvalid="Name is mandatory and must be unique"
                           validated={validation}
                           style={{ width: 280 }}
+                          isRequired={true}
                         >
                           <TextInput
                             type="text"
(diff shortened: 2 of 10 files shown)


kelvah (approved):

packages/pmml-editor/src/editor/components/Outputs/molecules/OutputFieldEditRow.tsx
```diff
@@ -186,14 +184,19 @@ const OutputFieldEditRow = (props: OutputFieldEditRowProps) => {
...
                     });
                   }}
                   onBlur={e => {
                     if (name?.valid) {
                       onCommit({
                         name: name.value as FieldName
                       });
+                    } else {
+                      setName({
+                        value: outputField?.name,
```
sjamboro, line 198: @manstis I am able to insert an empty string as output name (data field name) without any validation error. Please check CharacteristicsTableEditRow.tsx where I cannot insert an empty name of characteristic.

Step to reproduce:
- delete string in TextField
- press Enter
- confirm change (click somewhere)

image

sjamboro (commented):
@manstis just for sure ... this issue is only about consistency of DataField and OutputField? ... File which is upload and contains an empty name of DF, OF or Characteristicis (like `<DataField name="" optype="continuous" dataType="double"/>`) is not a scope of this issue.

packages/pmml-editor/src/editor/components/Outputs/molecules/OutputFieldEditRow.tsx
manstis: Good catch @sjamboro. This should not be possible. I will take a look.

packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
```diff
@@ -203,9 +203,10 @@ const DataTypeItem = (props: DataTypeItemProps) => {
                         <FormGroup
                           fieldId="name"
                           label="Name"
-                          helperTextInvalid="Name already used by another Data Type"
+                          helperTextInvalid="Name is mandatory and must be unique"
```
sjamboro, line 211: This does not look same
image

Maybe you should add here 'helperTextInvalidIcon={<ExclamationCircleIcon />}' to be same as Outputs

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
```diff
@@ -127,6 +127,9 @@ const DataDictionaryContainer = (props: DataDictionaryContainerProps) => {
 
   const dataTypeNameValidation = (dataTypeName: string) => {
     let isValid = true;
+    if (dataTypeName.trim().length === 0) {
+      return false;
+    }
```
sjamboro, line 132: @manstis I am confused here. Why do not you trim outputs? OutputsContainer use `const validations = useMemo` and interface. DataDictionaryContainer use `const validations = useRef` and does not contain interface and you changed `dataTypeNameValidation`.

packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
sjamboro: @manstis same code is placed in three different class/modul/files. Is possible to prepare one class/modul/file for similar behaviour of DF and OF? You have to change this message in three files. It is hard to find such bug as reviewer.

sjamboro: @manstis based on this image only labels are different and maybe properties.

tiagobento (approved):

packages/pmml-editor/src/editor/components/Outputs/molecules/OutputFieldEditRow.tsx
manstis: @kelvah How did you want to handle this? ATM `DataTypeItem`'s are contained with a `Form` that saves the updated `DataType` when submitted (by the User pressing enter). This does not validate the name but uses the last `onChange` value (i.e. it could be a duplicate name or, as @sjamboro shows, an empty value). This is different to both `Outputs` and `Characteristics` that have no `Form` submission handling. I ask since you wrote the DD stuff and might prefer to make the other places consistent with DD vs DD the others (obviously, if we want to keep the `Form` submission I'll fix the issue with saving invalid values).

packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
manstis: I have made the three (`DataField`, `OutputField` and `Characteristic`) consistent (icons and place-holders).

Label re-use etc will be handled by https://issues.redhat.com/browse/FAI-269.

packages/pmml-editor/src/editor/components/Outputs/molecules/OutputFieldEditRow.tsx
kelvah: @manstis I would prevent form submission by pressing enter.

manstis: Fixed.

packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
manstis: Fixed.

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
kelvah: > OutputsContainer use const validations = useMemo and interface. DataDictionaryContainer use const validations = useRef

Hi @sjamboro, I worked on Data Dictionary (DD) in the past so I can answer to this. I used `useRef` on purpose. DD is different from Outputs regarding validation. While editing a DD field some of its properties can become conditionally required, meaning the global validation is triggered (i.e. for strings with ordinal op-type). When that happens we have a warning sign on the field and a top level warning message (the one saying `Warning alert: Some items are invalid and need attention`). The top level warning has the side effect of pushing down the list of fields, so to avoid having the field moving while editing it, which could be annoying, I used `useRef` to not display the top level message until the user exits the editing of a specific field.

manstis: @sjamboro I believe @kelvah has explained the reason for the difference.

No further action is required.

manstis:
@sjamboro This is ready for re-review. Thank-you.

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
sjamboro: @kelvah and @manstis thank you

sjamboro (approved):

packages/pmml-editor/src/editor/components/Outputs/molecules/OutputFieldEditRow.tsx
sjamboro: @manstis it seems that problem is solved

PR #446, opened 2021-03-11
Update Patternfly - Add script to help on future updates

ljmotta:
This PR adds new scripts to make it easier to update the Patternfly version.

The `package.json` has three new scripts:
 - `update` - Update Patternfly version, link dependencies, build and update snapshots;
 - `update:deps` - Update Patternfly version, link dependencies, and build;
 - `update:snaps` - Update snapshots;

The snapshots update should be used with **_caution_**. It is meant to help with minor changes, such as changes that don't affect the UI and don't remove the necessity to manually check for any breaking change on the Channels.


tiagobento (commented):
Thanks for the PR, @ljmotta !

Could we rename `--snapshots` to `--test-snapshots-only`? Also, `--dependencies` should be `--dependencies-only`, WDYT?

tiagobento (approved):

PR #447, opened 2021-03-12
KOGITO-4442: Task Console Form breaks when task has no data assignments

pefernan:
This PR avoids breaking the FormRenderer when the form schema doesn't have properties. In runtimes this may happen if the task doesn't have data assignments.

### DIFF
--- a/packages/form-renderer/package.json
+++ b/packages/form-renderer/package.json
@@ -21,7 +21,9 @@
     "ajv": "6.12.2",
     "uniforms": "3.0.0",
     "uniforms-bridge-json-schema": "3.0.0",
-    "uniforms-patternfly": "4.1.2",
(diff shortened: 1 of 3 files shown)


packages/form-renderer/package.json
```diff
@@ -21,7 +21,8 @@
     "ajv": "6.12.2",
     "uniforms": "3.0.0",
     "uniforms-bridge-json-schema": "3.0.0",
-    "uniforms-patternfly": "4.1.2"
+    "uniforms-patternfly": "4.1.2",
+    "lodash": "^4.14.2"
```
tiagobento, line 27: Not sure this one should be here. It's only used in test code, so no reason for people consuming this package to download `lodash` transitively. Can you please move it to `devDependencies`?

tiagobento (approved):

PR #448, opened 2021-03-12
KOGITO-4436: Allow component to be embedded in a div as an alternative to the iframe

paulovmr:


manstis:
Out of interest; how will this work? The BC legacy editors inject a ton of resources into the `IFRAME`'s document; which presumably is not possible when using the `DIV`. I am just curious.

tiagobento:
@manstis Baby steps, Michael!! 

Just kidding :)

This PR is a first step towards a very sophisticated way of loading and sharing modules among different components. This first version does not do any of that, of course. This PR is allowing components (Kogito Views) to be embedded on an simple div and not know about it, since the whole communication remains done through our Channel/Enveope abstraction.

The changes @paulovmr did are infrasctructure-related and will not be used in DMN/BPMN/SceSim/PMML Editors for now. Although, at least the way I see it, the PMML Editor could be a very good first candidate, given its dependencies on React and PatternFly 4, which are already present at the Online Editor.

Anyways, hope that helps!

@paulovmr can give more details about the intrincacies of the code.

packages/envelope/src/embedded/EmbeddedEnvelopeFactory.tsx
```diff
@@ -77,6 +100,10 @@ export function EmbeddedEnvelopeFactory<
 
     useConnectedEnvelopeServer<ApiToProvide>(envelopeServer, props.api);
 
-    return <iframe ref={iframeRef} src={props.envelopePath} style={containerStyles} title="X" />;
+    if (props.config.containerType === ContainerType.DIV) {
+      return <div ref={divRef} id={envelopeServer.id} />;
```
tiagobento: Is this "id" still necessary?

tiagobento (commented):
@paulovmr Great work! Left a minor comment inline.

packages/envelope/src/embedded/EmbeddedEnvelopeFactory.tsx
paulovmr: Not anymore, I've just removed it.

paulovmr:
@manstis I believe @tiagobento already explained most of it. Basically this PR improves the message routing between components. With the iframe, we only had to worry about sending and receive messages using the `window` scope. Once we have multiple components in the same `window`, we need to track the messages destinations and only process them in the targeted component.

tiagobento (approved):

ederign (approved):

PR #409, opened 2021-02-03
FAI-362: Score Cards: Integrate with VSCode channel

manstis:
See https://issues.redhat.com/browse/FAI-362

@tiagobento I'm going to need some help here please.

Screenshot from 2021-02-03 11-04-43

**css**

The PMML Editor "host" page used for the "showcase" webapp does not have any CSS links/references. PatternFly seems to inject the CSS into the "host" HTML page programmatically on load or something. Come to mention it the PMML editor does not have an `index.html` page itself and only exists as a React component.

Screenshot from 2021-02-03 12-14-32

It appears as though it is not being injected into the HTML page created in `vscode-extension/KogitoEditor.setupWebviewContent()`. Do you have any ideas how I can solve this? 

**Monaco web-worker**

The web-worker required for Monaco auto-complete does not appear to be loaded either. Possibly due to the same reason as CSS above. DMN uses Monaco too. Is there anything specific that needs to be done for VSCode? Running the PMML Editor showcase from `pmml-editor/dist/index.html` has the webworker running fine (but I had to allow the browser to load resources from _file_ paths).

**Keyboard shortcut bindings**

I was expecting undo/redo from the VSCode toolbar to work "out of the box". The PMML Editor implements `Editor.undo()` and `Editor.redo()`. Am I missing something? The PMML Editor does not have any other keyboard shortcuts (outside of _normal_ browser/HTML bindings, e.g. tab, enter etc).

**window.confirm(..)**

The PMML Editor uses `window.confirm(....)`. This appears not to work inside VSCode, but I suspect this is by (Microsoft's) design and hence we'd need to move to PatternFly modals. Can you please confirm?

**Notification API**

Is this merged yet? How would I bind to it from a React editor?

PR #453, opened 2021-03-17
Improve VS Code Extension Integration Tests

tiagobento:
@domhanak While debugging the integration tests on #409, I found some points of improvement on the way we run our VS Code Extension IT tests. Can you please take a look? cc @caponetto 

What I did:
1. Removed the call to `yarn run compile`, since we only run the IT tests after running `build:prod`, which generates a `.vsix` file, that's used for executing the tests.
2. Added `rm -rf ./test-resource`. For some reason, when the tests failed, the existence of this file was blocking a re-run, so removing it before running fixes the problem
3. Removed the custom reporter on mocha configuration. I don't think we still need to report the errors on this JUnit format, right? So to make it easier to see the errors on our CI jobs, using the default reporter might save us some time!

**NOTE:** I noticed that some times the tests were haging before executing. This might be what's causing the flaky behaviour on our CI as well! What I noticed here locally is that the `before` part is executing before our extension is *active* on the testing VS Code window. Do you know if there's a way to wait for the extensions to be fully active before beginning the tests?

### DIFF
--- a/packages/vscode-extension-pack-kogito-kie-editors/.mocharc.json
+++ b/packages/vscode-extension-pack-kogito-kie-editors/.mocharc.json
@@ -1,1 +1,7 @@
-{}
+{
+    "retries": 1,
+    "reporter":  "xunit",
+    "reporterOptions": {
+        "output": "./it-tests/results/junit.xml"
+    }
+}
--- a/packages/vscode-extension-pack-kogito-kie-editors/.mocharc.json
+++ b/packages/vscode-extension-pack-kogito-kie-editors/.mocharc.json
@@ -1,7 +1,1 @@
-{
-    "retries": 1,
-    "reporter":  "xunit",
-    "reporterOptions": {
-        "output": "./it-tests/results/junit.xml"
-    }
-}
+{}
--- a/packages/vscode-extension-pack-kogito-kie-editors/package.json
+++ b/packages/vscode-extension-pack-kogito-kie-editors/package.json
@@ -110,7 +110,7 @@
     "compile": "webpack",
     "watch": "webpack",
     "test": "echo 'No tests to run.'",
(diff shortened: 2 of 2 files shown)


caponetto (approved):
LGTM

tiagobento:
@domhanak Can you please take a look at this PR?

domhanak (changes requested):
Please keep the reporter we need it to report test results to QE channels, otherwise looks good. 

Thanks for the cleanup, forgot to remove the unnecessary compilation in my separation PR.

domhanak:
@tiagobento what do you mean by fully active? Is there a time frame where the extensions is installed but not active? AFAIK we activate our extension `onCustomEditor` events, do we do that on StartUp event too? I couldn't find the code to confirm though. 
I would expect it to not be active because we activate it when we open the custom editor. Am I missing something?

The `before` part only opens a folder in VSCode and has nothing to do with the extension functionality, so I am not sure I follow.

tiagobento:
@domhanak Can you please share more about how you use this report? I'd like to keep it but it looks like it's not possible to have multiple reporters, and with the logs not showing up on the console during the builds, it's a little bit hard to understand what caused a test failure.

About the extension being active, I believe when you "install" an extension o VS Code, it takes a while for it to unpack and create the necessary directories and whatnot, so there's a little bit of a gap between installing and being able to use whatever it is the extension provides. In our case, if we open a "dmn" file, for example, without the extension being active, it will open as text.

domhanak:
@tiagobento The report can for example be seen here - https://github.com/kiegroup/kogito-tooling/pull/426/checks?check_run_id=2146639443. Download the `junit-kie-editors-vscode......zip` in artifacts file and check the report. 
No need to scroll in the Github Actions logs for the error.
This report is also to be used for test execution reporting purposes, we can discuss off-pr.
I think that we can remove the reporter config, but we need to keep the output to file.

I get back to you on that startup issue, I think I might have some solutions, but need to test it.

Edit: fixed typos.

tiagobento:
@domhanak Thanks! Reverted the reporters change. Let's do that fix on another PR.

domhanak (approved):

PR #455, opened 2021-03-18
Fix imports and PMML Editor package

tiagobento:
After releasing 0.8.4 and trying to use its packages to update `kie-tooling-store`, I faced issues with the way the some classes were being imported.  I also noticed that the imports for the PMML Editor wouldn't work, since they faced the same issue. This PR fixes these problems.

Tested `yarn start` without building the showcase first and it works, so I removed this script.

Also tested the VS Code Extension and it worked normally on both `yarn build:fast` and `yarn build:prod`.

### DIFF
--- a/packages/pmml-editor/README.md
+++ b/packages/pmml-editor/README.md
@@ -10,6 +10,4 @@ Nothing should be considered concrete at this point.
 
 In order to run:
 
-`yarn run build:showcase`
-
 `yarn run start`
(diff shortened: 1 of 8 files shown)


tiagobento:
cc @manstis

ederign (approved):

manstis:
@tiagobento @ederign The only problem with this PR is that the "showcase" is now included in the distribution. Previously, with the old `webpack.config.js` the "showcase" was only included for the "development" build and "production" was only the editor. Preferably the "showcase" should not be part of the distribution IMO.

tiagobento:
Yep, I noticed that. The showcase artifacts go to `dist`, while the editor artifacts go to `dist/editor`. So by changing the `file` property on `package.json`, I'm able to not publish the showcase artifacts to NPM.

PR #457, opened 2021-03-19
Fix PMML Editor CSS

tiagobento:
I noticed that patternfly.css was loading one extra time when the extension was built with the fixed import, so I tracked it down to the envelope importing it as `.scss` and the PMML Editor importing it as `.css`. I figured webpack treated them as different modules, therefore, loaded it twice. This is all theoretical of course, but it worked :)

Also, I changed the imports on the custom `.scss` files to be the same as in `styles.scss` of the Envelope. Couldn't find anything broken after this change too.

Please take a look at the VS Code extension on the CI artifacts! If everything is okay, we can do a new release first thing next week.

### DIFF
--- a/packages/editor/src/envelope/EditorEnvelopeView.tsx
+++ b/packages/editor/src/envelope/EditorEnvelopeView.tsx
@@ -17,9 +17,7 @@
 import * as React from "react";
 import { Editor } from "../api";
 import { LoadingScreen } from "./LoadingScreen";
-import "@patternfly/patternfly/base/patternfly-variables.css";
-import "@patternfly/patternfly/patternfly-addons.scss";
-import "@patternfly/patternfly/patternfly.scss";
+import "@patternfly/patternfly/patternfly.css";
 import { KeyBindingsHelpOverlay } from "./KeyBindingsHelpOverlay";
 import { useCallback, useImperativeHandle, useState } from "react";
 
--- a/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/AttributesTable.scss
+++ b/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/AttributesTable.scss
@@ -13,7 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@import "../../../../../../../node_modules/@patternfly/patternfly/patternfly.css";
+@import "../../../../../../../node_modules/@patternfly/patternfly/sass-utilities/all";
+@import "../../../../../../../node_modules/@patternfly/patternfly/base/variables";
 
 .attributes {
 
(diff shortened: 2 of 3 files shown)


manstis (approved):
It seems to solve the CSS problem. Thank-you.

I notice we still have an issue loading the fonts. The `WebView` debug console shows some errors relating to loading the fonts. It can however wait until next week. The old `webpack.config` had a rule explicitly for fonts.. IIRC it was for a reason.. so may be worth reinstating. 

I think we should try to analyse this PR works.

- Is it not better to only import PatternFly's `scss` as this is general the composition of others?
- IIRC the PMML editor does not override any PatternFly `css` without use of additional selectors. I therefore don't fully understand why loading the PF `css` once, twice or 10 times should affect the `css` for PMML. 
- If there is a genuine limitation it needs to be documented.

manstis:
Hey @kelvah I'd like to know your opinion on this.

It improves the situation for sure; and therefore I think it good to merge.

However I'd like to understand _why_ it works and _what_ we'd need to be wary of in the future.

That said subsequent investigative work could be the subject of a new JIRA.

manstis:
These are the only references to Patternfly CSS I can find in `pmml-editor`.

Screenshot from 2021-03-22 10-10-19

Interestingly two of them are what @tiagobento changes in this PR and the third is what @tiagobento has changed in `EditorEnvelopeView.tsx` (so both now import the CSS and not the SCSS). I am trying changing the `pmml-editor` to import `@patternfly/patternfly/patternfly.scss` (same as `EditorEnvelopeView.tsx`) and investigate why we're importing something in the other two files (and if they can be made consistent with `EditorEnvelopeView.tsx` too).

manstis (changes requested):
I propose this be closed and replaced with https://github.com/kiegroup/kogito-tooling/pull/458.

Thank-you @tiagobento for doing the hard (analysis) bit.

kelvah:
@manstis I think you're right. The `patternfly.css` import inside `AttributesTable.scss` and `OutputFieldsTable.scss` is a repetition and should be removed.  #458 is rightly doing so.

Aligning the base style import inside `EditorEnvelopeView.tsx` and `pmml-editor/src/editor/index.ts` is solving the problem but I think it works because webpack is changing only the bundling order of those styles. We still have the style duplication inside the VSCode extension, which is not ideal. I guess the base patternfly styles should be imported only once and at the highest level (which I suppose is the `editor` package). This poses some questions about how the style structure should be organized. Probably we should use peerDependecies for PF at lower levels (like pmml-editor). Also it's strange that the `editor` is importing styles while it doesn't list PF as a dependency. I think we can continue the discussion elsewhere and keep #458 as a fix for now.

tiagobento:
Closed in favor of #458

PR #458, opened 2021-03-22
Fix PMML CSS inside the editor package.

manstis:
@tiagobento @kelvah I propose this PR in preference of #457

This keeps the changes "inside the editor". The conclusion being: import the same packages as used by the container (envelope).

### DIFF
--- a/packages/pmml-editor/src/editor/index.ts
+++ b/packages/pmml-editor/src/editor/index.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import "@patternfly/patternfly/base/patternfly-variables.css";
 import "@patternfly/patternfly/patternfly.scss";
 
 export * from "./PMMLEditor";
--- a/packages/pmml-editor/src/editor/index.ts
+++ b/packages/pmml-editor/src/editor/index.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import "@patternfly/patternfly/patternfly.css";
+import "@patternfly/patternfly/base/patternfly-variables.css";
+import "@patternfly/patternfly/patternfly.scss";
 
 export * from "./PMMLEditor";
 export * from "./PMMLEditorFactory";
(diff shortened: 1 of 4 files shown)


manstis:
This fixes the CSS issues.

There is still a problem loading fonts but I will create a JIRA for it.

Screenshot from 2021-03-22 10-58-04

Fonts load OK in a regular browser (i.e. not Electron not in VSCode).

Screenshot from 2021-03-22 11-00-01

manstis:
See https://issues.redhat.com/browse/FAI-446

I suspect it is not going to be a lot of fun to fix PF font loading in VSCode.

packages/pmml-editor/src/editor/index.ts
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import "@patternfly/patternfly/patternfly.css";
+import "@patternfly/patternfly/base/patternfly-variables.css";
```
kelvah: Not sure why we need this. Variables are already included in the base stylesheet.

kelvah (approved):

packages/pmml-editor/src/editor/index.ts
manstis: Removed :+1:

tiagobento (approved):
@manstis No problem! If that fixes the issue, we can use this one instead of #457. I'll close the other one. Approving!

tiagobento:
@manstis Looks like GH Actions froze. Can you push an empty commit?

manstis:
@tiagobento This keeps failing on different integration tests; sometimes Mac, sometimes Windows.

Do you recall where I can see the real logs? I found some under GH's Artifacts dropdown and they showed time-outs.

I'm inclined to merge "as is" given the failures appear not to be consistent. WDYT?

Once this is merged I guess I'll need a new point release of `kogito-tooling` and Marketplace.

tiagobento:
Merging as the failures are certainly not related. 

@tomasdavidorg @jstastny-cz These test failures seem to be intermittent. Can you please take a look? They might be a good resource for finding the issues.

PR #444, opened 2021-03-11
KOGITO-4654: Format codebase and check formatting as part of full downstream build 

jomarko:
The formatting changes were split into commits per 'package' to facilitate review.

PR #465, opened 2021-03-26
KOGITO-4654: Format whole codebase by prettier

jomarko:
Formatted as 'yarn prettier --write .'
    
For more details see:
- https://issues.redhat.com/browse/KOGITO-4654


jomarko:
Replacement of #444

jomarko:
@tiagobento reformatted using existing rules as requested.

packages/backend/src/vscode/commands/runTestScenarioCommand.ts
```diff
@@ -75,7 +75,7 @@ async function run(
         message: response.message!,
         severity: "WARNING",
         type: "ALERT",
-        path: ""
+        path: "",
```
manstis, line 78: I guess this is "standard".. but we'd normally remove hanging commas on the last item.

If this "..is the way.." then so be it.. but it is definitely not something I'd type normally.

packages/desktop/src/__tests__/webview/__snapshots__/App.test.tsx.snap
```diff
@@ -144,7 +143,6 @@ exports[`invalid file type alert alert is closed after 3000ms 1`] = `
               <a
                 aria-current="page"
                 class="pf-c-nav__link pf-m-current"
-                tabindex="-1"
```
manstis: I worry about the automatic removal of `tabindex="-1"` as it can have a purpose.

packages/backend/src/vscode/commands/runTestScenarioCommand.ts
tiagobento: @jomarko Indeed. In fact, I don't see my Prettier execution adding trailing commas anywhere.. See 
image

Not sure how that happened on this PR, to be honest. Can you please share the exact commands you used to format the code?

packages/desktop/src/__tests__/webview/__snapshots__/App.test.tsx.snap
tiagobento: Not sure how your Prettier was able to format that too. Here's the error I get.
image

jomarko:
@tiagobento the exact command is shared in description - https://github.com/kiegroup/kogito-tooling/pull/465#issue-601642082

@manstis trailing commas and __snapshots__ should be excluded now

tiagobento:
@jomarko I can't run the command you shared. See 

```
11:38:19 @ kogito-tooling/: yarn prettier --write .
yarn run v1.22.10
$ /Users/tiagobento/redhat/kogito-tooling/node_modules/.bin/prettier --write .
[error] No matching files. Patterns tried: . !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,svn,hg}/**
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Also, what changed to not include trailing commas?

.github/actions/env-to-json/index.js
```diff
@@ -20,7 +20,7 @@ const fs = require("fs");
 
 function filterObjectProperties(object, prefix) {
   return Object.keys(object)
-    .filter(key => key.startsWith(prefix))
+    .filter((key) => key.startsWith(prefix))
```
tiagobento, line 23: @jomarko This is something that I don't get when formatting locally. 
image

Not sure what's causing these differences..

jomarko: that is caused by **arrowParens** option. https://prettier.io/docs/en/options.html#arrow-function-parentheses, included it now into PR.

jomarko:
@tiagobento you and Michael Asked me to not include trailing commas. https://github.com/kiegroup/kogito-tooling/pull/465#discussion_r602325385. It is this option https://prettier.io/docs/en/options.html#trailing-commas

manstis:
> @tiagobento you and Michael Asked me to not include trailing commas. [#465 (comment)](https://github.com/kiegroup/kogito-tooling/pull/465#discussion_r602325385). It is this option https://prettier.io/docs/en/options.html#trailing-commas

Thank-you @jomarko.

I didn't [ask](https://github.com/kiegroup/kogito-tooling/pull/465#discussion_r602325385) for trailing commas to be removed; I just said it seemed strange and not something we'd normally do. The link you shared states the default setting was explicitly "...changed from `none` to `es5` in v2.0.0.." and therefore perhaps trailing commas should be added?

I have spent a lot of time in the past debating code formatting. I'm therefore happy to go with whatever.

The important aspect is that we all use a formatter and all use the same format.

manstis (approved):

jomarko:
@manstis sorry for my bad wording. I meant, I agree that adding comma after last field of json object is sth not handy for most developers, including me.

However I opened PR with default seetings, where as you said, was changed to have them since 2.0.0. However If the team do not want the, I believe we do not need them. 

I am not javascript expert, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas says they are used when adding item into object and having clean diffs in version-control. If there is not other, more significant meaning (I do not know, support of some special browser), then I tend to think, we do nto need to use trailing commas.

tiagobento:
I personally do not care too much about the code format too, the reason we chose Prettier in the first place was for us to not have these long formatting talks. What I do care about though, is to keep this "clean slate" PR as small as possible. Also, we need to make sure that people will be able to follow the formatting rules being applied. So a few points:

1. I cannot run the command you shared. What am I doing wrong?
2. Why is the code changed in places where it shouldn't? We shouldn't be tweaking Prettier options. We have our configuration file with a `printWidth` configuration and that's it.
3. How is your Prettier capable of formatting every single file extension the repo where mine isn't? What can be the difference between our setup?

jomarko:
@tiagobento yes. thank you for questions.
1. @tomasdavidorg investigated that may be related to version of prettier, it is increased now, could you please to to run it again? I increased to 2.2.1
2. and 3. has probably the same rootcause, that we need to figure out, do you format your code directly from IDE please?
In case of yes, what IDE, what plugin installed, what command to start that plugin?
In case of no, how do you start formatting please?

tomasdavidorg (approved):
Looks good to me now. Thanks @jomarko

tiagobento:
@jomarko I can now run the command with the updated version.

In WebStorm, I simply run the shortcut to format with Prettier. On VS Code, I have the Prettier extension installed. I just noticed that WebStorm is formatting it differently from VS Code, though. Will comment here when I figure out what's happening.

tiagobento:
@jomarko When running the command, I noticed that we're formatting ignored files. We shouldn't do that, as it takes a very long time to do so. One thing I tried here was making `.prettierignore` be a symlink to `.gitignore`. Got better results, but still there are separate `.gitignore` files on the packages, so probably unifying them into one would be better. WDYT?

tiagobento:
Ok, WebStorm needs to be restarted to load changes on `prettier.config.js`. Since I changed branches with it open, it wasn't working. Go figure ¯\\_(ツ)_/¯. Also, it excludes some extensions from formatting, so one thing to keep in mind when using WebStorm.

Looks like originally you were using a globally installed Prettier 2, while the project was using Prettier 1.9, that's what caused the confusion. As @manstis already observed on this PR, somethings changed from versions 1.9 to 2.

So to summarise:
1. About the `arrowParens` config. First available in v1.9.0, default value changed from `avoid` to `always` in v2.0.0. See https://prettier.io/docs/en/options.html#arrow-function-parentheses
1. About the `trailingComma` config. Default value changed from `none` to `es5` in v2.0.0. See https://prettier.io/docs/en/options.html#trailing-commas

Keeping with our strategy to have a the minimum possible configuration for Prettier, and now that we understand where these discrepancies with the current formatting came from, I'd say we should remove the custom `arrowParens` and `trailingComma` from our Prettier configuration file and go with whatever is the default on the most recent Prettier. We'll only keep `printWidth` as previously discussed.

@jomarko  IMHO, two things need to be solved before merging this PR.
1. *.md files still have incorrect changes. Testing quickly here I found that if you put the text inside backticks (`), Prettier leaves it alone :)
2. `.prettierignore` must be improved. I like the idea of making it be a symlink to `.gitignore`, but we need to unify all `.gitignore` files into one for this to work. WDYT?

jomarko:
@tiagobento thank you.

1. could you please clarify more the invalid changes you see in the changed MD files?
2. applied

tiagobento:
@jomarko See the environment variables command for the Editors paths on the root README.md. Prettier added some `\` characters there.

jomarko:
@tiagobento oh, well spotted.  However I believe it is not issue, this is how the github renders the result:
Screenshot from 2021-03-31 14-56-50

jomarko:
@jstastny-cz @tomasdavidorg hello, sorry for interrupting you. This PR has two fails
- online-editor tests:windows - there is a fail in creating PMML, @tomasdavidorg what is your experience please?, is that test stable otherwise and fail is caused by this PR?
- editors-standalone:mac - there is a fail in dmn readonly, however from video to me it seems just some need of timeout increase @jstastny-cz what is you experience please? was that test stable on mac before this PR?

tomasdavidorg:
@jomarko I think this is just random fail but could you please rerun it. You can do it by `git commit --allow-empty -m "Trigger Build"`. In the screenshot I see the loading popup didn't disappear, so the machine was probably very slow.

tiagobento:
@jomarko Yeah... on WebStorm it renders with the slashes... Can you change it to have the (`)? I think it's better to read/edit too. Simple change!

@tomasdavidorg Yeah.. we're seeing too many random failures recently :/ Can you take a look?

jomarko:
@tiagobento applied your requirement about `export EXTERNAL_RESOURCE_PATH*`

domhanak (approved):

jomarko:
@tiagobento @ederign what is your opinion please? could we please proceed with merge? Are the some unresolved points?

jomarko:
@ederign @tiagobento another friendly reminder, builds are green and there are no conflicts now.

Could you please review? If you are still out of resources, could you please delegate to other peer? Thank you.

tiagobento (approved):

PR #477, opened 2021-04-12
KOGITO-4892: Fix the separator replace code

caponetto:
This PR fixes a bug where custom tasks cannot be found from a WID in a BPMN file (VS Code - Windows OS only).
See [KOGITO-4892](https://issues.redhat.com/browse/KOGITO-4892)

This code only replaces the first separator:
```js
"test/foo/bar/see/the/separators".replace("/", "\\")
```
See:
separators

According to the documentation:
replace

Thus, we need to use a regex to make it work properly.

### DIFF
--- a/packages/vscode-extension/src/KogitoEditableDocument.ts
+++ b/packages/vscode-extension/src/KogitoEditableDocument.ts
@@ -61,7 +61,7 @@ export class KogitoEditableDocument implements CustomDocument {
   get relativePath() {
     // For some reason, `asRelativePath` always returns paths with the '/' separator,
     // so on Windows, we need to replace it to the correct one, which is '\'.
-    return vscode.workspace.asRelativePath(this.uri).replace("/", nodePath.sep);
+    return vscode.workspace.asRelativePath(this.uri).replace(/\//g, nodePath.sep);
   }
 
   get fileExtension() {


ederign (approved):

caponetto:
Hi @domhanak, could you please validate this fix? (or re-assign accordingly)

domhanak:
@caponetto Sure, could you please add a unit test for this method?

domhanak (changes requested):
Whole class `KogitoEditableDocument` does not have unit tests.

I think I uncovered another issue but after deeper investigation I think that this change has no effect on said issue.
The issue is related to the use of `// comments` in work item definitions. I ll consult with Kirill :) 

Please add the unit test so that we have this path resolution covered really early in the test cycle.

I tested the scenario reported in KOGITO-4892 and that is indeed fixed on Windows machines. :heavy_check_mark:

caponetto:
Hi @domhanak, thanks for the validation!
Regarding the unit tests, please see https://github.com/kiegroup/kogito-tooling/pull/391#issuecomment-765356311.

domhanak:
@caponetto I disagree. We have a bug in our codebase, we must have a test that verifies it for the future. 

Checking this on E2E level takes significant amount of time compared to unit test that verify we apply replace with correct arguments on what `vscode.workspace.asRelativePath(this.uri)` can return.

domhanak (approved):
Circling back through my assigned PR's and this will be covered by  #480 which will add test for scenarios that this PR should fix. That kinda makes my last comment invalid so let's merge. :)

PR #478, opened 2021-04-12
KOGITO-4915: Standalone editors setContent implementation should receive path and content

paulovmr:


domhanak (approved):

domhanak:
@paulovmr Where should I check for DMN, BPMN specific handling of these? It seems that one uses path, content - toher uses content, path

domhanak:
pardon me , but I do not see any unit tests. Consider adding one that verifies `createEditor` method returns what we expect

ederign (approved):

PR #482, opened 2021-04-23
KOGITO-4654 Check the code is formatted during CI build

jomarko:
This PR follows changes done by:
- kiegroup#465

For more details see:
- https://issues.redhat.com/browse/KOGITO-4654

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,13 @@ jobs:
         yarn: [1.19.1]
 
     steps:
+      # https://github.com/prettier/prettier/issues/7825
+      # https://github.com/actions/checkout/issues/135
+      - name: Set git to use LF. Needed due to running action on widnows machines.
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
+
       - uses: actions/checkout@v2
       - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,10 @@ jobs:
           [ "0" == "$(git diff yarn.lock | wc -l)" ]
         shell: bash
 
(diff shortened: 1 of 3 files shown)


tomasdavidorg:
I would rather do the check in a separate workflow so the FDB or other builds are not blocked by this. WDYT?

tomasdavidorg:
Chrome fails are caused by update of Chrome browser to version 90. Fix is here https://github.com/kiegroup/kogito-tooling/pull/483

Maybe we could also consider separation of the `yarn.lock` check from the FBD and CI builds later. @tiagobento WDYT?

tiagobento:
Hi @jomarko @tomasdavidorg. Thanks for submitting this PR. However, I agree with @tomasdavidorg that the FDB Is not the best place to run it. We should include this check on our "CI - Default" workflows. Also, as you're adding a script to check the formatting, why not add one to actually to the formatting too? :) I'd say `yarn format` and `yarn format:check` are good names.

jomarko:
Thank you for quick feedback, incorporating changes now.

jomarko:
@tomasdavidorg if I have a look on the `ci.yml`, there is a step `Check yarn.lock md5`. it seems to me be a duplicate of `pull_request_full_downstream_ci.yml` - `Check yarn.lock`. Should I remove `Check yarn.lock` step?

tiagobento:
@jomarko I think you can remove the yarn.lock check from FDB, yes. There's no reason for this to be duplicated.

tiagobento (approved):

tiagobento:
@jomarko The CI is breaking on the new check. Looks like some files were not formatted?

jomarko:
@tiagobento there seems to be specific issue on windows. mac and ubuntu passed `format:check`. I tried to fix with the last commit.

I didn't find un-formatted files even after rebasing with master.

tomasdavidorg (approved):

PR #490, opened 2021-04-27
Update Editors to 7.53.0.Final

tiagobento:

PR #491, opened 2021-04-28
Fix tests with 7.53.Final: Apply workaround for RHPAM-3606

domhanak:
https://issues.redhat.com/browse/RHPAM-3606

cc @tiagobento 

Workaround is that we do not test with explorer when process has unknown custom tasks.


domhanak:
#490 related

domhanak:
FDB failed on appformer please restart it

tiagobento (approved):

PR #497, opened 2021-05-07
KOGITO-4686: Cannot open DMN/BPMN files from command line with VS Code

tiagobento:
See https://issues.redhat.com/browse/KOGITO-4686


ederign (approved):

caponetto (approved):
I couldn't validate the fix on VS Code v1.56 once it is opening the file in the Text Editor, not ours. I've tried other extensions and the behavior is the same.

On VS Code Insiders, on the other hand, I was able to reproduce the issue and this PR fixes it!
So I believe there is something wrong with VS Code v1.56 (Ubuntu).

tiagobento:
Yeah.. just confirmed it's happening on macOS too. :/

tiagobento:
According to https://github.com/microsoft/vscode/issues/123491 we need to wait a next release of VS Code to have this fixed.

tiagobento:
@domhanak You should be able to test this PR on VS Code Insiders, as the fix is already there.

tiagobento:
Looks like VS Code 1.56.1 is out, but the fix is not there on this release. That means we'll probably have to wait until 1.57.0.

domhanak:
This breaks the it tests for vscode. They can no longer find the the envelope div using `*[id="envelope-app"]`. Should the check be removed from tests?

domhanak:
#508 should fix the tests for good imho

domhanak:
Please ping me if I am stalling this, I am waiting for 1.57 release to verify this with it.

tiagobento:
@domhanak We can wait for 1.57.0 to validate this properly, no rush.

tiagobento:
Tests will probably break because of VS Code's 1.57.0 Workspace Trust feature. After #533 they'll likely pass again :)

domhanak:
@tiagobento I tested this with 1.57.0 and run into issue where I ended up with text editor opening instead of BPMN/DMN/SCESIM editors.
How to reproduce:
1. Install the extension and close all instances of VSCode.
2. Execute `code nameOfFile.bpmn/dmn/scesim` 
3. A text editor is opened instead of expected editor

Workaround: Have a VSCode instance already open, the command will open new EditorView in the running instance and loads the respective editor.
I manage to reproduce this 100% when there is no BPMN/DMN/SCESIM file in cache (not sure about this, please correct me if I am wrong). 
1. Basically, open a folder that has no open kogito editors and close vscode. 
2. After that proceed to use `code new.dmn` to open new instance
3. A text editor is open instead.

NOTE: Any subsequent executions of `code xxx.dmn/bpmn/scesim` open the expected kogito editor

Looks like this is a corner case when the VSCode instance is open for the first time on users environment. Is this something we can possibly fix here or should I create a new jira?

domhanak:
@tiagobento friendly reminder for my above comment.

domhanak (commented):
I tested this, you can see results in my comment. Nothing sever, just needs clarification.

tiagobento:
@domhanak Hi Dominik, thanks for testing it. I'll reply as soon as I can.

domhanak:
cc @tiagobento

domhanak (approved):

tiagobento:
@domhanak Ok, so after a couple of months (:D) I revisited this PR.

Scenarios I tested: (I'm on macOS)

- With VS Code completely closed:
1.  `code existing-file.dmn`
1.  `code new-file.dmn`


.
- With VS Code already open:
1.  `code existing-file.dmn`
1.  `code new-file.dmn`

For some reason, this doesn't work with BPMN files, but DMN, SceSim, and PMML are fine.

Let's create a separate Jira just for BPMN?

domhanak:
@tiagobento Hmm we should probably investigate why BPMN is affected and DMN not.

tiagobento:
@domhanak Whatever it is, seems that it's not related to the VS Code extension itself, but rather with the BPMN Editor. The infrastructure is prepared for this new use with this PR, so no reason to block merging because BPMN doesn't open, as it seems and issue with the Editor itself.

domhanak:
@tiagobento I already approved it so nothing is blocking the merge :)

thiagoelg:
Can https://github.com/kiegroup/kogito-tooling/pull/779 be a fix for the BPMN files not loading?

PR #515, opened 2021-05-21
Helpers should wait for elements before finding them

domhanak:
Uses drivers wait() function and until object to wait for elements in test helper methods.

WIP


domhanak:
Closing for now.

PR #516, opened 2021-05-21
KOGITO-3171: Do not show DMN guided tour for invalid models

jomarko:
For more details see:
- https://issues.redhat.com/browse/KOGITO-3171


jomarko:
Keeping in draft state, going to add cypress test for the change.

karreiro (approved):
Thank you, @jomarko. LGTM and works fine as well :-)

domhanak (approved):
LGTM, I see the instability in VSCode on ubuntu, I ll adress it in #515

jomarko:
@tomasdavidorg could you please advice how to proceed in this situation? I am:
- creating new asset in `SAMPLES_URL`
- and also using that asset in test

Should I firstly create separate PR, that just add an asset there?

tomasdavidorg:
@jomarko I see two options:

1. Create a separate PR for the broken DMN and rebase this PR after it is merged.
2. OR you can move the test to the UploadFileTest and use the broken DMN from the fixtures directory

jomarko:
@tomasdavidorg thank you, I moved it

packages/online-editor/it-tests/integration/UploadFileTest.ts
```diff
@@ -168,6 +168,30 @@ describe("Upload file test", () => {
...
+    });
+
+    // check editor title name
+    cy.get("[aria-label='Edit file name']").should("have.value", "testModelBroken");
+
+    // DMN guided tour dialog can't be shown for invalid models
+    cy.get("[data-kgt-close='true']").should("not.exist");
+
+    cy.get("[data-ouia-component-id='OUIA-Generated-Alert-danger-1']").contains(
+      "Error opening file. You can edit it as text and reopen the diagram after you've fixed it."
```
tiagobento: I don't think it's a good idea to verify the contents of the alert, to be honest. Probably given a specific data-testid or data-ouiaId to this specific alert?

packages/online-editor/it-tests/integration/UploadFileTest.ts
```diff
@@ -168,6 +168,30 @@ describe("Upload file test", () => {
     cy.get("#app p").should("be.visible").should("contain.text", "Welcome to Business Modeler!");
   });
 
+  it("should upload broken DMN file", () => {
```
tiagobento: This one should probably be "DMN Guided Tour popup shouldn't appear when opening broken file".

packages/online-editor/it-tests/integration/UploadFileTest.ts
```diff
@@ -168,6 +168,30 @@ describe("Upload file test", () => {
...
+    cy.get("[class='pf-c-brand']").within(($logo) => {
+      expect($logo.attr("src")).contain("dmn");
+      expect($logo.attr("alt")).contain("dmn");
+    });
+
+    // check editor title name
+    cy.get("[aria-label='Edit file name']").should("have.value", "testModelBroken");
+
+    // DMN guided tour dialog can't be shown for invalid models
+    cy.get("[data-kgt-close='true']").should("not.exist");
```
tiagobento: This selector is not very indicative of a Guided Tour component. Probably add `data-testid` or `data-ouiaId`?

packages/online-editor/it-tests/integration/UploadFileTest.ts
jomarko: could you please clarify more what you mean? I use already `data-ouia-component-id`

packages/online-editor/it-tests/integration/UploadFileTest.ts
jomarko: I am fine with that, will apply

packages/online-editor/it-tests/integration/UploadFileTest.ts
jomarko: I think I got it

packages/online-editor/it-tests/integration/UploadFileTest.ts
jomarko: The issue I see here is we use `ModalBoxCloseButton` patternfly component, that doesn't export `ouiaId` prop, while its nested patternfly `Button` component has such prop, unfortunately auto generated, as no value is passed from `ModalBoxCloseButton`.


So I think if I would change to `<ModalBoxCloseButton data-ouia-component-id="dmn-guided-tour-close" onClose={onCloseAction} />` still the auto generated `ouia-id` will be used.

Next I think if I would change `data-kgt-close` to `data-testid` in `GuidedTourContent.tsx` for example, it would end up with two `data-` attributes with testing purpose, what is also not ideal and basically the same as current state.

Ideally I would like to set `ouiaId` for `ModalBoxCloseButton` directly, but as React newbie, not sure how to do it.

I think options that I am able to do
- rename `data-kgt-close` to sth better
- replace `data-kgt-close` with `data-testid`
- add `data-ouia-component-id` to some parent div of the tour and than have locators like `[@data-ouia-component-id='dmn-guided-tour']/button[contains('close')]`

What would you suggest @tiagobento ?

tiagobento: I think 
- add data-ouia-component-id to some parent div of the tour and than have locators like [@data-ouia-component-id='dmn-guided-tour']/button[contains('close')]

would be the best option.

jomarko:
@tiagobento changes applied

tiagobento (approved):

PR #513, opened 2021-05-18
KOGITO-3560: Kogito Tooling infrastructure refresh and alignment - Part 1

tiagobento:
Fixed issues on this PR:
```
KOGITO-4760 | Upgrade to Webpack 5
KOGITO-4761 | Upgrade to latest TypeScript 
KOGITO-4762 | Upgrade to latest LTS Node, Yarn and Lerna versions
KOGITO-5027 | Rename build:fast to build:dev 
KOGITO-5028 | Align React versions between libraries and kogito-tooling
KOGITO-5030 | Rename yarn run init to yarn bootstrap
KOGITO-5034 | Improve dev source-maps and IDE code navigation
KOGITO-4764 | Create a git hook to automatically lint and format the code when committing
```

This is a fairly big PR, so here's a breakdown of everything it includes:
- Configure ESLint and remove TSLint.
- Bump the majority of our dependencies, including Node 16, Webpack 5, TypeScript 4, Electron 12, Yarn 1.22, Spectron 14, React 17.
- Install Lerna along with Yarn during CI to not rely on `npx`.
- Change `yarn run init` to `yarn bootstrap`.
- Change `build:fast` to `build:dev`.
- Introduce `WEBPACK_minimize`, `WEBPACK_TS_LOADER_transpileOnly` environment variables to tweak the build and make it faster (specially for CI).
- Introduce the `run-script-if` package with a handy node executable to run a script depending on a value of an environment variable. Using it on `yarn lint` and `yarn test` to be able to skip those steps and make faster builds (specially for CI). This is now possible by setting the `KOGITO_TOOLING_BUILD_lint` and `KOGITO_TOOLING_BUILD_test` environment variables.
- Change `README.md` to have a list of all the custom environment variables we have on the Kogito Tooling build. All of them have defaults, so after cloning the repository, `yarn bootstrap && yarn build:{dev,prod}` should work normally.
- Use cached `node_modules` dir on `it-test.yml` Workflow.
- Remove Windows build of desktop from `it-test.yml` as there's a bug on Spectron that prevents the tests from correctly running. I preferred to do that than to have a test continuously failing.
- Stop using `npx` and started using `yarn` instead. This makes us be sure that we're not depending on uninstalled dependencies, as `npx` will fetch those on-demand, while `yarn` won't.
- Fix some `VSCode` strings as they should be `VS Code`.
[...]

PR #508, opened 2021-05-17
KOGITO-5134: Fix VSCode integration tests editor waiting mechanism

domhanak:
@tiagobento not sure about the `vsce` upgrade please advise on that one.

Issue: https://issues.redhat.com/browse/KOGITO-5134

Ready for review, I would like to keep the loading explicit like this as it proved to be the most stable but feel free to suggest improvements.

PR #519, opened 2021-06-01
Improve VS Code IT tests and remove `vscode-extension-tester-native`

tiagobento:
Worked fine locally. Let's see if it behaves well on the CI as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,6 @@
     "url-loader": "^4.1.1",
     "vsce": "^1.88.0",
     "vscode-extension-tester": "^4.0.3",
-    "vscode-extension-tester-native": "^3.0.2",
     "vscode-test": "^1.5.2",
     "vscode-uitests-tooling": "^2.0.2",
     "watch": "^1.0.2",
--- a/packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
+++ b/packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
@@ -50,7 +50,7 @@ export default class VSCodeTestHelper {
   /**
    * Loading timeout for editors.
    */
-  private readonly EDITOR_LOADING_TIMEOUT: number = 10000;
+  private readonly EDITOR_LOADING_TIMEOUT: number = 60000;
 
   /**
    * Handle for VSCode workbench.
--- a/packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
+++ b/packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
@@ -95,6 +95,7 @@ export default class VSCodeTestHelper {
    * @returns a promise that resolves to a SideBarView of the openned folder
    */
   public openFolder = async (absolutePath: string): Promise<SideBarView> => {
+    await sleep(5000);
     await this.workbench.executeCommand("Extest: Open Folder");
     const inputBox = await InputBox.create();
     await inputBox.setText(absolutePath);
@@ -146,7 +147,7 @@ export default class VSCodeTestHelper {
     // const input = await InputBox.create();
     // await input.selectQuickPick('KIE Kogito Editors');
     const webview = new WebView(this.workbench.getEditorView(), By.linkText(fileName));
-    await sleep(10000);
+    await this.waitUntilKogitoEditorIsLoaded(webview);
     return webview;
   };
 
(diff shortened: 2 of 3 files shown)


packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
```diff
@@ -95,6 +95,7 @@ export default class VSCodeTestHelper {
    * @returns a promise that resolves to a SideBarView of the openned folder
    */
   public openFolder = async (absolutePath: string): Promise<SideBarView> => {
+    await sleep(5000);
```
domhanak, line 98: I am against this, adding sleeps is not a way to go.

packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
```diff
@@ -146,7 +147,7 @@ export default class VSCodeTestHelper {
     // const input = await InputBox.create();
     // await input.selectQuickPick('KIE Kogito Editors');
     const webview = new WebView(this.workbench.getEditorView(), By.linkText(fileName));
-    await sleep(10000);
```
domhanak, line 149: how is this `sleep` here? I removed the sleep in #508

packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
domhanak: I think we can introduce a simple wait instead. Using driver to wait for some specific workbench element like we do when opening the file in test. Wdyt?

packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
tiagobento: I put it back because otherwise the tests weren't passing on #513.

packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
tiagobento: I agree. This was a test to see if we're missing a waiting mechanism at this specific part of the tests (the very beginning), and looks like we do... Also, please check a wait for 2s I put after the loading screen disappears, without that the tests fail as well. I think we could wait for the keyboard shortcuts icon to appear or something like that, WDYT?

tiagobento:
@domhanak Most important on this PR is that I'm removing `vscode-extension-tester-native`, as it was not being used anywhere and it fetched a transitive dependency of 150MB.

domhanak:
@tiagobento Sure, but still it is a `devDependency` so I do not see any harm done there. Or am I missing something from node/yarn world? :)

packages/vscode-extension-pack-kogito-kie-editors/it-tests/helpers/VSCodeTestHelper.ts
domhanak: Oh cool, thanks for the heads up. I was a little bit confused :D

package.json
```diff
@@ -112,7 +112,6 @@
     "url-loader": "^4.1.1",
     "vsce": "^1.88.0",
     "vscode-extension-tester": "^4.0.3",
-    "vscode-extension-tester-native": "^3.0.2",
     "vscode-test": "^1.5.2",
     "vscode-uitests-tooling": "^2.0.2",
```
domhanak, line 116: This has been removed in #508 why is it back?

domhanak:
@tiagobento Could you please send a PR just to remove the dependency and keep juggling the waits in this PR to not mix things? 
Should I expected some changes here or is it ready?

package.json
tiagobento: IDK TBH, must have been a merge conflict poorly resolved by me :D

tiagobento:
@domhanak If you don't mind leaving the wait(5000) on the beginning of the tests until we have a better solution, I think we can merge it. Otherwise we can close the PR and work on a definitive fix. I won't be able to do that in the near future, though, as I'm working on other stuff. WDYT?

About the devDependency, the problem is that it's a very big dependency (~150MB), and we don't need it at all. Also, it takes a while to "build" using `node-gyp` during `yarn boostrap`. Eder also reported that `nut-js` (a transitive dependency of vscode-extension-tester-native) doesn't work on macOS arm64, so another reason to remove it.

package.json
domhanak: Ok, let me know if you are fixing it here or there is a need for new PR. :)

domhanak:
@tiagobento Understand, thanks for the insight. Nothing against its removal, it is not used anymore so let's get rid of it.

domhanak (approved):
As discussed in PR, the extra `wait()` in specific area adds a bit of stability so let's keep them there for now.
Thanks for the cleanup of unnecessary deps

PR #530, opened 2021-06-08
Include Lerna version requirement for building the project on README.

tiagobento:


README.md
```diff
@@ -41,6 +41,8 @@ In Kogito Tooling [releases page](https://github.com/kiegroup/kogito-tooling/rel
 
 ## Build from source
 
+> **NOTE:** We recommend using Node 16 when developing this project. Our CI uses Node `16.2.0`, Yarn `1.22.10`, and Lerna `4.0.0`.
```
mareknovotny, line 44: I can't get version 1.22.10 of yarn installed with standard npm command.

The latest possible version is 1.22.5 with `npm install -g yarn`
see https://classic.yarnpkg.com/en/docs/install#debian-stable

mareknovotny: Lerna is included in package.json https://github.com/kiegroup/kogito-tooling/blob/master/package.json#L84, ~~so I don't think this requires to be mentioned, but good to know~~

mareknovotny: so on clean git cloned repository it fails even if you have installed NodeJs 16.2.0 and yarn 1.22.5:
```
 (master) yarn --version
1.22.5
mnovotny@monster-T580:~/projects/git/KOGITO/kogito-tooling$ 
 (master) node --version
v16.2.0

(master) yarn bootstrap
yarn run v1.22.5
$ lerna bootstrap
/bin/sh: 1: lerna: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Ginxo: we have the same issue on nightly. Since `npx` was removed from scripts, `lerna` is required to be installed on the system. I don't know what's the purpose for that but it seems to be more complicated from usability POV. From developer perspective I just want to execute `yarn/npm install/bootstrap/whatever` in order to start using a repo, which is a kind of standard to every JS project.

mareknovotny:
@tiagobento i sent PR https://github.com/kiegroup/kogito-tooling/pull/531 to your branch for this PR, so you can now see what I talked about and how i expect the help for anybody trying to build it from source should be.

README.md
tiagobento: The tool responsible for downloading dependencies on a monorepo is not Yarn, but instead, it's Lerna (which uses Yarn behind the scenes, as it's configured). When using `npx` on a package that's not installed (globally or locally), `npm` fetches it and executes it once, and this step was causing some of our CI jobs to timeout. IDK how GH Actions handles npx downloads, but I'm guessing there's something different there. By having Lerna previously installed, we didn't observe this problem anymore.

To be honest, I think Lerna should not be on our package.json, as it's a tool that needs to be installed prior to bootstrapping the repository.

tiagobento:
@mareknovotny Thanks Marek, good suggestions. Merged the PR. As soon as I got your approval here, I'll merge this one too.

mareknovotny (approved):

PR #533, opened 2021-06-10
Disable Workspace Trust on VS Code Extension IT tests

tiagobento:
VS Code 1.57.0 is out and the new Workspaces Trust feature opens a popup when VS Code opens a new workspace. This caused our IT tests to fail. Since we trust the folders we're opening during the VS Code Extensions IT tests, disabling Workspaces Trust is not a problem.


caponetto (approved):

domhanak (approved):
Thanks @tiagobento !

PR #543, opened 2021-06-28
KOGITO-5058: Change default branch to `main`

tiagobento:


.github/workflows/prerelease.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - name: Create Draft Release
         id: create_release
-        uses: actions/create-release@master
+        uses: actions/create-release@v1.1.4
```
domhanak, line 28: can we actually control this? imho the action should always be `master` so that we avoid issues in the future. It is not our fault they still use master branch, is it?. Not sure we want to change these. Could you please clarify?.

tiagobento: Depending on mutable versions is not a good practice, as someone who's not us might affect the execution of our workflows. When we created these workflows, probably there were no tags available yet.

domhanak: Oh sure, did not think about it that way. Thanks for clarification, much appreciated. :)

ederign (approved):

PR #556, opened 2021-07-13
BAQE-1821: Introduce mocha-multi-reporters to allow test management

domhanak:
Adds mocha-multi-reportes devDependency of 1.5.1 version.
Configures xunit and spec reporters for vscode IT tests.
Xunit outputs results to XML file

Like this we will have console outputs and also XML report to consume.


jstastny-cz (approved):
LGTM.

tiagobento (approved):

PR #562, opened 2021-07-28
KOGITO-2313: Do not check for number of properties

domhanak:
This check is too strict and destabilizes the test suite
@jstastny-cz quick fix for KOGITO-2313 that adds metadata attributes to all activities.


jstastny-cz (approved):
LGTM.

tiagobento (approved):

PR #565, opened 2021-07-29
FAI-544: Score Cards: Model Setup switches not working

manstis:
See https://issues.redhat.com/browse/FAI-544

The issue was:

1. The same `div` was [used](https://github.com/kiegroup/kogito-tooling/blob/main/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CorePropertiesTable.tsx#L155-L163) to wrap both the "read-only" and "editing" content. The `onClick` handler called `event.stopPropagation()` that prevented the `Switch` from receiving click events.

2. A further consequence of using the same `div` is that if I removed the call to `event.stopPropagation()` then the event first triggered the move to edit state but was also then received by child elements and, since they were not the `div` with the "Click Outside" registered handler the handler was invoked and cancelled the move to edit mode; thus rendering the whole table inoperable.

I have moved the `onClick` handler registration (to enter edit mode) to a `div` used for the "read only" view and the "Click Outside" callback to a different `div` used for edit mode. This is consistent with how our editable lists work - where we have one component for "read only" and another for "editing". 

All problems are resolved :1st_place_medal:

### DIFF
--- a/packages/pmml-editor/tests/editor/components/EditorScorecard/organisms/CorePropertiesTable.test.tsx
+++ b/packages/pmml-editor/tests/editor/components/EditorScorecard/organisms/CorePropertiesTable.test.tsx
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { render } from "@testing-library/react";
+import * as React from "react";
+import { CorePropertiesTable } from "../../../../../src/editor/components/EditorScorecard/organisms";
+import { Operation, OperationContext } from "../../../../../src/editor/components/EditorScorecard";
+
+const commit = jest.fn(() => {
+  /*NOP*/
+});
+
+describe("CorePropertiesTable", () => {
+  test("render::Editing", () => {
+    const component = (
+      <OperationContext.Provider
+        value={{
+          activeOperation: Operation.UPDATE_CORE,
+          setActiveOperation: (operation) => {
+            /*NOP*/
+          },
+        }}
+      >
+        <CorePropertiesTable
+          modelIndex={0}
+          isScorable={true}
+          functionName={"regression"}
+          algorithmName={"algorithmName"}
+          baselineScore={1}
+          isBaselineScoreDisabled={false}
+          baselineMethod={"other"}
+          initialScore={2}
+          areReasonCodesUsed={true}
+          reasonCodeAlgorithm={"pointsBelow"}
+          commit={commit}
+        />
+      </OperationContext.Provider>
+    );
+
+    const { getByTestId, rerender } = render(component);
+    const container = getByTestId("core-properties-table");
+    expect(container.children[0].className.toString()).toContain("editable-item");
+    expect(container).toMatchSnapshot();
+
+    container.click();
+    rerender(component);
+    const containerEditing = getByTestId("core-properties-table");
+    expect(containerEditing.children[0].className.toString()).toContain("editable-item--editing");
+    expect(containerEditing).toMatchSnapshot();
+  });
+
+  test("render::Not editing", async () => {
+    const component = (
+      <CorePropertiesTable
+        modelIndex={0}
+        isScorable={true}
+        functionName={"regression"}
+        algorithmName={"algorithmName"}
+        baselineScore={1}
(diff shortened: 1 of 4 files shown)


manstis:
@tiagobento Is the macos build failure of `vscode-extension-pack-kogito-kie-editors` a known issue?

It's failed twice now.

tiagobento:
@manstis I guess you can say this is known... and VS Code keeps updating regularly... we have waves of failures on VS Code Extension IT tests... not ideal. But overall, tests are much less stable on macOS and Windows, for some reason. @domhanak Do you have any idea what can be causing these failures?

kelvah (approved):
LGTM

packages/pmml-editor/tests/editor/components/EditorScorecard/organisms/CorePropertiesTable.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+        />
+      </OperationContext.Provider>
+    );
+
+    const { getByTestId, rerender } = render(component);
+    const container = getByTestId("core-properties-table");
+    expect(container.children[0].className.toString()).toContain("editable-item");
+    expect(container).toMatchSnapshot();
+
+    container.click();
```
sjamboro, line 61: @manstis IMHO, this test just click somewhere in Model Setup and check that it is in the edit mode. It does not test if an user can modify "Is scorable" and "Use Reason Code".

sjamboro:
@manstis it seems that MacOS does not show predicate in `Wrong Data Field in Predicate` (action: open predicate from uploaded file). Do you think that it should be caused by delay? It seems that test `Rewrite predicate to compound` passed even if there is same assertion.

manstis:
Hi @sjamboro I've pushed a new commit with improved tests.

It's possible the timeout could be increased for MacOS; the failure report is:
```
@kogito-tooling/pmml-editor:   1 failing
@kogito-tooling/pmml-editor:   1) Predicate Test
@kogito-tooling/pmml-editor:        Use predefined predicate
@kogito-tooling/pmml-editor:          Wrong Data Field in Predicate:
@kogito-tooling/pmml-editor:       Timed out retrying after 4000ms
```
Was you able to access a video of the failing test? 

The log also states:
```
@kogito-tooling/pmml-editor:   (Screenshots)
@kogito-tooling/pmml-editor:   -  /Users/runner/work/kogito-tooling/kogito-tooling/packages/pmml-editor/dist-it-te     (1132x644)
@kogito-tooling/pmml-editor:      sts/screenshots/predicate.ts/Predicate Test -- Use predefined predicate -- Wrong               
@kogito-tooling/pmml-editor:       Data Field in Predicate (failed).png                                                          
@kogito-tooling/pmml-editor:   (Video)
@kogito-tooling/pmml-editor:   -  Started processing:  Compressing to 32 CRF                                                     
@kogito-tooling/pmml-editor:     Compression progress:  100%
@kogito-tooling/pmml-editor:   -  Finished processing: /Users/runner/work/kogito-tooling/kogito-tooling/packages/p   (10 seconds)
@kogito-tooling/pmml-editor:                           mml-editor/dist-it-tests/videos/predicate.ts.mp4
```

sjamboro:
@manstis in artefacts

sjamboro:
@manstis now another test is failing on mac. Should I create new jira to increase timeout before you merge this PR?

On windows, online-editor tests are failing (timeouted).

manstis:
@sjamboro If it's something I can configure in this PR I'm happy to do so.

Where would I need to set a timeout parameter to a larger value?

sjamboro:
@manstis in https://github.com/kiegroup/kogito-tooling/blob/main/packages/pmml-editor/it-tests/cypress.json add `"defaultCommandTimeout": 10000`

More details are here: https://docs.cypress.io/guides/references/configuration#Timeouts

sjamboro:
@manstis **timeouts for online-editor** are defined in code. For example see https://github.com/kiegroup/kogito-tooling/blob/main/packages/online-editor/it-tests/support/commands.ts#L39


I would rerun PR checks

manstis:
Thanks @sjamboro. Timeout increased... let's see what happens now!

sjamboro (approved):

tiagobento:
@sjamboro @manstis Timeouts are usually not the issue. GitHub Actions machines are slow, but not to the point where it takes minutes to load something. When a timeout happens, it's usually an indicator that the application is in different state than the tests are expecting, and since none of the two knows about that, they keep waiting on each other for something to change. And a timeout happens. 

In the past, I've seen a lot of *very* specific corner cases that are only caught by automated tests, and this is very hard to catch upfront. (e.g. I fixed a test where the reproducer was to click on my mouse button very fast at the "Save" button on the online editor). 

There's little to none difference between Windows, macOS and Linux tests in the browser tests. VS Code extension is the most problematic and we never had a stable-enough test suite on Windows and macOS for it. Please feel free to accept the PR if the tests are passing in one of the three platforms. 

We need to invest more time in stabilizing the tests. We can even consider disabling the test suites on macOS and Windows for the time being, as we're not benefiting from them existing right now.

@domhanak @tomasdavidorg @jstastny-cz @caponetto Please tell me if you agree and I'll send a PR leaving the IT tests running only on Linux.

manstis:
@tiagobento @ederign I think this PR is as "green" as we're going to get it.

tiagobento (approved):

PR #579, opened 2021-08-06
KOGITO-4107: Set Cypress baseUrl in package.json in Standalone Editors ITs

jstastny-cz:
https://issues.redhat.com/browse/KOGITO-4107

Goal here is to remove dependencies on internal packages from kie-editors-standalone/it-tests. This then allows to easily run it-tests also against already published packages, just by providing separate package.json and webpack config. This is useful in CI environment to verify published bits.


jstastny-cz:
@tiagobento could you please advise on how to make the build-env resolve properly on windows in cy:* scripts?

packages/kie-editors-standalone/it-tests/src/components/FileLoader.tsx
```diff
@@ -16,13 +16,23 @@
...
+  TEXT = "text",
+  BINARY = "binary",
+}
+
+export type ResourceContent = {
+  path: string;
+  content?: string;
+  type: ContentType;
+};
+
```
tiagobento: This duplication is not good.. Why is this necessary?

packages/kie-editors-standalone/package.json
```diff
@@ -26,8 +26,8 @@
     "build:prod": "yarn lint && yarn test && rimraf dist && webpack --config webpack.build-resources.config.js && yarn run build:preprocessor && webpack --config webpack.package-resources.config.js && yarn test:it",
     "build:productization": "yarn run build:prod",
     "start": "webpack serve --host 0.0.0.0 --config webpack.package-resources.config.js",
-    "cy:run": "yarn cypress run -b chrome --project it-tests",
-    "cy:open": "yarn cypress open --project it-tests",
+    "cy:run": "yarn cypress run -b chrome --project it-tests --config \"baseUrl=http://localhost:$(build-env standaloneEditors.dev.port)\"",
+    "cy:open": "yarn cypress open --project it-tests --config \"baseUrl=http://localhost:$(build-env standaloneEditors.dev.port)\"",
```
tiagobento: Great solution. However, this still depends on `build-env` to work..

tiagobento (changes requested):
@jstastny-cz Hi Jan, thanks for the PR! If the goal is to test published versions of `@kogito-tooling/kie-editors-standalone`, I think the best approach would be:

1. Cloning the tag of the published package you want to test.
2. Build this tag with `lerna run build:prod --scope=@kogito-tooling/kie-editors-standalone --include-dependencies`
3. Make the changes you need on `package.json` and `webpack.config.js` so that we actually point to the published version, and not the version we just built.
4. Run the tests.

Then you won't have to worry about the discrepancies between the published version you're testing and `main`. WDYT?

If the goal is to run `main` tests against a published version to catch discrepancies and/or breaking changes, then I think it you should be checking out `main` instead of a specific tag and doing the exact same steps described above.

Please tell me if I'm missing something.

tiagobento:
Answering your question about Windows. GitHub Actions runs commands by default on `pwsh`, but `yarn` runs them on CMD. So..

`yarn build:prod` script runs on `pwsh`, then the contents of the `build:prod` scripts run on `CMD`.

`CMD`, however, doesn't support sub-expressions syntax with `$(expr)`, but `pwsh` does. To use `pwsh`, You can copy the approach we did on our Maven packages. See https://github.com/kiegroup/kogito-tooling/blob/main/packages/backend-extended-services/package.json#L18

packages/kie-editors-standalone/package.json
jstastny-cz: Correct, but it's easy to reuse the spec files this way - the test specs have no scent of it now. I just provide different package.json and webpack.js config to use a pre-defined port (so env config is scoped just to configuration files, not tests themselves).

packages/kie-editors-standalone/it-tests/src/components/FileLoader.tsx
jstastny-cz: This is a copy of an internal type (previously imported from common api), it's used in the testing app internally (to handle file upload/download).
Why the use of it? In testing app there is a FileLoader component, which holds the models imported. For easier invocation of the standalone editors API endpoints, it uses the same structures as the implementation.

jstastny-cz:
@tiagobento thanks for your insights.

The build approach you've proposed I'm a bit concerned about due to the fact that it expects invoking the it-tests in both cases from same folder, thus sharing node_modules directory. I currently run the tests from kie-editors-standalone/it-tests, so running the tests really in isolation, so it on purpose has no access to parent dir `../node_modules` directory, to prevent using different version than the tested one by mistake.
And from recent experience, I'd like to refrain from close alignment of the configuration with the current upstream project. So the only shared part here would be just the testing app, tests and cypress config. While in my case using the `--config baseUrl=XY` gives me opportunity to easily override this from my custom package.json.

So first I would like to try the powershell approach you've mentioned. Then I am happy to discuss the rest for sure, we can even set up a call where I can walk you through the config I use.

tiagobento:
@jstastny-cz As we spoke, this PR will be reassessed once #582 is merged.

jstastny-cz:
@tiagobento I've limited the changes just to pass the baseUrl configuration property from package.json. With that goes the issue you've mentioned with powershell vs. CMD, so I've applied the workaround you've proposed. Ready for review again.

tiagobento (approved):

PR #582, opened 2021-08-10
Reorganize packages to fix Standalone Editors transitive type definitions

tiagobento:


jstastny-cz (approved):
Changes look alright. I leave to @Ginxo to assess if this fits the release workflow.

PR #597, opened 2021-08-18
PMML Editor: Remove relative paths to node_modules on SCSS files

tiagobento:
Relative paths break `kie-tooling-store`, as the dir structure is different.

### DIFF
--- a/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/AttributesTable.scss
+++ b/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/AttributesTable.scss
@@ -13,8 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@import "../../../../../../../node_modules/@patternfly/patternfly/sass-utilities/all";
-@import "../../../../../../../node_modules/@patternfly/patternfly/base/variables";
+@import "@patternfly/patternfly/sass-utilities/all";
+@import "@patternfly/patternfly/base/variables";
 
 .attributes {
   &__header {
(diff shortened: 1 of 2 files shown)


packages/pmml-editor/src/editor/components/EditorScorecard/organisms/AttributesTable.scss
```diff
@@ -13,8 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@import "../../../../../../../node_modules/@patternfly/patternfly/sass-utilities/all";
-@import "../../../../../../../node_modules/@patternfly/patternfly/base/variables";
+@import "@patternfly/patternfly/sass-utilities/all";
+@import "@patternfly/patternfly/base/variables";
```
kelvah: I don't understand the purpose of these imports to be honest :D
The base variables are already included inside PF base styles and I don't see any sass utility usage.
I see they were added to solve some package build issue though, so they probably have a reason to be that is beyond me.

If they have to stay, could you add a `~` a the beginning of the import path? Otherwise IDEs cannot resolve them. Like: `@import "~@patternfly/patternfly...";`

kelvah (commented):
Hi @tiagobento! Left a comment for you.

packages/pmml-editor/src/editor/components/EditorScorecard/organisms/AttributesTable.scss
tiagobento: @kelvah Great suggestion. Updated the paths to include `~`. Worked well on IDEs and on `kie-tooling-store` as well.

tiagobento:
@kelvah This is ready for review again :)

kelvah (approved):

PR #602, opened 2021-08-20
KOGITO-5721: Parse date objects on formInputs

caponetto:
See [KOGITO-5721](https://issues.redhat.com/browse/KOGITO-5721)

The form only accepts `date-time` as objects, not strings.
Since `JSON.parse` does not transform date strings to date objects, we need to provide a function that does such a thing.

Keep in mind that, the "time" part is still not being filled up in the form, but this is probably related to [KOGITO-5722](https://issues.redhat.com/browse/KOGITO-5722).

### Before:
KOGITO-5721-before

### After:
KOGITO-5721

### DIFF
--- a/packages/online-editor/src/__tests__/common/utils.test.ts
+++ b/packages/online-editor/src/__tests__/common/utils.test.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { extractEditorFileExtensionFromUrl } from "../../common/utils";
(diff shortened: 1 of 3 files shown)


tiagobento:
@tomasdavidorg Can you please take a look at this one?

ljmotta (approved):
Great solution!

tomasdavidorg (approved):

PR #605, opened 2021-08-23
FAI-546: ScoreCard Model Setup Test

sjamboro:
See: https://issues.redhat.com/browse/FAI-546

PR contains a simple UI test suite which modify basics model attributes.

### DIFF
--- a/packages/pmml-editor/it-tests/integration/scorecard-model.ts
+++ b/packages/pmml-editor/it-tests/integration/scorecard-model.ts
@@ -25,7 +25,7 @@ describe("Scorecard Model Test", () => {
     cy.ouiaId("model-name").click();
     cy.ouiaId("set-model-name").clear().type("Unique Scorecard Name");
 
-    cy.ouiaType("filler").click();
+    cy.ouiaType("filler").first().click();
 
     cy.ouiaId("model-name").should("be.visible").should("contain", "Unique Scorecard Name");
     cy.buttonPMML()
@@ -40,7 +40,7 @@ describe("Scorecard Model Test", () => {
     cy.ouiaId("model-name").click();
     cy.ouiaId("set-model-name").clear().type("Unique Scorecard Name");
 
-    cy.ouiaType("filler").click();
+    cy.ouiaType("filler").first().click();
 
     cy.ouiaId("model-name").click();
     cy.ouiaId("set-model-name").clear().type("Something{esc}");
@@ -61,7 +61,7 @@ describe("Scorecard Model Test", () => {
     cy.ouiaId("is-scorable").click();
     cy.ouiaId("is-scorable").find("input").should("not.to.be.checked");
 
-    cy.ouiaType("filler").click();
+    cy.ouiaType("filler").first().click();
 
     cy.ouiaType("model-setup-overview")
       .find("[data-ouia-component-type=model-property]")
@@ -91,7 +91,7 @@ describe("Scorecard Model Test", () => {
     cy.ouiaId("is-scorable").click();
     cy.ouiaId("is-scorable").find("input").should("be.checked");
 
-    cy.ouiaType("filler").click();
+    cy.ouiaType("filler").first().click();
 
     cy.ouiaType("model-setup-overview")
       .find("[data-ouia-component-type=model-property]")
@@ -121,7 +121,7 @@ describe("Scorecard Model Test", () => {
     cy.ouiaId("core-reasonCodeAlgorithm").find("button").should("be.disabled");
     cy.ouiaId("core-baselineMethod").find("button").should("be.disabled");
 
-    cy.ouiaType("filler").click();
+    cy.ouiaType("filler").first().click();
 
     cy.ouiaType("model-setup-overview")
       .find("[data-ouia-component-type=model-property]")
@@ -149,7 +149,7 @@ describe("Scorecard Model Test", () => {
     cy.ouiaId("core-reasonCodeAlgorithm").find("button").should("be.enabled");
     cy.ouiaId("core-baselineMethod").find("button").should("be.enabled");
 
-    cy.ouiaType("filler").click();
+    cy.ouiaType("filler").first().click();
 
     cy.ouiaType("model-setup-overview")
       .find("[data-ouia-component-type=model-property]")
@@ -185,7 +185,7 @@ describe("Scorecard Model Test", () => {
     cy.ouiaType("edit-model-setup").should("be.visible");
     cy.ouiaId("core-functionName").find("button").should("be.disabled").should("contain", "regression");
 
-    cy.ouiaType("filler").click();
+    cy.ouiaType("filler").first().click();
 
     cy.ouiaType("model-setup-overview")
       .find("[data-ouia-component-type=model-property]")
@@ -211,7 +211,7 @@ describe("Scorecard Model Test", () => {
     cy.ouiaType("edit-model-setup").should("be.visible");
     cy.ouiaId("initial-score").type("-15.5");
 
-    cy.ouiaType("filler").click();
+    cy.ouiaType("filler").first().click();
 
     cy.ouiaType("model-setup-overview")
       .find("[data-ouia-component-type=model-property]")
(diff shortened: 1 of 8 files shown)


tomasdavidorg (approved):

packages/pmml-editor/it-tests/integration/scorecard-model.ts
```diff
@@ -0,0 +1,490 @@
...
+      });
+
+    cy.ouiaType("model-setup-overview")
+      .find("[data-ouia-component-type=invalid-label]")
+      .should(($label) => {
+        expect($label).to.have.length(1);
+        expect($label[0]).to.have.text("Baseline score:\u00A0Missing");
+      });
+
+    /**
```
tomasdavidorg: Just a comment: Why do you use JSDoc comments `/**` instead of regular multi line comments `/*`?

sjamboro: Thanks, I should have used `/*` instead of `/**`. At the end I have decided to remove this comment and mention it in JS Doc comment of whole test suite.

sjamboro:
@manstis @tiagobento @ederign can you review and merge this PR? All checks have passed.

packages/pmml-editor/it-tests/integration/scorecard-model.ts
```diff
@@ -0,0 +1,497 @@
...
+        expect($label[1]).to.have.text("Function:\u00A0regression");
+        expect($label[2]).to.have.text("Use Reason Codes:\u00A0Yes");
+        expect($label[3]).to.have.text("Reason Code Algorithm:\u00A0pointsBelow");
+        /*
+         * The label contains different string when is invalid:
+         * - invalid label - "Baseline score:"
+         * - valid label - "Baseline Score:"
+         * TODO: reach an agreement about functionality.
+         */
+        expect($label[4]).to.have.text("Baseline Score:\u00A0-15.5");
```
sjamboro, line 277: @manstis: Can I correct a text of invalid label to "Baseline Score"? (to unify invalid and valid text of labels) It seems that other labels contain capital characters.

packages/pmml-editor/it-tests/integration/scorecard-model.ts
```diff
@@ -0,0 +1,497 @@
...
+      .should(($label) => {
+        expect($label).to.have.length(6);
+        expect($label[0]).to.have.text("Is Scorable:\u00A0Yes");
+        expect($label[1]).to.have.text("Function:\u00A0regression");
+        /*
+         * This attribute is not required by PMML definition.
+         * If its value is the empty string then the particular label should be removed.
+         * TODO: reach an agreement about functionality.
+         */
+        expect($label[2]).to.have.text("Algorithm:\u00A0");
```
sjamboro, line 351: @manstis: What is the expected behavior of the "Algorithm" attribute? Other labels disappear when its text is empty.

tiagobento:
Is this good to merge? I see there are comments unresolved.

packages/pmml-editor/it-tests/integration/scorecard-model.ts
manstis: Hi @sjamboro Yes you can do this as part of this PR, if you like.

The "Characteristics" section of the editor should also probably be corrected as it only capitalises the first letter.

Screenshot from 2021-09-02 10-07-49

I had a quick check of the "Data Dictionary", "Mining Schema" and "Outputs". They seem to capitalise each word too, but there might be some edge cases you can catch and also correct.

Of course, none of these corrections are really in scope of this PR.. so feel free to make a JIRA for me instead if you prefer.

packages/pmml-editor/it-tests/integration/scorecard-model.ts
manstis: Hi @sjamboro the _expected_ behaviour is to be consistent with other optional labels; i.e. if the property has no value then do not render it when in "view" mode (i.e. we don't show the coloured Pattern Fly Label component). When in "edit" mode we always show a UI component.

packages/pmml-editor/it-tests/integration/scorecard-model.ts
sjamboro: @manstis ok, I am going to correct labels in model setup here. I will focus on these issues during cypress test preparation and unify all labels.

packages/pmml-editor/it-tests/integration/scorecard-model.ts
sjamboro: @manstis so Attribute label has to be fix. Can I correct it here?

manstis: @sjamboro The editor looks OK (on `kogito-group/kogito-tooling/main`).. see [here](https://user-images.githubusercontent.com/497544/131816806-ff5393b1-93d4-43da-bd71-c32e7ae89151.png). The "Algorithm" attribute is not shown when in "view" mode. Is the new test finding it somewhere?! Let me checkout this branch and check.

manstis: @sjamboro OK, so I can replicate a "Algorithm" label with no text.. enter a value in the edit mode and then delete it. The value becomes an empty `String` and no longer `undefined`. We only check for if the value is `undefined` when rendering the component and hence it is rendered along with an empty `String` as it's value.

There are probably other places we write back an empty `String` to the model so a new JIRA to fix it across the whole editor is probably the best solution (and this test can be fixed then too).. or fix here. I don't mind. Your choice!

manstis: This specific case can either be fixed [here](https://github.com/kiegroup/kogito-tooling/blob/main/packages/pmml-editor/src/editor/reducers/ScorecardReducer.ts#L88) (by setting to `undefined` if an empty `String`) or [here](https://github.com/kiegroup/kogito-tooling/blob/main/packages/pmml-editor/src/editor/components/EditorScorecard/organisms/CorePropertiesTable.tsx#L268) (by setting the value passed to `setAlgorithmName(..)` to `undefined` if `e === undefined`).

manstis: @sjamboro Oh, BTW, please let me know what approach you decide to do (fix on this PR or a new JIRA etc) so I can pick up and fix if and as required. I won't make a JIRA until I hear what you want to do to avoid us duplicating it.

manstis:
@tiagobento Please do not merge at the moment.

packages/pmml-editor/it-tests/integration/scorecard-model.ts
sjamboro: @manstis please check it. I have corrected the invalid label to "Baseline Score" and removed TODO comment. All test suites passed on my laptop.

packages/pmml-editor/it-tests/integration/scorecard-model.ts
```diff
@@ -0,0 +1,491 @@
...
```
sjamboro, line 351: @manstis You have better knowledge of code. Please create new JIRA where you fix it. There is suggestion how you could fix the "Attribute 'Algorithm' - set and unset" test when new behaviour will be implemented:
```suggestion
        expect($label).to.have.length(5);
        expect($label[0]).to.have.text("Is Scorable:\u00A0Yes");
        expect($label[1]).to.have.text("Function:\u00A0regression");
        expect($label[2]).to.have.text("Use Reason Codes:\u00A0Yes");
        expect($label[3]).to.have.text("Reason Code Algorithm:\u00A0pointsBelow");
        expect($label[4]).to.have.text("Baseline Method:\u00A0other");
```

manstis: OK, see https://issues.redhat.com/browse/FAI-558. I'll work on it next week.

manstis (approved):
Thank-you for your patience @sjamboro.

manstis:
@ederign @tiagobento This is OK to merge whenever you like now.

PR #614, opened 2021-09-02
BAQE-1828: Modify npm scripts to use settings.xml with maven builds

domhanak:
Adds new variable that can be set during builds, defaults to settings in .m2 directory of user's home. 

NOTE
This does not set the repository URL or version that wrapper will be using as having a correct settings.xml overrides following properties.
These can be set by using `-Dmaven=3.8.2` to set correct maven and if we use `export MVNW_REPOURL=path` the correct repo will be used.
One culprit for Jenkins is that `maven` is a ENV variable there already but has `maven-3.8.2` value.


domhanak:
Closing this for now, will reopen after we find the best solution.

PR #618, opened 2021-09-07
KOGITO-5838: Update vscode-extension-tester to 4.2.0

domhanak:
Raise version to 4.2.0.
Modify the way we open folders according to docs


jomarko (approved):
@domhanak thank you, this change seems to be reasonable to me.

tiagobento (approved):

PR #619, opened 2021-09-08
FAI-529: Score Cards: Data Dictionary: Remove duplication of delete icons

manstis:
See https://issues.redhat.com/browse/FAI-529

@sjamboro What type of tests would you like to see for what is largely a CSS related change?

Characteristics are covered by the `predicate.ts` integration test.

I'll see if I can put together some form of unit test.


packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaAddFields/MiningSchemaAddFields.tsx
```diff
@@ -13,9 +13,10 @@ import { MiningSchemaOption } from "../MiningSchemaContainer/MiningSchemaContain
 interface MiningSchemaAddFieldsProps {
   options: MiningSchemaOption[];
   onAdd: (fields: string[]) => void;
+  isDisabled: boolean;
```
manstis, line 16: @kelvah I disable the "Add (Mining) Field" dropdown and button when in "edit mode" for consistency with what already happens with the "Data Dictionary" and "Outputs" dialogs.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaFields/MiningSchemaFields.tsx
```diff
@@ -97,63 +92,118 @@ const MiningSchemaItem = (props: MiningSchemaFieldProps) => {
...
-          <SplitItem>
-            <Button variant="plain" onClick={deleteField}>
-              <TrashIcon />
-            </Button>
-          </SplitItem>
-        </Split>
-      </section>
-    </li>
+    <>
+      {index === editing && (
```
manstis, line 111: @kelvah I changed rendering of a "Mining Schema Field" so that the "delete" icon is not shown when in "edit mode" for consistency with what already happens for "Data Dictionary", "Outputs" and "Characteristics".

tiagobento:
Reminder: Our CI doesn't run on draft PRs.

manstis:
@sjamboro @kelvah I've added unit tests. This PR is ready for review. Thank-you.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaFields/MiningSchemaFields.tsx
kelvah: I think you are right about bringing some consistency to the Mining Schema section.
I was uncertain about how to deal with it at the beginning because it has some differences.
Anyway, I think it's better to keep the same behavior. We will eventually revisit the topic when we'll work on the dialog-to-tab change.

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
```diff
@@ -80,8 +82,22 @@ const DataDictionaryContainer = (props: DataDictionaryContainerProps) => {
...
 
-  const handleDelete = (index: number) => {
+  const handleDelete = (index: number, interaction: Interaction) => {
     onDelete(index);
+    if (interaction === "mouse") {
+      //If the DataTypeItem was deleted by clicking on the delete icon we need to blur
+      //the element otherwise the CSS :focus-within persists on the deleted element.
+      if (document.activeElement instanceof HTMLElement) {
+        document.activeElement?.blur();
+      }
```
kelvah: @manstis your solution is already working fine but I was trying to understand what caused the issue.

If one item gets deleted, it _should_ be removed from the DOM, so there shouldn't be any focused delete buttons :thinking: 

I think that using only the `index` as `key` [for DataTypeItems](https://github.com/kiegroup/kogito-tooling/pull/619/files#diff-f4eb22a5a613dc20e41674456831cb8e40672143d671349299ef568afac901e4R257) is causing some re-render trouble. When deleting the item n.3, it's the item n.4 that gets actually removed from the DOM, while the item at position `3` gets only its properties updated (with the values of the ex n.4, which is moving up in the list). 

So a possible solution could be to use something different as key, like concatenating index and item name, in order to force the re-render on the interested subjects. What you think?

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
```diff
@@ -80,8 +82,22 @@ const DataDictionaryContainer = (props: DataDictionaryContainerProps) => {
...
+    } else if (interaction === "keyboard") {
+      //If the DataTypeItem was deleted by pressing enter on the delete icon when focused
+      //we need to set the focus to the next DataTypeItem. The index of the _next_ item
+      //is identical to the index of the deleted item.
+      setDataTypeFocusIndex(index);
```
kelvah, line 98: This is a really nice idea :+1:

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
```diff
@@ -160,8 +176,18 @@ const DataDictionaryContainer = (props: DataDictionaryContainerProps) => {
...
   }, [dataDictionary, editing]);
 
+  //Set the focus on a DataTypeItem as required
+  useEffect(() => {
+    if (dataTypeFocusIndex !== undefined) {
+      document.querySelector<HTMLElement>(`#data-type-item-n${dataTypeFocusIndex}`)?.focus();
+    }
+    return () => {
+      setDataTypeFocusIndex(undefined);
+    };
```
kelvah: I think we have no interest in resetting the focus index when the component gets unmounted. Or maybe I am missing something?

packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
```diff
@@ -174,14 +174,16 @@ const DataTypeItem = (props: DataTypeItemProps) => {
 
   return (
     <article
+      id={`data-type-item-n${index}`}
```
kelvah, line 177: Really small detail, sorry. We have both an id and a class with the same name. Do we need both?

kelvah (commented):
Thank you @manstis. I've left you some comments.

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
manstis: See https://issues.redhat.com/browse/FAI-570 as discussed.

Ideally we need a primary key for model collection elements but we might be able to find alternative approaches.

packages/pmml-editor/src/editor/components/DataDictionary/DataTypeItem/DataTypeItem.tsx
manstis: I'll keep the `id` as we don't use CSS consistently across the different "lists" of items.

Keeping the `id` makes the implementation of the behaviour for this JIRA/PR consistent.

packages/pmml-editor/src/editor/components/DataDictionary/DataDictionaryContainer/DataDictionaryContainer.tsx
manstis: Removed from PR.

kelvah (approved):

packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributesTableAction.tsx
```diff
@@ -17,30 +17,41 @@ import * as React from "react";
...
+          className="attribute-item__delete"
+          ouiaId="delete-attribute"
           variant="plain"
-          onClick={onDelete}
-          onKeyDown={(e) => {
-            if (e.key === "Enter") {
-              onDelete(e);
+          onClick={(e) => handleDelete(e, "mouse")}
+          onKeyDown={(event) => {
+            if (event.key === "Enter") {
```
sjamboro, line 51: @manstis  I understand the idea: when I have focused this button and pressed enter then I delete item. In my point of view is better just "click" on button. It is very difficult to select this button to perform "Keydown event" - "enter". When I click on the delete button I delete it. I need to perform "left click" + "enter" or press severel time "tab" (focus attribute, focus predicate, focus delete predicate, focus detele attribute) and then "enter" to delete. Do you have some hint how use keyboard easily?

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributesTableRow.tsx
```diff
@@ -71,19 +72,25 @@ export const AttributesTableRow = (props: AttributesTableRowProps) => {
...
+      id={`attribute-n${attributeIndex}`}
+      data-testid={`attribute-n${attributeIndex}`}
+      className={"attribute-item"}
       onClick={() => onEdit()}
       onKeyDown={(e) => {
         if (e.key === "Enter") {
-          e.preventDefault();
-          e.stopPropagation();
-          onEdit();
+          handleEdit(e);
```
sjamboro, line 89: @manstis should be possible to press "delete" to perform ```handleDelete(event, "keyboard")``` and focus another item on the list?

manstis: No @sjamboro it should not be possible to press `delete` in this PR. Pressing `enter` when a HTML element has the focus is the standard way to replicate "click events" with the keyboard (see how any other button behaves on a web site). We could raise a JIRA for UX to comment on keyboard shortcuts of which `delete` deleting a row would be a consideration.

packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributesTableAction.tsx
manstis: @sjamboro Many users will no doubt "click" on the icon with the mouse pointer. However to maximise accessiblity for users with disabilities we need to also provide support for screen readers and keyboard operation. I agree it does make writing an integration test an interesting exercise however I do something similar in my unit tests.

See [here](https://github.com/kiegroup/kogito-tooling/pull/619/files#diff-70a00a6eca3cbc329def8b9db40e45d30a3c42d13397f3b406c852ba6108b18bR154) for an example test. I have to first [focus the row](https://github.com/kiegroup/kogito-tooling/pull/619/files#diff-70a00a6eca3cbc329def8b9db40e45d30a3c42d13397f3b406c852ba6108b18bR177-R180) then [re-render the component](https://github.com/kiegroup/kogito-tooling/pull/619/files#diff-70a00a6eca3cbc329def8b9db40e45d30a3c42d13397f3b406c852ba6108b18bR177-R180) (so the delete icon is visible) and then [find the delete icon](https://github.com/kiegroup/kogito-tooling/pull/619/files#diff-70a00a6eca3cbc329def8b9db40e45d30a3c42d13397f3b406c852ba6108b18bR200-R201) and emulate a [keyboard event](https://github.com/kiegroup/kogito-tooling/pull/619/files#diff-70a00a6eca3cbc329def8b9db40e45d30a3c42d13397f3b406c852ba6108b18bR202). The focus/keyboard event should be possible with an integration test but you would not have to re-render the component as it'd be running "live" under Cypress.

sjamboro: @manstis you mention important thing about the "delete" button visibility. We have to focus the row element. Is touchscreen supported or not? Do you have experience how it looks on touchscreen? Based on [this side](https://www.prowebdesign.ro/how-to-deal-with-hover-on-touch-screen-devices/) I am afraid that the user with touchscreen cannot delete any row. The "delete" icon is visible only if the row is focused or hovered.

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributesTableRow.tsx
sjamboro: @manstis I do not if there is missunderstood. I have no doub about pressing `enter` instead of `click`. That is correct. I mean allow `enter` and `delete` key. `enter` to edit row and `delete` to delete row. I understand if you need the new JIRA. I was not sure about scope of this JIRA.

packages/pmml-editor/tests/editor/components/EditorScorecard/organisms/AttributesTable.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+            onCommit={onCommit}
+          />
+        </OperationContext.Provider>
+      </Provider>
+    );
+    const container = getByTestId("attributes-table");
+    expect(container).toMatchSnapshot();
+
+    expect(getByTestId("attribute-n0")).not.toBeUndefined();
+    expect(getByTestId("attribute-n1")).not.toBeUndefined();
```
sjamboro, line 84: @manstis can you assert number of row or "attribute-n2" to be null?

sjamboro:
@manstis I am going to check that the delete button is visible and accesible via cypress test. I want to create new JIRAs (DD, Mining, Outpust, Characteristics). Your tests are ok. It would be nice to check the content of the row ("field1" and "field2") and missing icon without hover or focus.

packages/pmml-editor/src/editor/components/EditorScorecard/atoms/AttributesTableAction.tsx
manstis: See https://issues.redhat.com/browse/FAI-577. This comment does highlight a flaw in this whole approach.

packages/pmml-editor/src/editor/components/EditorScorecard/molecules/AttributesTableRow.tsx
manstis: See https://issues.redhat.com/browse/FAI-577. We could add support for `[enter]` putting a row into "edit mode".

sjamboro (approved):
Reviewed. Thank you @manstis. Related cypress tests: [FAI-578](https://issues.redhat.com/browse/FAI-578), [FAI-579](https://issues.redhat.com/browse/FAI-579), [FAI-580](https://issues.redhat.com/browse/FAI-580), [FAI-581](https://issues.redhat.com/browse/FAI-581)

manstis:
@tiagobento @ederign This PR is ready for your eyes now :+1:

tiagobento (approved):

PR #621, opened 2021-09-09
KOGITO-5428: Introduce DMN Runner cypress test

jomarko:


domhanak:
@tiagobento In the logs of `CI :: Monorepo / monorepo-pr-ci (ubuntu-latest) (pull_request)` I can see we are building packages like `kie-editors-standalone`, `desktop` and more. I thought we should be building only packages that area affected by these changes. Is there a bug in the behavior or is this because there is a changes in `build-env` package? Thanks for clarification in advance!

domhanak (changes requested):
Looks good to me, we can merge once the test is passing. Currently there is an issue, also visible on CI now.
```
 1) DMN Runner Test
       Test DMN Runner on DMN sample:
     AssertionError: Timed out retrying after 10000ms: expected '<div.pf-c-description-list__text>' to be 'visible'

This element `<div.pf-c-description-list__text>` is not visible because its content is being clipped by one of its parent elements, which has a CSS property of overflow: `hidden`, `scroll` or `auto`
```

tomasdavidorg (approved):
Looks good to me when it is green 🟢.

jomarko:
@domhanak, thank you for the review, I have updated the assertion in last commit, lets see what our CI automation results are.

@tomasdavidorg thank you for the review too.

domhanak (approved):
Locally it passes for me. Thanks @jomarko for new test! Great work

tiagobento (approved):

PR #636, opened 2021-10-04
AF-2893: Move Dashbuilder Runtime to Kogito Tooling

jesuino:
This PR:

- Moves DashBuilder to Kogito Tooling and make it part of the build.

- Creates the release workflow for Dashbuilder

- Creates base images for Dashbuilder Authoring and Dashbuilder Runtime.


ederign:
@tiagobento can you please take a look at this?

tiagobento (approved):

sthundat:
Hi William,

Thanks for migrating the Dashbuilder Runtime to Kogito Tooling.

Overall it looks good to me. There is an observation during the testing:

Upon setting up dashbuilder authoring locally, the war file `dashbuilder-authoring.war` which is generated using the command `yarn build:dev` is not working.
As we had discussed, this could be because of the compilation issue with GWT.

The war file generated using command `mvn clean install -DskipTests -Dgwt.compiler.skip=true` is working fine and I have verified the same.


In nutshell,locally I have tested the binaries(wars and jars) and everything works fine!

Thanks,
Savitha

tiagobento:
@sthundat Hi Savitha, thanks for the review! The WARs generated from our CI are working correctly, so maybe it's possible that this is an issue with your local configuration?

sthundat:
> @sthundat Hi Savitha, thanks for the review! The WARs generated from our CI are working correctly, so maybe it's possible that this is an issue with your local configuration?

Hi @tiagobento ,Thanks for your reply.
 @jesuino would  you consider this is a configuration issue on my local?
 I remember you have mentioned it is missing GWT compilation. Please correct me if i am wrong.

Thanks,
Savitha

jesuino:
Thank you so much for your review @sthundat 

>> Upon setting up dashbuilder authoring locally, the war file dashbuilder-authoring.war which is generated using the command yarn build:dev is not working.

This indeed skip GWT compilation, but it was on purpose because the dev build was very slow.

THe suggested approach for dashbuilder development is:

* Run build:fast from yarn
* In packages/dashbuilder, then run `mvn gwt:run` on DB Authoring;
* on DB Runtime we have to build from `dashbuilder-runtime-parent` using `mvn clean install` and then run the JAR from `dashbuilder-runtime-app/target` - this will be like this until we have a pure client DB Runtime because we have no GWT adapter for Quarkus.

sthundat:
Hi @jesuino ,
Thanks for the clarification!In this case I can approve the PR.
Anyway great work William!
Thanks,
Savitha

sthundat (approved):

Issue #425, opened 2021-02-18
Using embedded editors


manstis:
Hi,

Please see https://github.com/kiegroup/kogito-apps/pull/654

I am unable to embed the `DmnEditor` following the instructions I can find.

Can you please assist.

tiagobento:
Hi @manstis ! What are the instructions you're using? From I see on your PR, the problem is likely the `origin: ''` parameter you're passing to the DMN Standalone Editor. Please don't pass any `origin` so that the default values are used.

manstis:
Great thank-you @tiagobento. `origin` removed and all good. Works fine.

PR #638, opened 2021-10-07
 kie-editors-standalone README: remove origin parameter from the DmnEditor example

tbouffard:
Using the `origin: ''` parameter doesn't work. The default works most of the time.

This should prevent users from encountering issue like #425

### DIFF
--- a/packages/kie-editors-standalone/README.md
+++ b/packages/kie-editors-standalone/README.md
@@ -27,7 +27,6 @@ const editor = DmnEditor.open({
   container: document.getElementById("dmn-editor-container"),
   initialContent: Promise.resolve(""),
   readOnly: false,
-  origin: "",
   resources: new Map([
     [
       "MyIncludedModel.dmn",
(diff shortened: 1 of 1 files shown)


packages/kie-editors-standalone/README.md
```diff
@@ -27,7 +27,6 @@ const editor = DmnEditor.open({
   container: document.getElementById("dmn-editor-container"),
   initialContent: Promise.resolve(""),
   readOnly: false,
-  origin: "",
```
tbouffard, line 30: Note that due to this removal, the example no longer uses the entire list of available parameters that is described later in the document.
We could think about renaming `Parameters description` into `Available parameters` or something else to point this out.

ljmotta: I think `Availabe parameters` would do the work. Could you make this change?

ljmotta (commented):
Thanks for this PR @tbouffard ! I've just made a comment inline.

packages/kie-editors-standalone/README.md
tbouffard: Done in d068193e20393ea51cede63f45a7b2f6c55f8290

ljmotta (approved):
LGTM! Thanks again @tbouffard

tiagobento (approved):

PR #642, opened 2021-10-13
FAI 579 - Mining Schema (PMML Editor test Suite)

sjamboro:
https://issues.redhat.com/browse/FAI-579

Goals:
Checks that
- editor reflected changes in DataFields
- buttons "Add Fields(s)" and "Add All Fields" works properly
- is possible to delete mining field
- is possible to set/unset some properties of mining field (Usage Type and Outliers)
- is possible to view properties in overview mode
- PMML editor prevents an user to create invalid PMML source (disabled and enabled

Not Goals:
- check that PMML is valid
- cover all properties (see FAI-631 and FAI-632)


packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaFields/MiningSchemaFields.tsx
```diff
@@ -63,7 +72,7 @@ interface MiningSchemaFieldProps {
   onCancel: () => void;
 }
 
-const MiningSchemaItem = (props: MiningSchemaFieldProps) => {
+const MiningSchemaItem: React.FC<MiningSchemaFieldProps> = (props) => {
```
manstis, line 75: Just for my information; why do we need to change this to be an explicit `FunctionalComponent`?

manstis (approved):
LGTM (I do like the Cypress tests format).

I asked a question for my understanding; but nothing wrong with the PR IMO.

packages/pmml-editor/src/editor/components/MiningSchema/MiningSchemaFields/MiningSchemaFields.tsx
sjamboro: I am sorry I forgot to reverse this change here. I tried to make it Ouia compilant, but I do not have ouia packages in kogito-tooling. On other hand, one warning disappear.

packages/pmml-editor/it-tests/integration/mining-schema.ts
```diff
@@ -0,0 +1,464 @@
...
+   * Mining schema cannot be defined without any Data Field in the Data Dictionary.
+   */
+  it("No DD defined", () => {
+    cy.newButtonPMML().click();
+    cy.buttonMiningSchema().click();
+
+    cy.ouiaId("mining-toolbar")
+      .should("be.visible")
+      .within(() => {
+        cy.ouiaId("select-mining-field").find("div").first().should("have.class", "pf-m-disabled");
```
kelvah: Just wondering, is it fine to rely to PatternFly internal classes for end-to-end tests?
They are out of our control and they could eventually change after a PF update (although it's unlikely).

In this specific case the test could look for the `Select` toggle. It's a regular `<button>` and it could be targeted using the [`toggleId` prop](https://www.patternfly.org/v4/components/select/#select).
Maybe I am just worrying too much :slightly_smiling_face: 
What you think?

kelvah (approved):
LGTM.

I only left a question out of curiosity.

packages/pmml-editor/it-tests/integration/mining-schema.ts
sjamboro: @kelvah @tiagobento First my motivation was to check if the select is disabled because there is no data field to choose. My first option was `cy.ouiaId("select-mining-field").should("be.disabled");` but it did not work.

I can focus on the "disabled" attribute. It is directly in input and button 
`cy.ouiaId("select-mining-field").find("input#toggleId").should("be.disabled");cy.ouiaId("select-mining-field").find("button#toogleId").should("be.disabled");`

We decited to use PF so its functionality should be tested by developers of PF. I decided to check if we set this component as disabled and I do not care how PF do it (using attribute disabled, lost event handlers, replace by another element). I do not know if names of classes ("pf-m-disabled"), tags ("<Select>") or properties ("toggleId") are stable in PatternFly. @kelvah you call it as "internal class". Where can I find this piece of information? I understand that internal name is a problem.

By the way there is another occurence of `pf-m-disabled` when the option is disabled. When I study DOM in this case I am not able to find another option.

My suggestion: can we discuss it in new JIRA? I will add some occurence of `pf-m-disable` to the new JIRA and try to find a way how to avoid it.

kelvah: I think `Select` is difficult to test because it does not render its `<input>` element at all when it's disabled.
That's strange honestly. I would have expected a disabled input.

Anyway, what I was suggesting is to use the `toggleId` prop, like:
```
<Select
   ...
   ouiaId="select-mining-field"
   toggleId="select-mining-field-toggle"
>
```

Doing so it would be possible to test if the selection is disabled with:
`cy.ouiaId("select-mining-field").find("#select-mining-field-toggle").should("be.disabled")`

This way we are using a selector provided by us.
I called `pf-m-disabled` an internal class because it's used by PF to style the component and it's not configurable. It can be considered an internal implementation that could change without notice.

No problem for me if you want to address the topic in a new JIRA :slightly_smiling_face:

sjamboro: @kelvah I have fixed the reported issue of using "pf-m-disabled" class. Other occurences of this class should be resolved by https://issues.redhat.com/browse/FAI-641.

sjamboro: @kelvah @tiagobento Is possible to merge this PR?

tiagobento (approved):

PR #644, opened 2021-10-18
KOGITO-5974: Stunner - Integrate editors JS / TS API with the Standalone Editors API

handreyrc:
Hi @romartin and @tiagobento, please could you review this PR?

Thanks!

PR #603, opened 2021-08-20
kie-editors-standalone: update options description in README

tbouffard:
`origin`: add default when using the file protocol
`readOnly`: it is also available in the BPMN Editor

**Rationale**
- origin: [BPMN](https://github.com/kiegroup/kogito-tooling/blob/0.12.0/packages/kie-editors-standalone/src/bpmn/index.ts#L32) and [DMN](https://github.com/kiegroup/kogito-tooling/blob/0.12.0/packages/kie-editors-standalone/src/dmn/index.ts#L32)
- readOnly: [EnvelopeServer](https://github.com/kiegroup/kogito-tooling/blob/0.12.0/packages/kie-editors-standalone/src/bpmn/index.ts#L47) and [Channel](https://github.com/kiegroup/kogito-tooling/blob/0.12.0/packages/kie-editors-standalone/src/bpmn/index.ts#L84)

PR #645, opened 2021-10-24
 kie-editors-standalone README: update the setContent signature

tbouffard:
The `setContent` has 2 parameters as of v0.9.1. So add the new `path` parameter in the README to avoid misleading users.

**Notes**
This breaking changes was marked as a fix in the 0.9.1 release notes (https://github.com/kiegroup/kogito-tooling/releases/tag/0.9.1, _KOGITO-4915 - Standalone editors setContent implementation should receive path and content_), so it was quite hard to notice.
At least, I didn't notice it could break the usage of the method after quickly checking the release notes, especially in a patch version. I was implicitly assuming that `kie-editors-standalone` was following semantic versioning.

I have been fooled myself in the past (https://github.com/process-analytics/bpmn-visualization-examples/issues/217#issuecomment-942061749), so this doc update should help others to not loose time.
If you can make such changes more visible in the future, this will be nice 😸

### DIFF
--- a/packages/kie-editors-standalone/README.md
+++ b/packages/kie-editors-standalone/README.md
@@ -55,7 +55,7 @@ Available parameters:
 The returned object will contain the methods needed to manipulate the Editor:
 
 - `getContent(): Promise<string>`: Returns a Promise containing the Editor content.
-- `setContent(content: string): Promise<void>`: Sets the content of the Editor. The returning Promise will be rejected if setting the content fails.
+- `setContent(path: string, content: string): Promise<void>`: Sets the content of the Editor. The returning Promise will be rejected if setting the content fails.
 - `getPreview(): Promise<string>`: Returns a Promise containing the SVG string of the current diagram.
 - `subscribeToContentChanges(callback: (isDirty: boolean) => void): (isDirty: boolean) => void`: Setup a callback to be called on every content change in the Editor. Returns the same callback to be used for unsubscription.
 - `unsubscribeToContentChanges(callback: (isDirty: boolean) => void): void`: Unsubscribes the passed callback from content changes.


packages/kie-editors-standalone/README.md
```diff
@@ -55,7 +55,7 @@ Available parameters:
 The returned object will contain the methods needed to manipulate the Editor:
 
 - `getContent(): Promise<string>`: Returns a Promise containing the Editor content.
-- `setContent(content: string): Promise<void>`: Sets the content of the Editor. The returning Promise will be rejected if setting the content fails.
+- `setContent(path: string, content: string): Promise<void>`: Sets the content of the Editor. The returning Promise will be rejected if setting the content fails.
```
tbouffard, line 58: ❓ I don't know what to explain about the `path` parameter.
The method seems to work when passing `null`, `undefined` or whatever string. I don't know where the value is used next in the editor. 
Can a reviewer suggest anything?

tiagobento:
Hi @tbouffard! Thanks for the PR. Indeed, I see the problems you described. We've been trying to keep retro-compatibility when we can, but as we didn't reach 1.0.0 yet, sometimes we end up breaking APIs. We should've done it in a minor release tough, not in a patch, as you said. Good catch.

Your changes look good! For completeness, the whole public API of the Standalone Editors are described by these two types: https://github.com/kiegroup/kogito-tooling/blob/00204e7a4c884132a3f520b297e66ddf7114ae0d/packages/kie-editors-standalone/src/common/Editor.ts#L24-L41

packages/kie-editors-standalone/README.md
tiagobento: @romartin @yesamer ^

tbouffard:
@tiagobento thanks for the feedback and the link to the editor API types.

Keeping a README in sync with the API is always a challenge. The current Pull Request is the third (after #603 and #638) that I created to align the documentation with the implementation. 
I've just seen upcoming changes in the editors API in #644 and for now, they are not added to the README (I haven't seen review comments requesting to update the README). So for sure, the README will be outdated in the future.

To reduce the maintenance, it will probably be better to only provide minimal information in the README and generate the documentation from the source code. This is out of the topic of this Pull Request, but it should help both the kogito-tooling development team and the users 😸

jomarko:
Hello team,

I realized, we have this documented also in `kogito-editors-java` [1], do you think updates are needed also there?

[1]
https://github.com/kiegroup/kogito-editors-java/search?l=Markdown&q=setcontent

romartin:
@jomarko good catch, although those usages of the setContent in the README files are not really intended for community documentation, it's just a way for developers to know that they can use it. Anyway, I see the scenario simulation readmes are not up to date with the double argument of the setContent method, so let's see if @yesamer wants to update it or just drop it..

packages/kie-editors-standalone/README.md
romartin: Hey @tbouffard @tiagobento 

Let you know that the arguments for the `setContent` method changed, as @tbouffard describes, because of https://issues.redhat.com/browse/KOGITO-631.

The goal for this `path` argument, which can be `null`, is to keep in syn  the process id and the filename. For instance, how the VSCode plugin works, when creating a new process (`.bpmn` file), is by calling the setContent method in the process editor, and passing the filename (of the just created bpmn file), and an empty string as raw content. This way, the editor "detects" it's a new process (as content is empty), but it can generate its process identifier by using the filename the user just created.

There exist different reasons behind this, also a few constraints, you can see more details at KOGITO-631, but main reasons are:
- keep in sync filename <-> process id
- generating process id's that are valid java identifiers (in order to properly run the processes in the kogito engine)
- subprocess / decision navigation (asset navigation) (work in progress)

romartin: @tbouffard  feel free to quickly test this behavior, and the use of the `path` argument. by creating a new `.bpmn` file in vscode, by giving it a concrete name. Then once the process editor opens, open the properties dock panel (east side panel), and look at the process id field

domhanak: @tbouffard The `path` should be a relative path to the editors resources. So if there is a process in the root of your resources the path should just be name of the process.
@romartin @paulovmr please correct me if I am wrong.

domhanak (approved):
@tbouffard LGTM, thanks for fixing the README. You can clarify the path operator (see my comment) and we can merge imho.

yesamer:
@romartin I agree with you, that tutorial was intended to support internal development in the very early stage of kogito development.

tiagobento (approved):

tbouffard:
Thanks for the information. The pull request has already been merged so I won't be able to update the README here. I will see if I can contribute again in the future.

PR #660, opened 2021-11-03
AF-2913: Support External DataSets in DashBuilder

jesuino:
Everything is explained in JIRA:

https://issues.redhat.com/browse/AF-2913

Basically it is support for external datasets that can be loaded from a HTTP URL or from a File URL.


jesuino:
To create a dataset use Dashbuilder Authoring go to DataSets and select "External".

image

Suggested test scenarios:

1) Load a simple dataset from a file.

1.a) create a file with some data:

```
[
    ["A", 10],
    ["A", 20],
    ["B", 30],
    ["B", 10],
]
```

image

1.b) Create a new External DataSet and point to the created file using the following URL format: file://{path to the dataset file};

1.c) Check the dataset columns type, it must be of types LABEL and NUMBER, also the columns should be generated (Column 0 and Column 1).

1.d) Do the same, but with a dataset that contains the columns information. Make sure to change the first column type from LABEL to TEXT, this way we can see that the columns declaration is respected and also the columns IDs:

```
{
   "columns":[
      {  "id":"CL1",  "type":"TEXT" },
      { "id":"CL2", "type":"NUMBER"}     
   ],
   "values":[
    ["A", 10],
    ["A", 20],
    ["B", 30],
    ["B", 10],
]
}
```
image


2) Load the dataset from an URL
2.a) Use python webserver to expose a directory through HTTP and then use the file URL during an external dataset creation;

3) Create a dashboard and run on Dashbuilder Runtime

3.a) Using the first dataset from step 1, create a dashboard (it may be a page with a simple bar chart)
image

3.b) Use custom export to export a ZIP with the dataset and the page;
3.c) Start DashBuilder Runtime and import the ZIP, it should work;
3.d) Stop DB Runtime, use the system property "dashbuilder.dataset.%s.url" - replace %s with the dataset name - to point to a new dataset JSON - add more rows to this new definition or change the rows existing - but do not change columns definition (type and ID) or it will result in an expected error;

4) Test Cache
4.a) Go to the dataset definition and turn on cache, make sure it updates every few seconds:

image

4.b) Go the page created in step 3 and setup refresh:

image

4.c) Change the dataset (add more rows to it) and wait the page to update - it should update with the new rows:

https://user-images.githubusercontent.com/359820/140787909-e8cdf6c6-591c-44c3-95eb-621c5c1ce130.mp4

sthundat:
Hi William,
I have tried below scenarios for external dataset verification
1.External dataset loaded from a file URL
2.External dataset loaded from a http resource by creating a micro service
Both scenarios works fine! Thank you
Regards,
Savitha

sthundat (approved):

tiagobento (approved):

PR #664, opened 2021-11-10
FAI-578: Data Dictionary - PMML editor - integration test

sjamboro:
FAI:578: https://issues.redhat.com/browse/FAI-578

Goal: create test suite which cover user interaction with Data Dictionary editor
Non-goal: check all properties of DF

### DIFF
--- a/packages/pmml-editor/it-tests/integration/data-dictionary.ts
+++ b/packages/pmml-editor/it-tests/integration/data-dictionary.ts
@@ -1,0 +1,517 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as buildEnv from "@kogito-tooling/build-env";
+
+describe("Data Dictionary Test", () => {
+  beforeEach(() => {
+    cy.visit(`http://localhost:${buildEnv.pmmlEditor.dev.port}/`);
+  });
+
+  it("Empty DD", () => {
+    cy.newButtonPMML().click();
+    cy.buttonDataDictionary().click();
+
+    cy.ouiaId("dd-toolbar")
+      .should("be.visible")
+      .within(() => {
+        cy.ouiaId("add-data-type").should("not.be.disabled").should("be.visible");
+        cy.ouiaId("add-multiple-data-type").should("not.be.disabled").should("be.visible");
+        cy.ouiaId("order-toggle").should("not.be.disabled").should("be.visible");
+      });
+    cy.ouiaId("no-data-fields-title").should("be.visible").should("have.text", "No Data Fields Defined");
+  });
+
+  it("Add multiple Fields", () => {
+    cy.newButtonPMML().click();
+    cy.buttonDataDictionary().click();
+
+    cy.ouiaId("add-multiple-data-type").click();
+    cy.ouiaId("multiple-data-types").should("be.visible").type("aaa\nbbb\n");
+    cy.ouiaId("add-them").click();
+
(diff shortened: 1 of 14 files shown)


sjamboro:
Jenkins retest this

sjamboro:
@manstis @kelvah @tiagobento can you review this PR?

packages/pmml-editor/it-tests/integration/data-dictionary.ts
```diff
@@ -0,0 +1,517 @@
...
+        '  <DataDictionary numberOfFields="2">' +
+          '    <DataField name="aaa" optype="categorical" dataType="string"/>' +
+          '    <DataField name="bbb" optype="categorical" dataType="string"/>' +
+          "  </DataDictionary>"
+      )
+      .ouiaId("pmml-modal-confirm")
+      .click();
+    /*
+     * Current implementation allows use string "test\ntest" without any warning
+     * TODO: discuss this topic with developers
```
manstis, line 74: @kelvah FYI.

manstis, line 74: @kelvah FYI.

packages/pmml-editor/it-tests/integration/data-dictionary.ts
```diff
@@ -0,0 +1,517 @@
...
+        /*
+         * Unset property
+         */
+        cy.ouia("edit-props", "link-label").click();
+        cy.ouia("df-props", "editor-body").within(() => {
+          cy.ouiaId("val-0").find("button").click();
+          cy.ouiaId("val-0").find("button").click();
+          cy.ouiaId("val-0").find("button").click();
+          cy.ouiaId("val-0").find("button").click();
+          //PMML Spec says that there can be zero Values, but editor does not allow to delete the last item.
```
manstis, line 316: @kelvah FYI. @sjamboro and myself were discussing this. We should probably allow deletion of the last item.

packages/pmml-editor/it-tests/integration/data-dictionary.ts
```diff
@@ -0,0 +1,517 @@
...
+        });
+        cy.ouiaId("back-to-DFs").click();
+        cy.ouiaId("dd-toolbar").click();
+
+        cy.ouia("dt-string", "dd-type-item")
+          .within(() => {
+            cy.ouia("Invalid Value", "data-props-label")
+              .should("exist")
+              .should("to.have.text", "Invalid Value:\u00A0invalid");
+            //XSD allows several invalid values. PMML editors allows just one.
```
manstis, line 391: Yeeeees... and we concluded that the PMML Specification only supports one `Value` with `property=invalid`

manstis (approved):
LGTM. Thank-you @sjamboro.

I would like to understand why `ouiaId` is used in some places but `data-ouia-component-id` in others and the purpose of `data-ouia-component-type`. Can you please explain?

packages/pmml-editor/it-tests/integration/data-dictionary.ts
kelvah: I've created a ticket to track it: [FAI-680](https://issues.redhat.com/browse/FAI-680)

packages/pmml-editor/it-tests/integration/data-dictionary.ts
kelvah: I have to disagree?
I recollect having a similar discussion in the past.
I would treat the two things separately: the spec and what the user sees in the UI. It's true the spec says it's possible to have 0 values defined. But values are used for other properties too (like invalid and missing value).
In our specific scenario that means having no constraints.
When the user wants to add constraints (using the constraint type selection) then he has to provide at least one value. Having the constraints option selected without values makes little sense. Unless we want to automatically clear the constraints type selection when deleting the last value, which could be undesirable when I'm just updating all the values for example.

manstis: > I have to disagree?

I don't know; do you have to disagree?

Joking aside. I can see your point: the UI can be "better than the spec" and prevent Users making mistakes. If they've elected to have "Value" constraints, as you say, it makes no sense having zero defined. We have to support opening a file with zero "Values" defined (as the User could have done something daft in a text editor etc) but we can prevent them from creating such a scenario with our editor.

Fair point. Fair play.

tiagobento:
@manstis AFAIK, some components in PatternFly are "OUIA-Compliant" [1], and some are not. Those which are, have the `ouiaId` property, that not only accepts custom values, but also has a default generated value. That's at least my understanding :)

[1] https://www.patternfly.org/v4/developer-resources/open-ui-automation/

tiagobento (approved):

kelvah (approved):

PR #681, opened 2021-11-17
KOGITO-5074: Multi-file support in .new

caponetto:
This PR introduces the new Multi-file capabilities to the `online-editor` package.
A more detailed description of this task can be found in the Jira and its related epics:

https://issues.redhat.com/browse/KOGITO-5984

---

A working version of the features introduced in this PR can be found here:

https://kiegroup.github.io/kogito-online-staging/multifile-v7

---

Other changes on this PR, necessary to have the main features on `online-editor` implemented.
- Chrome Extension: Changed icons
- Chrome Extension: Removed feature to open single files on Online Editor. Now you can open the entire repository by pressing a button. Everything else is the same.
- Added two new packages that re-build the contents of `feel-input-component` and `boxed-expression-component`. These packages are used on `unitables`, which is a new package that implements `uniforms` as a table. The default implementation is custom-made for DMN schemas, but it is possible to make it generic so that in the future other types of schemas are supported.
- `dmn-json-schema-generator` was incorporated by `dmn-dev-sandbox-quarkus`, which is the app that runs the DMN models when we deploy them to OpenShift.
- `online-editor` webapp now runs on `https`. This is necessary to test the DMN Dev Sandbox feature locally.
- Bumped Quarkus to 2.4.0.Final and Kogito to 1.12.0.Final.
- Removed `frontend-maven-plugin` from `dashbuilder-js`, as it was causing more problems and it is not necessary, to be honest.
- Upgraded PatternFly. Some snapshot tests needed to be updated because this new version adds OUIA attributes by default. Also, some locators in IT tests changed to be compliant with the new `ouiaId` prop for some PF components.
- Some libraries were upgraded as well. (e.g. `react-router`)
- The DMN Dev Sandbox app now supported multiple files, and it's possible to switch between them after the deploy.
- Some minor infrastructure changes on `editor`, like renames and refs.
- Ignoring Keyboard Shortcuts for an entire DOM tree is now possible with the `useElementsThatStopKeyboardEventsPropagation` hook. We use that on the DMN Runner table.
- Changed the default name of new files from `"new-file"` to `"Untitled"`.
- New patches for:
   - LightningFS: We actually have opened a PR directly on the library, but we got no feedback. We'll be using a patched version until we decide what to do.
   - LightningFS types: To reflect the features we added.
   - Feel Input: Changing imports
   - Boxed Expression: Changing imports


ederign (approved):
Tiago/Guilherme, I did some tests with the URL provided and looks like most of the things are working as expected. I was able to complete all workflows that I tried.

Just found some two NOT BLOCKER issues that I discussed with you on PVT.

We should also fix the tests before merging it. otherwise great job Guilherme/Tiago and team.

tiagobento (approved):

PR #691, opened 2021-11-22
Add new "Shared value" mechanism on `EnvelopeBusMessageManager`

tiagobento:
This PR introduces the concept of a "Shared value" along with an example of how to use it at the `TodoListView` component.

Shared values are a way of sharing data between Channels and Envelopes, and allow for a much more React-friendly API than the requests/response model. When an API provides a Shared value, this data is immediately available for anyone consuming this shared value to use.

When a Shared value is a React state, it's very simple to use.

Transforming a state in a Shared value:
```jsx
// Create any state
const [value, setValue] = useState();

// Bind it to a Shared value
useStateAsSharedValue(value, setValue, apiToProvide.shared.mySharedValue)
```

Consuming a Shared value:
```jsx
// Consume a Shared value just like you would do with any regular state
const [value, setValue] = useSharedValue(apiToConsume.shared.mySharedValue)
```

---
### Example on `TodoListView`

https://user-images.githubusercontent.com/1584568/142877831-d266d30c-2bd3-42c0-8a2a-566769151609.mov

---

### Also on this PR:
- Added the `@kogito-tooling-examples` scope to the multi-package live-reload loader, and excluded `@kogito-tooling/kie-editors-standalone`
- Changed the notifications mechanism to allow for `.send` instead of simply calling a method.
- Improved the generic EmbeddedEnvelope mechanism. Now you can use a handy `RefForwardingEmbeddedEnvelope` to create your fully typed EmbeddedEnvelope. This is much more flexible and doesn't have the problems that the higher-order component had.
- Reviewed the dependencies on the hooks of many of our examples.
- Renamed `EnvelopeBusController` to `EnvelopeClient`. These names are far from ideal, but at least now it's more clear that `EnvelopeServer` and `EnvelopeClient` make a pair.
- Changed the Envelope mechanism to actually render the view DURING an init request. This prevents the view from trying to react the EnvelopeServer without it being properly set yet.


tomasdavidorg (approved):
Looks good to me. 
Just one note: The README of the examples is outdated:

1. List of the examples is not complete and the links do not work.
2. `yarn run init` there is no such command.
3. `yarn run build:fast` should be `yarn run build:dev`
4. After `yarn run start` app is running on port `:9100` (not `9001`).

tiagobento:
@tomasdavidorg Thanks a lot for the review. I created https://issues.redhat.com/browse/KOGITO-6325 to track this problem with the examples.

ederign (approved):
Great work Tiago, really nice improvement.

PR #699, opened 2021-11-26
KOGITO-6214: Settings for exporting SVG files in VSCode Extensions

thiagoelg:
# JIRA
https://issues.redhat.com/browse/KOGITO-6214

# On this PR
Introduces two new settings for each of the DMN and BPMN Editor VSCode extensions:
- `kogito.{dmn/bpmn}.svgFilenameTemplate`
Is the template to be used for the exported SVG filename and defaults to `${fileBasenameNoExtension}-svg.svg`.
- `kogito.{dmn/bpmn}.svgFilePath`
Is the path where the exported SVG should be saved and defaults to `${fileDirname}`.

Screen Shot 2021-11-26 at 17 19 00

They can use different tokens to refer to the workspace folder, the .bpmn/.dmn file name, path, and extension, according to this table:

| Variable                       | Example                                   |
| ------------------------------ | ----------------------------------------- |
| **${workspaceFolder}**         | `/home/your-username/your-project`        |
| **${fileDirname}**             | `/home/your-username/your-project/folder` |
| **${fileExtname}**             | `.ext`                                    |
| **${fileBasename}**            | `file.ext`                                |
| **${fileBasenameNoExtension}** | `file`                                    |

These variables were based on the ones already available in VSCode for the `launch.json` and `tasks.json` files. ([VSCode: Variables reference](https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables-examples))

The idea is to, whenever it gets implemented in VSCode, be able to use the predefined variables instead of replicating them on our own, without breaking the user's settings. (Related issues: https://github.com/microsoft/vscode/issues/2809 and https://github.com/microsoft/vscode/issues/46471)

# Also on this PR

The new and existing settings (`runOnSave`) were changed to support the use of markdown descriptions and input fields (without the need to edit the `settings.json` file directly). ([VSCode: Configuration schema](https://code.visualstudio.com/api/references/contribution-points#Configuration-schema))

### DIFF
--- a/packages/vscode-extension/src/generateSvg.ts
+++ b/packages/vscode-extension/src/generateSvg.ts
@@ -66,8 +66,8 @@ export async function generateSvg(args: {
       })
     : undefined;
 
-  const tokens: Record<SettingsValueInterpolationToken, string | undefined> = {
-    "${workspaceFolder}": workspace?.uri.fsPath,
+  const tokens: Record<SettingsValueInterpolationToken, string> = {
+    "${workspaceFolder}": workspace?.uri.fsPath ?? parsedPath.dir,
     "${fileDirname}": parsedPath.dir,
     "${fileExtname}": parsedPath.ext,
     "${fileBasename}": parsedPath.base,
--- a/packages/vscode-extension/src/generateSvg.ts
+++ b/packages/vscode-extension/src/generateSvg.ts
@@ -30,10 +30,10 @@ type SettingsValueInterpolationToken =
   | "${fileBasename}"
   | "${fileBasenameNoExtension}";
 
-function interpolateSettingsValue(args: { tokens: Record<string, string | undefined>; value: string }) {
+function interpolateSettingsValue(args: { tokens: Record<string, string>; value: string }) {
   const { tokens, value } = args;
   return Object.entries(tokens).reduce(
-    (result, [tokenName, tokenValue]) => result.replaceAll(tokenName, tokenValue ?? ""),
+    (result, [tokenName, tokenValue]) => result.replaceAll(tokenName, tokenValue),
     value
   );
 }
--- a/packages/vscode-extension/src/generateSvg.ts
+++ b/packages/vscode-extension/src/generateSvg.ts
@@ -23,12 +23,18 @@ import { I18n } from "@kie-tooling-core/i18n/dist/core";
 
 const encoder = new TextEncoder();
 
-type FilePathToken = "workspaceFolder" | "fileDirname" | "fileExtname" | "fileBasename" | "fileBasenameNoExtension";
+type SettingsValueInterpolationToken =
+  | "${workspaceFolder}"
+  | "${fileDirname}"
+  | "${fileExtname}"
+  | "${fileBasename}"
+  | "${fileBasenameNoExtension}";
 
-function tokenReplace(tokens: Record<string, string | undefined>, text: string) {
-  return Object.entries(tokens).reduce<string>(
-    (result, [token, value]) => result.replace(RegExp(`\\$\{${token}}`, "ig"), value ?? ""),
-    text
+function interpolateSettingsValue(args: { tokens: Record<string, string | undefined>; value: string }) {
+  const { tokens, value } = args;
+  return Object.entries(tokens).reduce(
+    (result, [tokenName, tokenValue]) => result.replaceAll(tokenName, tokenValue ?? ""),
+    value
   );
 }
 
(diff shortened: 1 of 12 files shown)


packages/vscode-extension/src/generateSvg.ts
```diff
@@ -43,9 +58,42 @@ export async function generateSvg(args: {
     return;
   }
 
+  const workspace = vscode.workspace.workspaceFolders?.length ? vscode.workspace.workspaceFolders[0] : null;
   const parsedPath = __path.parse(editor.document.uri.fsPath);
-  const svgFileName = `${parsedPath.name}-svg.svg`;
-  const svgUri = editor.document.uri.with({ path: __path.join(parsedPath.dir, svgFileName) });
+
+  const tokens: Record<SettingsValueInterpolationToken, string | undefined> = {
+    "${workspaceFolder}": workspace?.uri.fsPath,
```
tiagobento: Is it possible to know what's VS Code's default behavior for when there's no Workspace open? Also, if I have more than one folder at my Workspace.. won't that break? I mean, is it possible to know to which `workspaceFolder` does a file belong? Seems to me that `${workspaceFolder}` should have a different value depending on the file we're generating the SVG for.

packages/vscode-extension/src/generateSvg.ts
```diff
@@ -43,9 +58,42 @@ export async function generateSvg(args: {
     return;
   }
 
+  const workspace = vscode.workspace.workspaceFolders?.length ? vscode.workspace.workspaceFolders[0] : null;
```
tiagobento: Why use `null` and not `undefined`?

packages/vscode-extension/src/generateSvg.ts
```diff
@@ -43,9 +58,42 @@ export async function generateSvg(args: {
...
+  if (__path.parse(svgFilenameTemplate).dir) {
+    vscode.window.showErrorMessage(
+      `The kogito.${fileExtensionNoDot}.svgFilenameTemplate setting should be a valid filename, without a path prefix. Current value: ${svgFilenameTemplate}`
+    );
+    return;
+  }
+
+  const svgFileName = interpolateSettingsValue({
+    tokens,
+    value: svgFilenameTemplate.length ? svgFilenameTemplate : "${fileBasenameNoExtension}-svg.svg",
```
tiagobento:
```suggestion
    value: svgFilenameTemplate? svgFilenameTemplate : "${fileBasenameNoExtension}-svg.svg",
```

Would that be better? I mean, `""` and `0` are falsy values, so it would be the same, but with less code? 😝

packages/vscode-extension/src/generateSvg.ts
```diff
@@ -43,9 +58,42 @@ export async function generateSvg(args: {
...
+  });
+  const svgFilePath = interpolateSettingsValue({
+    tokens,
+    value: svgFilePathTemplate.length ? svgFilePathTemplate : "${fileDirname}",
```
tiagobento: Same as above.

tiagobento (commented):
@thiagoelg Congrats for the first PR!! This is looking very good. Left some comments inline before approving, can you please take a look? Thanks!!

packages/vscode-extension/src/generateSvg.ts
thiagoelg: No particular reason, but I've changed to `undefined` now to keep it in line with the `.find(...)` results

packages/vscode-extension/src/generateSvg.ts
thiagoelg: I couldn't figure out what VSCode does with the `${workspaceFolder}` variable when there's no workspace open because this variable was meant to be used with `tasks.json` and `launch.json`, both files that require a workspace to exist.

In this implementation `${workspaceFolder}` defaults to `undefined` without a workspace. Should we change that? Maybe default to `fileDirname`?

I tested in a workspace with multiple folders and it indeed will always save in the first one, even if it's not the one with the `.dmn/bpmn` file. Couldn't find a way to get the file's workspace directly, so I implemented a small check in the array of workspaces to get the correct one.

packages/vscode-extension/src/generateSvg.ts
thiagoelg: Sure!

tiagobento (approved):

packages/vscode-extension/src/generateSvg.ts
tiagobento: Ok, looks better now! Thanks for this change!!

I approved the PR, but I instantly remembered what you wrote here. Yes, I think we should default `${workspaceFolder}` to be `fileDirname`. This will also make everything simpler by changing the type to `Record<<SettingsValueInterpolationToken, string>`, also avoiding the need on `interpolateSettingsValue` of treating undefined values, as all tokens will *always* have a value.

tiagobento (changes requested):
See comment above.

tiagobento (approved):

ederign (approved):
@thiagoelg thanks for this, in special, the care to fix the descriptions!

domhanak:
Could this be covered by some unit tests? feel free to ping me on gchat to clarify.

domhanak (approved):
Works very well. :)
I was able to set `${workspaceFolder}/src/main/resources/META-INF/` for both BPMN and DMN editors and extensions properly stored SVG files on the expected location. 
Same goes for different format name.
Tested only on Linux only though.

I ll leave the unit tests up to you.
Acceptance test will be implemented as part of https://issues.redhat.com/browse/KOGITO-6347

thiagoelg:
Thank you everyone for the review!

@domhanak  I've added a simple unit test to the `interpolateSettingsValue` function

PR #726, opened 2021-12-23
Disable DMN Runner integration test temporarily

ljmotta:


tiagobento:
@ljmotta Let's disable this test temporarily? We can fix it without rush and bring the CI back to green. WDYT?

ljmotta:
@tiagobento I'm trying a new aproach, if it fails again I guess we will need to skip this test as you suggested.

tiagobento (approved):

tiagobento:
@ljmotta Thanks. Let's work on fixing it on another PR.

PR #729, opened 2021-12-28
KOGITO-5522: Incorporate `kogito-editors-java`

tiagobento:
This PR incorporates the `kogito-editors-java` repository into the kogito-tooling monorepo.

The entire repository was placed inside the `stunner-editors` directory.

The directory structure is kept identical, except for the `kogito-editors-js` package, which was hoisted into the top level of the monorepo. This avoids nested lerna builds and encourages reusability and enhances the separation of concerns of our packages. The names are kept, with the exception of `dmn-loader`, which is now called `stunner-editors-dmn-loader`.

I also removed the `yarn  build:dev` and `yarn build:prod` commands, as we should begin to choose what we want to build. Two new commands were created -- `yarn build:dev:until [pkg-name]` and `yarn build:prod:until [pkg-name]`.

Of course building everything is still supported with `lerna run build:dev` and `lerna run build:prod`.

Some changes were made to the way the `DMN Loader` package is configured. The `wire` env var was renamed to `DMN_LOADER__outputPath` and is now managed by `build-env`, just like all the other env vars we have.

Maven commands still work normally inside `packages/stunner-editors`.

The envisioned workflow for Editor developers is

1. yarn bootstrap
1. yarn build:dev:until @kogito-tooling/stunner-editors
1. mvn clean gwt:run -f [bpmn/dmn/scesim webapp path]
1. For instructions on how to enable "live-reloading" of React components, refer to DMN Loader's README.

Something that also changed is how we load CSS for React components inside the DMN Editor. The CSS was separate from the JS bundle in production builds. Although this the ideal thing to do, it gives us some challenges due to the discrepancy between dev and prod builds. For this reason, I changed the DMN Loader to use our default Webpack configuration, so everything works the same now. This is an important optimization issue and we're aware that this could be considered a "step back", however, having everything on the same repository is far more important at this time.

I already contacted Jozef and Dominik about the QE builds/checks we have, so this PR can still be updated once I have more information.


tiagobento:
⚠️ **_NOTE:_** DO NOT SQUASH ⚠️

domhanak:
@tiagobento I examined the `kie-editors-standalone` it-tests failures and something is not right. The videos suggest that `No commands were run in this test` or that the `cy.loadEditors` is not a function. My suspicion is that there is an issue with config and most probably something overrides the config of `kie-editors-standalone` it tests.

tiagobento:
@domhanak Yes.. I'm investigating it.

jomarko (changes requested):
Tests in `./packages/boxed-expression-component/it-tests/integration/` were not started as part of `https://github.com/kiegroup/kogito-tooling/runs/4722385526?check_suite_focus=true` 

We need to run these tests as part of CI builds.

tiagobento:
@jomarko They are not running on `kogito-editors-java` as well. I tried enabling them, but they were not stable enough yet.

jomarko:
@tiagobento I can not agree, there were part of regular CI build on `kogito-editors-java` https://github.com/kiegroup/kogito-editors-java/pull/173/files

tiagobento:
@jomarko Ok, you're right. Let's enable them.

README.md
```diff
@@ -106,3 +88,29 @@ The Kogito Tooling project contains several applications. To develop each one of
...
+  - Located at `packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-kogito-runtime`.
+  - Run `mvn clean gwt:run` to start.
+
+- DMN
+
+  - Located at `packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-testing`.
+  - Run `mvn clean gwt:run` to start.
+  - If you want to enable live-reloading capabilities of the React components that are part of the DMN Editor, follow [these steps](./packages/stunner-editors/docs/live-reload-dmn-loader.md).
+
+- SceSim
```
yesamer: I suggest to use `Test Scenario` instead of `SceSim` as name of the Editor

yesamer:
Thank you for this huge work. After the merge, I think it's crucial to inform the team about the new commands you introduced and their usage, in order to avoid full compilation in local builds.

yesamer (approved):
@tiagobento Thank you for this huge work. After the merge, I think it's crucial to inform the team about the new commands you introduced and their usage, in order to avoid full compilation in local builds.

jomarko:
Thank you @tiagobento for bringing back most of `./boxed-expression-component/it-tests/integration/`. I reported one more jira for `stunner-editors` integration tests https://issues.redhat.com/browse/KOGITO-6518

jomarko (approved):

romartin (approved):

ederign (approved):

PR #737, opened 2022-01-04
Fix dependency mismatches

caponetto:


tiagobento (approved):

tiagobento:
@caponetto Thanks! Looks like this started happening because `syncpack` released a new version with a breaking change. Can you please fix this version on our `setup-env` action so it doesn't happen anymore?

tiagobento (approved):

PR #744, opened 2022-01-10
KOGITO-6513: add issues template

r00ta:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,65 @@
+name: Bug Report
+description: Report a bug in Kogito
+labels: kind/bug
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report! 
+        The preferred method to file a bug is [our JIRA board](https://issues.redhat.com/projects/KOGITO). You are still allowed to submit a GitHub issue, but you may experience delays in the feedback.
```
danielezonca, line 9: I'm not sure about this. As far as I remember Kogito tooling team was considering to use only GitHub Issues for community issues.
@porcelli @ederign 
Can you please confirm?

tiagobento:
I don't think this `bug_report` applies here.. It's very Maven/Java-centric and most of the times we actually don't need to know so many details about the "build environment" of the person using our software, given that they can simply be consuming our tools on https://sandbox.kie.org or using the VS Code extension. The other files look ok.

.github/ISSUE_TEMPLATE/bug_report.yml
ederign: Daniele, there is no final decision on this. What I asked that time was just to keep the GitHub issues open.

PR #780, opened 2022-01-19
Make packages be built in parallel on the workflows

caponetto:
Removing the `concurrency 1` so packages can be built in parallel on the workflows.
Also, I needed to tweak some args to avoid memory-related build failures.
With these changes, I've observed an improvement of ~20% on the build times.

**Note**: Adding the `wip` label to the PR since I'm still doing additional tests.


caponetto:
Looks like flaky tests tend to fail more often when running in parallel.

tiagobento (approved):

tiagobento:
Merging this one as the CI seems stable. Should we find any problems in the future, we'll re-evaluate.

PR #811, opened 2022-01-27
Create new env var to customize KIE Sandbox's URL in IT tests

tomasdavidorg:
If `CYPRESS_ONLINE_EDITOR_URL` env. variable is set, it will be used as entry point for the Cypress integration tests.

### DIFF
--- a/packages/build-env/index.js
+++ b/packages/build-env/index.js
@@ -97,6 +97,11 @@ const ENV_VARS = {
     default: undefined,
     description: "",
   },
+  ONLINE_EDITOR__url: {
+    name: "ONLINE_EDITOR__url",
+    default: undefined,
+    description: "",
+  },
   DMN_DEV_SANDBOX__baseImageRegistry: {
     name: "DMN_DEV_SANDBOX__baseImageRegistry",
     default: "quay.io",
@@ -317,6 +322,7 @@ module.exports = {
 
   onlineEditor: {
     dev: {
+      urlProperty: ENV_VARS.ONLINE_EDITOR__url.name,
       port: 9001,
     },
     gtmId: getOrDefault(ENV_VARS.ONLINE_EDITOR__gtmId),
--- a/packages/build-env/index.js
+++ b/packages/build-env/index.js
@@ -97,9 +97,9 @@ const ENV_VARS = {
     default: undefined,
     description: "",
   },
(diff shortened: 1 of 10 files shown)


tomasdavidorg:
@tiagobento `process.env` does not work in the cypress tests but there is a special tool for that, see https://docs.cypress.io/guides/guides/environment-variables

tiagobento:
Instead of adding a new loose environment variable, can you please create a new reference on `build-env`? That's the purpose of this package.

tiagobento (changes requested):
Please see the comment above.

tomasdavidorg:
@tiagobento I tried to use the `build-env` but the only way how it works with the cypress is exporting the `ONLINE_EDITOR__url` property name to `onlineEditor.dev.urlProperty` and calling `Cypress.env(buildEnv.onlineEditor.dev.urlProperty)` to get content of the property.

`Cypress.env("SOME_PROP")` usually expect that `CYPRESS_SOME_PROP` is set but if I set `config.env = process.env;` in plugins/index.ts then the `CYPRESS_` prefix is not needed. https://docs.cypress.io/guides/guides/environment-variables#Option-3-CYPRESS_

I was not able to use directly the `ONLINE_EDITOR__url`. Cypress cannot see process.env and therefore the
```
const get = (envVar) => process.env[envVar.name];
const getOrDefault = (envVar) => get(envVar) ?? envVar.default;
get(ENV_VARS.ONLINE_EDITOR__url)
```  
is always `undefined`

But maybe I missed something.

tiagobento:
@tomasdavidorg Ok, understand. Thanks for the detailed explanation. In this case I would advise that you follow the same approach we have on the `kie-standalone-editors` package. See https://github.com/kiegroup/kie-tools/blob/main/packages/kie-editors-standalone/package.json#L30-L35

tomasdavidorg:
@tiagobento Ok, I will try to update it.

tiagobento:
@tomasdavidorg Probably a good name for this new env var would be `ONLINE_EDIOTR__cypressUrl`

tomasdavidorg:
@tiagobento please take a look again.

tiagobento (approved):

tiagobento:
@tomasdavidorg Looks great, thanks a lot!

PR #825, opened 2022-02-01
KOGITO-3655: Implement cells keyboard navigation and editing

fantonangeli:
This adds the keyboard navigation to the DMN Boxed Expression Editor
https://issues.redhat.com/browse/KOGITO-3655


packages/boxed-expression-component/src/components/LogicTypeSelector/LogicTypeSelector.tsx
```diff
@@ -99,6 +99,8 @@ export const LogicTypeSelector: React.FunctionComponent<LogicTypeSelectorProps>
     targetElement,
   } = useContextMenuHandler(boxedExpression.editorRef?.current ?? document);
 
+  const globalContext = useContext(BoxedExpressionGlobalContext);
```
vpellegrino: No need to call `useContext` since we are already calling a function that returns it: `useBoxedExpression()`

packages/boxed-expression-component/src/components/LogicTypeSelector/LogicTypeSelector.tsx
```diff
@@ -153,8 +155,10 @@ export const LogicTypeSelector: React.FunctionComponent<LogicTypeSelectorProps>
       boxedExpression.boxedExpressionEditorGWTService?.notifyUserAction();
       const selectedLogicType = itemId as LogicType;
       onLogicTypeUpdating(selectedLogicType);
+      globalContext.setIsContextMenuOpen(false);
+      console.log("onLogicTypeSelect");
```
vpellegrino: Please, remove `console.log`

packages/boxed-expression-component/src/components/LogicTypeSelector/LogicTypeSelector.tsx
```diff
@@ -211,6 +215,10 @@ export const LogicTypeSelector: React.FunctionComponent<LogicTypeSelectorProps>
     return !selectedExpression.noClearAction && contextMenuVisibility && clickedOnAllowedTableSection;
   }, [contextMenuVisibility, selectedExpression.noClearAction, targetElement]);
 
+  useEffect(() => {
```
vpellegrino: Probably we don't need to set a callback for `useEffect` (which requires a re-render), and maybe you can change `setIsContextMenuOpen` inside the custom hook `useContextMenuHandler`

packages/boxed-expression-component/src/components/Table/EditableCell.tsx
```diff
@@ -50,6 +50,7 @@ export function EditableCell({ value, rowIndex, columnId, onCellUpdate, readOnly
   const [mode, setMode] = useState(READ_MODE);
   const [cellHeight, setCellHeight] = useState(CELL_LINE_HEIGHT * 3);
   const [preview, setPreview] = useState<string>("");
+  // const wrapper = useRef<HTMLDivElement>(null);
```
vpellegrino: Please, remove commented code

packages/boxed-expression-component/src/components/Table/EditableCell.tsx
```diff
@@ -90,6 +91,9 @@ export function EditableCell({ value, rowIndex, columnId, onCellUpdate, readOnly
     return `editable-cell ${selectedClass} ${mode}`;
   }, [isSelected, mode]);
 
+  /**
+   * @deprecated KOGITO-3655
+   */
   const onFocus = useCallback(() => {
```
vpellegrino: If this function is no more used by production code, I'm ok for removing it

packages/boxed-expression-component/src/components/Table/Table.css
```diff
@@ -94,6 +94,17 @@
   border: 1px solid var(--pf-global--palette--black-300);
 }
 
+.expression-container .table-component table tbody td:focus {
+  border: 2px solid #438cb1;
+  outline: none !important;
```
vpellegrino, line 99: Why `!important` needed?

packages/boxed-expression-component/src/components/Table/TableBody.tsx
```diff
@@ -46,9 +61,49 @@ export const TableBody: React.FunctionComponent<TableBodyProps> = ({
   getColumnKey,
   onColumnsUpdate,
   tdProps,
+  showTableHandler,
 }) => {
   const headerVisibilityMemo = useMemo(() => headerVisibility ?? TableHeaderVisibility.Full, [headerVisibility]);
 
+  const globalContext = useContext(BoxedExpressionGlobalContext);
```
vpellegrino: Same as above, there is `useBoxedExpression` function

packages/boxed-expression-component/src/components/Table/TableHandler.tsx
```diff
@@ -77,6 +77,8 @@ export const TableHandler: React.FunctionComponent<TableHandlerProps> = ({
   const [selectedColumn, setSelectedColumn] = useState(lastSelectedColumn.placeholderOf || lastSelectedColumn);
   const [selectedRowIndex, setSelectedRowIndex] = useState(lastSelectedRowIndex);
 
+  const globalContext = useContext(BoxedExpressionGlobalContext);
```
vpellegrino: Same as above, you can use `useBoxedExpression`

vpellegrino (changes requested):
@fantonangeli thanks for your first PR! 
Amazing to see how fast you did get to contribute here!

I just performed the code check (refer to inline comments). Then I'll perform some manual tests to try out the feature!

In general, I love that you added it-tests! However I think that unit tests are important too, and we should make the test suite consistent with your changes (especially functions added in `FocusUtils.ts`). WDYT?

packages/boxed-expression-component/it-tests/integration/context_expression_spec.ts
```diff
@@ -123,8 +123,12 @@ describe("Context Expression Tests", () => {
       expect($outputs.eq(3)).to.contain("output-2");
     });
   });
+});
+
+describe("Nested Relations", () => {
```
jomarko: Could we please rename, to make more clear in logs we are still in context expression tests:
```ts
describe("Context Expression Tests :: Nested Relations", () => {
```

packages/boxed-expression-component/src/components/Table/Table.tsx
```diff
@@ -318,6 +319,8 @@ export const Table: React.FunctionComponent<TableProps> = ({
     [defaultCell, readOnlyCells]
   );
 
+  const commonCellProps = {};
```
ljmotta: Unused constant.

packages/boxed-expression-component/it-tests/integration/context_expression_spec.ts
```diff
@@ -146,12 +150,68 @@ describe("Context Expression Tests", () => {
     // insert one column left
     cy.ouiaType("expression-column-header-cell-info").contains("column-1").rightclick();
     cy.contains("Insert left").click({ force: true });
+  });
 
+  it("Define nested Relation", () => {
```
jomarko: Actually we define nothing in this test, could we please rename the test like "Check" rather than "Define"?

packages/boxed-expression-component/src/components/Table/Table.tsx
```diff
@@ -247,6 +247,7 @@ export const Table: React.FunctionComponent<TableProps> = ({
       setTableHandlerTarget(target);
       globalContext.currentlyOpenedHandlerCallback?.(false);
       setShowTableHandler(true);
+      globalContext.setIsContextMenuOpen(true);
```
ljmotta: Could you please replace the remaning `globalContext` usages with the `boxedExpression` (line 91)? So we can remove the `useContext` from this file.
```ts
#L101 const globalContext = useContext(BoxedExpressionGlobalContext);
```

packages/boxed-expression-component/it-tests/integration/context_expression_spec.ts
```diff
@@ -146,12 +150,68 @@ describe("Context Expression Tests", () => {
...
+      .type("{downarrow}{leftarrow}");
+
+    cy.get(".context-expression > div > table > tbody > tr:eq(1) td:eq(0)")
+      .should("be.focused")
+      .type("{rightarrow}{rightarrow}{rightarrow}{downarrow}");
+
+    cy.get(".context-expression > div > table > tbody > tr:eq(2) td:eq(2)").should("be.focused").type("{uparrow}");
+
+    cy.get(".context-expression > div > table > tbody > tr:eq(1) td:eq(2)").should("be.focused");
+  });
```
jomarko: Thank you for this test, however I think we do not test keyboard navigation in nested relation, but just in context entries.

My understanding is, to test nested relation, the focused cell should be one of "nested 1", "nested 2" or "nested 3", like attached. Created manually, not as result of your tests.

Screenshot from 2022-02-02 14-19-14

fantonangeli:
Thank you @vpellegrino for your encouragement and your comments.
I will answer and fix your points. 

Regarding `FocusUtils.ts`: I didn't write unit tests because uses querySelector() and there is very little logic inside, so I focused more on testing it with the e2e tests. 
But maybe some unit test makes sense and I can write them.

packages/boxed-expression-component/it-tests/integration/decision_table_expression_spec.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as buildEnv from "@kie-tools/build-env";
+
+describe("Decision Table Keyboard Navigation Tests", () => {
```
jomarko, line 143: Please do not introduce new file. Please use `decision_table_expression.ts` and rename it to `decision_table_expression_spec.ts`. The missing "_spec" in the name is probably some historical mistake.

jomarko (commented):
@fantonangeli thank you for the PR. As Valentino wrote, I also appreciate you added cypress tests.

Some comments to them inline. One general question. The locators you introduced in cypress tests are heavily index based, like:
```
parent child:eq(index_1) another_child:qe(index_2)
```

Such locators is not easy to understand on the first look. While comparing with test "Keyboard navigation" in "relation_expression_spec.ts" there you use locators referring either to cell text content, or some attributes, that gives at least some hint, what area of editor is tested. Do you think can we improve this somehow?

### Manual review
#### Online-Editors always saving
If you check the @kie-tools/online-editor, each arrow key press causes save operation is triggered, what is an performance issue I believe. Do you think can we handle it somehow? You can notice it if you will check the "Saved" indicator below during work with the editor.
Screenshot from 2022-02-02 13-29-21

#### Copy and Paste scenario
New DMN Expression editor support a copy and paste operation. The problem is that the paste will not work, if user navigates to the demanded cell by arrow keys. If the target cell is selected by mouse click, the paste still works. Do you think can we address this as part of this PR, Or we need to do it separately?

packages/boxed-expression-component/src/components/Table/Table.css
fantonangeli: The `outline: none` avoids this:
image

The `important` is actually not really needed, but I want to make sure that the default ":focus-visible" behaviour of the browser doesn't affect the cell in cross-browser uses. Firefox is not affected by this, only Chrome

fantonangeli:
Thank you for your comments, @jomarko.
I understand the issue of using indexes, I will try to rewrite my tests making more use of `ouiaId()` and `contains()` now to make it more readable.

**Online-Editors always saving**
Thanks a lot for noticing it, this will be my priority.

**Copy and Paste scenario**
I can open the context menu if the user press enter on the first cell of a row. Do you agree?
I can do it in this PR.

packages/boxed-expression-component/it-tests/integration/context_expression_spec.ts
fantonangeli: I do the keyboard navigation tests for the nested table in the test that run after this test. Here I was testing that the navigation on the parent is going well.
I should rename this test to make it clear I'm only testing the parent.

jomarko:
@fantonangeli thank you.

Regarding *Copy and Paste scenario* I think we do not understand each other. Please try to do folowing using new boxed expression editor
- open a decision table that
- drag a rectangle by mouse over input column cells , to select them
- press CTRL+C
- click the first output column cell to select it
- press CTRL+V 

input column cell values will be pasted correctly into outputput column cells. The problem is, that values will be not pasted, if user select the output cell by keyboard arrow keys, instead of mouse click. We can have a call to demonstrate better if you wish.

fantonangeli:
Hi @jomarko
I had a look at the code, my suggestion would be to do the copy&paste with a different Jira issue and a different PR.
Because other interactions are involved there. 
Please, let me know your opinion

jomarko:
@fantonangeli sure, I created https://issues.redhat.com/browse/KOGITO-6719

packages/boxed-expression-component/src/components/Table/common/FocusUtils.ts
```diff
@@ -39,3 +39,130 @@ export const focusNextTextArea = (currentTextArea: HTMLTextAreaElement | null) =
...
+/**
+ * Focus Upper Cell of a react-table. Works from any element inside a cell or a cell itself.
+ *
+ * @param currentEl the crrent element
+ * @param rowIndex the current row index
+ * @returns
+ */
+export const focusUpperCell = (currentEl: HTMLElement | null, rowIndex: number): void => {
+  const currCell = <HTMLTableCellElement>getParentCell(currentEl);
+  const currBody = currCell.closest("tbody");
```
ljmotta: Please, make it consistent with the `currentEl`. Use `currentCell` and `currentBody`

packages/boxed-expression-component/tests/components/test-utils.tsx
```diff
@@ -112,7 +112,22 @@ export function usingTestingBoxedExpressionI18nContext(
     ctx: usedCtx,
     wrapper: (
       <I18nDictionariesProvider defaults={usedCtx.defaults} dictionaries={usedCtx.dictionaries} ctx={usedCtx.ctx}>
-        {usedCtx.children}
+        <BoxedExpressionGlobalContext.Provider
```
ljmotta: We should keep the contexts independent. This context is already available by using the `usingTestingBoxedExpressionProviderContext` or `wrapComponentInContext` methods, no need to pass it here.

packages/boxed-expression-component/src/components/Table/common/FocusUtils.ts
```diff
@@ -39,3 +39,130 @@ export const focusNextTextArea = (currentTextArea: HTMLTextAreaElement | null) =
...
+/**
+ * Focus Lower Cell of a react-table. Works from any element inside a cell or a cell itself.
+ *
+ * @param currentEl the crrent element
+ * @param rowIndex the current row index
+ * @returns
+ */
+export const focusLowerCell = (currentEl: HTMLElement | null, rowIndex: number): void => {
+  const currCell = <HTMLTableCellElement>getParentCell(currentEl);
+  const currBody = currCell.closest("tbody");
```
ljmotta: Same as above

ljmotta (commented):
Thanks for this PR @fantonangeli! I've made some comments inline. 

Also, I've made some manual tests and I got a regression on the `tab` funcionality for the DMN Runner Table. I'm not sure if this is intended for the Boxed Expression Editor, @vpellegrino could you check it, please?

## DMN Runner Table

### before this PR

It is necessary to use the `tab` key one time to start editing the next cell

https://user-images.githubusercontent.com/24302289/152366596-cdf0b96f-34f8-484a-97bb-6d173d8c5739.mp4

### after this PR

It is necessary to use the `tab` key two times to start editing the next cell

https://user-images.githubusercontent.com/24302289/152366702-87e7499a-71bf-42bd-8b5d-dace66fca728.mp4


## Boxed Expression Component

### before this PR

The `tab` select the next cell and make it editable.

https://user-images.githubusercontent.com/24302289/152366943-da501a7d-1542-4fb7-9aa9-15e0bd6ed28c.mp4

### after this PR

The `tab` select the next cell and it's not possible to edit it unless you click on it, or press `enter`.

https://user-images.githubusercontent.com/24302289/152366978-6312e3cd-0e26-442c-8add-611a6b96cc37.mp4

fantonangeli:
Hi @ljmotta,
thanks for your review and your suggestions.

Regarding the keyboard navigation, I followed the old version, replicating its behaviour. 
At that version, we only have the arrow keys navigation and in the new one, we only have the tab navigation for textareas and not all cells.

In the Jira issue https://issues.redhat.com/browse/KOGITO-3655 we want to navigate regular cells with arrow keys and tabs.
If I keep the previous behaviour for tab keys we will have 2 different behaviours:
- arrow keys can navigate every cells, then press enter to enter edit mode. Same as old editor
- tab keys can navigate only textarea cells going directly in edit mode

After this task there is https://issues.redhat.com/browse/KOGITO-6608
In that Issue I will implement the keyboard navigation for header cells. The tab key will have to navigate on that cells, then press enter to open the context menu (differently from textarea cells).

My suggestion would be to keep the same behaviour for tabs and arrow keys, like a Google Spresheet.
Please let me know your viewpoint.

vpellegrino:
I got @ljmotta point: after the user pressed the `tab`, it was immediately possible to edit the next cell.

I also got the point from @fantonangeli about replicating what was happening in the previous editor. For sure we need to be consistent about the experience with `tab` and `arrow` keys.

Let's wait for @jomarko to see what is his opinion about the expected behavior. Maybe we can keep as a first stage the previous editor's experience and then trace an improvement for having a stylesheet experience as @ljmotta has described.

ljmotta:
@fantonangeli Thanks for your clarification.

If the idea is to replicate the stylesheet experience, we should make it possible to edit a cell by pressing any "input" key if it was already selected. With this PR we have two different behaviors for the selected cell:
 - If it was used a `click` to select a cell, it's possible to edit it using any "input" key;
 - if it was used an `arrow` or `tab` key to select a cell, It's necessary to use the `enter` key to be able to edit it.

Also, a stylesheet uses two behaviors for tabs and arrows.
 - arrows and tabs to move around if it's **not** on edit mode.
 - tabs to move even if it's on edit mode.

@vpellegrino The actual behavior (before this PR) is already similar to the stylesheet experience. The `tab` and `click` already work as expected. My two cents is that I believe we would be doing a regression by changing it to the old editor behavior and then updating it again afterward.

yesamer:
@fantonangeli @jomarko @vpellegrino @ljmotta  Let's tackle this together guys, in order to agree the best implementation. 
Quick call / chat tomorrow?

ljmotta:
@yesamer Looks good to me.

fantonangeli:
Just keeping a note here that today we updated the points in the Jira task
https://issues.redhat.com/browse/KOGITO-3655

vpellegrino (approved):
### Code
From the code perspective, this PR looks really good!

### Edit
Seems good to me!

### Navigation
The navigation with the keyboard is really improved! I just tried it and it looks fun. I just found out that:
1. Once I am on an empty expression and I press `enter` on the "Select expression" cell, it is not possible to navigate the "select logic type" menu. Not sure if it should be covered by this task (maybe not), but it would be really nice to have it working.
2. For the _Function (PMML)_ expression is not possible to open the PMML literal expression menu with `enter` button:
image
Really side comment. In my opinion, it does not deserve to block this PR.

### Final considerations
I am really happy that you started to contribute to this project. This task represents a good improvement from a user-experience perspective. Thanks for leveraging it!

fantonangeli:
Thank you @vpellegrino 

Thanks a lot for your points:

I agree with point 1, but because there are different types of context menus and select input, I think it should be covered by a different PR. Maybe I'm wrong but this also looks a bit "out of scope" for this task.
Before that PR it can be better to finish [KOGITO-3655](https://issues.redhat.com/browse/KOGITO-3655) that I already started.

Regarding point 2: I didn't know this particular case about PMML. I will do it with priority because I noticed a bug in this case: if you open the select input, the keyboard navigation it's not blocked. 
Because of this bug, I convert now this PR into a draft and fix it.

Thanks again for your reviews.

jomarko:
@fantonangeli thank you for updates, I found this point I think worth to discuss. As the PR is openend quite a long time I think also valid, if you say should be handled separately later. 

I tried this simple scenario:
- select cell
- edit its value
- press Tab to confirm changes and move to other cell
- Start edit **just by keyboard** again

If I compare this behavior directly in DMN expression vs DMN Runner. It behaves differently.

In my screenshot:
- I Typed 651 into FICO column
- Pressed Tab
- Tried to start edit Age by (Enter, just start typing) but it didn't start edit mode. Surprisingly repeated Tab press started Age edit mode

Is that expected?

Screenshot from 2022-02-16 12-11-11

fantonangeli:
Hi @jomarko 
After some investigation, I found the bug: the `keyboard-shortcuts` package catches the keyboard events preventing the table to listen to that event. 
Because my knowledge of the full project is still limited, I don't know the implication of that particular line of code. I will need to ask someone.

There is still work to do for the keyboard navigation to the DMNRunner on the `unitable` package, mostly to interact with its inputs. 

I think we can in the meantime merge this PR that it's related to the `boxed-expression-component` package and then work to the DMNRunner which seems a separate work.

Do you agree? I would also ask the opinion of @vpellegrino and @yesamer

vpellegrino:
@fantonangeli what you are saying sounds fine from my point of view.
Just waiting for @ljmotta, for his point of view about https://github.com/kiegroup/kie-tools/pull/825#issuecomment-1041381036

ljmotta:
@fantonangeli The `keyboard-shortcuts` is an API that handles keyboard events in the `envelope` (the Editor). To summarize, it catches the events and bubbles them up to the `channel` (in this case, the KIE Sandbox). It would be strange that the bug origin is in the `keyboard-shortcuts` as the DMN Runner Table resides in the `channel`. Also, if you have any questions regarding the codebase, feel free to reach me. I'll do my best to answer them. :grinning: 

After performing some tests, I found that the bug cause is on the `TableBody`, but I couldn't tell where exactly. It's possibly missing a custom behavior for the DMN Runner Table on the `tdBaseProps`.

Regarding improving the navigation for the DMN Runner Table, I think it's ok to have another PR.

jomarko (commented):
I didn't went trough each possible cell of each possible expression type, rather tried scenarios including some other areas of sandbox and found this:
- https://issues.redhat.com/browse/KOGITO-6791
- https://issues.redhat.com/browse/KOGITO-6402

In the VS Code Extension my sanity checks will be done tomorrow.

fantonangeli:
@ljmotta 
What I found is the line 38 of "kie-tools/packages/keyboard-shortcuts/src/channel/Hooks.tsx" 
`  const stopPropagation = (ev: KeyboardEvent) => {
    ev.stopPropagation();
  };
`
Commenting on it I saw that TableBody was receiving the keyboard events. But commenting it, probably will create some bug elsewhere and we need to do it in a different way.

ljmotta:
@fantonangeli 
Just correcting myself. The `keyboard-shortcuts` catch the keyboard events from the `channel` and passes them down to the `envelope` (previously, I wrote the wrong flow).

Indeed, this line is stopping the propagation of all `keydown`, `keyup`, and `keypress` events inside the `selector` nodes. The expected behavior was to stop only events from the `channel` to the `envelope`. This was done to avoid shortcuts being fired from inside the table. So you're right. Removing it will create some bugs :sweat_smile:. I've created a [JIRA](https://issues.redhat.com/browse/KOGITO-6792) to address this issue.

Still, regarding the `keyboard-shortcuts`, the `IGNORED_TAGS` constant shouldn't contain `layout` tags, such as `TD`. It will disable keyboard shortcuts from all `TD` tags in the `channel`. We can have a `TD` element on another component, and the keyboard shortcuts will not fire.

I took a deeper look into the tab issue. Removing the suggested line will fire the `onKeyDown` method from the `tdBaseProps`, but it will not work. After some testing, I've removed the `tabIndex` property from the `tdBaseProps`, and it worked even with the `stopPropagation`. I'm not sure why the property is required, but I believe we can disable it for the DMN Runner Table, WDYT?

Also, while doing these tests, `yarn start --env live` wasn't working. Firstly due to a miss typing. I can open a PR to your branch providing the fix if it's ok. Secondly, due to the casting using the `<>` operator. In this project, we use the `as Type` to cast because of the conflict with React tags.

fantonangeli:
@ljmotta 
Thank you for your very interesting and useful comment!

"have a TD element on another component"... do you mean in a separated component? 

I put `TD` in `IGNORED_TAGS` because `keyboard-shortcuts` was catching the arrow keys and continuously saving. Would you some suggestions to do it in a different way?

I used `TABINDEX` on `TD` tags to make them focusable, otherwise, I cannot use the `:focus` CSS selector.

Good to know about `as Type` casting, thanks.
Thanks again for your comments, please feel free to open a PR on my branch or if it's just about the castings I can do it easily ;-)

ljmotta:
@fantonangeli 
Yes, in any other component on the "KIE Sandbox" or any other `channel` that uses the `keyboard-shortcuts` to send keyboard events to the `envelope`.

I'm going to investigate why this is happening. As far I see, this doesn't happen on the `Stable` version. Another thing that I've observed is the arrows are moving the selected node behind the Boxed Editor. This is why the "save" action is triggered.

What I suggest to fix the DMN Runner Table tab issue is to pass down the `isRunnerTable` property on the `BoxedExpressionGlobalContext`, and consume it on the `TableBody`. Then remove the `tabIndex` property on the `tdBaseProps` in case `isRunnerTable === true`.

fantonangeli:
@ljmotta
Thanks a lot for your help. 
If I remove the `tabIndex` for the DMN Runner, we will not have the blue border for the selected cell.
Is that ok for that particular case?

jomarko:
@vpellegrino @fantonangeli @yesamer @danielzhe hello.

I am sorry. I struggle to check VS Code extension as I promised here https://github.com/kiegroup/kie-tools/pull/825#pullrequestreview-884816232.

I am not able to open any DMN file.

I tried to install vsix:
- downloaded from github action CI result of this PR
- built locally on main branch as: `yarn build:prod:until vscode-extension-pack-kogito-kie-editors`
- built locally of kie-tools#825 PR branch as: `yarn build:prod:until vscode-extension-pack-kogito-kie-editors`


Is it only me or do we have a blocker please?

Screenshot from 2022-02-17 07-53-25

vpellegrino:
@jomarko do you have the chance to try the `vsix` produced from the `main` branch?

jomarko (approved):
vsix issue nto related to this pr, confirmed on private communication channel. approving

fantonangeli:
I fixed the problem for the tab in the DMNRunner. But there is still the need to do the full keyboard navigation for the DMNRunner in a different PR. I used the `tabIndex` set to '-1'.
@ljmotta I converted all the castings in FocusUtils.ts.
For the future, we can evaluate to set the `tsconfig.json` with: `no-angle-bracket-type-assertion`. 
https://palantir.github.io/tslint/rules/no-angle-bracket-type-assertion/

ljmotta:
@fantonangeli
After some research, I understood what is happening with the shortcuts. I found this enlightening [blog post](https://dev.to/romaintrotard/under-the-hood-of-event-listeners-in-react-4g01) explaining how React events work under the hood.  :tada:

The `keyboard-shortcuts` utilizes event listeners attached directly to the DOM. These events are going to be triggered before the `onKeyDown` prop because React sends the events to the root node and then sends them to the `target` node.

So what we have is a project limitation. To remove the `TD` from the `IGNORED_TAGS`, we need to use events attached directly to the DOM instead of utilizing the `onKeyDown` prop. With this in place, we can isolate all key events from the Boxed Expression Editor using the same hook that is used on the DMN Runner Table. I will open a PR to your branch with a possible solution, feel free to change anything you want.

Another thing that I've realized with this PR is when we select a cell using a mouse, we select the `textarea` if it's a `data` cell of the `Relation Expression` or `Decision Table Expression`, we select an `input` on the `DMN Runner Table`. Utilizing the navigation, it always selects the `TD` of the next cell and not what is inside of it, like the mouse `click`. This can look harmless, but this will change the functionality of the "Tab" key on the DMN Runner Table.

fantonangeli:
@ljmotta 
Thank you very much for your help. 
This solution was very difficult to find. I used the logic on TD listening because I have many different cases and different actions to manage. It was also not easy for me to communicate from the TD to the component inside the cell, due to the structure of the `react-table`. 
Unfortunately, I didn't know the DMN Runner was using our `Table` component and this PR would have affected that.
After this PR I will have to manage also the header cells keyboard navigation with colspans/rowspans, and this will add more cases to the list.
With this in mind, I wanted to keep the code as simple as possible with all the logic called from `TableBody` component.
I'm open to evaluating together other solutions.
Thanks again for your PR, I only changed the command to test the TAB key with Cypress.

packages/boxed-expression-component/it-tests/integration/decision_table_expression_spec.ts
```diff
@@ -0,0 +1,319 @@
...
+    cy.contains("td", /cell 4/)
+      .should("be.focused")
+      .click({ force: true })
+      .type("{downarrow}{downarrow}{downarrow}{downarrow}");
+
+    // check the cell 12 is focused
+    cy.contains("td", /cell 12/).should("be.focused");
+  });
+
+  it("Edit cells appending text selecting Td", function () {
```
jomarko: what is the reason we are not consistent for `it` please?
- it("text", () => {
- it("text", function() {

fantonangeli: For sure an editor code snippet that I didn't see, but no tech reasons.
I will fix it together with other possible modification

tiagobento:
Is this good to merge? @ljmotta @yesamer @vpellegrino

yesamer (approved):
@fantonangeli Thank you, LGTM. Well done!

ljmotta:
@fantonangeli 
Indeed, it doesn't look like a simple solution, and the project is very complex, with many dependencies and packages. So don't worry about not knowing everything, you will catch it in time.

Regarding the DMN Runner Table, it is still one downgrade to be fixed. Now that the `onKeyDown` method is used by the DMN Runner Table, the Tab shortcut stopped working one more time, it focuses on the next cell, but it's not possible to edit it. This happened because the "focus" methods use exclusively `td` elements for the "nextSibling". Maybe, it's necessary to check the type of element that triggered the action and focus on the next sibling that is equal to the `target`.

fantonangeli:
@ljmotta 
To temporarily fix this downgrade, I wrote a simple option to disable the Keyoard Navigation from the `DMNRunner`.
You can test it in the last commit.
After this PR is approved we can continue working on the Keyoard Navigation on the `DMNRunner` in the issue below with another PR.
https://issues.redhat.com/browse/KOGITO-6791
Peek 2022-02-23 15-55

yesamer:
@fantonangeli Can you please resolve the below conflict?

ljmotta:
@fantonangeli Thanks a lot for this fix.

ljmotta (approved):

vpellegrino (approved):
Fine for me.

yesamer:
@fantonangeli can I merge this?

fantonangeli:
@yesamer yes, please. 
From my viewpoint, it's ready to be merged.

PR #828, opened 2022-02-01
Mutually exclusive decision to run `CI` or `CI :: Full`

tiagobento:
With GitHub Action's `paths` and `paths-ignore`, we can't properly do that.

This PR changes this decision to a manual shell check, which give us more flexibility and saves resources.


tiagobento:
Note that the checks won't behave properly on this PR, as `merge-pr-changes` action comes from `main`.

thiagoelg (approved):

PR #829, opened 2022-02-02
Fix CI workflows

tiagobento:
The `!` operator was in the wrong place, causing `CI :: Full` to run every time.

I'm also fixing the `--since` flag on the normal CI.


caponetto (approved):

tiagobento:
Merging this as without this fix, all other PRs are impacted.

PR #847, opened 2022-02-07
KOGITO-6745: Extend VS Code Channel API to enable support to custom implementations

pefernan:


pefernan:
jenkins restest this please

packages/vscode-extension/src/index.ts
```diff
@@ -64,7 +67,8 @@ export async function startExtension(args: {
     vsCodeNotificationsApi,
     vsCodeJavaCodeCompletionChannelApi,
     args.viewType,
-    vsCodeI18n
+    vsCodeI18n,
+    args?.channelApiProducer
```
tiagobento:
```suggestion
    args.channelApiProducer
```

packages/vscode-extension/src/KogitoEditorChannelApiProducer.ts
```diff
@@ -0,0 +1,55 @@
...
+import { BackendProxy } from "@kie-tools-core/backend/dist/api";
+import { NotificationsApi } from "@kie-tools-core/notifications/dist/api";
+import { JavaCodeCompletionApi } from "@kie-tools-core/vscode-java-code-completion/dist/api";
+import { I18n } from "@kie-tools-core/i18n/dist/core";
+import { VsCodeI18n } from "./i18n";
+import { Uri } from "vscode";
+
+/**
+ * Produces instances of KogitoEditorChannelApi to be used if we want to provide the extension a channel api with
+ * custom features.
```
tiagobento:
```suggestion
 * Produces instances of KogitoEditorChannelApi to be used if we want to provide the extension a Channel API with
 * custom features.
```

tiagobento (changes requested):
I don't know if this is in scope for this PR, but I feel like `startExtension` should receive a generic parameter now, like T extends KogitoEditorChannelApi, and it should return a way to communicate with the Editor that is there (e.g. the envelopeApi). This generic parameter would be extended to the new Producer interface too.

If we want to provide custom features for specific Editors, this PR is adding a good way of providing methods to the Editors, but we're not able to consume them still. Being able to consume a custom Envelope API would also be a good improvement. WDYT?

tiagobento (approved):

paulovmr (approved):

tiagobento:
After speaking with Pere, we decided that it's going to be too much for us to have this type of generic behavior right now, so this PR is enough for now.

PR #848, opened 2022-02-08
KOGITO-6669: Search Service for Domain Objects

danielzhe:
**The problem:**
When you select a cell in the the new boxed expression edition, you must to select the Domain Object related to the cell in the GWT/Java world in order to populate the properties panel in the right side of the editor.

**The solution:**
When a cell is selected in the boxed expression editor, its communicates the GWT/Java world about that passing the UUID of the selected cell.

This PR searches for the respective Domain Object in the GWT/Java world and select it.

In order to make the test easily while the integration with new boxed expression editor is not done, I added a search box in the DMN testing webapp:
image
*This search box can be removed once the new boxed expression editor have its integration with the search service implemented.*

This search box works when you are in boxed expression editor. It just select the item with the UUID that you searched, so you can check if the search is working by accessing the properties panel at the right and see if the item that you searched is selected. Also noticed that the rectangle around the selected item in the boxed expression is not triggered by this mechanism.

If a item not found or a blank one is searched, the selection is cleared.

For the integration with boxed expression editor, a method called  "selectObject" was created in the file `BoxedExpressionService`. @vpellegrino can change the name of this method to another one if he wants when the integration was implemented.

Demo video of the featuring working: https://issues.redhat.com/secure/attachment/12686691/12686691_search+domain+object.mp4

### DIFF
--- a/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
+++ b/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2022 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.dmn.api.definition.model;
+
+import org.kie.workbench.common.stunner.core.domainobject.DomainObject;
+
+/**
+ * Defines a class that contains {@link DomainObject} that is searchable.
+ */
+public interface HasDomainObject {
+
+    /**
+     * Find a {@link DomainObject} with the given UUID.
+     * @param uuid The UUID.
+     * @return The {@link DomainObject} or null if none is found.
+     */
+    DomainObject findDomainObject(final String uuid);
+}
--- a/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
+++ b/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
@@ -16,6 +16,8 @@
 
 package org.kie.workbench.common.dmn.api.definition.model;
 
+import java.util.Optional;
+
 import org.kie.workbench.common.stunner.core.domainobject.DomainObject;
 
 /**
@@ -26,7 +28,7 @@ public interface HasDomainObject {
     /**
      * Find a {@link DomainObject} with the given UUID.
      * @param uuid The UUID.
-     * @return The {@link DomainObject} or null if none is found.
+     * @return The {@link DomainObject} or empty if none is found.
      */
-    DomainObject findDomainObject(final String uuid);
+    Optional<DomainObject> findDomainObject(final String uuid);
 }
--- a/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/Relation.java
+++ b/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/Relation.java
@@ -24,8 +24,10 @@ import org.kie.workbench.common.dmn.api.definition.HasTypeRef;
 import org.kie.workbench.common.dmn.api.property.dmn.Description;
 import org.kie.workbench.common.dmn.api.property.dmn.Id;
 import org.kie.workbench.common.dmn.api.property.dmn.QName;
+import org.kie.workbench.common.stunner.core.domainobject.DomainObject;
 import org.kie.workbench.common.stunner.core.util.HashUtil;
 
+import static org.kie.workbench.common.dmn.api.definition.model.common.DomainObjectSearcherHelper.find;
 import static org.kie.workbench.common.dmn.api.definition.model.common.HasTypeRefHelper.getFlatHasTypeRefs;
 
 @Portable
@@ -65,6 +67,11 @@ public class Relation extends Expression {
         clonedRelation.column = column.stream().map(InformationItem::copy).collect(Collectors.toList());
         clonedRelation.row = row.stream().map(List::copy).collect(Collectors.toList());
         return clonedRelation;
+    }
+
+    @Override
+    public DomainObject findDomainObject(final String uuid) {
+        return find(getRow(), uuid);
     }
 
     public java.util.List<InformationItem> getColumn() {
--- a/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/Relation.java
+++ b/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/Relation.java
@@ -16,6 +16,7 @@
 package org.kie.workbench.common.dmn.api.definition.model;
 
 import java.util.ArrayList;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -28,6 +29,7 @@ import org.kie.workbench.common.stunner.core.domainobject.DomainObject;
 import org.kie.workbench.common.stunner.core.util.HashUtil;
 
 import static org.kie.workbench.common.dmn.api.definition.model.common.DomainObjectSearcherHelper.find;
+import static org.kie.workbench.common.dmn.api.definition.model.common.DomainObjectSearcherHelper.getDomainObject;
 import static org.kie.workbench.common.dmn.api.definition.model.common.HasTypeRefHelper.getFlatHasTypeRefs;
 
 @Portable
@@ -71,6 +73,10 @@ public class Relation extends Expression {
 
     @Override
     public DomainObject findDomainObject(final String uuid) {
+        final DomainObject domainObject = getDomainObject(getColumn(), uuid);
+        if (!Objects.isNull(domainObject)) {
+            return domainObject;
+        }
         return find(getRow(), uuid);
     }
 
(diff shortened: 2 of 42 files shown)


jomarko:
@danielzhe can we address the one reported codesmell?

https://sonarcloud.io/project/issues?id=org.kie.kogito%3Akogito-tooling&open=AX7at86Sn30aBnrCQ8nI&pullRequest=848&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL

It is minor, but seems as pretty easy to address it. What do you think?

vpellegrino:
Awesome job @danielzhe! 
I am going to check out your branch and try to put things together! Thank you for this `searcher`!

jomarko (commented):
Thank you for this useful PR @danielzhe. 

Here are my manual review findings:

Using the new search:
- I was not able to find header cell of literal expression
- I was not able to find header cell of relation columns


Not sure if that is related to this PR:
- all rows of List expression seems to have same UUID after creation? as a consequence, when I changed a list entry form litteral expression to decision table, UUIDs were missing completely for the decision table 

Screenshot from 2022-02-09 09-25-03

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.kie.workbench.common.dmn.api.definition.model;
+
+import org.kie.workbench.common.stunner.core.domainobject.DomainObject;
+
+/**
+ * Defines a class that contains {@link DomainObject} that is searchable.
+ */
+public interface HasDomainObject {
```
vpellegrino, line 26: One architectural observation: do we really need to introduce this interface? 
I guess (but maybe I'm wrong), that all classes that need to implement the `findDomainObject` method are already extending/implementing some existing class/interface.
For example, we have `DomainObject` itself (stunner code, and we can avoid changing it, ok), but we also have `DMNElement`. WDYT?

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Find a {@link DomainObject} with the given UUID.
+     * @param uuid The UUID.
+     * @return The {@link DomainObject} or null if none is found.
+     */
+    DomainObject findDomainObject(final String uuid);
```
vpellegrino: An observation on the signature of this method. Why don't we return `Optional<DomainObject>` avoiding to return `null` when none DomainObject is found?

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/Relation.java
```diff
@@ -67,6 +69,11 @@ public Relation copy() {
         return clonedRelation;
     }
 
+    @Override
+    public DomainObject findDomainObject(final String uuid) {
+        return find(getRow(), uuid);
```
vpellegrino, line 79: are we missing search on columns?

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-runtime/src/main/java/org/kie/workbench/common/dmn/showcase/client/editor/DMNDiagramEditor.java
```diff
@@ -210,6 +210,13 @@ public boolean isReadOnly() {
         });
     }
 
+    public Promise<Void> searchDomainObject(String uuid) {
+        return promises.create((resolve, reject) -> {
+            //commands.getUndoSessionCommand().execute();
```
vpellegrino: is this comment wanted?

vpellegrino (changes requested):
Code is clean and well organized. I have few observations @danielzhe 
Thanks again for it!

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
danielzhe: The problem is that we don't want all `DMNElement` to have `findDomainObject` , neither all `DomainObject`. It is only a few classes that have this.

vpellegrino: Ok, I got your point. However, IMHO, `HasDomainObject` is a bit misleading because here we want to search for an object that **is** a `DomainObject`, **and** it also has a set of properties that the user can edit/populate.
@danielzhe @yesamer @jomarko WDYT about `IsDmnElementWithPropertySet` or `IsDomainObjectWithPropertySet`?

jomarko: as we are dependent on `org.kie.workbench.common.stunner.core.domainobject.DomainObject` i vote for `IsDomainObjectWithPropertySet`

vpellegrino:
Hi @danielzhe, it seems that the search is not executed on BKM nodes.
Can you please double-check?
bkm_node_search_not_working

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
danielzhe: Because the last time I did something like that to avoid null check, the reviewers told me to not do that. It was in parameter, but I was trying to avoid to refactor again.

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
danielzhe: Sorry, I can't agree. It is not always a class that is domain object. If we change the interface for `IsDomainObjectWithPropertySet` it will give the false impression that, for example, the class `Relation` is a DomainObject which it is not. @jomarko 

 `Relation`, for example, *is not* a domain object and we want so search *in* `Relation` for *DomainObjects* . @vpellegrino

danielzhe:
@jomarko As discussed in GChat, this issue https://github.com/kiegroup/kie-tools/pull/848#pullrequestreview-877075863 you reported is not related to this PR.

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
vpellegrino: If I remember correctly, that was a different case. You were passing an Optional as input parameter.
That is discouraged, but it is totally fine to return an Optional.

Refer to:
https://www.baeldung.com/java-optional-return#:~:text=The%20Optional%20type%20was%20introduced,fewer%20NullPointerExceptions%20in%20the%20applications.

https://rules.sonarsource.com/java/RSPEC-3553

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
vpellegrino: Ok @danielzhe thanks for your opinion.
Please let me know what name(s) you suggest and I'm sure we will reach a consensus 😊

danielzhe:
@vpellegrino Changes applied!

jomarko:
Gettin vever ending loading right now, is it only me please?
Screenshot from 2022-02-11 11-58-09

vpellegrino:
> Gettin vever ending loading right now, is it only me please? Screenshot from 2022-02-11 11-58-09

@jomarko it works for me. I suggest deleting every `.errai` folder and trying to rebuild the whole project.

jomarko (approved):
thank you @danielzhe

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
danielzhe: I don't understand why `HasDomainObject` is not a good name, because the points that you raised are not correct and the name tells exactly what it is: "a class that contains domain objects".
1. We **are not** looking for "for an object that is a DomainObject, and it also has a set of properties that the user can edit/populate." 
2. We are looking only for a class that "(...) has a set of properties that the user can edit/populate" and those "set of properties" are `DomainObject`'s **and/or** `HasDomainObject`.

In that context I think the name `HasDomainObject` It says that a class "has a domain object" and that's exactly what it is. 

The name you suggested are not correct because they are misleading given the idea that the classes that implements this interface are `DomainObjects` but they are not!

So if the issue is the words "domain object" we can call it for another thing like `HasEditableProperties` with the method `findEditableProperty`, but again this will give another issues because `Decision` and `BusinessKnowledgeModel` it is not "properties" and we want to find them too...

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
danielzhe: Ok, I gonna change it. Thank you! @vpellegrino  =)

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/Relation.java
danielzhe: Fixed!

yesamer:
@danielzhe Can you please rebase this one as well?

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/definition/model/HasDomainObject.java
vpellegrino: ok, @danielzhe

danielzhe:
Changes applied, @vpellegrino ! Using `Optional` now. =)

Also it is rebased, @yesamer

vpellegrino (approved):
LGTM,
Thank you @danielzhe

vpellegrino:
@yesamer can we have a merge, please? Thank you

yesamer (approved):

PR #850, opened 2022-02-11
KOGITO-6496: Remove unused/unnecessary dependencies with KIE V7 from `stunner-editors`

yesamer:
https://issues.redhat.com/browse/KOGITO-6496

The scope of the PR is to begin the removal of 7.x stream dependencies.

- kie-parent version `7.58.0.Final` defined in the parent, will no change in this PR. A ticket with the goal to refactor it is already created https://issues.redhat.com/browse/KOGITO-6378. 
- Updated `version.org.kie` to version `8.17.0.Beta`, which is the latest available.
- `kie-soup-bom` hasn't been ported to version 8.x. Thus, we still need to import it from 7.x. For this reason, I defined a new propriety `version.org.kie7`  set to `7.65.0.Final`. A few words about this module. Considering it is mostly used for util usages with static methods like (`checkEmpty`, `checkNull`, Sets and Lists support management), we can remove this dependency in the future moving these utility classes inside the `stunner-editor` project. Open for suggestions
- `narayana-integration-bom` despite this seems not actually used in our project, it is imported in the parent pom. It doesn't exists in 8.x stream, so I was forced to add it. I think this can be removed with https://issues.redhat.com/browse/KOGITO-6378
- `jbpm-bom` not used - removed
- `kie-bom` and `drools-bom` removed, as `kie-dmn-bom` contains all the dependencies we need. 
-  `kie-dmn-bom`: moved to `8.17.0.Beta`. In 8.x stream a dmn module was removed (`drools-workbench-models-datamodel-api`. Therefore, I removed all the parts of the code using that module: `AsyncPackageDataModelOracle` and `WorkItemWidget`. Both are not used in Kogito.

Note: I tried to import `kogito-kie-bom` with version `1.17.0.Final` instead of `kie-dmn-bom` `8.17.0.Beta`, but unfortunately it doesn't declare all the dmn dependencies we need here (eg. `kie-dmn-feel-gwt`). This means we need to import `kie-dmn-bom` even in this case. An alternative is to modify `kogito-kie-bom` declaring the missing deps, so we can import `kogito-kie-bom` with version `1.17.0.Final` instead of `kie-dmn-bom` `8.17.0.Beta`. Open to suggestions for this. In the meanwhile, I think this is the best option.


yesamer:
Please don't review this yet.

yesamer:
@tiagobento @romartin @jomarko Ready for the review. Thank you.

packages/stunner-editors/pom.xml
```diff
@@ -237,26 +238,18 @@
...
       <dependency>
-        <groupId>org.drools</groupId>
-        <artifactId>drools-bom</artifactId>
-        <version>${version.org.kie}</version>
+        <groupId>org.kie</groupId>
+        <artifactId>narayana-integration-bom</artifactId>
+        <version>${version.org.kie7}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
```
tiagobento, line 255: Shouldn't we remove those two as well? This is still coming from KIE V7.

yesamer: @tiagobento I explained in the PR description the reason I didn't remove them

tiagobento: @yesamer 🙈

Sorry Yeser, my bad. Let me read it carefully :)

jomarko (approved):
@yesamer thank you for the PR
- I was not able t check VSIX due to known issue, thus scesim was not checked at all
- I checked online-editor channel, there I tried especially to execute some model in DMN runner
- I think even more cleanup can be done, see my last two commits in #869, however to have reasonable PR size, better to merge after your PR is in probably

yesamer:
@jomarko Thank you for your contribution, I totally appreciate it. After we merge this, we should address the `kie-soup` removal in another ticket. Considering these are mostly utils methods, we can:
- Move the utility method inside a uberfire/appformer package which is reachable by all project leaf:
- Substitute the utilities method with java native API (eg. `checkNull` -> `Objects.isNull`

BTW, I rebased this PR with the vsix extension patch, so I expect vsix should work (and it works to me)

//cc @tiagobento

tiagobento:
Ok, after speaking with @yesamer, we agreed that this PR can go in and two more PRs should follow.

1. Removing kie-soup-bom and potentially duplicating the utility methods that we need.
2. Removing the parent POM and moving what's necessary to the parent POM of `stunner-editors`.

After this, `stunner-editors` will be its own thing, and not depend on any other streams, which frees the way for us to start on [KOGITO-6378](https://issues.redhat.com/browse/KOGITO-6378).

Thanks Yeser!

tiagobento (approved):

tiagobento:
Should we merge this?

yesamer:
@tiagobento Please proceed

PR #865, opened 2022-02-21
KOGITO-6808: Create a basic mechanism to open/close Widgets on top of the Monaco Editor for the Serverless Workflow Editor

tiagobento:


paulovmr (approved):

packages/serverless-workflow-editor/src/monaco/augmentation/completion/index.ts
```diff
@@ -15,73 +15,143 @@
...
-          language: augmentation.language,
-          monacoContext: {
-            model,
-            context,
-            token,
-            position,
-          },
-        });
-      }
+      const workflow = Specification.Workflow.fromSource(model.getValue());
```
pefernan: my only concern would be that parsing the editor content to a Workfoow here would make autocomplete not work if the editor content doesn't meet the SW specs, which may happen often during the edition time. For example adding new items in the functions array or similar...

tiagobento: Good point! Take a look at b46a16f

pefernan (approved):
+1

pefernan (approved):

PR #868, opened 2022-02-22
KOGITO-6807: Implement undo, redo and isDirty on the Serverless Workflow Editor

paulovmr:


packages/serverless-workflow-editor/src/editor/ServerlessWorkflowEditor.tsx
```diff
@@ -103,33 +114,46 @@ const RefForwardingServerlessWorkflowEditor: React.ForwardRefRenderFunction<
         },
       };
     },
-    [content]
+    [monacoEditorRef]
```
tiagobento: Refs don't need to be dependencies, as React guarantees that they're reference is unchanged.

packages/serverless-workflow-editor/src/editor/ServerlessWorkflowEditor.tsx
```diff
@@ -37,6 +39,13 @@ interface Props {
    */
   ready: () => void;
 
+  /**
+   * Delegation for KogitoEditorChannelApi.kogitoEditor_stateControlCommandUpdate(command) to signal to the Channel
+   * that the editor is performing an undo/redo operation. Increases the decoupling of the ServerlessWorkflowEditor
+   * from the Channel.
+   */
+  stateControlCommandUpdate: (command: StateControlCommand) => void;
```
tiagobento:
```suggestion
  onStateControlCommandUpdate: (command: StateControlCommand) => void;
```

packages/serverless-workflow-editor/src/editor/ServerlessWorkflowEditor.tsx
```diff
@@ -37,6 +39,13 @@ interface Props {
    */
   ready: () => void;
```
tiagobento: on
```suggestion
  onReady: () => void;
```

tiagobento: and `onNewEdit` too.

packages/serverless-workflow-editor/src/editor/ServerlessWorkflowEditor.tsx
```diff
@@ -103,33 +114,46 @@ const RefForwardingServerlessWorkflowEditor: React.ForwardRefRenderFunction<
...
   );
 
-  useEffect(() => {
-    props.ready();
-  }, [originalContent, props]);
+  const contentChanged = useCallback(
+    (newContent: string, operation?: MonacoEditorOperation) => {
+      if (operation === MonacoEditorOperation.EDIT) {
+        props.newEdit(new KogitoEdit(newContent));
+        monacoEditorRef.current?.pushUndoStop();
```
tiagobento: Do you need to call this here? I mean, this seems like a piece of logic that could be moved to inside the Monaco editor component at the same place you call `onContentChanged` with the operation equal to MonacoEditorOperation.EDIT. This way we avoid adding this `pushUndoStop` method to the ref interface and simplify things. WDYT?

packages/serverless-workflow-editor/src/editor/ServerlessWorkflowEditor.tsx
```diff
@@ -103,33 +114,46 @@ const RefForwardingServerlessWorkflowEditor: React.ForwardRefRenderFunction<
...
       };
     },
-    [content]
+    [monacoEditorRef]
   );
 
-  useEffect(() => {
-    props.ready();
-  }, [originalContent, props]);
+  const contentChanged = useCallback(
```
tiagobento:
```suggestion
  const onContentChanged = useCallback(
```

packages/serverless-workflow-editor/src/monaco/augmentation/MonacoEditorApi.ts
```diff
@@ -23,28 +23,53 @@ export interface MonacoEditorApi {
...
+  constructor(
+    content: string,
+    onContentChange: (content: string, operation: MonacoEditorOperation) => void,
+    augmentation: MonacoAugmentation
+  ) {
     this.model = editor.createModel(augmentation.language.getDefaultContent(content), augmentation.language.languageId);
-    this.model.onDidChangeContent((event) => onContentChange(this.model.getValue()));
+    this.model.onDidChangeContent((event) => {
+      if (!event.isUndoing && !event.isRedoing) {
+        onContentChange(this.model.getValue(), MonacoEditorOperation.EDIT);
```
tiagobento, line 55: Here I think you could call the `pushUndoStop`. Would that work?

packages/serverless-workflow-editor/src/monaco/augmentation/MonacoEditorApi.ts
```diff
@@ -23,28 +23,53 @@ export interface MonacoEditorApi {
...
+      }
+    });
+    this.onContentChange = onContentChange;
     this.augmentation = augmentation;
   }
 
   redo(): void {
+    this.editor?.focus();
     this.editor?.trigger("whatever...", "redo", null);
```
tiagobento:
```suggestion
    this.editor?.trigger("??????", "redo", null);
```

What's the first parameter here? I think we could use something better than `whatever` 😝

packages/serverless-workflow-editor/src/monaco/augmentation/MonacoEditorApi.ts
```diff
@@ -23,28 +23,53 @@ export interface MonacoEditorApi {
...
   }
 
   undo(): void {
+    this.editor?.focus();
     this.editor?.trigger("whatever...", "undo", null);
```
tiagobento: Same here.

packages/serverless-workflow-editor/src/monaco/augmentation/MonacoEditorApi.ts
```diff
@@ -57,7 +82,21 @@ export class DefaultMonacoEditor implements MonacoEditorApi {
...
+    });
+
+    this.editor.addCommand(KeyMod.CtrlCmd | KeyCode.KEY_Y, () => {
+      this.onContentChange(this.model.getValue(), MonacoEditorOperation.REDO);
+    });
+
+    this.editor.addCommand(KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.KEY_Z, () => {
+      this.onContentChange(this.model.getValue(), MonacoEditorOperation.REDO);
+    });
+  }
```
tiagobento: This is a bit controversial. I think we should base the decision on the current OS. What's the default keybinding for `undo` on VS Code for Windows, macOS and Linux? We should follow the same approach, I believe. Even better if we could get the associated key bindings with the `undo`/`redo` commands in VS Code.

tiagobento (changes requested):
Nice PR, @paulovmr!

Left some comments inline, please take a look at them.

tiagobento (approved):

pefernan (approved):
+1

PR #877, opened 2022-03-02
KOGITO-6841: Remove `kie-parent` from the `stunner-editors` package

yesamer:
JIRA: https://issues.redhat.com/browse/KOGITO-6841

This PR removes `kie-parent` as pom parent of the `stunner-editor`.
So, I moved all required dependencies and plugins from `kie-parent` to `stunned-editor` pom.

Note: We still need a dependency with 7.x branch for `kie-soup` dependency only. This will managed and removed in a future PR.


yesamer:
jenkins retest this please

yesamer:
@tiagobento @romartin @jomarko Ready for the review.

tiagobento (approved):

tiagobento:
@yesamer Thanks for this PR! As we discussed earlier, one more PR is to follow this one, completely removing the dependencies with `version.org.kie7`. Great job 🎉

jomarko (commented):
Hello team. From quality perspective, the kie-parent removal means we are loosing definition of full mvn profile:
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/main/pom.xml#L1927

My opinions is that we should declare the full profile in kie-tools. However we should probably refresh/simplify it. For example I do nto think javadoc-plugin should be continued to be part of full profile.

However spotbugs and enforcer seems to be important plugins of full profile. Even current codebase refers to things of full profile:
- <spotbugs.failOnViolation>true</spotbugs.failOnViolation> - stunner-editors/pom.xml
- <id>fullProfile</id> - kie-wb-common-stunner/pom.xml

Furthermore we should probably use full profile as part of CI, once we agreed with its set-up.

One more question, is there plan to somehow concentrate props dependencies management of stunner-editors and dashbuilder? For example there is inconsistency
- stunner-editors: <version.jacoco.plugin>0.8.5</version.jacoco.plugin>
- dashbuilder: refers to non existing property ${version.jacoco.plugin} - CI doesn't fail, as is used just by currently not used run-code-coverage profile

yesamer:
@jomarko Thank you for noticing that. I'll move the full profile too.

jomarko:
@yesamer thank you for answer. It is quite surprising for me. Are you saying we are not interested in any of spotbugs and enforcer checks? If that is a case also all configuration related to them should be removed. Not just profile itself.

yesamer:
@jomarko No, what I meant was: I'm considering to the declare the full profile in kie-tools, as you suggested.
Anyway, the enforcer works with the default profile too, I need to check which are the differences with the full profile. 
Spotbug is missing instead.

jomarko:
@yesamer ok, just ping me once ready to recheck. Thank you.

yesamer:
@jomarko According to the `full-profile` definition:

```
<!-- The execution's configuration is part of the pluginManagement. This piece here only makes sure the
                     execution is enabled (by specifying a phase) for full profile builds. -->
```
                     
So, it is just a way to be sure the enforcer is enabled, probably not necessary because I checked and the enforcer works with the default profile.
The java-doc I agree with you, it seems useless. But it is used somewhere in the default one. 
Spotbug is missing, as I said.
My proposal is to copy it 1:1 for the moment, I think there is space to optimize that pom (and child poms) in the future.

yesamer:
@jomarko `fullProfile` added. Some considerations:
- It is clear that profile was never used in last months, as it reports bugs in one the first analyzed module
`[ERROR] Failed to execute goal com.github.spotbugs:spotbugs-maven-plugin:3.1.8:check (spotbugs-check) on project reflections: failed with 19 bugs and 0 errors  -> [Help 1]`. Therefore, if we choose to use it in CI, we definitely need to plan an effort to resolve all the reported bugs.
- In `kie-parent`, `spotbugs-maven-plugin` depends on `kie-build-tool` --> https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/main/pom.xml#L1680, which is a 7.x assets. I didn't add that dependency, as the scope of this PR series is to remove any 7.x dependency, but this may results in a different behavior

romartin (approved):
Looks good! Thanks @yesamer

jomarko (approved):

PR #886, opened 2022-03-09
KOGITO-6861: Exporting SVGs to projects in different drives on Windows fails

thiagoelg:
# JIRA Issue
https://issues.redhat.com/browse/KOGITO-6861

# In this PR
- Add a forward slash to the `parsedPath.dir` to avoid concatenating the parsedPath to the editor's installation path on Windows when generating the SVG image

### DIFF
--- a/packages/vscode-extension/src/generateSvg.ts
+++ b/packages/vscode-extension/src/generateSvg.ts
@@ -60,10 +60,10 @@ export async function generateSvg(args: {
     return;
   }
 
-  const parsedPath = __path.parse(editor.document.uri.fsPath);
+  const parsedPath = __path.parse(editor.document.uri.path);
   const workspace = vscode.workspace.workspaceFolders?.length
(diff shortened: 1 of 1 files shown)


tomasdavidorg:
Hi @thiagoelg , I was able to export to svg:
image
and the svg is located in the same directory but after I clicked `Open SVG`, I got this error:
```
Unable to open 'aaa-svg.svg': Unable to resolve resource c:%5Caaa%5Caaa-svg.svg.
```
image

(Same error for C: drive or different drive)

thiagoelg:
@tomasdavidorg Nice find!

I think this should be fixed now, I stopped using `fsPath` and changed to `path` which follows unix like file paths patterns, so everything should be working.

I tested on Windows and MacOS, let me know if you find anything else, thanks!

tomasdavidorg (approved):
Thanks @thiagoelg Tested on Fedora and Windows and it looks good to me. 🙂

tiagobento (approved):

PR #903, opened 2022-03-24
KOGITO-6608: Implement header cells keyboard navigation and editing on the Boxed Expression component

fantonangeli:
This PR is to add the keyboard navigation to the header cells.

There are still some points to be done or discussed:

- Popover text field (Name) in the online editor only: esc key is not working with the focus on the text input. This depends on packages/keyboard-shortcuts/src/channel/Hooks.tsx:39 
- Any click inside a PopoverMenu removes the blue border on a cell because it removes the focus from a cell.
- Literal Expression has no keyboard navigation. This is because it doesn't use an HTML Table and needs the rewriting of the component.

I'm not sure if all these points have priority or not. 
Please let me know your opinion and in case I can manage them in a different PR.

https://issues.redhat.com/browse/KOGITO-6608

Peek 2022-03-24 10-01


packages/boxed-expression-component/src/components/Table/Table.tsx
```diff
@@ -376,6 +387,57 @@ export const Table: React.FunctionComponent<TableProps> = ({
...
+        }
+
+        //prevent the parent cell catch this event if there is a nested table
+        if (e.currentTarget !== getParentCell(e.target as HTMLElement)) {
+          return;
+        }
+
+        if (boxedExpression.isContextMenuOpen) {
+          e.preventDefault();
+          if (key === "Escape") {
```
ljmotta: WDYT about creating an enum on `FocusUtils.ts` with all navigation keys?

packages/boxed-expression-component/src/components/Table/common/TableUtils.ts
```diff
@@ -35,18 +45,119 @@ export const getCellCoordinates: (cell: Element | undefined | null) => { x: numb
...
+
+/**
+ * Get the table header's rows henght.
+ *
+ * @param tableInstance the tableInstance
+ * @param skipLastHeaderGroup true to skip last header group
+ * @return the number of rows, 0 otherwise
+ */
+export const getHeaderRowsLenght = (tableInstance: TableInstance, skipLastHeaderGroup: boolean): number => {
+  if (!tableInstance || typeof skipLastHeaderGroup === "undefined" || !tableInstance.headerGroups) {
```
ljmotta: The type of `skipLastHeaderGroup` should't be `skipLastHeaderGroup?: boolean`?

ljmotta (commented):
@fantonangeli Thanks for this PR! Looks very good. I've just made some minor comments inline. Also, a minor issue, this PR broke a functionality of the DMN Runner Table. Deleting a Decision Node makes the output table disappear. I couldn't figure out why, but I've already prepared a fix for it. I'll open a PR to your branch.

packages/boxed-expression-component/it-tests/integration/decision_table_expression_spec.ts
```diff
@@ -317,4 +340,42 @@ describe("Decision Table Keyboard Navigation Tests", () => {
...
+      cy.get("@annotationCell").should("have.text", "annotation-1 edited");
+
+      cy.get("@annotationCell").should("be.focused").get("input").should("not.exist");
+    });
+
+    it("Cancel editing", () => {
+      // from the annotation cell, edit the text then press esc to cancel editing
+      cy.get("@annotationInput").type("not to save text{esc}");
+
+      cy.get("@annotationCell").should("not.contain.text", "not to save text");
```
jomarko: could we please also assert what text is contained?

packages/boxed-expression-component/src/components/Table/Table.tsx
jomarko: not sure if `FocusUtisl.ts` is the best place, but yes, extract this Key comparator logic to some place, that is better unit testable could be useful.

packages/boxed-expression-component/tests/components/Table/common/FocusUtils.test.tsx
```diff
@@ -0,0 +1,548 @@
...
+
+    jest.resetAllMocks();
+  });
+
+  describe("getParentCell tests", () => {
+    it("should fail", () => {
+      // @ts-ignore
+      expect(getParentCell()).toBeNull();
+      expect(getParentCell(null)).toBeNull();
+    });
```
jomarko: could you please all `it("should fail", () => {` tests?

jomarko (commented):
Thank you for the PR.

_Popover text field (Name) in the online editor only: esc key is not working with the focus on the text input. This depends on packages/keyboard-shortcuts/src/channel/Hooks.tsx:39_ - is from my point of view quite important for better user experience. Generally the interaction with _Name and Data Type_ popover with keyboard is limited, I was not able to:
- change values there, always needed also to click in popover
- close the popover after interacted with it (i.e. not closing it immediately after its opening), always needed to use also mouse

Not reporting any jira due to lack of my resources for providing kie-tools reviews.

packages/boxed-expression-component/src/components/Table/Table.tsx
fantonangeli: One more option is to create a small class that relies on regex.test() to check also insensitive case. Eg.
```typescript
class KeyboadKeys {
  static isEscape(key: string) {
    return /escape/i.test(key);
  }
}
```

Regarding the place, `kie-tools/packages/keyboard-shortcuts/src/envelope/DefaultKeyboardShortcutsService.ts:46` has already a `KEY_CODES` object that is missing a few keys like 'tab' and 'enter', but I don't know the implication of modifying `KEY_CODES`.

fantonangeli:
@jomarko 
Created jira "[DMN Designer] interaction with Name and Data Type popover with keyboard"
https://issues.redhat.com/browse/KOGITO-6918

packages/boxed-expression-component/src/components/EditExpressionMenu/EditTextInline.tsx
```diff
@@ -15,50 +15,80 @@
...
         setToggle(true);
+        onToggle(true);
       }
     },
-    []
+    [onKeyDown, onCancel, onToggle]
   );
 
+  useEffect(() => {
+    // Typescript don't accept the conversion between DOM event and React event
```
vpellegrino: have you tried to convert first to `unknown` and then to the desired type?

jomarko (approved):

yesamer (approved):

yesamer:
@fantonangeli LGTM, thank you!

packages/boxed-expression-component/src/components/Table/Table.tsx
ljmotta: I don't see a problem modifying the `KEY_CODES`, but the `DefaultKeyboardShortcutService` is not the right place either. Its purpose is to forward shortcuts to the `envelope`. Taking this into account, I don't think it's a good idea to create a dependency between the `boxed-expression-component` and the `keyboard-shortcuts`. Regarding the class, I guess a `NavigationKeys` enum would suffice, as we don't use other variations.

packages/boxed-expression-component/src/components/Table/EditableCell.tsx
```diff
@@ -186,10 +179,10 @@ export function EditableCell({ value, rowIndex, columnId, onCellUpdate, readOnly
       if (isTab) {
         if (!event.shiftKey) {
           //this setTimeout fixes the focus outside of the table when the suggestions opens
-          setTimeout(() => focusNextDataCell(textarea.current, rowIndex), 0);
+          setTimeout(() => focusNextCell(textarea.current, 1, false), 0);
```
vpellegrino: Wondering if it makes sense to have a function with a boolean parameter (that is always `false` in our case). Why don't have a separate method?

packages/boxed-expression-component/src/components/Table/Table.tsx
```diff
@@ -376,6 +388,62 @@ export const Table: React.FunctionComponent<TableProps> = ({
     [getRowKey]
   );
 
+  /**
+   * Function to be executed when a key has been pressed on a cell
+   * @param rowIndex the index of the row
+   */
+  const onCellKeyDown = useCallback(
```
vpellegrino, line 439: Well, I think that this function is performing too many actions... why don't you modularize it by creating multiple methods to call? In that way, comments would not be needed

packages/boxed-expression-component/src/components/Table/TableHeader.tsx
```diff
@@ -286,3 +289,175 @@ export const TableHeader: React.FunctionComponent<TableHeaderProps> = ({
...
+  onHeaderClick: (columnKey: string) => () => void;
+  onHorizontalResizeStop: (column: Column, columnWidth: number) => void;
+  renderHeaderCellInfo: (
+    column: Column,
+    columnIndex: number,
+    onAnnotationCellToggle?: (isReadMode: boolean) => void
+  ) => React.ReactElement;
+}
+
+const RenderResizableHeaderCell = ({
```
vpellegrino: why not use a separate file for this component?

packages/boxed-expression-component/src/components/Table/common/TableUtils.ts
```diff
@@ -35,8 +45,38 @@ export const getCellCoordinates: (cell: Element | undefined | null) => { x: numb
...
+
+  return DEFAULT;
+};
+
+/**
+ * Fetch cell coordinates, counting all cells from the table header.
+ * Note: supports the colspan.
+ *
+ */
+export const getFullCellCoordinates: (cell: Element | undefined | null) => TableCellCoordinates = (cell) => {
```
vpellegrino: Wondering how many times this function is executed since it iterates on every single cell...
I am a bit worried about performance on large tables. Aren't we able to compute coordinates directly on the template where the cell is declared?

packages/boxed-expression-component/src/components/Table/common/TableUtils.ts
```diff
@@ -50,3 +90,60 @@ export const getCellTableId: (cell: Element | undefined | null) => string = (cel
...
+ * @param cellCoordinates the cell coordinates. set x or y = -1 to get the last element.
+ * @returns the table cell, null otherwise
+ */
+export const getCellByCoordinates = (
+  table: HTMLTableElement,
+  cellCoordinates: TableCellCoordinates
+): HTMLTableCellElement | null => {
+  const { x, y } = cellCoordinates || {};
+
+  if (!table || x === undefined || x < -1 || y === undefined || y < -1 || table.rows.length <= y) {
```
vpellegrino: this condition looks a bit complex to me. Can you simplify it by using intermediate variables (with meaningful names)?

vpellegrino (commented):
@fantonangeli the overall architecture looks good to me. Thanks for this huge feature.
I added a few comments. Please let me know what you think.

packages/boxed-expression-component/src/components/Table/EditableCell.tsx
fantonangeli: In this case, is false because the event is fired from the `Tab` key, but when `focusPrevCell/focusNextCell` is fired from an `ArrowLeft/ArrowRight` key needs to be true (default value).
I didn't separate the logic for Arrows and Tabs because the logic is mostly the same. 
Do you think it's better to separate the Tab/Arrow logic from those functions?

packages/boxed-expression-component/src/components/Table/common/TableUtils.ts
fantonangeli: I can change the logic this way for better performance, but the improvement will affect only the navigation. 
This is because `getFullCellCoordinates` is called only during the keyboard navigation. WDYT?

packages/boxed-expression-component/src/components/Table/EditableCell.tsx
vpellegrino: My opinion is that we can have two separate methods to call (one for `true` and another for `false`) just for having two separate entry points, but internally they can share partially or not the same logic/method.

packages/boxed-expression-component/src/components/Table/common/TableUtils.ts
vpellegrino: Well. Ideally, instead of computing coordinates whenever the keyboard navigation occurs, we should pre-compute all coordinates every time the table changes and save them on the cell itself.
This is only my point of view for having a better user experience and void freezing the browser while navigating a big table.
We can have a trade-off on this and it doesn't necessarily need to be done right now.

fantonangeli:
All comments are closed and from my side, this PR can be merged

yesamer:
@vpellegrino @ljmotta Can you please finalize this PR review? Thx!

vpellegrino (approved):
LGTM, thanks @fantonangeli

ljmotta (approved):
LGTM

PR #908, opened 2022-03-29
KOGITO-6417: [DMN Designer] Undo/redo in Boxed Expression Editor for editable cells is not working properly

danielzhe:
See: https://issues.redhat.com/browse/KOGITO-6417

TL;DR: go to "the solution" at the end of the text.

A quick explanation about how the undo/redo works and talks between TS/JavaScript (where the editor is implemented) and GWT/Java world (where the undo/redo is implemented):
- The undo/redo works by saving the status of the current Expression every time an user action is performed. 
- This notification from TS/JavaScript world to GWT/Java world is made by the method `notifyUserAction()`.
- The method `onCellUpdate` passes the data from TS/JavaScript world to GWT/Java world.

So when user presses CTRL+Z (or click on "Undo"), the saved status is restored and all changes performed after the `notifyUserAction()` is discarded.

**The issue:**
SCENARIO 1: If user double clicks on the cell and types, everything works. No issue in this scenario.
SCENARIO 2: If user starts to type without double-clicking the cell, the first letter is not considered an user action, only the subsequent typed letters. This is the issue reported in the ticket.

**The cause:**
We was calling `notifyUserAction` in the `triggerReadMode`. The `triggerReadMode` reports to GWT world the changes made by calling the `onCellUpdate`, so we was calling the `notifyUserAction` before `onCellUpdate` here. It makes sense.

But after implementation of the feel autocomplete to the new Boxed Expression Editor, a new scenario of communication between TS/Js-GTW/Java world was introduced. The method `onTextAreaChange` used by the feel autocomplete also needs to call `onCellUpdate`. 
The solution could be call `notifyUserAction` in the `onTextAreaChange`, but then we will have a double call to `notifyUserAction`.

But remember: if user starts to edit double-clicking the cell it does not call `onTextAreaChange`. 

**The solution :**
Change the moment that `notifyUserAction` is called.
This PR change it from `triggerReadMode` before update to GWT world to `triggerEditMode` before starts everything!
Also it changes the order of the call in the `onTextAreaChange` to call the `triggerEditMode` before the update of the cell content (`onCellUpdate`).

@yesamer

### DIFF
--- a/packages/boxed-expression-component/showcase/index.tsx
+++ b/packages/boxed-expression-component/showcase/index.tsx
@@ -34,9 +34,60 @@ import {
 } from "../src";
 import { Button, Modal } from "@patternfly/react-core";
 import { PenIcon } from "@patternfly/react-icons";
-import { dataTypes, pmmlParams } from "../tests/components/test-utils";
 import "../src/components/BoxedExpressionEditor/base-no-reset-wrapped.css";
 import ReactJson from "react-json-view";
+
+/**
+ * Constants copied from tests to fix debugger
+ */
+export const dataTypes = [
+  { typeRef: "Undefined", name: "<Undefined>", isCustom: false },
+  { typeRef: "Any", name: "Any", isCustom: false },
+  { typeRef: "Boolean", name: "boolean", isCustom: false },
+  { typeRef: "Context", name: "context", isCustom: false },
+  { typeRef: "Date", name: "date", isCustom: false },
+  { typeRef: "DateTime", name: "date and time", isCustom: false },
+  { typeRef: "DateTimeDuration", name: "days and time duration", isCustom: false },
+  { typeRef: "Number", name: "number", isCustom: false },
+  { typeRef: "String", name: "string", isCustom: false },
+  { typeRef: "Time", name: "time", isCustom: false },
+  { typeRef: "YearsMonthsDuration", name: "years and months duration", isCustom: false },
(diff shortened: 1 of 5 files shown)


yesamer (approved):
@danielzhe LGTM

jomarko (commented):
I can not confirm the fix works. I downloaded vsix form the PR artifacts. I started edit cell without double click, then Undo, but first added letter remained.

https://issues.redhat.com/secure/attachment/12703374/undo.webm

danielzhe:
@jomarko I guess you found a new issue: CTRL+Z without leaving the edit mode, inside the cell while editing. That's another case. I gonna try to handle it in this PR.

danielzhe:
@jomarko Changes applied! @ljmotta provided a fix for the specific case that you found (undo/redo inside the text box without leaving it). The VSIX including his changes and my changes are here: https://drive.google.com/file/d/1qtfeNJ1pW7YQ9VqZm8Kf39QG-SQG8xus/view?usp=sharing

@yesamer fyi ^^^^^^

Also, @ljmotta kindly fixed an issue regarding the debugger! Ty, Luiz!

yesamer (approved):

jomarko (approved):
@danielzhe I didn't found issue during manual check. Thank you for updates.

Observation. PR doesn't contain any test, what increases the risk, it will be broken in future by other PRs.

PR #914, opened 2022-03-31
KOGITO-6859: [DMN Editor] Enabling Import Java Classes feature

yesamer:
JIRA: https://issues.redhat.com/browse/KOGITO-6859

This PR enables a _first_ working version of the "Enabling Import Java Classes" feature.

**Changes introduced in this PR**

- JavaCodeCompletion service: The service API has been fixed. The reason was the `DelegateHandler` `executeCommand` methods returned a `CompletableFuture<?>` instead of a simple `Object` class. With this change applied, the service works without the previously reported exception.
- JavaCodeCompletion service: Updated pom dependencies and removed `gson` which is already imported via `tycho`
- JavaCodeCompletion service: Accessors retrieved in `GetAccessorsHandler` returns Simple name java classes as Type, while an FQCN is required for our scope. (eg. Prior to this PR, returned type was `Book`, instead of the required `com.package.Book`.
- DMN Editor: The "Import Java Classes" button is now shown in the editor, in the Data Type tab.
- ImportJavaClasses react component: Updates to improve the integration with the service, in particular for the Fields service. Some accessors need to be filtered out (eg. `getClass()`, `serialVersionUID`, the name reshaped (eg. remove `get` prefix from getters method`)
- ImportJavaClasses react component: Improved Tooltip management in the search box 
- ImportJavaClasses react component: Tests have been restored

**What is missing (next steps)**
- Lists and Arrays are currently not supported. I will manage it in a future PR.
- The `@KieActivator` annotation requires an improvement, as not a realistic solution in the current status (more details in the `How to Test` section.

**How to test it**
It requires:
- A Generated `vsix` file with the changes introduced here
- [Language Support for Java(TM) by Red Hat vscode plugin](https://marketplace.visualstudio.com/items?itemName=redhat.java)
- A project with Java classes
- A Java class marked with `@KieActivator` annotation. Unfortunatetly, at current status, we need to declare the annotation in our project. As mentioned in `What is missing` section, this is **temporary** and will be improved in a future PR. 
Steps are: 
1. Create a `org.kie.api.project` package in your project 
2. Create a `KieActivator` class under that package
3. Copy and paste this class declaration on it:
```
package org.kie.api.project;

import java.lang.annotation.Retention;
import java.lang.annotation.Target;

import javax.inject.Qualifier;

import static java.lang.annotation.ElementType.TYPE;
import static java.lang.annotation.RetentionPolicy.RUNTIME;

@Qualifier
@Target({ TYPE })
@Retention(RUNTIME)
public @interface KieActivator {

}
```
As a result, you should see this:

Screenshot from 2022-04-08 14-32-11

4. Add the `@KieActivator` annotation in one Java Class (eg. here I created a `Test.java` marked with the annotation)
Screenshot from 2022-04-08 14-37-35

5. Now you are ready to use the feature. As shown in the below video

Screencast-from-04-08-2022-03_40_12-PM


adrielparedes (approved):
Java part is ok, lets wait for @caponetto or @tiagobento to review the TS part, approved!

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClassesWizardSecondStep.tsx
```diff
@@ -49,19 +71,31 @@ export const ImportJavaClassesWizardSecondStep = ({
     return new JavaField(name, type, dmnTypeRef);
   }, []);
 
+  const isInDenyList = useCallback((accessorName: string) => {
+    console.log(accessorName);
```
tiagobento:
```suggestion
```

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClassesWizardSecondStep.tsx
```diff
@@ -49,19 +71,31 @@ export const ImportJavaClassesWizardSecondStep = ({
...
+    const denyList = ["getClass()", "serialVersionUID"];
```
tiagobento: Maybe this can be a constant?

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClassesWizardSecondStep.tsx
```diff
@@ -41,6 +41,28 @@ export const ImportJavaClassesWizardSecondStep = ({
...
+  const renameAccessorName = useCallback(
+    (originalName: string) => {
+      if (isGetterMethod(originalName)) {
+        const name = originalName.substring(originalName.indexOf(getPrefix) + 3, originalName.lastIndexOf("("));
+        return name.charAt(0).toLowerCase() + name.slice(1);
+      } else {
+        return originalName;
+      }
+    },
+    [isGetterMethod]
```
tiagobento: How does it work for boolean getters? The conversion is that they start with `is`, right? I think we could take that into account as well?

packages/import-java-classes-component/src/components/ImportJavaClasses/model/JavaClassUtils.ts
```diff
@@ -14,6 +14,14 @@
...
+ * It removes the class package and the generic type (if present)
+ * @param javaClassName
+ */
 export const getJavaClassSimpleName = (javaClassName: string) => {
-  return javaClassName.split(".").pop()!;
+  let javaClassSimpleName = javaClassName;
+  if (javaClassName.includes("<")) {
+    javaClassSimpleName = javaClassSimpleName.split("<")[0]!;
+  }
+  return javaClassSimpleName.split(".").pop()!;
```
tiagobento, line 27: Note that this won't work for static inner classes, where the `.` is part of the class simple name. Don't need to solve that, just add a comment maybe?

packages/vscode-java-code-completion-extension-plugin/vscode-java-code-completion-extension-plugin-core/src/main/java/org/kogito/core/internal/handlers/GetAccessorsHandler.java
```diff
@@ -80,7 +81,14 @@ protected GetPublicResult getAccessor(CompletionItem item, String fqcn) {
...
             String[] label = item.getLabel().split(":");
             result.setAccessor(label[0].trim());
-            result.setType(label[1].trim());
+            String type = label[1].trim();
+            Map<String,String> data = (Map<String, String>) item.getData();
+            String fqcnType = data.get(DATA_FIELD_SIGNATURE);
+            if (fqcnType != null && fqcnType.contains(")L")) {
+                type = fqcnType.split("\\)L")[1];
+                type = type.replaceAll(";$", "");
+            }
```
tiagobento, line 91: This is for arrays and other primitives, right? I think we could add a comment here as well for us to know what this code is about.

tiagobento (approved):
@yesamer Yay!! Glad to see this is working now. Left some minor comments inline. Don't need to solve all fo them, just pointing out things I think can be reported as "bugs" in the future.

Anyway, great great job @yesamer!

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClassesWizardSecondStep.tsx
yesamer: Yes, we can easily add `is` to manage boolean getters as well, good point.
My concern about this is: `get / is` prefixes are not enough to define a getter method.
These prefixes are often used for NOT-getters methods, so this logic could produce wrong fields in some cases.
Unfortunately, I can't retrieve other information to improve this heuristic from the LSP service, thus this limitation will remain in the functionality.

packages/import-java-classes-component/src/components/ImportJavaClasses/model/JavaClassUtils.ts
yesamer: I can add a comment, sure. To solve the case you mentioned, I think we need to retrieve the class package.

packages/vscode-java-code-completion-extension-plugin/vscode-java-code-completion-extension-plugin-core/src/main/java/org/kogito/core/internal/handlers/GetAccessorsHandler.java
yesamer: No, that is to retrieve the Full Qualified Class name of the getter method. I'll put a comment to clarify

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClassesWizardSecondStep.tsx
yesamer: Done!

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClassesWizardSecondStep.tsx
yesamer: Done!

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClassesWizardSecondStep.tsx
yesamer: `is` prefix management added.

packages/import-java-classes-component/src/components/ImportJavaClasses/model/JavaClassUtils.ts
yesamer: Comment added.

packages/vscode-java-code-completion-extension-plugin/vscode-java-code-completion-extension-plugin-core/src/main/java/org/kogito/core/internal/handlers/GetAccessorsHandler.java
yesamer: Comment added. In Unit Tests additional examples are avaialble.

yesamer:
@domhanak Hello, did you plan to test this?

PR #915, opened 2022-04-04
KOGITO-6915: [DMN Editor] add visual regression tests to cypress

fantonangeli:
TO BE MERGED AFTER: https://github.com/kiegroup/kie-tools/pull/903

Closes: https://issues.redhat.com/browse/KOGITO-6915

Command used to install 'cypress-image-snapshot ':
```
npx lerna add cypress-image-snapshot --scope=@kie-tools/boxed-expression-component --dev
````

### CODE REVIEW

**Review by jomarko — COMMENTED** [review]
Thank you for a PR @fantonangeli 
- Regression tests data cell focus.snap.png - I would use different name. it is data cell, but still, it is somehow special, user allowed to edit in in very limited way, there is even sync mechanism, I would name it rather like context entry leading cell, context entry key cell, what do you think?
- Literal Expression Tests -- Regression tests.snap.png - no focus there, is it expected?

Postponing more detailed review once the linked PR is merged and this is converted from draft to regular PR

**Comment by fantonangeli** [comment]
Thank you @jomarko for your review. 
Regarding the filenames I just let Cypress choose the filename automatically according to the test name. This way we don't need to see the code to see which file is connected to a failing test. But I can configure the filename in the options of `cy.matchImageSnapshot();` call.

Regarding the Literal Expressions I didn't set a focus because, even if they have the focus working with the CSS, they are not rendered with an HTML Table and they have no keyboard navigation at the moment.

**Comment by jomarko** [comment]
@yesamer @fantonangeli it seems CI is stuck.

**Comment by fantonangeli** [comment]
@jomarko yes the CI doesn't pass, but it works well locally. I also tried to increase the threshold of image diff, but it still fail in CI.
I think I will have to evaluate a different Cypress library.

**Comment by jomarko** [comment]
@fantonangeli maybe the screenshots needs to be generated on exactly the same machine, where the test run? I am using fedora, What is your local set-up? I can try to run the tests locally on my machine if we have different machines.

**Comment by fantonangeli** [comment]
Thank you @jomarko for your great idea! I used another laptop with Arch to understand more about the differences between the snapshots generated with the RHEL and tried the configuration. In fact, they were just slightly different.
Now it should pass in the CI as well.

File: packages/boxed-expression-component/it-tests/support/commands.ts [hunk 1]
```diff
@@ -39,6 +39,17 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+import { addMatchImageSnapshotCommand } from "cypress-image-snapshot/command";
+
+addMatchImageSnapshotCommand({
+  customSnapshotsDir: "../it-tests/cypress/snapshots",
+  blackout: [".updated-json"],
```
jomarko, line 47: @fantonangeli just to understand, is the trick to ignore the "textarea" with json?

fantonangeli: Yes @jomarko, this way I hide the JSON drawing a black box over it. 
This is necessary because the JSON is always different and also that part is not relevant for the tests.

jomarko (approved):
@fantonangeli thank you for this contribution!

tiagobento (approved):
@fantonangeli Great addition! Nice example on how to test complex UIs. Can be useful for Stunner-based editors testing as well. cc @jomarko @domhanak @tomasdavidorg @barboras7 @Kbowers-redhat

yesamer:
Thank you @fantonangeli, excellent contribution.

yesamer (approved):

PR #916, opened 2022-04-04
KOGITO-6932: DMN editor: text annotation is not saved correctly when creating it by copy and paste

danielzhe:
Ticket: https://issues.redhat.com/browse/KOGITO-6932

**Cause of the issue and fix:**
The issue was caused because the ID of the cloned `TextAnnotation` was not being updated.
The `DMNDeepCloneProcess` receive two parameters: the `source` and the `target`. 
It calls the super method that clones everything in the first level (even references to the same object, so a inner object in `source` and `target` would be the same). 

The `DMNDeepCloneProcess` then resolves those cloned references to a news instances, clone values that super method can't clone, and **sets a new ID for the `target`**.

A new ID for the cloned item should be set in `cloneDRGElementBasicInfo`.

But `TextAnnotation` it is an exception and it is not a DRGElement, so that's why the new ID was not being set.

### DIFF
--- a/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
+++ b/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
@@ -29,6 +29,7 @@ import com.google.gwt.regexp.shared.RegExp;
 import org.kie.workbench.common.dmn.api.definition.HasText;
 import org.kie.workbench.common.dmn.api.definition.HasVariable;
 import org.kie.workbench.common.dmn.api.definition.model.BusinessKnowledgeModel;
+import org.kie.workbench.common.dmn.api.definition.model.DMNElement;
 import org.kie.workbench.common.dmn.api.definition.model.DRGElement;
 import org.kie.workbench.common.dmn.api.definition.model.Decision;
 import org.kie.workbench.common.dmn.api.definition.model.Expression;
@@ -99,6 +100,10 @@ public class DMNDeepCloneProcess extends DeepCloneProcess implements IDeepCloneP
             cloneDRGElementBasicInfo((DRGElement) source, (DRGElement) target);
         }
 
+        if (source instanceof DMNElement) {
+            cloneDMNElement((DMNElement) source, (DMNElement) target);
+        }
+
         if (source instanceof HasText) {
             cloneTextElementBasicInfo((HasText) source, (HasText) target);
         }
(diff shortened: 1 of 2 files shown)


packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
```diff
@@ -99,6 +100,10 @@ public DMNDeepCloneProcess(final FactoryManager factoryManager,
             cloneDRGElementBasicInfo((DRGElement) source, (DRGElement) target);
         }
 
+        if (source instanceof DMNElement) {
+            cloneDMNElement((DMNElement) source, (DMNElement) target);
+        }
+
```
jomarko, line 106: We have this logic for cloning now:
```java
        super.clone(source, target);

        if (source instanceof DRGElement) {
            cloneDRGElementBasicInfo((DRGElement) source, (DRGElement) target);
        }

        if (source instanceof DMNElement) {
            cloneDMNElement((DMNElement) source, (DMNElement) target);
        }

        if (source instanceof HasText) {
            cloneTextElementBasicInfo((HasText) source, (HasText) target);
        }

        if (source instanceof HasVariable) {
            final IsInformationItem sourceVariable = ((HasVariable) source).getVariable();
            final IsInformationItem targetVariable = ((HasVariable) target).getVariable();
            cloneTypeRefInfo(sourceVariable, targetVariable);
        }

        if (source instanceof Decision) {
            cloneDecision((Decision) source, (Decision) target);
        }

        if (source instanceof BusinessKnowledgeModel) {
            cloneBusinessKnowledgeModel((BusinessKnowledgeModel) source, (BusinessKnowledgeModel) target);
        }

        return target;
```

However I do not see a place, where the TextAnnotation `textFormat` is cloned.

jomarko (commented):
Thank you for the PR @danielzhe . The build failed. I didn't check manually. One comment to code.

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
danielzhe: Sorry, I don't understand your question. The only change this PR do is that now we set a new ID to `DMNElement` which includes `TextAnnotation`.

jomarko (approved):

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/commands/clone/DMNDeepCloneProcess.java
jomarko: ok

yesamer (approved):
@danielzhe LGTM, thank you. Can you please check if this issue reproduces in BC too and, in case, backport it?

PR #917, opened 2022-04-04
KOGITO-6928: Expression in Assignments that contains comma is saved with incorrect value when reopening a process

Josephblt:
Fixed problem with deserialization. The comma was used inside the expression and as a separator for the full assignments string. That caused the problem.

JIRA: [KOGITO-6928](https://issues.redhat.com/browse/KOGITO-6928)

VS Code: [VSCode Extension](https://drive.google.com/file/d/1Y2KhRHxfJeHdMQXg9d1HBM_g0bjL4y8u/view?usp=sharing)

### DIFF
--- a/packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/model/AssignmentData.java
+++ b/packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/model/AssignmentData.java
@@ -245,7 +245,7 @@ public class AssignmentData {
     public void setAssignments(final String sAssignments) {
         assignments.clear();
         if (sAssignments != null && !sAssignments.isEmpty()) {
-            String[] as = sAssignments.split(",");
+            String[] as = sAssignments.split(",(?=\\[din])|,(?=\\[dout])");
             for (String a : as) {
                 if (!a.isEmpty()) {
                     Assignment ass = Assignment.deserialize(this,
--- a/packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/test/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/model/AssignmentDataTest.java
+++ b/packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/test/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/model/AssignmentDataTest.java
@@ -16,6 +16,8 @@
 
 package org.kie.workbench.common.stunner.bpmn.client.forms.fields.model;
 
+import java.util.List;
+
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


Josephblt:
Hello @romartin, @domhanak. This is ready to test. Keep in mind that this is a blocker issue. Thanks.

packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/main/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/model/AssignmentData.java
```diff
@@ -245,7 +245,7 @@ public String getAssignmentsString() {
     public void setAssignments(final String sAssignments) {
         assignments.clear();
         if (sAssignments != null && !sAssignments.isEmpty()) {
-            String[] as = sAssignments.split(",");
+            String[] as = sAssignments.split(",(?=\\[din])|,(?=\\[dout])");
```
romartin, line 248: @Josephblt I am not 100% sure if that's correct, because as far as I remember since we support generics in the assignment type input box, this is necessary in order to separate the comma for generics, eg: when using a `Map<K, V>`.

@inodeman can you please review this? As far as I remember it was you who implemented this regexp, is that correct? Does this change break the generics thing?

Thanks guys!

Josephblt: Well, this regex is used to split assignments only. In this part of the unmarshalling, they are already processed and they come bundled together in a single string. The problem is that expressions like the one you mentioned will be identified as a separator. This look-ahead regex will mitigate this misidentification.

packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-client/src/test/java/org/kie/workbench/common/stunner/bpmn/client/forms/fields/model/AssignmentDataTest.java
```diff
@@ -32,11 +34,34 @@ public void setUp() throws Exception {
...
+
     @After
     public void tearDown() {
         super.tearDown();
     }
 
+    @Test
+    public void testSetAssignments() {
+        AssignmentData assignmentData = new AssignmentData();
+        String assignment1 = "function.getValue(par1,par1)";
```
domhanak, line 47: @Josephblt thinking about value like `function.getValue(par1,par2,par3)` OR `function.getValue(par1,par2.Cpar, par3)`

Could we please expand the values tested here?

Josephblt: Yes, I can. No problem. But those test cases don't add too much. The problem here was we were misidentifying commas (,) as assignment separators. So just adding any test with a comma will do. Unless we have an assignment with an expression containing ",[din]" or ",[dout]" we won't have any issues.

inodeman:
Hi @romartin I think since this fixes an already existing deserealization problem, there should be no problem.

romartin (approved):
Ok thanks for the updates guys! Approving, looks good to me!

domhanak (approved):
Firstly, the fact that we represent all assignments as one string split by `,` is not ideal and may lead to further issues if for some reason the expression contains `[din]` `[dout]` allthough unlikely, it still leaves big chunk of inputs that can lead to bugs in editor behavior.
I suggest improving the code to use class for an Assignment and then processing `List<Assignment> assignments` here. 
Please consider, not mandatory for merging.

For testing.
I was able to verify that the behavior is now correct and users process is not malformed
when they use expressions ins `Source` of an assignments.
Here is a screenshot of the Assignments modal used for testing. Let me know if you find something missing.
Screenshot from 2022-04-08 09-20-40

Editor is working with assignments as expected after this fix so I am approving.
As this is a blocker, I will be creating a regression IT test in VSCode as part of https://issues.redhat.com/browse/KOGITO-6965.

Thank @Josephblt  for the fix and quick responses. I'll pinged you to discuss the
string problem offline.

PR #929, opened 2022-04-08
KOGITO-6918: [DMN Editor] interaction with name and data type popover with keyboard

fantonangeli:
Fixes: https://issues.redhat.com/browse/KOGITO-6918

Peek 2022-04-07 16-51


jomarko (approved):
Thank you for big amount of tests @fantonangeli

vpellegrino (approved):
Looks good to me, thanks @fantonangeli

packages/boxed-expression-component/src/components/PopoverMenu/PopoverMenu.tsx
```diff
@@ -64,19 +72,24 @@ export const PopoverMenu: React.FunctionComponent<PopoverMenuProps> = ({
   className,
   hasAutoWidth,
   minWidth,
+  isVisible = false,
   onHide = () => {},
   onCancel = () => {},
+  onShown = () => {},
 }: PopoverMenuProps) => {
+  const triggerManually = isVisible !== null;
```
ljmotta: `isVisible` have a default value and it's a boolean, this will not always be true?

packages/boxed-expression-component/src/components/PopoverMenu/PopoverMenu.tsx
```diff
@@ -64,19 +72,24 @@ export const PopoverMenu: React.FunctionComponent<PopoverMenuProps> = ({
...
   const { setIsContextMenuOpen } = useBoxedExpression();
-  const [isVisible] = useState(false);
+  const [isPopoverVisible, setIsPopoverVisible] = useState(false);
```
ljmotta: If `triggerManually` is `true` then we're going to have two states controlling the same thing (the `isVisible` on the parent component and the `isPopoverVisible` on PopoverMenu). This will cause a render delay. This happens because `isVisible` needs to be updated to trigger the `useEffect` and update the `isPopoverVisible`. To fix this issue, you can make a forwarding ref and expose `isPopoverVisible` and `setIsPopoverVisible` through the ref, using the `useImperativeHandle` hook.

ljmotta (changes requested):
Hi @fantonangeli , thanks for this PR! I've made some comments inline.

packages/boxed-expression-component/src/components/PopoverMenu/PopoverMenu.tsx
fantonangeli: Thank you for your interesting comments @ljmotta , I will try to follow your suggestion and update this PR next week.

fantonangeli: Hi @ljmotta, I updated the code to use `useImperativeHandle` and checked with a counter in the `PopoverMenu` component to see how many times is re-rendered, but I don't see a difference for the first rendering. 
I only see the difference when the user closes the popover and it is saving only one rendering.

ljmotta (approved):
LGTM!

yesamer (approved):

PR #938, opened 2022-04-11
KOGITO-6972: Add zoom control buttons to the mermaid generated SVG

paulovmr:


tiagobento (approved):

tomasdavidorg (commented):
Hi @paulovmr 
One thing I noticed: The Generated SVG (the file after save) also contains the controls. Should it be there? (Also it looks like the SVG generation is currently broken - texts are not visible). Otherwise it looks good to me.
Screenshot from 2022-04-20 09-27-16

paulovmr:
@tomasdavidorg Thanks for the review, I've removed the zoom controls from the exported SVG now. As for the SVG being broken, it uses `foreignObject`s, so you have to open it in a browser to see it correctly.

tomasdavidorg (approved):
Thanks @paulovmr. Looks good to me 🙂

PR #939, opened 2022-04-11
KOGITO-6958: DashBuilder YAML Editor

jesuino:
Introduces a new editor for Dashbuilder in YML format. It uses monaco editor and is based on the existing serverless editor.


tiagobento:
@jesuino Hi William! Looks like there's some linting errors happening on the new package. Can you please take a look?

jesuino:
Hello @barboras7 

To run the editor first build in root directory the following command:

```
yarn build:prod:until @kie-tools/dashbuilder-editor
```

It will take almost one hour to run (if running for the first time). then you can start the editor:

```
cd packages/dashbuilder-editor
yarn start
```
Then access the editor at: `http://localhost:9007`

You can try any YAML from our guide and it should work: https://www.dashbuilder.org/docs/#chap-dashbuilder-yaml-guides

barboras7:
Hi @jesuino, 
thanks for the information. My testing was actually in progress already, I am sharing my findings below (there might be some more, I still do have some scenarios in my head that I want to try :slightly_smiling_face: ). 

Also, I am aware that some of the issues I am sharing below are not directly caused by this PR but I see this as a good place to start the discussion. I can open a Jira for them if we agree on that.

1. There's a typo in the imports description, see attached screenshot (dashbuilderś instead of dashbuilders)
Screenshot from 2022-04-25 14-26-01
2. The editor width is set to default every time the user sets preview to true and then back to false (as a user, I want to have the editor width I set during my entire interaction with the editor)
3. If a user uploads a yaml file, it is displayed in the preview but the raw content is not available in the editor (which means that the user can not edit the yaml file to their needs)
Screenshot from 2022-04-25 14-21-00
4. If a user is working with a large yaml file, it is difficult to find the issue with the content of the yaml file (it would be useful to add the number of the first incorrect row into the notification)
Screenshot from 2022-04-25 14-15-53
5. I prepared some HTML code in business central and tested it in Dashbuilder YAML editor. The code was not read properly as it is probably not supported. Should all HTML options from business central be supported in the YAML editor? 
Screenshot from 2022-04-25 14-16-26

Please, let me know what you think :slightly_smiling_face:

jesuino:
Hello @barboras7 ,

Thanks for the feedback. Responses inline:
>> There's a typo in the imports description, see attached screenshot (dashbuilderś instead of dashbuilders)

Thanks! We fixed the typo and started our work to have modes on dashbuilder runtime, where if Dashbuilder is in mode "editor" so it should not allow upload. We also added a default content so users can start their own dashboard from there:

https://drive.google.com/file/d/1zsQNJgo2n2146vgbjN2dM0weWQa9lMTM/view

>> The editor width is set to default every time the user sets preview to true and then back to false (as a user, I want to have the editor width I set during my entire interaction with the editor)

I completely agree with you, but this is tough because the resizable panel component does not hold the width information in the same format that it uses to set the initial width. Hence I don't think it is possible ATM. 


>> If a user uploads a yaml file, it is displayed in the preview but the raw content is not available in the editor (which means that the user can not edit the yaml file to their needs)

Uploaded should be disabled and we did it in my latest commit.

>> If a user is working with a large yaml file, it is difficult to find the issue with the content of the yaml file (it would be useful to add the number of the first incorrect row into the notification)


>> I prepared some HTML code in business central and tested it in Dashbuilder YAML editor. The code was not read properly as it is probably not supported. Should all HTML options from business central be supported in the YAML editor?

The editor should soon become not compatible with the same editor from BC. Regarding the HTML it does not work because some parts are lost during the YAML to JSON conversion and some reserved characters are not converted correctly. A way to workaround this is by using multiple line in YAML:

image

Here's the code I tried:
```
pages:
  - components:
      - html: >-
          <p style="color:red; font-size: xx-large">Hello World</p>
```

Please let me know if you have any question!

packages/dashbuilder/dashbuilder-runtime-parent/dashbuilder-runtime-client/src/main/resources/org/dashbuilder/client/resources/i18n/AppConstants.properties
```diff
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-EmptyScreenView.title=Dashboards are not available
-EmptyScreenView.subTitle=Dashboards were not imported. You can import by uploading a supported file or by posting content using dashbuilder\u015B editor APIs.
+EmptyScreenView.title=No content to show
+EmptyScreenView.subTitle=Dashboards were not imported. You can import by uploading a supported file or by posting content using dashbuilder's editor APIs.
 UploadWidgetView.importButton=Import Dashboards
 RuntimeScreenView.title=Welcome to Dashbuilder Runtimes
```
barboras7, line 19: I see in this file many properties that are already outdated. IMHO, it would be good to get rid of these (and probably also some code) as they are no longer useful. It doesn't necessarily have to be a part of this PR, I can create a Jira for that but it seems like a good opportunity to avoid having lots of unused code :slightly_smiling_face:

barboras7:
Hi @jesuino,

thank you for your responses. I am adding one suggestion into the PR review, not necessarily needed to be fixed now. 

Also, you forgot to answer this concern from the first review:
`If a user is working with a large yaml file, it is difficult to find the issue with the content of the yaml file (it would be useful to add the number of the first incorrect row into the notification)
`
Would this be possible? Now, or in the future? 

Thank you,
Barbora

tiagobento (approved):

jesuino:
Hello @barboras7 

Sorry, I missed the point regarding the error message.

I sent a change which I was working on which was a new screen for error: 

https://user-images.githubusercontent.com/359820/165824076-f0661b01-a8ab-44c3-b2cb-823cc142c1a4.mp4

Now, the error messed could be messed sometimes:

image

this is something that will be solved when we (hopefully soon) solve with schema validation. See how it works with SWF editor.

image


Regarding not used properties, yes, we should focus on it. I opened a JIRA for leftovers[1], but still didn't have the time to focus on it.

Thanks!

[1] https://issues.redhat.com/browse/KOGITO-6882

barboras7 (approved):
Looks good, thanks :tada:

PR #943, opened 2022-04-13
KOGITO-6974: SWF Diagram Editor - Line control points are enabled in view only mode, but they should not

inodeman:
Hi @romartin changes as requested, can you review the PR? Thanks

### DIFF
--- a/packages/serverless-workflow-diagram-editor/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/ControlPointControlImpl.java
+++ b/packages/serverless-workflow-diagram-editor/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/ControlPointControlImpl.java
@@ -40,6 +40,7 @@ import org.kie.workbench.common.stunner.core.client.command.CanvasCommand;
 import org.kie.workbench.common.stunner.core.client.command.CanvasCommandFactory;
 import org.kie.workbench.common.stunner.core.client.command.CanvasCommandManager;
 import org.kie.workbench.common.stunner.core.client.command.CanvasViolation;
+import org.kie.workbench.common.stunner.core.client.shape.Shape;
 import org.kie.workbench.common.stunner.core.command.CommandResult;
 import org.kie.workbench.common.stunner.core.command.util.CommandUtils;
 import org.kie.workbench.common.stunner.core.graph.Edge;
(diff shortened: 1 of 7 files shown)


domhanak:
Just thinking out loud, could we reuse the read-only mechanism from bpmn/dmn/scesim?

inodeman:
Hi @domhanak yes, that is the idea, after this gets merged, we will apply it to stunner on bpmn/dmn editors

romartin:
Please @inodeman let's make sure that the lienzo-webapp (showcase) is behaving as before with the line CPs. Once the user moves on top of some connector, CPs should still be displayed as before.

domhanak:
@inodeman @romartin Hello, I found one small issue that is probably related. The control points are off but I can still move the lines using area selection. Please check the video. Let me know if we are fixing it here.

https://user-images.githubusercontent.com/5937251/165701848-d9cee003-0642-4089-b6cf-46e249026ee1.mp4

domhanak (changes requested):
Pending question that needs to be resolved.

romartin:
oh good catch @domhanak , let's see if we can also fix during this PR, thanks!

romartin:
@inodeman @domhanak 

This issue is present due to the support for multiple selection, as the user is able create a "selection area" and select multiple lelents. Sooo considering actual milestone, which is readOnly (viewer) mode, there is no need at all for multiple selection. We will just support single selection, so once clicking on a single state. 

This way I suggest to just disable the selection multiple for now, and enable just "single" selection. This way this issue will go (for now), and then, let's report it for fixing in the next milestone, once edit is enabled. I think it's a pretty balanced solution

Soooo @inodeman can you please try to disable multiple selection for the viewer, and use only single selection approach? I have been quickly looking and I think ti's just one line change --> Instead of using `MultipleSelection` in the `DefaultViewerSession#init` (see [1] ) , change it by `org.kie.workbench.common.stunner.core.client.canvas.controls.select.SingleSelection` . This way, the viewer will inject the SingleSelection component instead of multiple ,so this way this bug should be gone for now.

Please @inodeman give it a try and let me know if that works
In this case, **I'll create a ticket for existing issue** that @domhanak reported, by postponed for next milestone.

BTW Also @inodeman this PR needs merge/rebase please, can you update it so all checks are back to green? Thanks

Thanks guys!

[1] https://github.com/kiegroup/kie-tools/blob/main/packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/session/impl/DefaultViewerSession.java#L62

inodeman:
Hi @romartin your proposed solution worked out great, ready for review, thanks!

romartin:
Hey @domhanak 

Disabled the multiple selection, this way the issue does not apply. Actually we don't have plans to support multiple seleciton, because we are implementing a selection binding mechanism but only works when a single node selected.

Soo ready to test again!
Thanks!

romartin (approved):

domhanak (approved):

PR #953, opened 2022-04-27
Update test dependencies in `stunner-editors`

yesamer:


romartin (approved):
Thanks @yesamer . Usually I am aware when upgrading `mockito` because our lienzo testing framework relies on it, but if checks are passing means tests are passing as well. So looks good, approving

yesamer:
@romartin Consider `3.x` is the old branch. Now they are releasing `4.x` version. Here they removed a lot of deprecated API. At this moment I updated it to the last `3.x` version, in the future we should evaluate updating to `4.x`, but this could require more effort. Thx!

PR #955, opened 2022-04-27
Make InsecureSkipVerify default to false

thiagoelg:
Solves https://github.com/kiegroup/kie-tools/security/code-scanning/1 and https://issues.redhat.com/browse/KOGITO-5758

"The field InsecureSkipVerify controls whether a TLS client verifies the server's certificate chain and host name. If set to true, the client will accept any certificate and any host name in that certificate, making it susceptible to man-in-the-middle attacks."

Making it default to false to avoid this problem, but adding a config to make it viable for cases like pointed here: https://github.com/kiegroup/kogito-tooling-go/pull/21/files#r694221801

Also, a new systray option was added to allow/disallow this behavior during runtime.


tiagobento:
@caponetto Can you please take a look at this one?

caponetto (approved):
I've tested both a local version and a deployed one. LGTM!
Thanks, @thiagoelg!

PR #956, opened 2022-04-27
KOGITO-7067: [DMN Designer] DMN Editor: Slowness while opening large DMN file

danielzhe:
Ticket: https://issues.redhat.com/browse/KOGITO-7067

**The issue:**

Running a profiling tool, I found out that the event RefreshDecisionComponents is fired [here](https://github.com/kiegroup/kie-tools/blob/main/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/graph/DMNElementsSynchronizer.java#L81) for every loaded node.

This event is handled [here](https://github.com/kiegroup/kie-tools/blob/4466b9cd1ea5eb9869fbbba3270e2e3b805cbc7c/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/docks/navigator/DecisionNavigatorPresenter.java#L112).

It means that for every node inserted on the canvas, during the loading process, the [refreshComponentsView()](https://github.com/kiegroup/kie-tools/blob/4466b9cd1ea5eb9869fbbba3270e2e3b805cbc7c/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/docks/navigator/DecisionNavigatorPresenter.java#L148)is performed.

The Decision Components panel contains all the nodes in the canvas or imported from another model, so they can be duplicated.

This event is used to refresh the Decision Components with the changes every time a node is update.

**The fix:**
During the loading process of the model, we don't need to handle the RefreshDecisionComponents for each node. We only need a single refresh at the end of the loading process.
This PR suspend the refresh during the loading process and resumes it after the loading is done.

**The result:**
Doing that change I reduced the load of a large DMN file from 3:53 (3 minutes and 53 seconds) to 0:33 (33 seconds) in my local machine.

### DIFF
--- a/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/docks/navigator/DecisionNavigatorPresenter.java
+++ b/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/docks/navigator/DecisionNavigatorPresenter.java
@@ -63,6 +63,8 @@ public class DecisionNavigatorPresenter extends AbstractActivity {
 
     private DMNDiagramsSession dmnDiagramsSession;
 
+    private boolean isRefreshComponentsViewSuspended;
+
     double latestDeferred = 0;
 
     @Inject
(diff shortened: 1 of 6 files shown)


yesamer (approved):
@danielzhe Great work! In my local, it took 12.68 seconds to complete. Awesome :1st_place_medal:

jomarko (approved):
@danielzhe perfect work!

Seems as worth of backport into kie-wb-common.

PR #958, opened 2022-04-27
Dependabot alert #165 and #166: Upgrade async package

thiagoelg:
Solves https://github.com/kiegroup/kie-tools/security/dependabot/165 and https://github.com/kiegroup/kie-tools/security/dependabot/166


domhanak:
@thiagoelg Hi there I do not see any package.json changes, this won't solve it in a long run.

tiagobento:
@domhanak We don't need to change the `package.json`, as versions are locked by `yarn.lock`. The changes made here were bumping patch releases, which is aligned with what is defined in our `package.json`. Remember that our dependencies do not "automatically update" even though we're using `^`-prefixed versions. That's the whole purpose of `yarn.lock.`

tiagobento (approved):

PR #960, opened 2022-04-29
Fix for `windows-ci-workflows`

yesamer:
It seems that using `windows-lastest` as OS build results in build instability.
Recently, `windows-latest` was updated to `window-2022` os.
Forcing the previous one, `windows-2019` should remove the instability

References: https://github.com/hashicorp/terraform-cdk/issues/1564

### DIFF
--- a/.github/workflows/monorepo_pr_ci.yml
+++ b/.github/workflows/monorepo_pr_ci.yml
@@ -14,11 +14,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


tiagobento:
@yesamer Just remembered that we have `windows-latest` in other places. Can you please take a look at the other jobs? I think for the Staging and Release jobs we still have `windows-latest`.

yesamer:
@tiagobento It seems `windows-lastest` works now, closing as no longer necessary

PR #964, opened 2022-05-02
KOGITO-4172: [DMN Designer] New Boxed Expression editor - User must be able to create new lines by using shortcuts

fantonangeli:
Fixes: https://issues.redhat.com/browse/KOGITO-4172

Pressing CTRL+ENTER in edit mode on the focused cell creates a new line. Pressing ENTER move the focus on the cell below.

Notes:
- kie-tools/packages/feel-input-component/src/Monaco/FeelEditorService.ts line 84: I didn't find a better way to check if the `Suggest Widget` is open. This check is needed to allow the selection of the suggestion when user press enter
- kie-tools/packages/boxed-expression-component/src/components/Table/EditableCell.tsx: Because I could not trigger the `onKeyDown` in the unit test, I only tested the new behaviour in the e2e tests.

Any suggestion to fix these 2 points in a better way is welcome.

kogito-4172


jomarko (commented):
@fantonangeli thank oyu for the PR an tests. There are two observations from my site.
- The CTRL+ENTER support doesn't work in DMN Runner (the table mode)
- Copy and paste of area, where a cell contains multiline doesn't work correctly [1]. In the first picture I selected two input column cells, copied them in second picture I tried to paste the copied data into output column.

Please let me know if some of ^ is addressable  as part of this PR, or if I should report a jira for future. If the severity of ^ is low form your point of view. I will not report a jira.

[1]
Screenshot from 2022-05-03 09-35-04

Screenshot from 2022-05-03 09-35-11

packages/feel-input-component/src/FeelInput/FeelInput.tsx
```diff
@@ -75,6 +79,15 @@ export const FeelInput = React.forwardRef<FeelInputRef, FeelInputProps>(
...
       setMonacoValue: (newValue: string) => FeelEditorService.getStandaloneEditor()?.setValue(newValue),
+      insertNewLineToMonaco: () => {
+        const editor = FeelEditorService.getStandaloneEditor();
+
+        if (!editor) {
+          return;
+        }
+
+        editor.trigger("keyboard", "type", { text: "\n" });
+      },
```
ljmotta: You could use an one-liner here.
`() => FeelEditorService.getStandaloneEditor()?.trigger("keyboard", "type", { text: "\n" });`

packages/feel-input-component/src/Monaco/FeelEditorService.ts
```diff
@@ -76,6 +76,14 @@ export class FeelEditorService {
     return this.getEditorBuilder().standaloneEditor !== undefined;
   }
 
+  /**
+   * Check If Suggest Widget is open. NOTE: you may need to call it with a setTimeout to 100, to wait for the animation to complete.
+   * @returns true if Suggest Widget is open, false otherwise.
+   */
+  static isSuggestWidgetOpen(): boolean {
+    return !!(this.getStandaloneEditor() as any)?._contentWidgets["editor.widget.suggestWidget"]?.position;
```
ljmotta, line 84: Maybe I've missed, but I couldn't find `_contentWidgets` property on the editor interfaces or in the docs. Could you please clarify? Thanks

packages/boxed-expression-component/src/components/Table/EditableCell.tsx
```diff
@@ -162,21 +165,30 @@ export function EditableCell({ value, rowIndex, columnId, onCellUpdate, readOnly
...
     (event: Monaco.IKeyboardEvent, newValue: string) => {
-      const key = event?.code.toLowerCase() ?? "";
-      const isModKey = event.altKey || event.ctrlKey || event.shiftKey;
-      const isEnter = isModKey && key === "enter";
-      const isTab = key === "tab";
-      const isEsc = !!key.match("esc");
+      const key = event?.code ?? "";
+      const isEnter = NavigationKeysUtils.isEnter(key);
+      const isTab = NavigationKeysUtils.isTab(key);
+      const isEsc = NavigationKeysUtils.isEscape(key);
```
ljmotta, line 168: Very good!

ljmotta (commented):
Thanks for this PR @fantonangeli ! Nice job! I've made some comments inline.

packages/feel-input-component/src/Monaco/FeelEditorService.ts
fantonangeli: @ljmotta I didn't find an "official" solution in the Monaco editor to check "Suggest Widget is open".

I tried different solutions and what worked better for me was this:
https://github.com/Microsoft/monaco-editor/issues/520#issuecomment-316642892
I just preferred to use `.position` instead `.isVisible` after a few tries, because with the 2nd one I didn't always received the right result.

A different way can be done just with vanilla JS with something like `document.querySelector()`

packages/boxed-expression-component/src/components/Table/EditableCell.tsx
fantonangeli: Thank you

packages/feel-input-component/src/Monaco/FeelEditorService.ts
ljmotta: I was more concerned about the `any` cast. We will need to keep track of `monaco` future release as `_contentWidget` is a private property and is more likely to be changed. This caveat would also be present with vanilla JS, so nothing we can do about it. I don't have a problem with this solution.

fantonangeli:
@jomarko Thank you for your tests. Below are my answers to your points:


-   The CTRL+ENTER support doesn't work in DMN Runner (the table mode):
     
     The `DMNRunnerTable` uses the package `@kie-tools/unitables` and renders the text cell with an `<input type="text">`, that doesn't support the multi line, so we need a `<textarea>`.
     To fix this I need to modify `kie-tools/packages/unitables/src/uniforms/AutoField.tsx`, that can have implications.
     Before doing this I would like to ask the opinion of @ljmotta. 

-   Copy and paste of area, where a cell contains multiline doesn't work correctly [1]. In the first picture I selected two input column cells, copied them in second picture I tried to paste the copied data into output column.
      
      I can fix this point in this PR.

packages/feel-input-component/src/Monaco/FeelEditorService.ts
fantonangeli: I created a jira to keep track of this: https://issues.redhat.com/browse/KOGITO-7094

ljmotta:
@fantonangeli The `<input>` comes from the `uniforms-patternfly` `TextField`. So to make this change, we could create a custom field and use it on the unitables `AutoField` instead of the default `TextField`. I think we can make this change another PR. I've create a JIRA for it: https://issues.redhat.com/browse/KOGITO-7104 .

jomarko:
@fantonangeli not sure why, it seems build was skipped, i.e. vsix file was not generated. Could we re-run complete CI? or provide vsix file so I can recheck please?

fantonangeli:
@jomarko the build has skipped because I put this PR temporarily in "draft", to finalize the copy&paste.
I will change it to "ready for review" when everything works on my side.

fantonangeli:
@jomarko for the copy & paste with multiline values I created a different Jira to track the change
https://issues.redhat.com/browse/KOGITO-7109

ljmotta (approved):
@fantonangeli Thanks for this PR! The code looks great.

jomarko (approved):
Thank you @fantonangeli

yesamer (approved):

PR #966, opened 2022-05-02
KOGITO-6934: Modify Dashbuilder Runtime to support client deployments

jesuino:
With this change dashbuilder can accept client content and run on client only.

It also formalize the runtime modes:

* APP: This is the default and the legacy mode, where it tries to retrieve information and load datasets from server. If this mode is on, but no backend is available, then the mode will change to EDITOR.
* EDITOR: This is for using dashbuilder as an editor, with dynamic content loaded via the API.
* CLIENT: This is the new mode where users can provide their own YML files to be loaded in Dashbuilder. In this mode there's no request to the backend and it expect users to provide YML files. If not file is provided, then users can upload their YML, which is lost after the page is refreshed.

Users can configure dashbuilder using the file `setup.js`. Here's a client setup where dashbuilder looks for dashboards in path `dashboards` and we map two dashboards:

 ``` 
dashbuilder = {

    path: 'dashboards',
    dashboards: ["dashboard1.yml", "dashboard2.yml"]
}
 ```
 
 To make dashbuilder as an editor:
 
  ```
dashbuilder = { mode: "EDITOR" }
 ```


jesuino:
Hello @barboras7  could you please review this PR?

Thanks!

barboras7:
Hi @jesuino,
could you please advise me on how to use setup.js file to set the editor mode? 
I might have missed this information when I was not working on Dashbuilder :slightly_smiling_face:

jesuino:
Hello @barboras7 

This is a new feature added by this PR. It allow us to use Dashbuilder Runtime as an editor. We have been doing it since it was only client mode, but this PR formalize the modes which dashbuilder runtime client can run.

To turn it into an editor use setup.js as:
```
dashbuilder = {mode: "EDITOR" }
```
Then you can send content to dashbuilder. The steps are:
* Build dashbuilder-runtime-client;
* Modify setup.js as mentioned above
* Open `dashbuilder-runtime-client/target/dashbuilder-runtime-client/index.html` on a browser and try this in the Developer Tools console and you should see a bar chart:

```
window.postMessage(`datasets:
- uuid: products
  content: >-
            [
              ["Computers", "Printer", 7, 4],
              ["Computers", "Laptop", 3, 2],
              ["Electronics", "Camera", 10, 7],
              ["Electronics", "Headphones", 5, 9]
            ]
  columns:
    - id: Section
      type: LABEL
    - id: Product
      type: LABEL
    - id: Quantity1
      type: NUMBER
    - id: Quantity2
      type: NUMBER
pages:
- properties:
    text-align: center

  components:
    - properties:
          border: solid   1px
      settings:
        type: BARCHART
        chart:
            bgColor: FEFEFE
            width: '800'
            height: '400'
            
            zoom: 'true'
            margin:
                right: '50'
                top: '50'
            legend:
                show: 'true'
                position: 'bottom'
            grid:
                x: 'false'
                y: 'false'
        dataSetLookup:
            uuid: products
            group:
            - columnGroup:
                source: Product
              groupFunctions:
                - source: Product
                - source: Quantity1
                  function: SUM
                - source: Quantity2
                  function: SUM
`, null)
```
Please let me know if you have any question

barboras7:
Hi @jesuino,
my issue was actually to understand where to put the `setup.js` file so that it is considered when starting the Dashbuilder Runtime, but I figured out based on the code from the PR :slightly_smiling_face: 
Anyway, I should be done with the review tomorrow. At this moment, I just would like to share one concern/suggestion. 
When the editor is running in the `CLIENT` mode, on the left side, there's this example code and on the right side, there's the upload page/uploaded model. I think it might be quite confusing that the "editor" shows some code and the model shown is based on another code. My suggestion would be to either set the `preview=true` as the default, hide the 'editor' completely from the UI when using `CLIENT` mode or load the source code of model (and make it read-only for example). WDYT?

jesuino:
Hello @barboras7 

Thanks for clarification, sorry I didn't get it the first time.

>> When the editor is running in the CLIENT mode, on the left side, there's this example code and on the right side, there's the upload page/uploaded model. I think it might be quite confusing that the "editor" shows some code and the model shown is based on another code. My suggestion would be to either set the preview=true as the default, hide the 'editor' completely from the UI when using CLIENT mode or load the source code of model (and make it read-only for example). WDYT?

The package `dashbuilder-editor` should not use Dashbuilder runtime in `CLIENT` mode.

The `CLIENT` is supposed to be used with static YML files. We plan to have a better way to deploy YAML files in the future, then the concept of modes will be internal only. Users will only see dashbuilder inside some channel (VSCode, Dev Sandbox or Chrome extension) or when using some of our tooling to deploy the dashboard (from sandbox itself, a NPM package or a base container image)

Thanks!

barboras7:
Everything looks good to me, thanks @jesuino! :+1:

barboras7 (approved):

tiagobento (approved):

PR #990, opened 2022-05-12
Optimize pnpm filters on Monorepo CI

tiagobento:
This PR makes a better filtering of the dependencies that need to be built on the `CI :: Monorepo` workflow.


caponetto (approved):
LGTM

PR #999, opened 2022-05-23
KOGITO-7162 : SWF Viewer - Package the app by using obfuscated style

inodeman:
Hi @romartin can you review this PR?


romartin (approved):
Thanks @inodeman , good catch here. 

Honestly I am not sure why the static fields are causing this issue, anyway it's true that there is not need for these fields to be non-static, as the roles for every node will no change during execution, so this fixes the issue also is better at memory consumption. So looks a good change for me to do! Approving.

Thanks!

romartin:
Hey @domhanak you're the tester for this ticket, although I don't think there is much to do with QE here. This is just an implementation fix that has no impact on the runtime, but it was prevernting @caponetto to move forward with the vscode integration.

So myabe just asking @caponetto to approve this is good for me, but I prefer to check with double you as well @domhanak 

Thanks guys

caponetto (approved):
LGTM!

PR #1002, opened 2022-05-25
KOGITO-6975: Additional JSON marshalling tasks

Josephblt:
**JIRA:** [KOGITO-6975 [SWF Editor] - Additional JSON marshalling tasks](https://issues.redhat.com/browse/KOGITO-6975)

JSON Object node support for start, transition, end
Removed logic from definitions.
Adjusted default spacing for JSON pretty-printing.

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/main/webapp/test.html
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/main/webapp/test.html
@@ -538,6 +538,49 @@
             },
           ],
         },
+
+        {
+          id: "startTransitionEndObjects",
+          version: "1.0",
+          specVersion: "0.8",
+          name: "Start, Transition and End as Objects",
+          description: "Start, Transition and End as Objects",
+          start: {
+            stateName: "Init State",
+            schedule: "2020-03-20T09:00:00Z/2020-03-20T15:00:00Z",
+          },
+          states: [
+            {
+              name: "Init State",
+              type: "inject",
+              end: false,
+              transition: {
+                nextState: "Next State",
+              },
+              onErrors: [
+                {
+                  errorRef: "Some Error",
+                  end: false,
+                  transition: {
+                    nextState: "Error State",
+                  },
+                },
+              ],
+            },
+            {
+              name: "Error State",
+              type: "inject",
+              end: true,
+            },
+            {
+              name: "Next State",
+              type: "inject",
+              end: {
+                terminate: true,
+              },
+            },
+          ],
+        },
       ];
     </script>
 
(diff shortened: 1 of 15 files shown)


packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/main/webapp/test.html
```diff
@@ -538,6 +538,49 @@
...
+            },
+            {
+              name: "Error State",
+              type: "inject",
+              end: true,
+            },
+            {
+              name: "Next State",
+              type: "inject",
+              end: {
```
domhanak, line 578: according to specification, users are able to also use:
```
end : {
    terminate: true,
    produceEvents: [{
        eventRef: "provisioningCompleteEvent",
        data: "${ .provisionedOrders }"
    }],
    compensate: true,
    continueAs: "newWorkflowID"
}
```
does it make sense to use everything possible here so that we know we can handle these cases too w/o  braking the view/format/json after marshalling/unmarshalling roundtrip?

romartin (approved):

romartin:
Hey @Josephblt 

Can you please try to move this forward? There is some pending Q from @domhanak . 

Thanks!

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/main/webapp/test.html
Josephblt: Hello, @domhanak. Yes, most of the fields here are not yet implemented so they have no effect. The unmarshalling works fine and does not raise errors. The problem is with marshalling. It removes the unused fields for some reason. But anyway, since the editor is read-only this should not be a problem. So I vote to proceed with the review.

domhanak: @Josephblt well I am not sure I agree, the fact we are removing some data from the file when marshalling happens is not ideal. What will happen once we integrate to VSCode? The issue will be there too, the save operation might be problematic, let me know, I ll report this anyway once I finish the review. Monday probably.

Josephblt: Hello @domhanak. When we do the VSCode integration, yes this can be a problem. But, for now, we do not need to worry.

caponetto: Guys, I'm out of context here but keep in mind that VS Code integration is ongoing.
So the fewer issues the better.
/cc @romartin

domhanak (approved):
So the main use case is working, however I see a*critical* if not *blocker* issue related to marshaling roundtrip in this PR.
See: https://issues.redhat.com/browse/KOGITO-7331 cc @caponetto  

Approving based on gchat clarifications and agreement that above issue will be resolved in follow up improvements.

romartin:
Thanks for the feedback @domhanak , also let you know that we already started work on https://issues.redhat.com/browse/KOGITO-7331 :+1: 

BTW @caponetto you're also a reviewer, should I wait for your approval or can we move forward? 

Thanks guys

caponetto:
@romartin feel free to merge it :)

caponetto (approved):

PR #1007, opened 2022-05-26
KOGITO-7196: SWF Viewer - Replace reset viewport button by autofit

inodeman:
Hi @romartin can you review this PR?

### DIFF
--- a/packages/serverless-workflow-diagram-editor/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/components/mediators/ZoomLevelSelector.java
+++ b/packages/serverless-workflow-diagram-editor/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/components/mediators/ZoomLevelSelector.java
@@ -46,14 +46,14 @@ public class ZoomLevelSelector implements IsWidget {
     }
 
     private final View view;
-    private Command onReset;
+    private Command onScaleToFit;
     private Command onDecreaseLevel;
     private Command onIncreaseLevel;
 
     @Inject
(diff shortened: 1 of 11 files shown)


romartin (approved):
Thanks @inodeman , looks good to me!

romartin:
hey @domhanak , do you want to take a look at this? Thx in advance!

domhanak (changes requested):
HI @inodeman @romartin I noticed an issue with this. Here is my review.

**Use cases**

- Button is there and does what it should do but the action is not working properly 100% of the time. See third point.
- When I use the autofit button without moving the scrollbars everything works nicely.
- When I use the autofit button when I move the scrollbars first the autofit, fits the image into wrong place. See picture:
Screenshot from 2022-06-01 10-06-23

**Here are the steps to reproduce:**
1. Open the `Event Based Switch Example`
2. Zoom in using the widget so that you can move the scrollbars (both or one, issue happens in both cases)
3. Move the Scrollbars into different position. ( I move the all the way to the right in the screenshot)
4. Use autofit
5. Notice the image not centered and fitted.

**Expected results**
The scrollbars are also reset into initial position & image is fitted properly.

I also noticed that the canvas can not be navigated using mouse and I can only use the scrollbars, let me know what is the plan about this. I think it should be reported for enhancing later.

romartin:
thanks for the feedback @domhanak , good catches. ( Hopefully we'll be able to automate all those checks sooner than later, right? hehe :) )

Please @inodeman waiting for your answers or fixes too :)

Thanks guys

inodeman:
Hi @domhanak @romartin looking into it, should have a fix soon

inodeman:
Hi @domhanak @romartin fixed the issue about Scrollbars not resetting, about the issue about not being able to navigate using the mouse I think that will have to be another issue/discussion, let me know what you think

domhanak (approved):
Thanks, LGTM now :smile:

tiagobento (approved):

PR #1017, opened 2022-06-05
KOGITO-7141: Stop using Lerna for `yarn update-version-to`

RishiRajAnand:


RishiRajAnand:
As per discussion with @tiagobento , we'll continue using Lerna since its stewardship is being passed on to Nrwl and community can expect can long term support with a clear roadmap.

PR #1018, opened 2022-06-06
KOGITO-7227: SWF Viewer - Error Handling

Josephblt:
**JIRA:** [KOGITO-7227 - SWF Viewer: Error Handling](https://issues.redhat.com/browse/KOGITO-7227)

The error page is not shown after the canvas is loaded.
Marshaller errors are now caught and properly treated.
Error messages are shown in the editor for proper clarification.

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/main/java/org/kie/workbench/common/stunner/sw/client/editor/DiagramEditor.java
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/main/java/org/kie/workbench/common/stunner/sw/client/editor/DiagramEditor.java
@@ -41,7 +41,6 @@ import org.kie.workbench.common.stunner.core.client.service.ServiceCallback;
 import org.kie.workbench.common.stunner.core.client.shape.Shape;
 import org.kie.workbench.common.stunner.core.client.util.WindowJSType;
 import org.kie.workbench.common.stunner.core.diagram.Diagram;
-import org.kie.workbench.common.stunner.core.diagram.DiagramParsingException;
 import org.kie.workbench.common.stunner.core.diagram.Metadata;
 import org.kie.workbench.common.stunner.sw.client.services.ClientDiagramService;
 import org.kie.workbench.common.stunner.sw.client.services.IncrementalMarshaller;
@@ -112,7 +111,6 @@ public class DiagramEditor {
     }
 
     public Promise<Void> setContent(final String path, final String value) {
-        close();
         return promises.create((success, failure) -> {
             diagramService.transform(path,
                                      value,
@@ -158,7 +156,7 @@ public class DiagramEditor {
 
                                          @Override
                                          public void onError(final ClientRuntimeError error) {
-                                             stunnerEditor.handleError(new ClientRuntimeError(new DiagramParsingException()));
+                                             stunnerEditor.handleError(error);
                                              failure.onInvoke(error);
                                          }
                                      });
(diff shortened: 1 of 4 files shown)


Josephblt:
Hello @romartin. Please assign reviewers for this. Thanks.

domhanak:
@Josephblt Looks good the error message were helpful, only thing I notice is that in some cases the error is `a.b is undefined`
This happens when we have empty `states: []` or when there is something required missing like state is missing its `name`
Is this something that should be part of this PR or is it out of scope? I expected the message to say something like: _missing required attribute `name` for state on line 26_

Josephblt:
Hello @domhanak. That wasn't part of the scope. These are raw errors and this is the expected behavior. But I agree that mapping errors to their exact location (line, column) and showing user-friendly messages are great ideas. I suggest creating another JIRA if there isn't one already. I'm copying @romartin as well, to see what he thinks. Thanks.

domhanak (approved):
LGTM. I see no reason to delay, we can discuss the possibility of the suggestion on weekly call. :)

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/main/java/org/kie/workbench/common/stunner/sw/client/editor/DiagramEditor.java
```diff
@@ -62,6 +62,7 @@ public class DiagramEditor {
     private final IncrementalMarshaller incrementalMarshaller;
     private final CanvasFileExport canvasFileExport;
 
+    private boolean canvasInitialized = false;
```
romartin: hey @Josephblt , just a quick suggestion before merging. I think there is no need for this `canvasInitialized` flag, I think it's redundant, and in fact, I see no usages for it in the code, maybe it's something can be dropped? 
Anyway, if you need that flag information, it should be not necessary a new boolean flag there -> in order to see if the editor has been already initialized, what about  just calling `stunnerEditor.isClosed()`? 
Please let me know if it makes sense for you. 
Thx

romartin (commented):
Hey @Josephblt , all looks good to me, thanks, great work! Anyway, please just waiting for your feedback in a single comment, let's see if we can simplify a bit more :)
Thx

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/main/java/org/kie/workbench/common/stunner/sw/client/editor/DiagramEditor.java
Josephblt: Hello @romartin. The isClosed flag does not work well in this case. If you take a look at the set content structure it is possible for it to be true but the canvas is not showing yet. For example, a file that parses OK but has an unmatching state name reference. That is why I have to wait for the canvas to be initialized flag it and follow from there. I admit we could drop that canvasInitialized flag if we remove the error messages in the editor. Otherwise, we have to use it. Please let me know what you think. Thanks.

Josephblt: Hello @romartin. The isClosed flag does not work well in this case. If you take a look at the set content structure it is possible for it to be true but the canvas is not showing yet. For example, a file that parses OK but has an unmatching state name reference. That is why I have to wait for the canvas to be initialized flag it and follow from there. I admit we could drop that canvasInitialized flag if we remove the error messages in the editor. Otherwise, we have to use it. Please let me know what you think. Thanks.

romartin (approved):
Hey @Josephblt , looks great now! Thansk for that small refactoring, it results now much more readable and easy to understand. :+1: 
Great stuff

PR #1049, opened 2022-06-24
KOGITO-7089: [yard Editor] Create VSCode Extension for yard

yesamer:
https://issues.redhat.com/browse/KOGITO-7089

First version of the VSCode extension for yard

Screenshot from 2022-06-29 11-13-08


packages/vscode-extension-yard-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -0,0 +1,130 @@
...
+      return;
+    }
+
+    await maybeOpenAsDiagramIfSwf({
+      configuration: args.configuration,
+      textEditor: vscode.window.activeTextEditor,
+      active: false,
+    });
+  }
+}
```
tiagobento, line 130: @yesamer I think we could extract this to avoid the duplication. The mechanisms will be always the same for YARD and SWF, so no point in duplicating it, as I don't expect they will evolve separately. Same with some fo the configurations.. Can you please try and see if those fit into the `vscode-extension` package? If not, feel free to create a new one :D

yesamer: @tiagobento It makes sense, I'll check that. Thx u!

yesamer: @tiagobento I created a ticket for that https://issues.redhat.com/browse/KOGITO-7474

packages/vscode-extension-yard-editor/src/extension/extension.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const configuration = new YardVsCodeExtensionConfiguration();
+
+  const kieToolsEditorStore = await KieToolsVsCodeExtensions.startExtension({
+    editorDocumentType: "text",
+    extensionName: "kie-group.vscode-extension-yard-editor",
+    context: context,
+    viewType: WEBVIEW_EDITOR_VIEW_TYPE,
+    generateSvgCommandId: COMMAND_IDS.getPreviewSvg,
+    silentlyGenerateSvgCommandId: COMMAND_IDS.silentlyGetPreviewSvg,
```
yesamer, line 46: @tiagobento @caponetto Added these IDs because they are mandatory - even if currently yard doesn't use SVG generation command. Should we think to make these optional?

caponetto: Considering that YARD and PMML editors are being forced to pass these ids even not supporting these commands, I think it would be nice to make them optional e register them only if command ids are provided.

caponetto (approved):
LGTM!

packages/vscode-extension-yard-editor/src/extension/extension.ts
tiagobento: +1 to make them optional.

yesamer: https://issues.redhat.com/browse/KOGITO-7491

PR #1053, opened 2022-06-28
[NOISSUE] Correctly use config and live setting for insecureSkipVerify

thiagoelg:
Extended Services app was only reading the `insecureSkipVerify` flag from the config file, not the live setting (from the systray menu). 

Now it correctly reflects both, by starting the proxy with the config from the yaml file, then reading the live setting from the systray menu.


tiagobento (approved):

caponetto (approved):
I just tested with OpenShift Sandbox and another OpenShift cluster that has a self-signed certificate.  The option is now working as expected. Thank you, @thiagoelg!

ederign (approved):

PR #1055, opened 2022-06-30
KOGITO-7477: Better error handling VS Code Service Registry integration

pefernan:
Since the we cannot make our VS Code SWEditor extension depend on "RH Authentication" extension (so it can work in vscode.dev) I just made some changes to fix the login issues if "RH Authentication" isn't installed (and enabled):

1. Warn the user if he has a registry that requires RH Auth via warning when the registry is initialized. This could also be triggered when the registry is re-initialized due to a settings change or when the vscode extensions registry changes (install/uninstall/activated/enabled..).

image

2. When a  Service Registry Client is initialized, if needs RH Auth but the extension isn't present it will be created but the login button won't appear in the `functions` codelenses


packages/vscode-extension-serverless-workflow-editor/src/extension/serviceCatalog/serviceRegistry/ServiceRegistriesStore.ts
```diff
@@ -140,4 +153,22 @@ export class ServiceRegistriesStore {
   private notifyRefresh() {
     this.subscriptions.forEach((subscription) => subscription(this.storedServices));
   }
+
+  private checkSettingsAuthProviders(registriesSettings: SwfServiceRegistriesSettings) {
+    if (this.authProviderFactory.isRHAuthEnabled) {
+      return;
+    }
+
+    const needsRHAccount = registriesSettings?.registries?.find(
```
tiagobento:
```suggestion
    const needsRedHatAccount = registriesSettings?.registries?.find(
```

or 

```suggestion
    const needsRhAccount = registriesSettings?.registries?.find(
```

We do not use upper-case acronyms, as they can be very confusing when put together. (e.g. `RHTHTTPCDIService` This is a stupid name but you get the point 😝 )

This is a convention of our codebase, but there's no definitive answer, of course. See this discussion: https://stackoverflow.com/questions/8277355/naming-convention-for-upper-case-abbreviations

packages/vscode-extension-serverless-workflow-editor/src/extension/serviceCatalog/serviceRegistry/auth/AuthProviderFactory.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+import * as vscode from "vscode";
+import { AuthProviderType } from "@kie-tools/serverless-workflow-service-catalog/dist/api";
+import { AuthProvider } from "./AuthProvider";
+import { RHCCAuthProvider } from "./RHCCAuthProvider";
+import { NoOpAuthProvider } from "./NoOpAuthProvider";
+
+export class AuthProviderFactory {
+  private _rhAuthEnabled: boolean;
```
tiagobento: Do we really need this here? I mean, why can't we always do this -> `vscode.extensions.getExtension("redhat.vscode-redhat-account") !== undefined;`?

This variable here looks like a state we don't want to manage..

tiagobento (changes requested):
@pefernan Nice PR!

Can you please take a look at the comments? I think we can remove a mutable variable from the `AuthProviderFactory` class.

packages/vscode-extension-serverless-workflow-editor/src/extension/serviceCatalog/serviceRegistry/auth/AuthProviderFactory.ts
pefernan: @tiagobento I'm keeping this state to be able to listen to extensions change (install/uninstall/disable) [here](https://github.com/kiegroup/kie-tools/pull/1055/files/d13c41ef4f02ea7d44317a9777a5538df1bf2223#diff-b05525786f27acb9980ccbf816d1824ed1a580d30dbff866f5f84c8888f39710R36). This way we can be aware if there are changes on the extension and if the change affects the RH Auth Extension. The problem is that the event triggered doesn't give any clue about the change that happened

tiagobento: Ok, I get it.. so you're using it to compare if the extension went from NOT AVAILABLE to AVAILABLE, right? But then I think the thing is a little bit more complex.. What happens if I go from NOT AVAILABLE to AVAILABLE to NOT AVAILABLE again? I mean, we're not calling a `deinit` method on `ServiceRegistriesStore`.

If we don't have a deinit process, then why do we need an init as well? I mean, something looks odd with this approach.. can you please give more details?

pefernan: yea on every change that affects the RH Auth extension we re-initialize the store.

`init` actually does the `deinit` (clear) before doing the real `init` (load the registry clients from the extension settings), maybe it could be renamed to `reinit` if you will.

paulovmr (approved):

packages/vscode-extension-serverless-workflow-editor/src/extension/serviceCatalog/serviceRegistry/auth/RedHatAuthenticationStore.ts
```diff
@@ -0,0 +1,62 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as vscode from "vscode";
+
+export class RedHatAuthenticationStore implements vscode.Disposable {
+  private static instance = new RedHatAuthenticationStore();
```
tiagobento: @pefernan Can we please avoid having a static singleton? I much rather have this as an instance created at `extension.ts` and passed down as a parameter.

packages/vscode-extension-serverless-workflow-editor/src/extension/serviceCatalog/serviceRegistry/auth/RedHatAuthenticationStore.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  private readonly subscriptions = new Set<() => void>();
+  private readonly disposeCallback: vscode.Disposable;
```
tiagobento:
```suggestion
  private readonly rhAuthExtensionChangeCallback: vscode.Disposable;
```

packages/vscode-extension-serverless-workflow-editor/src/extension/serviceCatalog/serviceRegistry/auth/RedHatAuthenticationStore.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  private _rhAuthEnabled: boolean;
```
tiagobento:
```suggestion
  private _rhAuthExtensionEnabled: boolean;
```

packages/vscode-extension-serverless-workflow-editor/src/extension/serviceCatalog/serviceRegistry/auth/RedHatAuthenticationStore.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public dispose() {
+    this.subscriptions.clear();
+    this.disposeCallback.dispose();
+  }
+
+  private getCurrentRHExtensionEnabled(): boolean {
+    return vscode.extensions.getExtension("redhat.vscode-redhat-account") !== undefined;
+  }
```
tiagobento: Still using capitalized acronym.

packages/vscode-extension-serverless-workflow-editor/src/extension/serviceCatalog/serviceRegistry/auth/RedHatAuthenticationStore.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  public static get() {
+    if (this.instance === undefined) {
+      this.instance = new RedHatAuthenticationStore();
+    }
+    return this.instance;
+  }
```
tiagobento: Can be removed once this is not a static singleton anymore.

tiagobento (changes requested):
Good to go except for some conventions around names and the static singleton.

@pefernan Can you please double-check that everywhere we reference the "Red Hat Authentication" extension, we use the same terminology? Right now we have "Red Hat Auth", "Red Hat Authentication", "Red Hat Authentication Extension" etc.

I personally prefer the full name, "Red Hat Authentication Extension".

tiagobento (approved):

PR #1057, opened 2022-07-04
upgrade gson to version 2.8.9

rgdoliveira:
com.google.code.gson:gson before 2.8.9 is vulnerable to Deserialization of Untrusted Data via the
writeReplace() method in internal classes, which may lead to DoS attacks.


yesamer (approved):
Thank you, @rgdoliveira

PR #1058, opened 2022-07-04
KOGITO-7476: [yard Editor] Setup Tests for yard VSCode extension

yesamer:
In this PR, I introduced integration tests for the yard vs-code extension.
To run it locally, you need to set up the test-it global variable to `true`
`export KIE_TOOLS_BUILD_testIT=true`

Then, you can run `yarn test:it` or `yarn build:prod`

Please note that some assets (eg. `VSCodeTestHelper.ts`) are a duplication utils class already present in `SWF` and `kogito-kie-editors` extensions it-tests. The idea for the future is to create a module for our extensions that contains common utils class
(with @domhanak )


packages/vscode-extension-yard-editor/package.json
```diff
@@ -16,7 +16,9 @@
     "pack:prod": "vsce package --githubBranch main  --yarn -o ./dist/vscode_extension_yard_editor_$npm_package_version.vsix",
     "compile": "webpack --env dev",
     "watch": "webpack --env dev",
-    "run:webmode": "pnpm vscode-test-web --browserType=chromium --extensionDevelopmentPath=. --version=stable"
+    "run:webmode": "pnpm vscode-test-web --browserType=chromium --extensionDevelopmentPath=. --version=stable",
+    "test:it": "pnpm run-script-if --bool \"$(build-env global.build.testIT)\" --then \"pnpm test:it:clean\" \"cpr it-tests/resources it-tests-tmp/resources\" \"tsc --project tsconfig.it-tests.json --skipLibCheck --sourceMap true\" \"extest setup-and-run --yarn -u -e ./test-resources -o ./it-tests/settings.json out/*.test.js\"",
+    "test:it:clean": "rimraf ./dist-it-tests && rimraf ./test-resources && rimraf ./out && rimraf ./it-tests-tmp && rimraf test-resou && rimraf *.vsix"
```
tiagobento, line 21: @yesamer Can we try and enable these tests as part of the build?

yesamer: @tiagobento Undoubtedly we can (and we must) do it. That is part of a strategy we are designing with @domhanak to improve our test assets. And we should do the same for `SWF` and `kogito-kie-editors`.

yesamer: @tiagobento Modified `"build:prod"` with `"rimraf dist && webpack && pnpm pack:prod && pnpm test:it` 
I will create a PR with the same changes for the `SWF` and `kogito-kie-editors` extensions.
I remember you all that these tests will run only if `build-env global.build.testIT` variable is set to true. Currently, only for our nightly runs these are enabled.

tiagobento: @yesamer Yeah, we disabled those because they were showing very frequent false negatives. Let's try and make sure that those tests are stable enough to be part of the build.

tiagobento (approved):

PR #1063, opened 2022-07-05
KOGITO-7491: Setting SVG commands ID as optional at Editor creation

yesamer:
JIRA: https://issues.redhat.com/browse/KOGITO-7491

Currently `startExtension` requires SVG commands ID. These are useful in case the editor supports SVG creation. `yard` doesn't require it, so I put these as optional parameters.


tiagobento (approved):

tiagobento:
@yesamer I believe we could apply the same changes for the PMML Editor extension, as it also doesn't have SVG generation.

PR #1090, opened 2022-07-13
Fix deploy without kogito.yml file

ljmotta:
- Remove the requirement of a kogito.yml file to deploy the service.
- Fix the Windows file paths.
- Remove the "kie-sandbox" from the Staging Knative Workflow plugins artifacts.

### DIFF
--- a/packages/kn-plugin-workflow/pkg/common/helper.go
+++ b/packages/kn-plugin-workflow/pkg/common/helper.go
@@ -28,14 +28,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func GetOsCommand(command string) string {
-	var osCommand = "./" + command
-	if runtime.GOOS == "windows" {
-		osCommand = ".\\" + command
-	}
-	return osCommand
-}
-
 func RunCommand(command *exec.Cmd, verbose bool, commandName string, friendlyMessages []string) error {
 	stdout, _ := command.StdoutPipe()
 	stderr, _ := command.StderrPipe()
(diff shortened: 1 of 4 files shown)


paulovmr (approved):

packages/kn-plugin-workflow/pkg/common/helper.go
```diff
@@ -21,12 +21,21 @@ import (
...
 	"time"
 
 	"github.com/briandowns/spinner"
 	"github.com/spf13/cobra"
 )
 
+func GetOsCommand(command string) string {
+	var osCommand = "./" + command
+	if runtime.GOOS == "windows" {
+		osCommand = ".\\" + command
```
barboras7: I am not sure if I understand this correctly, however based on what I read here https://springhow.com/maven-wrapper-explained/, the mvnw command for windows should be in `mvnw.cmd` format. Based on our code, it's .\mvnw. Is it correct?

ljmotta: Yes, it's correct. Strange, we've tested the `mvnw` for windows and we didn't have any problems. I'm going to make the change to use the `mvnw.cmd`.

packages/kn-plugin-workflow/pkg/common/helper.go
```diff
@@ -100,3 +101,11 @@ func DefaultTemplatedHelp(cmd *cobra.Command, args []string) {
 		fmt.Fprintf(cmd.ErrOrStderr(), "unable to display help text: %v", err)
 	}
 }
+
+func GetOsCommand(command string) string {
+	var osCommand = "./" + command
+	if runtime.GOOS == "windows" {
+		osCommand = ".\\" + command
```
barboras7, line 108: This code returns `.\mvnw.cmd` for Windows, but based on what I found (for example in the article I posted here before) I would expect `mvnw.cmd`. Have you tested manually that the code works for Windows?

ljmotta: Yes, I have. The `.\` says that the `mvnw.cmd` is in the current directory. This way works in the PowerShell (it doesn't work without it) and in the Command Prompt.

barboras7 (approved):
Looks good, thanks @ljmotta 👍

PR #1102, opened 2022-07-15
KOGITO-7365:  [SWF Editor] Defining an API to get the node name of text line selected

fantonangeli:
Issue: https://issues.redhat.com/browse/KOGITO-7365

Special thanks to @tiagobento, who helped me to fix a channel communication issue.

Currently, it only shows a `console.log()` with the `nodeName` and the `documentUri` in the VSCode Extension.

KOGITO-7365


packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -183,4 +201,40 @@ export async function setupDiagramEditorControls(args: {
       active: false,
     });
   }
+
+  vscode.window.onDidChangeTextEditorSelection((e) => {
+    // prevent kogitoSwfLanguageService__moveCursorToNode to fire this event
+    if (e.kind === vscode.TextEditorSelectionChangeKind.Command) {
+      return;
+    }
```
tiagobento: I think we could include a new guard clause here to ignore this event if the file is not SWF.

packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -183,4 +201,40 @@ export async function setupDiagramEditorControls(args: {
...
+
+    const uri = e.textEditor.document.uri;
+    const offset = e.textEditor.document.offsetAt(e.selections[0].active);
+
+    const swfOffsetsApi = initSwfOffetsApi(e.textEditor.document);
+
+    if (!swfOffsetsApi) {
+      return;
+    }
```
tiagobento: Then here, we can always return a SwfOffsetsApi implementation, never undefined.

packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -22,15 +22,41 @@ import {
...
 
 function isSwf(textDocument: vscode.TextDocument) {
   return getFileLanguage(textDocument.fileName) !== null;
 }
 
+function initSwfOffetsApi(textDocument: vscode.TextDocument): SwfJsonOffsets | SwfYamlOffsets | undefined {
+  if (!textDocument) {
+    console.debug("TextDocument not found");
+    return;
+  }
```
tiagobento: No need to test that, as the type system doesn't allow undefined values.

packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -22,15 +22,41 @@ import {
...
+
+  const fileLanguage = getFileLanguage(textDocument.fileName);
+
+  if (!fileLanguage) {
+    return;
+  }
```
tiagobento:
```suggestion
  const fileLanguage = getFileLanguageOrThrow(textDocument.fileName);
```

packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -22,15 +22,41 @@ import {
...
 
-let swfOffsetsApi: SwfOffsetsApi | undefined = undefined;
+const swfJsonOffsets = new SwfJsonOffsets();
+const swfYamlOffsets = new SwfYamlOffsets();
 
 function isSwf(textDocument: vscode.TextDocument) {
   return getFileLanguage(textDocument.fileName) !== null;
 }
 
+function initSwfOffetsApi(textDocument: vscode.TextDocument): SwfJsonOffsets | SwfYamlOffsets | undefined {
```
tiagobento:
```suggestion
function initSwfOffsetsApi(textDocument: vscode.TextDocument): SwfJsonOffsets | SwfYamlOffsets {
```

packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -135,20 +161,12 @@ export async function setupDiagramEditorControls(args: {
...
-        }
+        const swfOffsetsApi = initSwfOffetsApi(textEditor.document);
 
-        const editorContent = textEditor.document.getText();
-
-        if (!swfOffsetsApi || documentUri !== swfOffsetsApi.documentUri) {
-          swfOffsetsApi =
-            fileLanguage === FileLanguage.JSON ? new SwfJsonOffsets(documentUri) : new SwfYamlOffsets(documentUri);
+        if (!swfOffsetsApi) {
+          return;
```
tiagobento: Can be removed once the signature changes.

packages/serverless-workflow-text-editor/src/editor/textEditor/SwfTextEditorController.ts
```diff
@@ -171,6 +177,30 @@ export class SwfTextEditorController implements SwfTextEditorApi {
     this.editor?.setPosition(targetPosition);
   }
 
+  public highlightNodeOnDiagram(event: editor.IEditorMouseEvent): void {
```
tiagobento: I don't think this method's name is very good. Our SWF Text Editor can work without a Diagram companion, so, in theory, it doesn't even know what a Diagram is :)

packages/serverless-workflow-text-editor/src/editor/textEditor/SwfTextEditorController.ts
```diff
@@ -51,19 +51,19 @@ export interface SwfTextEditorInstance {
...
-
   constructor(
     content: string,
     private readonly onContentChange: (content: string, operation: SwfTextEditorOperation) => void,
     private readonly language: FileLanguage,
     private readonly operatingSystem: OperatingSystem | undefined,
     private readonly isReadOnly: boolean,
     private readonly setValidationErrors: (errors: editor.IMarker[]) => void,
-    private readonly fileName: string = ""
+    private readonly onEditorSelectedNode: (nodeName: string) => void
```
tiagobento: Maybe rename to `onNodeSelected`?

tiagobento (changes requested):
@fantonangeli This is looking good! Very nice to see the interactions between the text and diagram Editor for SWF getting into shape :)

First, let me say that not all the comments I'm going to make here are necessarily related to your changes, but since you're touching things that are close enough to what I'm going to mention, I think we can use the opportunity to do some boy scouting and improve the code :)

------

Looking at this PR, I just noticed that there's some confusion regarding the API interfaces and where the methods live.

- `kogitoSwfLanguageService__moveCursorToNode` should only exist at `ServerlessWorkflowDiagramEditorChannelApi`
- `kogitoSwfLanguageService__highlightNode` should only exist at `ServerlessWorkflowDiagramEditorEnvelopeApi`

The LanguageService API should *not* contain those methods, as these are UI features, not related to language support. Can you please revisit all the definitions of these two methods to make sure that they only exist at the right places?

Also, I think it's time for a cleanup at the VS Code extension API implementations:
- `packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowEditorChannelApiImpl.ts` should be renamed to `ServerlessWorkflowDiagramEditorChannelApiImpl.ts`, as it's only used by the SWF Diagram Editor.
- `packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowEditorChannelApiProducer.ts` should be renamed to `ServerlessWorkflowDiagramEditorChannelApiProducer`, as it's only used for the SWF Diagram Editor.

Note that `ServerlessWorkflowDiagramEditorChannelApiImpl` won't need the SwfLanguageServiceApi anymore, as the methods I mentioned above will be moved somewhere else.

I noticed that `ServerlessWorkflowCombinedEditorChannelApi` is not implementing the `SwfLanguageServiceApi`, but it should.

Last thing I noticed is that on the SWF Combined Editor, the `moveCursorToNode` is forwarding the call to its Channel, where in fact it should handle it and communicate the SWF Text Editor that this occurred. Same thing for the `highlightNode` method, which should not forward, but instead communicate the Diagram that the selection changed on the SWF Text Editor.

-------

Please let me know if you have any questions. Let's try and clean the APIs a little bit. I understand the nature of what we're doing is very complex, but we can try and keep this to a minimum whenever possible :D

packages/serverless-workflow-combined-editor/src/impl/ServerlessWorkflowDiagramEditorChannelApiImpl.ts
```diff
@@ -0,0 +1,105 @@
...
+    return this.defaultApiImpl.kogitoWorkspace_resourceListRequest(request);
+  }
+
+  public kogitoEditor_theme(): SharedValueProvider<EditorTheme> {
+    return this.defaultApiImpl.kogitoEditor_theme();
+  }
+
+  public kogitoSwfDiagramEditor__onNodeSelected(args: { nodeName: string; documentUri?: string }): void {
+    return this.textEditorEnvelopeApi?.notifications.kogitoSwfTextEditor__moveCursorToNode.send(args);
+  }
```
tiagobento, line 104: 🎉

packages/serverless-workflow-combined-editor/src/editor/ServerlessWorkflowCombinedEditor.tsx
```diff
@@ -265,11 +275,26 @@ const RefForwardingServerlessWorkflowCombinedEditor: ForwardRefRenderFunction<
...
+    useSwfDiagramEditorChannelApi({
+      channelApi:
+        editorEnvelopeCtx.channelApi as unknown as MessageBusClientApi<ServerlessWorkflowDiagramEditorChannelApi>,
+      locale: props.locale,
+      embeddedEditorFile: embeddedDiagramEditorFile,
+      onEditorReady: onDiagramEditorReady,
+      getSwfTextEditorEnvelopeApi: () =>
+        textEditor?.getEnvelopeServer()
+          .envelopeApi as unknown as MessageBusClientApi<ServerlessWorkflowTextEditorEnvelopeApi>,
+    });
```
tiagobento: Here, you should be memoizing the parameter object with `useMemo`, as it will always trigger the inner hooks otherwise, given that a new object will be created everything this executes. Same goes for the `getSwfTextEditorEnvelopeApi` lambda. This should have its own memoization. Same goes for the `useSwfTextEditorChannelApi` hook.

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramApi.ts
```diff
@@ -19,8 +19,8 @@
  */
 export interface DiagramApi {
   /**
-   * Move the cursor in the text editor to a specified node
-   * @param stateName the name of the target node
+   * Event fired when a node is selected
+   * @param nodeName the name node
```
tiagobento:
```suggestion
   * @param nodeName the name of the node
```

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramService.ts
```diff
@@ -27,13 +27,13 @@ export class DiagramService implements DiagramApi {
   ) {}
 
   /**
-   * Move the cursor in the text editor to a specified node
-   * @param nodeName the name of the target node
+   * Event fired when a node is selected
+   * @param nodeName the name node
```
tiagobento: I think this is redundant? Maybe leave this just at the interface.

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramService.ts
```diff
@@ -27,13 +27,13 @@ export class DiagramService implements DiagramApi {
...
    */
-  public moveCursorToNode(nodeName: string) {
+  public onNodeSelected(nodeName: string) {
     if (!nodeName) {
       return;
     }
```
tiagobento: `string` can never be undefined or null. Are you checking for empty strings? I mean, this can be ambiguous.

packages/serverless-workflow-diagram-editor-envelope/src/envelope/ServerlessWorkflowDiagramEditorEnvelopeApiImpl.ts
```diff
@@ -122,4 +122,8 @@ export class ServerlessWorkflowDiagramEditorEnvelopeApiImpl
     }
     return editor.centerNode(uuid);
   }
+
+  public kogitoSwfDiagramEditor__highlightNode(args: { nodeName: string; documentUri?: string }) {
+    console.log("Received kogitoSwfDiagramEditor__highlightNode with args:", args);
+  }
```
tiagobento, line 128: This is not going to be implemented as part of this PR?

packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowDiagramEditorChannelApiImpl.ts
```diff
@@ -172,13 +178,17 @@ export class ServerlessWorkflowEditorChannelApiImpl implements ServerlessWorkflo
...
+  public kogitoSwfDiagramEditor__onNodeSelected(args: { nodeName: string; documentUri?: string }): void {
+    this.swfTextEditorEnvelopeApiImpl.kogitoSwfTextEditor__moveCursorToNode({
       ...args,
       documentUri: this.editor.document.document.uri.path,
     });
   }
 
+  public kogitoSwfTextEditor__onSelectionChanged(args: { nodeName: string; documentUri?: string }): void {
+    this.diagramEditorEnvelopeApi?.notifications.kogitoSwfDiagramEditor__highlightNode.send(args);
+  }
```
tiagobento: 🎉

packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as vscode from "vscode";
+import { COMMAND_IDS } from "./commandIds";
+
+export interface ServerlessWorkflowTextEditorEnvelopeApi {
+  kogitoSwfTextEditor__moveCursorToNode(args: { nodeName: string; documentUri?: string }): void;
+}
```
tiagobento: Do we really need to duplicate this interface here? I'm thinking maybe we can inline this whole file, as it's not very complex anyway.

packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export class ServerlessWorkflowTextEditorEnvelopeApiImpl implements ServerlessWorkflowTextEditorEnvelopeApi {
+  public kogitoSwfTextEditor__moveCursorToNode(args: { nodeName: string; documentUri: string }): void {
+    vscode.commands.executeCommand(COMMAND_IDS.moveCursorToNode, args);
```
tiagobento: Somehow I didn't see that on my other reviews, but I don't think using VS Code's commands is the best solution for this. I feel like we're loosely coupling two features that should be more explicitly dependent on the code. VS Code commands, as a general rule, should be user-executable using the Commands prompt.. that's not the case here.. Do you think we can refactor this in a future PR?

packages/serverless-workflow-text-editor/src/editor/textEditor/SwfTextEditorController.ts
```diff
@@ -171,6 +176,30 @@ export class SwfTextEditorController implements SwfTextEditorApi {
...
+      return;
+    }
+
+    const offset = this.editor?.getModel()?.getOffsetAt(position);
+
+    if (!offset) {
+      return;
+    }
+
+    this.swfOffsetsApi.parseContent(this.getContent());
```
tiagobento, line 190: This was a trigger for me, as I think it's encapsulating mutable state. The reuse of `SwfOffsetsApi` instances opens a dangerous precedent that I would like to avoid. Do you think we can refactor this in a future PR?

packages/serverless-workflow-text-editor/src/editor/textEditor/SwfTextEditorController.ts
```diff
@@ -171,6 +176,30 @@ export class SwfTextEditorController implements SwfTextEditorApi {
...
+
+    const nodeName = this.swfOffsetsApi.getStateNameFromOffset(offset);
+
+    if (!nodeName) {
+      return;
+    }
+
+    this.onSelectionChanged(nodeName);
+  }
```
tiagobento: Just out of personal preference and style, I think to use negative space to indicate a separation between sections of a function. Here, here's how I would format this code:
```suggestion
    const position = event.target.position;
    if (!position) {
      return;
    }

    const offset = this.editor?.getModel()?.getOffsetAt(position);
    if (!offset) {
      return;
    }

    this.swfOffsetsApi.parseContent(this.getContent());

    const nodeName = this.swfOffsetsApi.getStateNameFromOffset(offset);
    if (!nodeName) {
      return;
    }

    this.onSelectionChanged(nodeName);
  }
```

tiagobento: This is not mandatory, it's more like a style alignment hint. Feel free to ignore this if you will :)

packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -22,15 +22,44 @@ import {
...
+  const fileLanguage = getFileLanguageOrThrow(textDocument.fileName);
+
+  const editorContent = textDocument.getText();
+
+  const swfOffsetsApi = fileLanguage === FileLanguage.JSON ? swfJsonOffsets : swfYamlOffsets;
+
+  swfOffsetsApi.parseContent(editorContent);
+
+  return swfOffsetsApi;
+}
```
tiagobento: Same comment here for the `SwfOffsetApi`s

packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
```diff
@@ -189,4 +206,35 @@ export async function setupDiagramEditorControls(args: {
...
+
+    if (!envelopeServer) {
+      return;
+    }
+
+    envelopeServer.envelopeApi.notifications.kogitoSwfDiagramEditor__highlightNode.send({
+      nodeName,
+      documentUri: uri.path,
+    });
+  });
```
tiagobento: I don't think this is the best place to put this. I would put this in `builtInVsCodeEditorSwfContributions.ts`. The `Diagram Editor Controls` referred by this file's name is not very descriptive, I know, but this was originally created to handle the opening of a new tab with the Diagram to the right of the built-in text-editor.

Also, I would move the `COMMAND_IDS.moveCursorToNode` registration there too.

Do you think we can rename this file to `setupDiagramEditorCompanionTab`? Maybe it's a better name :D

packages/serverless-workflow-text-editor/src/editor/textEditor/SwfTextEditorController.ts
```diff
@@ -171,6 +176,30 @@ export class SwfTextEditorController implements SwfTextEditorApi {
     this.editor?.setPosition(targetPosition);
   }
 
+  public onMouseDown(event: editor.IEditorMouseEvent): void {
```
tiagobento: Maybe rename this to `handleMouseDown`?

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramApi.ts
```diff
@@ -19,8 +19,8 @@
  */
 export interface DiagramApi {
```
tiagobento: Following the naming convention recently added to the `kie-bc-editors` package, this should be renamed to `DiagramExposedInteropApi`. See any other ExposedInteropApis for reference. Also, see that on `VsCodeServerlessWorkflowDiagramEditorFactory`, the exposed interop object `javaCodeCompletionService` property is typed with an interface that's not an ExposedInteropApi. Can you please extract `JavaCodeCompletionExposedInteropApi` to its own file and import it there?

packages/serverless-workflow-combined-editor/src/editor/hooks/useSwfTextEditorChannelApi.tsx
```diff
@@ -50,42 +50,35 @@ export function useCustomSwfChannelApi(args: {
...
   );
 
   const swfServiceCatalogChannelApiImpl = useMemo(
     () =>
       args.channelApi &&
       services &&
       serviceRegistriesSettings &&
-      new SwfServiceCatalogChannelApiImpl(args.channelApi, services, serviceRegistriesSettings),
+      new SwfServiceCatalogChannelApiImpl(
+        args.channelApi as MessageBusClientApi<SwfServiceCatalogChannelApi>,
```
tiagobento: Do you really need this cast here? I mean, `ServerlessWorkflowTextEditorChannelApi` implements `SwfServiceCatalogChannelApi`, doesn't it?

tiagobento (changes requested):
@fantonangeli Looking amazing!!! I found some things to update on this PR, but the code looks so much better already. I'm glad to see you're confident making chances and experimenting on the multiple layer of abstractions that we have. Very nice progress!

Let me know if you have questions about my comments. I think this is ready for review from others too, so please feel free to assign them :)

packages/serverless-workflow-combined-editor/src/editor/hooks/useSwfTextEditorChannelApi.tsx
```diff
@@ -18,25 +18,25 @@ import { EmbeddedEditorFile, StateControl } from "@kie-tools-core/editor/dist/ch
...
 import { useMemo } from "react";
+import { SwfServiceCatalogChannelApiImpl } from "../..";
 import { ServerlessWorkflowCombinedEditorChannelApi } from "../../api";
-import { SwfCombinedEditorChannelApiImpl } from "../../impl/SwfCombinedEditorChannelApiImpl";
-import { SwfFeatureToggleChannelApiImpl } from "../../impl/SwfFeatureToggleChannelApiImpl";
-import { SwfLanguageServiceChannelApiImpl } from "../../impl/SwfLanguageServiceChannelApiImpl";
-import { SwfServiceCatalogChannelApiImpl } from "../../impl/SwfServiceCatalogChannelApiImpl";
+import { ServerlessWorkflowTextEditorChannelApiImpl } from "../../impl/ServerlessWorkflowTextEditorChannelApiImpl";
 
-export function useCustomSwfChannelApi(args: {
```
caponetto, line 28: The goal of the `useCustomSwfChannelApi` hook was solely to make it reusable for text and diagram editors.
Since now we have `useSwfTextEditorChannelApi` and `useSwfDiagramEditorChannelApi`, maybe there is no point in having their code in hooks anymore.

packages/serverless-workflow-combined-editor/src/api/ServerlessWorkflowCombinedEditorChannelApi.ts
```diff
@@ -23,6 +24,4 @@ import { SwfFeatureToggleChannelApi } from "./SwfFeatureToggleChannelApi";
 export interface ServerlessWorkflowCombinedEditorChannelApi
   extends KogitoEditorChannelApi,
     SwfFeatureToggleChannelApi,
-    ServerlessWorkflowTextEditorChannelApi,
-    ServerlessWorkflowMermaidViewerChannelApi,
-    ServerlessWorkflowDiagramEditorChannelApi {}
```
caponetto, line 28: Please clean up unused imports.

caponetto (commented):
Hey @fantonangeli, really nice improvements!
I've left some inline comments. Please check if they make sense. :)
Also, I did a quick check on the SWF Sandbox, and looks like the text highlight after node selection is not working anymore. Can you please take a look at this?
Thanks!

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramService.ts
fantonangeli: This method is called by the GWT layer I can receive a non-valid `nodeName: null, undefined, empty string.
Because I do this check on the Java side as well, I can also skip this check. WDYT?

packages/serverless-workflow-diagram-editor-envelope/src/envelope/ServerlessWorkflowDiagramEditorEnvelopeApiImpl.ts
fantonangeli: Yes, this will be implemented in this Jira:
https://issues.redhat.com/browse/KOGITO-7366

packages/serverless-workflow-text-editor/src/editor/textEditor/SwfTextEditorController.ts
fantonangeli: I reused the `SwfOffsetsApi` to be able to cache the parsed result. In fact `parseContent()` return the memoized value, if the content is not changed. I can also refactor this in a future PR

packages/vscode-extension-serverless-workflow-editor/src/extension/setupDiagramEditorControls.ts
fantonangeli: I can also refactor this in a future PR

packages/serverless-workflow-combined-editor/src/editor/hooks/useSwfTextEditorChannelApi.tsx
fantonangeli: The cast is needed because `args.channelApi`type  is `MessageBusClientApi<ServerlessWorkflowTextEditorChannelApi>`

packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts
fantonangeli: I cannot reuse the same interface from `repos/kie-tools/packages/serverless-workflow-text-editor/src/api/ServerlessWorkflowTextEditorEnvelopeApi.ts` because the `ServerlessWorkflowTextEditorEnvelopeApi` extends `KogitoEditorEnvelopeApi`.
One way to reuse it, is to implement all `KogitoEditorEnvelopeApi` methods in `kie-tools/packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts`.
WDYT

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramService.ts
tiagobento: Ok, so I Java doesn't have a way to prevent `null` from being passed to a `String` type, which means that on the TypeScript perspective, this should be typed as `string | undefined | null`.

I'm not sure if GWT maps Java's `null` to JavaScript's `undefined` or `null`. @treblereel Would you know?

packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts
tiagobento: _"One way to reuse it, is to implement all KogitoEditorEnvelopeApi methods in kie-tools/packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts."_

Yep, that's what I mean by inlining this file. No need to have a separate interface for a method that's already declared in `ServerlessWorkflowTextEditorEnvelopeApi`.

packages/serverless-workflow-text-editor/src/editor/textEditor/SwfTextEditorController.ts
tiagobento: Yep, I understand. There are other ways to `hide` the cached value and make this class behave as if it didn't have one, though. Caching is a detail of the implementation of the `SwfOffsetsApi`, consumers of this class could simply ask for the parsed content as if it were being evaluated every time, but internally, you would cache the results when the content isn't changed. But let's do this refactoring in a future PR.

packages/serverless-workflow-combined-editor/src/editor/hooks/useSwfTextEditorChannelApi.tsx
tiagobento: Yep, what I mean is that `ServerlessWorkflowTextEditorChannelApi` is a superset of `SwfServiceCatalogChannelApi`, so there shouldn't be a problem assigning it directly. I tested with this simple code below and didn't get any errors.. Isn't that the same case?


```typescript
let a: MessageBusClientApi<KogitoEditorChannelApi> = {} as any;
let b: MessageBusClientApi<WorkspaceChannelApi> = {} as any;

b = a;
```

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramService.ts
treblereel: > Ok, so I Java doesn't have a way to prevent `null` from being passed to a `String` type, which means that on the TypeScript perspective, this should be typed as `string | undefined | null`.
> 
> I'm not sure if GWT maps Java's `null` to JavaScript's `undefined` or `null`. @treblereel Would you know?

If my memory serves me well, it must be `null`. There are few methos of Js that can help us in the case: 
* jsinterop.base.Js.typeof(...)
* jsinterop.base.Js.undefined()

to check if String can be null we can create a native method that do a simple check `console.log(check === null)`, so if we call it from Java like check(null), we ll see it's `null`.

So it's better to add a validation for such cases

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramApi.ts
fantonangeli: `JavaCodeCompletionApi` seems to be used by different packages, which I would avoid creating bugs.
Can we do it in a specific PR or creating a new `JavaCodeCompletionExposedInteropApi` and writing the JSdoc `@deprecated` comment to `JavaCodeCompletionApi`?

packages/serverless-workflow-combined-editor/src/editor/hooks/useSwfTextEditorChannelApi.tsx
fantonangeli: If you agree I would keep them separated from `ServerlessWorkflowCombinedEditor.tsx`, to keep it small and let `useSwfTextEditorChannelApi` and `useSwfDiagramEditorChannelApi` manage their own dependency, that can be more in the future. WDYT?

caponetto: Sounds good to me!

packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts
fantonangeli: @tiagobento If I embed `ServerlessWorkflowTextEditorEnvelopeApiImpl.ts` in `ServerlessWorkflowDiagramEditorChannelApiImpl.ts`, there the `Diagram Channel` will not call the `TextEditor Envelope` to move the cursor. 
Instead, `ServerlessWorkflowDiagramEditorChannelApiImpl.ts` will do the action directly without the separation between `Diagram Channel` and `TextEditor Envelope.
Is this ok?

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramApi.ts
tiagobento: Ok. Let's create a Jira to do it later then :) No need to deprecate.

packages/vscode-extension-serverless-workflow-editor/src/extension/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts
tiagobento: @fantonangeli It's not a problem at all. In fact, the way it is now is a little bit confusing, as you're calling the built-in VS Code text editor `Serverless Workflow Text Editor`, where it's not.

The VS Code implementation of `ServerlessWorkflowDiagramEditorChannelApi` does **not** interact with the serverless-workflow-text-editor component, because we don't use it on VS Code. That's why I think you should remove this class (`ServerlessWorkflowTextEditorEnvelopeApiImpl`) entirely and put the logic inside `ServerlessWorkflowDiagramEditorChannelApiImpl`. Makes sense?

packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramApi.ts
fantonangeli: Jira: https://issues.redhat.com/browse/KOGITO-7876

fantonangeli:
This issue is now ready for review

packages/vscode-java-code-completion/src/api/JavaCodeCompletionApi.ts
```diff
@@ -17,6 +17,7 @@
 import { JavaCodeCompletionAccessor } from "./JavaCodeCompletionAccessor";
 import { JavaCodeCompletionClass } from "./JavaCodeCompletionClass";
 
+/* TODO: JavaCodeCompletionApi: javaCodeCompletionService property is typed with an interface that's not an ExposedInteropApi. Can you please extract JavaCodeCompletionExposedInteropApi to its own file and import it there https://github.com/kiegroup/kie-tools/pull/1102#discussion_r935817679 */
```
caponetto: We have a JIRA for that, right? Please remove this comment.

packages/serverless-workflow-combined-editor/src/editor/ServerlessWorkflowCombinedEditor.tsx
```diff
@@ -265,12 +275,37 @@ const RefForwardingServerlessWorkflowCombinedEditor: ForwardRefRenderFunction<
...
+    }),
+    [editorEnvelopeCtx, embeddedDiagramEditorFile, onDiagramEditorReady, textEditor]
+  );
+
+  const useSwfTextEditorChannelApiArgs = useMemo(
+    () => ({
+      channelApi:
+        editorEnvelopeCtx.channelApi as unknown as MessageBusClientApi<ServerlessWorkflowTextEditorChannelApi>,
+      locale: props.locale,
+      embeddedEditorFile: embeddedDiagramEditorFile,
```
caponetto:
```suggestion
      embeddedEditorFile: embeddedTextEditorFile,
```

caponetto (commented):
Hi @fantonangeli, can you please check the Serverless Logic Web Tools?
The node selection still seems broken.

output

tiagobento (approved):
The code looks great to me! Amazing work @fantonangeli!!

I just merged #1202, maybe this will help verify that this PR is working as expected? Approving and ready to merge once the bugs are fixed and the CI is green.

fantonangeli:
@tiagobento 
Because in this PR I only show a `console.log()`, I have to listen/read to the DevTools Logs in the integration test. 
I tried to read those logs from the tests today but couldn't find a way.
Maybe we do it in the following PR?

caponetto (approved):
Thank you for this PR, @fantonangeli! Great work!

tomasdavidorg:
Hi @fantonangeli, @tiagobento I know about this PR but we had QE Face2Face last week. I will definitely look tomorrow. Sorry for the delay.

tomasdavidorg (approved):
Looks good to me.

tomasdavidorg:
One more thing: I tried this on Windows and the `documentUri` does not seem to be correct. 

`Received kogitoSwfDiagramEditor__highlightNode with args: {nodeName: 'GetCreditScore', documentUri: '/c:/Users/Tomas/test sw/swf-testing-folder-main/loanbroker.sw.json'}`

fantonangeli:
> One more thing: I tried this on Windows and the `documentUri` does not seem to be correct.
> 
> `Received kogitoSwfDiagramEditor__highlightNode with args: {nodeName: 'GetCreditScore', documentUri: '/c:/Users/Tomas/test sw/swf-testing-folder-main/loanbroker.sw.json'}`

Because it is is exactly the path I receive from VSCode API (e is the VSCode event), that I will use in the following PR to verify the diagram is the correct one:
`const uri = e.textEditor.document.uri;`
I think it should not cause issues because it is the URI representation of the path.

yesamer (approved):

PR #1105, opened 2022-07-16
KOGITO-7507: Improving `kogito-kie-editors` tests stability

yesamer:
In some cases, the test `Opens UserTask.bpmn file in BPMN Editor and test On Entry and On Exit actions` throws this error:
`stale element reference: element is not attached to the page document`
The reason is: `propertiesPanel` requires to be re-instantiated when the test changes the editor status.

A hint to apply when a new `it-tests` is introduced, is to run it locally multiple times. This is because these kinds of failures are not-deterministic.

### DIFF
--- a/packages/vscode-extension-pack-kogito-kie-editors/it-tests/extension-editors.test.ts
+++ b/packages/vscode-extension-pack-kogito-kie-editors/it-tests/extension-editors.test.ts
@@ -345,7 +345,7 @@ describe("Editors are loading properly", () => {
 
   it("Opens MultipleInstanceSubprocess.bpmn file in BPMN Editor and test Implementation/Execution value change", async function () {
     this.timeout(40000);
-    // Inicialization
+    // Initialization
     webview = await testHelper.openFileFromSidebar(MULTIPLE_INSTANCE_BPMN);
     await testHelper.switchWebviewToFrame(webview);
     const bpmnEditorTester = new BpmnEditorTestHelper(webview);
@@ -413,7 +413,7 @@ describe("Editors are loading properly", () => {
   });
 
   it("Opens UserTask.bpmn file in BPMN Editor and test On Entry and On Exit actions", async function () {
-    this.timeout(20000);
+    this.timeout(40000);
     webview = await testHelper.openFileFromSidebar(USER_TASK_BPMN);
     await testHelper.switchWebviewToFrame(webview);
     const bpmnEditorTester = new BpmnEditorTestHelper(webview);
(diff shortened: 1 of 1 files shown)


yesamer:
Lanched it 5 times.
Four of these were successfully
One failed:
`  1) Editors are loading properly
       Opens UserTask.bpmn file in BPMN Editor and test On Entry and On Exit actions:
     StaleElementReferenceError: stale element reference: element is not attached to the page document
  (Session info: chrome=98.0.4758.109)
      at Object.throwDecodedError (/home/runner/work/kie-tools/kie-tools/node_modules/.pnpm/selenium-webdriver@4.2.0/node_modules/selenium-webdriver/lib/error.js:522:15)
      at parseHttpResponse (/home/runner/work/kie-tools/kie-tools/node_modules/.pnpm/selenium-webdriver@4.2.0/node_modules/selenium-webdriver/lib/http.js:549:13)
      at Executor.execute (/home/runner/work/kie-tools/kie-tools/node_modules/.pnpm/selenium-webdriver@4.2.0/node_modules/selenium-webdriver/lib/http.js:475:28)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async Driver.execute (/home/runner/work/kie-tools/kie-tools/node_modules/.pnpm/selenium-webdriver@4.2.0/node_modules/selenium-webdriver/lib/webdriver.js:735:17)
  2) Editors are loading properly
       "after each" hook for "Opens UserTask.bpmn file in BPMN Editor and test On Entry and On Exit actions":
     NoSuchElementError: no such element: Unable to locate element: {"method":"css selector","selector":".monaco-dialog-box"}`

@domhanak Any idea to resolve this instability for that test?

domhanak:
@yesamer Agree with the hint. Not sure right now @zdrapela could you please take a look? Looks like a test you introduced has some issue over multiple runs, so there might be an issue.

@yesamer Today I noticed Stale Reference Errors on nightly jenkins too, so maybe something updated and we need to do so too? :D Checking, but from what I remember this is related to startup of the tests, which is handled by the framework.

yesamer:
@domhanak @tiagobento @caponetto @zdrapela 
I found the root cause of instability: a `chromedriver` `98.x` bug solved in version `103.x`
https://github.com/redhat-developer/vscode-extension-tester/issues/466

`vscode-extension-tester` team is trying to solve this and target the fix in the next `4.3.1` version.
Said so, I propose to disregard my PR, because not resolutive, and cohabit with this instability until the new version is released.

tiagobento:
Just a small note, as I didn't investigate it recently, but as far as I remember, we've been having issues with VS Code tests way before version 99 of the Chrome driver.

I'm +1 to wait for the fix, but keep in mind that not all instabilities may be solved after upgrading.

yesamer:
@tiagobento I am quite confident the new version will solve most of the failures. ;)
In case of other root causes, we will fix them too.

tiagobento:
@yesamer Ok! The more instabilities it fixes, the better! :)

Closing the PR.

PR #1112, opened 2022-07-20
KOGITO-7577: SWF Editor - Parameter named 'uuid' should be not null

hasys:
Hi @romartin, @ricardozanini,

The issue was that if Workflow missing name, diagram wasn't shown since Stunner uses Workflow name as a UUID for the root node.

The fix consist of two parts:
1. Workflow ID is used as UUID instead of Workflow Name
2. If Workflow ID is missing, than generate some random value to use it as UUID.

Please, tell me if it is OK to use Workflow ID as UUID and it will not break any backward compatibility (like automated tests or something else).

If you think it's better to keep Workflow Name as UUID of the root diagram node, I will revert this change and keep only second point (If Workflow Name is missing, generate some randome value to use it as UUID).

[VSIX for local tests](https://drive.google.com/file/d/1erUT_bBreE7Bi0Lr_Fb6H6GvTx0p8Tdz/view?usp=sharing)

Thank you!

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/definition/Workflow.java
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/definition/Workflow.java
@@ -60,6 +60,12 @@ public class Workflow {
     public String id;
 
     /**
+     *  Domain-specific workflow identifier
+     */
+    @Property
+    public String key;
+
+    /**
      * Workflow name.
      */
     @Property(meta = PropertyMetaTypes.NAME)
@@ -91,6 +97,15 @@ public class Workflow {
 
     public Workflow setId(String id) {
         this.id = id;
+        return this;
+    }
+
+    public String getKey() {
+        return key;
+    }
+
+    public Workflow setKey(String key) {
+        this.key = key;
         return this;
     }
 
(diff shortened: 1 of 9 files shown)


ricardozanini:
@hasys using the workflow ID is ideal since it's a required property defined by the [specification](https://github.com/serverlessworkflow/specification/blob/0.8.x/specification.md#workflow-definition-structure).

Although one can use the **key** instead, ID is not required in this case. I'd recommend using the same logic here. Name is an optional attribute just to a description for human eyes.

In a nutshell: id is required if key is not present and vice-versa.

romartin:
Thanks for the guidance @ricardozanini ! Good catch.

I am agree @hasys , let's use ID or key, depending on the one being present, it looks this way we'll properly follow the spec. On ther other hand, I think the key field is not yet in our client models, so maybe you have to add this string in the model classes, updated marshallers, and then do the logic that obtains one or another. Does it makes sense?

Apart from this ID/key change, everything looks good, good job! thanks @hasys

hasys:
Hello @ricardozanini, @romartin,

great catch! I changed the behavior to be closer to the spec:
1. Use `workflow.id` if it's not null or use `workflow.key` as UUID to generate diagram
2. Generate random UUID if both `workflow.id` and `workflow.key` are null to be able to show the diagram anyway.

I reported a [KOGITO-7622](https://issues.redhat.com/browse/KOGITO-7622) issue to provide some error notification if both `workflow.id` and `workflow.key` are empty.

Some notes from [KOGITO-7622](https://issues.redhat.com/browse/KOGITO-7622): there is already some sort of validation on VS Code plugin side which doesn't allow to open file if workflow missing both ID and KEY values see [KOGITO-7622](https://issues.redhat.com/browse/KOGITO-7622) for more details.

[Latest VSIX](https://drive.google.com/file/d/1av_z3b5RjqX0sNZ22N-V-lhOxGqzJ8Fo/view?usp=sharing)

Let me know, if it is as you expected. Thank you!

romartin (approved):
Thanks @hasys , looks good to me!

domhanak (approved):
LGTM, if you want you can add the exmaple that did not work to SWEditorSeleniumIT.java to verify the roundtrip of marshalling. I ll add it in my PR if you want to merge already.

hasys:
Hi @domhanak, initial issue was reproducible using unit tests, I think it should be enough. Thank you!

PR #1114, opened 2022-07-21
KOGITO-7579: Knative Workflow - Specify Kogito version

ljmotta:
## JIRAS
[KOGITO-7579: Specify Kogito version](https://issues.redhat.com/browse/KOGITO-7579)
[KOGITO-7485: macOS ARM build](https://issues.redhat.com/browse/KOGITO-7485)
[KOGITO-7573: Verify image name](https://issues.redhat.com/browse/KOGITO-7573)
[KOGITO-7607: Use Maven instead of Maven wrapper](https://issues.redhat.com/browse/KOGITO-7607)

## On this PR
This PR adds the possibility to tweak the Kogito and Quarkus versions. Also, it adds some general improvements:
- Removed the use of the `mvnw` and `mvnw.cmd`.
- Check image name during build.
- Build improvements:
  - Added macOS ARM cross-compilation.
  - Simplified staging and release build.

### Specify Kogito and Quarkus version
During the project creation, it's possible to pass a `--quarkus-version` or the `--kogito-version` flags in the `create` command to set a specific Quarkus version and a Kogito version for the project.

The `create` command now generates a `workflow.config.yml` file containing the versions, and those versions can be tweaked at any time.

#### New commands
- **version**
 Show the plugin version:
```bash
kn workflow version
```

- **config**
 Tweak the config file and apply those changes in the pom.xml:
```bash
kn workflow config --quarkus-version=2.10.0.Final --kogito-version=1.24.0.Final
```
Apply manual changes in the pom.xml:
```bash
kn workflow config --apply
```


ljmotta:
Hi @spolti ! I've added you as a reviewer as I think you can add insightful feedback.

@ricardozanini If you could take a look too, it would be great.

ricardozanini (commented):
Great work!

I might be missing something, but I'm missing the repository configuration. If the user sets the version like `quarkus-redhat-000x`, how will maven download it?

packages/kn-plugin-workflow/pkg/command/build.go
```diff
@@ -163,50 +168,71 @@ func runBuildConfig(cmd *cobra.Command) (cfg BuildConfig, err error) {
...
-	if err := common.RunCommand(
-		addExtension,
-		cfg.Verbose,
-		"build",
-		getAddExtensionFriendlyMessages(),
-	); err != nil {
-		return err
+		if err := common.RunExtensionCommand(
+			cfg.Verbose,
+			"quarkus:remove-extension",
```
spolti, line 210: is the log message correct? Adding extension while the command is removing it?

packages/kn-plugin-workflow/pkg/common/helper.go
```diff
@@ -65,6 +65,41 @@ func RunCommand(command *exec.Cmd, verbose bool, commandName string, friendlyMes
...
+
+	if err := RunExtensionCommand(verbose, "quarkus:add-extension", friendlyMessages, extensionsToAdd); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func RunExtensionCommand(verbose bool, extensionCommand string, friendlyMessages []string, extensions string) error {
+	command := exec.Command("mvn", extensionCommand, fmt.Sprintf("-Dextensions=%s", extensions))
```
spolti, line 95: how to check if the mvn is available in the path?

packages/kn-plugin-workflow/pkg/command/build.go
```diff
@@ -243,8 +269,17 @@ func runBuildImage(cfg BuildConfig) error {
 	return nil
 }
 
+func checkImageName(name string) (err error) {
+	matched, err := regexp.MatchString("[a-z]([-a-z0-9]*[a-z0-9])?", name)
+	if !matched {
+		fmt.Println("ERROR: Image name should match [a-z]([-a-z0-9]*[a-z0-9])?")
```
spolti: Do you mind to add a example about what this regex should match? Users might find it difficult to read regex.

packages/kn-plugin-workflow/pkg/command/config.go
```diff
@@ -0,0 +1,157 @@
...
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/kiegroup/kie-tools/kn-plugin-workflow/pkg/common"
+	"github.com/ory/viper"
+	"github.com/spf13/cobra"
+)
+
+type ConfigCmdConfig struct {
```
spolti, line 28: maybe rename to CmdConfig only?

packages/kn-plugin-workflow/pkg/command/config.go
```diff
@@ -0,0 +1,157 @@
...
+		common.GetVersionedExtension(common.KOGITO_ADDONS_QUARKUS_KNATIVE_EVENTING_EXTENSION, kogitoVersion),
+	); err != nil {
+		return err
+	}
+
+	fmt.Println("✅ Quarkus extensions were successfully updated in the pom.xml")
+	return nil
+}
+
+func runConfigCmdConfig(cmd *cobra.Command) (cfg ConfigCmdConfig, err error) {
```
spolti, line 128: maybe rename?

packages/kn-plugin-workflow/pkg/command/config.go
```diff
@@ -0,0 +1,157 @@
...
+
+		QuarkusVersion: viper.GetString("quarkus-version"),
+		KogitoVersion:  viper.GetString("kogito-version"),
+
+		Verbose: viper.GetBool("verbose"),
+	}
+	return
+}
+
+func getUpdateExtensionFriendlyMessages() []string {
```
spolti: isn't this duplicated?

packages/kn-plugin-workflow/pkg/command/config_workflow.go
```diff
@@ -0,0 +1,125 @@
...
+		fmt.Println("ERROR: writing the config yml file.")
+		return
+	}
+
+	fmt.Printf("Config file created on %s \n", configFilePath)
+	return
+}
+
+func ReadConfig(dependenciesVersion common.DependenciesVersion) (quarkusVersion string, kogitoVersion string, err error) {
+	configFile, err := ioutil.ReadFile(common.WORKFLOW_CONFIG_YML)
```
spolti, line 68: is it expected to be always at the same palce?

spolti (commented):
very nice, just a few things.

ederign:
" might be missing something, but I'm missing the repository configuration. If the user sets the version like quarkus-redhat-000x, how will maven download it?"

@ricardozanini , for now, to overcomplicate it, the user will have to have global settings configured. (This is probably the case if they use an RH distribution in their company).

packages/kn-plugin-workflow/pkg/command/build.go
ljmotta: Yes, it's. I'm going to add a comment in the code regarding this. I'm removing the "opposite" extension to add the chosen one. In the "Adding Jib Extension", I remove the Docker extension, and when "Adding Docker Extension", I remove the Jib extension. This procedure is part of the "Adding extension", and I'm hiding it from the user.

packages/kn-plugin-workflow/pkg/common/helper.go
ljmotta: Good catch. I should have checked the `mvn` and `Java` before the command starts with the `CheckJavaDependencies` function that is in the `check.go` file. I've forgotten to add this function call in the `config` command.

packages/kn-plugin-workflow/pkg/command/build.go
ljmotta: Sure

packages/kn-plugin-workflow/pkg/command/config.go
ljmotta: I've adopt the "<Command>CmdConfig" name pattern. So the other commands names are "BuildCmdConfig", "CreateCmdConfig", etc. `ConfigCmdConfig` it's not pretty, but it's what it is for this case.

packages/kn-plugin-workflow/pkg/command/config.go
ljmotta: No. I'm going to centralize all messages, I think it's going to be better to maintain it.

packages/kn-plugin-workflow/pkg/command/config_workflow.go
ljmotta: Yes

ljmotta:
@spolti Hi, I believe I've covered all raised topics, do you have any other additional comment?

@barboras7 Could you please take a look into this PR?

barboras7:
Hi @ljmotta,
sure. Sorry for the delay, Mondays are mostly meetings for me. I'm on it now!

packages/kn-plugin-workflow/cmd/main.go
```diff
@@ -20,12 +20,22 @@ import (
...
+
 func main() {
-	err := root.NewRootCommand().Execute()
-	if err != nil {
+	cfg := root.RootCmdConfig{
+		DependenciesVersion: common.DependenciesVersion{
+			QuarkusVersion: quarkusVersion,
+			KogitoVersion:  kogitoVersion,
+		},
+		Version: version,
```
barboras7: It might be difficult in the future to understand what `version` is. Would it be possible to use a more specific variable name?

packages/kn-plugin-workflow/pkg/command/build.go
```diff
@@ -163,50 +168,74 @@ func runBuildConfig(cmd *cobra.Command) (cfg BuildConfig, err error) {
...
+			cfg.Verbose,
+			common.GetFriendlyMessages("adding Quarkus extension"),
+			common.GetVersionedExtension(common.QUARKUS_CONTAINER_IMAGE_DOCKER, quarkusVersion),
+		); err != nil {
+			return err
+		}
 	}
 
-	fmt.Println("✅ Quarkus extension was successfully add to the project")
+	fmt.Println("✅ Quarkus extension was successfully add to the project pom.xml")
```
barboras7:
```suggestion
	fmt.Println("✅ Quarkus extension was successfully added to the project pom.xml")
```

packages/kn-plugin-workflow/pkg/command/config_workflow.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	config := Config{}
+	err = yaml.Unmarshal(configFile, &config)
+	if err != nil {
+		fmt.Printf("ERROR: unmarshaling the %s file.\n", common.WORKFLOW_CONFIG_YML)
+		return
+	}
+
+	// check if quarkus and kogito versions in the config file
```
barboras7:
```suggestion
	// check if quarkus and kogito versions are in the config file
```

barboras7 (commented):
Hi @ljmotta,
great work! I am leaving some minor comments below. 

Also, one thing I noticed when testing the plugin manually:
When I create a project using `kn workflow create --name myTesting`, I am able to do so. However, when I want to build this, I need to use `mytesting` instead of `myTesting` to succeed. Do you think you could add a validation for project creation to avoid such issues?

ljmotta:
@barboras7 Sure thing. I'm going to add the validation and make the suggested changes. Thanks for reviewing.

ljmotta:
@barboras7 Actually, after I've started making the changes, I've realized that adding this validation in the create command wouldn't be the ideal. The create command creates a local folder for the user and each user have their own organization. The build generates a container image, therefore has its own name limitations and patterns, and don't need to follow the folder/project name.

barboras7:
Thanks @ljmotta, I understand why you don't feel like the validation would be ideal. Maybe just a small suggestion - a log message at a warning level, when project name does not follow a correct image name pattern like `To build an image, you need to use an image name matching the following regex ....` or something similar. Anyway, that's just a suggestion. Great work! :partying_face:

barboras7 (approved):

spolti (approved):

tiagobento (approved):

PR #1116, opened 2022-07-21
Make `build-env` configurable per package

tiagobento:
All packages should now define their `env/index.js` file to be able to use `build-env`. This improves separation of concerns and avoids full builds being triggered when a new env var is added/changed/removed. Also, it's not possible for a package to access envs from other packages without making it explicit on `env/index.js`, which will require this package to list its dependent env as a dependency on its package.json.

Composing envs is easily done with the `composeEnv` function and only works when the packages have a strict dependency relationship.

Note that with this change, it's now possible to isolate and scope env configurations. If more than one package needs to consume a specific env variable, creating a new package to encapsulate this env configuration is encouraged.

---

Further improvements on separate PRs:
1. Define a naming pattern for our env vars
2. Improve error detection for unset variables
3. Creation of a script that generates a "report" documenting all the environment variables for a certain scope of the repository. (e.g. What are the possible env vars to configure when building `online-editor`?)


tiagobento:
Speaking with @ljmotta, we agreed that this PR can go before #1114 .

caponetto (approved):

PR #1121, opened 2022-07-25
KOGITO-7530: SWF Viewer can't visualize the workflow when transition is broken

hasys:
It worked fine when transition where missing, but can't visualize if transition was incorrect.

Now it can load the process and shows incorrect transition pointing to the top-left corner of the diagram.

Jira issue for further UX improvements: [KOGITO-7634](https://issues.redhat.com/browse/KOGITO-7634)

image

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/MessageCode.java
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/MessageCode.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2022 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.sw.marshall;
+
+// Message codes done in Interfaces instead of Enums for easier I18n later.
+public interface MessageCode {
+
+    String INVALID_TARGET_NAME = "Transition %s has invalid target.";
+
+}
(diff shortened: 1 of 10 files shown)


romartin:
Thanks @hasys 

As we previously discussed, could it be possbie to just not render the line, in case from/to are not properly set, and also show the error?

romartin (changes requested):
**Change request**: could it be possbie to just do not render the line, in case from/to are not properly set, and also show the error, as other marshalling errors? (ask @Josephblt if need for help here)

Thx!

hasys:
Hi @romartin, good points, applied:

now when transition has incorrect state name, no broken edges shown but also there is an error message on the top right corner, see below:

image

[VSIX for tests](https://drive.google.com/file/d/1QSCIn5juJy74mEl7GKjkgkllc7ZB9WVf/view?usp=sharing)

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/MessageCode.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.sw.marshall;
+
+// Message codes done in Interfaces instead of Enums for easier I18n later.
+public interface MessageCode {
+
+    String INVALID_TARGET_NAME = "Transition %s has invalid target.";
```
LuboTerifaj: %s is replaced with state name, not the transition. Would it make sense to you to change the message here?

LuboTerifaj (commented):
Hello @hasys ,

just one minor comment regarding message.

Otherwise it looks good to me.

Thank you!

hasys:
Hello @LuboTerifaj,

great catch, thank you! I changed the error message to:

`Transition from %s has invalid target.`

Does it make sense now?

LuboTerifaj (approved):
Hello @hasys , yes, it is better now. Thanks for the update.

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/MessageCode.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.sw.marshall;
+
+// Message codes done in Interfaces instead of Enums for easier I18n later.
+public interface MessageCode {
+
+    String INVALID_TARGET_NAME = "Transition from %s has invalid target.";
```
romartin, line 22: I expect in some early future to fix all our literals and move them to i18n. we have some missing. Although nothing lbokcing this PR, its' a separate task we have to as part of the backlog.

romartin (approved):
Nice work @hasys !

PR #1126, opened 2022-07-27
KOGITO-7663: Fix inconsistency in filename template description and default value

zdrapela:
Jira: https://issues.redhat.com/projects/KOGITO/issues/KOGITO-7663


tiagobento:
This pattern is inherited from the DMN and BPMN extensions, that made the generated SVG compatible with Business Central. 

I'm +1 to change it for SWF, though.

cc @ederign

ederign:
Hi @zdrapela this makes sense, but before merging this,  we need to double check if this is not somehow hardcoded on quarkus dev UI. @Sara4994 could you please confirm?

Sara4994:
@ederign just checked quarkus dev UI, it is not hardcoded.

domhanak:
@zdrapela confilicts, could you please report a follow-up JIRA to add some sort of integration tests into VSCode suite for the SVG gen?

domhanak:
@zdrapela now it looks like a formatting issue, please check

zdrapela:
@domhanak It was some issue with `prettier`, but now all checks are OK. 
Now I found out, that [README](https://github.com/kiegroup/kie-tools/blob/main/packages/vscode-extension-serverless-workflow-editor/README.md) mentions `${fileBasenameNoExtension}-svg.svg` as a default value (instead of current default `${fileBasename}.svg`). BPMN and DMN also use `${fileBasenameNoExtension}-svg.svg`, so it might be better to go that way.

domhanak (approved):
LGTM.

tiagobento:
Deferring to @ederign.

ederign:
@paulovmr would mind to take a look ?

paulovmr:
Hi @zdrapela , the `kogito-addons-quarkus-process-svg` extension looks at a file with the same name as the workflow definition to make the SVG available through a REST service, so we should use `${fileBasenameNoExtension}.svg` by default. If you could please change that on the README as well. Thanks!

paulovmr (changes requested):
Hi @zdrapela , the `kogito-addons-quarkus-process-svg` extension [looks](https://github.com/kiegroup/kogito-runtimes/blob/main/addons/common/process-svg/src/main/java/org/kie/kogito/svg/AbstractProcessSvgService.java#L67) at a file with the same name as the workflow definition to make the SVG available through a REST service, so we should use `${fileBasenameNoExtension}.svg` by default. If you could please change that on the README as well. Thanks!

zdrapela:
Hi @paulovmr, thank you for explaining. I changed that to `${fileBasenameNoExtension}.svg` in the default value (same as it was before this PR), in the description and in README.

paulovmr (approved):
Thanks, @zdrapela !

domhanak:
I think we can ignore the sonar check. It was broken works now and does not make sense for this PR.

yesamer (approved):

ederign (approved):

PR #1133, opened 2022-07-28
Ignoring non-source paths on CI

tiagobento:
This PR creates a mechanism to exclude non-source files and paths from the changes detection on our CI workflows.

The goal here is to encourage people to edit `README.md` files, update icons, images, and other non-source-related things that improve our code. People doing these kind of contributions should not be penalized with a useless build that takes hours to complete.

Here's is how it works:

- If only non-source files change on a PR, neither CIs will run, and we'll save a lot of time and resources.

- If source files change exclusively inside `packages` or `examples`, `CI :: Monorepo` will run, which will build only the affected packages.

- If at least one source file change at the root of the monorepo, only `CI :: Monorepo (full)` will run.

### DIFF
--- a/.github/supporting-files/ci/non-source-paths.txt
+++ b/.github/supporting-files/ci/non-source-paths.txt
@@ -1,0 +1,12 @@
+*.md
+*.icns
+*.ico
+*.gif
+*.jpg
+*.jpeg
+*.png
+*.svg
+docs
+gifs
+repo
+.github/CODEOWNERS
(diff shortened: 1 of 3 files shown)


ederign (approved):

tiagobento:
Ok, after some battles with bash, I got it :)

This is working as expected now.

caponetto (approved):

thiagoelg (approved):

PR #1135, opened 2022-07-28
Download Maven dependencies on bootstrap

tiagobento:
Bootstrap will take a little longer, but the builds will fail faster if something is wrong. The total build time won't change.


tiagobento:
Hmm... looks like `mvn dependency:resolve` is not able to detect local dependencies that are not built yet? I can't help but think that this is maybe caused by a wrong configuration from our side. @yesamer Would you know?

yesamer (commented):
@tiagobento For your goal, I think it's better to use `mvn dependency:go-offline`. With this command, you will download only the external dependencies required to build the project even if you are offline. I tested it in `stunned-editor` and works well. The only prerequisite is to use a recent version of `maven-dependency-plugin`. 
`mvn dependency:resolve` tries to resolve internal modules dependency too, that are not available in external repo or previously completed and installed.

Apologies if I directly modified your PR, my first intent was to suggest changes. However, it was useful to see the effect of my changes

EDIT: In `dashbuider` there's this dependency `io.netty:netty-transport-native-unix-common:jar:${os.detected.name}-${os.detected.arch}:4.1.59.Final` which is not able to find - investigating

yesamer:
Things are getting interesting https://github.com/netty/netty/issues/11272

tiagobento:
@yesamer Hey! Wow, thanks for really going for it :D No worries about committing directly here, actually, I think that's great 😄 

And about the error on the netty dependency.. maybe we don't need that anymore? cc @jesuino 

Also, an interesting question... how does it work currently? 🤔

ederign:
@tiagobento where the netty is used?

yesamer:
@tiagobento As I wrote you in chat, `netty` is not actually used according to `mvn dependency:analyzer`
Imported but not used.

yesamer:
Found a usage `import io.netty.util.internal.shaded.org.jctools.queues.MessagePassingQueue.Consumer` @ `DashboardExporter` is maybe required for that?

caponetto:
> Found a usage `import io.netty.util.internal.shaded.org.jctools.queues.MessagePassingQueue.Consumer` @ `DashboardExporter` is maybe required for that?

Maybe this Consumer should be `java.util.function.Consumer`.

ederign:
Yep probably. @jesuino can you please fix this? ?^

tiagobento:
Ok, so it looks like everything is working great ok Windows, but there's a weird failure on Ubuntu and macOS... very odd scenario 😝

ederign (approved):
LGTM. +1 to merge when build is green

tiagobento:
@ederign Thx!! @yesamer and I spoke today about this PR, and we agreed that we need to investigate a little bit more to see why it failed downloading dependencies for the Java Completion Plugin.. although it passed now, it doesn't seem stable enough for us to merge it.

We're going to try and find the root cause of those failures first.

tiagobento:
This is not stable enough to go in. Closing for now. Thanks @yesamer for all the help investigating this!

PR #1146, opened 2022-08-04
KOGITO-7471: Auto-layout for SWF

danielzhe:
This is the first version of the autolayout for SWF.
More improvements can and will be made in the future, especially in terms of usability when we had the layout information saved. So we can compare what was changed and don't perform a full layout in every refresh of the canvas.

image

image


tiagobento:
Hey @danielzhe, amazing work!! From the screen shots I can tell this is going very well :)

Quick question.. is the ELK-based mechanism for auto-layout going to be removed as part of this PR as well?

romartin:
Thanks @danielzhe !

Can you please rebase this branch, it is actually in conflicts.

Also notice delegating to @handreyrc the engineering review, as I'll be off during next weeks.

NOTE: IMO this should be only merging when NO RISKS are found, we should properly test this and ensure we don't break any example, etc. This way I understand we still have time for the properly verification and fixes, if necessay, and try to include this for DP2. Otherwise, if reaching deadlines but having still open issues here, I would not vote for merging. I'll be off so letting to you this decision during following weeks!
@handreyrc @tiagobento @ederign @LuboTerifaj ^^


Thanks!

romartin:
@danielzhe @handreyrc please remember to **drop ELK** from the bundle (JStypes, js files, etc)

romartin:
@handreyrc please as part of the review for this PR, can you please obtain some **performance metrics** for the autolayout? 

Notice we already have the metrics for the autolayout based on ELK [1], I expect this implementation faster, also as dropping the ELK bundle, there is no need for the 300ms the bundle takes to be parsed & evaluated by the browser. So I really expect better performance for this feature... let's see :) 

[1] https://issues.redhat.com/browse/KOGITO-7041?focusedCommentId=20572270&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20572270

danielzhe:
Rebase and ELK removed!

handreyrc:
Hey @danielzhe,

I tested this PR and I have some improvement points to suggest:

1- Transitions are being connected always to the center magnet and it causes trasitions to overlap some states when the workflow is too wide. It may look better if connections were made to top/bottom magnets of the states only. A setting could be implemented for that purpose.
2- The layout is Bottom-up. In this approach it is more likely that most of the states are concentrated in the bottom of the diagram also it makes the layer hierarchy unclear and a bit harder to follow. For example, the states "Compensate", "HandleError1" and "HandleError2" could be in the same layer as "Wait for Events" state. See below:
image
3- If a node is diconnected it is moved to the bottom of the diagram and the transitions may overlap other states and crosspath with other trasitions. Maybe changing the aproach to top-down and force disconected nodes to the right of the diagram may look better from the user point of view. It is how it works with the current ELK implementation. See below:
image

WDYT?

**Performance**
The performance in comparison to ELK is better. Actualy We have a great improvement here. It is fast!
image

By the way excelent job! 
Thanks

danielzhe:
Thank you for your review, @handreyrc ! I gonna do some changes based on your suggestions and ping you back when it is done!

danielzhe:
Hi, @handreyrc 
I did some changes as you requested. I think the result is much better now. Some points to consider:

1. The layout gives a false impression that it is bottom-up but it is not. This happens because of the way algorithm (Longest Path algorithm) that we are using to position nodes inside each layer. Consider the diagram that we are using. The algorithm works by calculating the cost to go from Start node to each one of the nodes. So, for example, the cost to go from Start to  "Handle Error 3" it is 6 (6 nodes that we have to cross to get there). And that's why the "Handle Error 3" node it is in layer 6. Probably the ELK is using another more complex approach to do the layering. I tried another algorithm but the result was worst than we have now.

2. I tried to force the disconnected nodes to go to the right but the result wasn't good, causing edges crossing:
Move right

3. The "End" node is always alone, so I think the UX is better now.

Let me know what you think.

LuboTerifaj (commented):
Hey @danielzhe,

the VSCode extension used either from https://github.com/kiegroup/kie-tools/suites/8030930123/artifacts/345128769 or built locally doesn't work.

If I open any serverless workflow file, the diagram doesn't load. There is the following error in console: `Error loading webview: Error: Could not register service workers: InvalidStateError: Failed to register a ServiceWorker: The document is in an invalid state..`.

Does it work on your machine?

Thanks!

caponetto:
Hey @LuboTerifaj, I have been facing this issue every once in a while but I don't think this is related to our extension.
Something that helps me is to completely close VS Code when this issue happens (ref https://stackoverflow.com/a/69316961).

LuboTerifaj:
Hello @caponetto , @danielzhe , the extension still doesn't work. If I open any serverless workflow file, the diagram cannot be opened. There is the error in console log: `Error opening 'fileName.sw.json'.` `in NotificationsAlerts.ts:42`.

caponetto:
> Hello @caponetto , @danielzhe , the extension still doesn't work. If I open any serverless workflow file, the diagram cannot be opened. There is the error in console log: `Error opening 'fileName.sw.json'.` `in NotificationsAlerts.ts:42`.

Yep, I can also reproduce this issue here with the latest artifact from CI.

danielzhe:
It's fixed now, @LuboTerifaj . Ready for review.

romartin:
Hey  guys, 

Sorry for late responses. I am starting the review process, will provide feedback during the following days.

 @danielzhe is there any vsix being uplodaded for the manual testing of this stuff?

domhanak:
@romartin Artifacts are in `Checks -> Full build -> Artifacts` OR 
https://github.com/kiegroup/kie-tools/actions/runs/3024908781#artifacts

domhanak:
@danielzhe Hi, could you please check if this test passes locally:
`SWEditorSeleniumIT.testProcessTransactionsExample:249->testExample:256->waitCanvasPanel:281 » Timeout`
The issue we observe in builds is that it does not load the canvasPanel. Asking if we shold investigate unstable test or not. Thanks

romartin:
Hey  @domhanak 

Thanks for the pointer! But I still can't find the `vsix` in the `Linux_build-artifacts` :( Where do you get the vsix from?? Thx!

domhanak:
@romartin ah I see, the build won't reach the package as this stunner package fails to build.

LuboTerifaj:
Hello @danielzhe , what's the status of this PR? Is it ready for testing?

romartin:
@danielzhe can you please rebase ? thx

romartin:
@danielzhe as already discussed in today's call, for now 2 things to do from your side please:

1. Rebase from main upstream
2. Bug Fix -> Assume there could not exist an initial start state (if not defined in the json).This way, multiple states can have no incoming transition, if no other incoming connectors from other states, and because they're not defined as start states.

Finally, once both above points are done, it's supposed the builds and tests will work, and so the vsix will be automatically generated. So we will be able to go ahead with another review iteration

Thanks!

danielzhe:
@romartin @LuboTerifaj Rebased.

romartin:
Hey  @danielzhe !

Thanks a lot for this work, looks and performs really good!

I had been testing and found a few edge cases, that would be nice, at least, considering how hard can be to tackle, and ikf necessary, fixing as well. 

1. When the wf's `start` attribute is pointing to a non existing node (for instance when a typo there), the start shape is on top, just floating, with no outgoing lines. Could we just NOT display the start shape in that case? The idea is: if the start node has no lines at all, no need to show it.
wrong-start

2. Same for end shape -> In case no states are end, the end shape just floats there... same as start, could we just not display the end shape as well?
no-end

3. When a state has some reference to a previous state, a line back to the state is created, but it is attached to the bottom magnet (of the source state), and also it not really looking nice IMO. Could we improve that in some way?  I would recommend using the top magnet, and maybe doing an orthogonal line back to the state, instead of  a line with just 1 point as it is now. OR maybe also a line with more points would work too
lines-back-to-previous

4. End shape when having lots of incoming lines -> This is already reported at https://issues.redhat.com/browse/KOGITO-7197 . Can we fix this in some way too? IMO this doesn't block the UI, but would be nice trying to find some quick win that makes this look better. We can ask UX (Missy) as well. WDYT?
end-incoming-lines

Will continue with more manual tests and provide any other feedback if so

Thanks again for this good work!! :+1:

LuboTerifaj (changes requested):
Hello @danielzhe ,

I have found the following issues:

- If there is odd number of states below Start node, the Start state is not in the center.  

- If you specify non-existing state as a start node, the start is shown in top left corner.

- If end node is not specified, or if it is specified, but it's not boolean, or it is false, it is shown in top left corner.

- All states without specified transition or end are shown in top left corner.

- The states with the transition to the first level of states are shown next to the Start node and the Start node is not in the center.

- The connections intersect the other nodes, or the connections overlap each other.

- The states connected with transition are sometimes shown in the same level, but they shouldn't.

- If there are two states with the same name, the diagram gets broken.


See some of the issues mentioned above in wrongly handled diagrams from Examples (https://github.com/serverlessworkflow/specification/tree/main/examples):
- book-lending
- car-vitals-checks-parent
- car-vitals-checks-sub-workflow
- continuing-as-a-new-execution
- filling-a-glass-of-water
- monitor-job
- reusing-function-and-event-definitions
- send-cloudEvent-on-workflow-completion

Thanks!

romartin:
Thanks @LuboTerifaj  for the review as well!

@danielzhe @LuboTerifaj I think guys that we should now take a look at the reported issues in this PR, and have some discussion about what's feasible, what can wait, or what still needs work. This way, also as @LuboTerifaj will be on PTO during next weeks, I have just proposed a quick call for tomorrow. Also added @domhanak in the call, so he can help on the review while Lubo will be off.

LuboTerifaj:
Hello @danielzhe , 

regarding the issues I commented above, there is one issue that might not be clear:

- The states connected with transition are sometimes shown in the same level, but they shouldn't.
  - The issue happens if the last state doesn't have end property. For example:
same-level

danielzhe:
VSIX: https://github.com/kiegroup/kie-tools/actions/runs/3185392984#artifacts

Changes applied! Thank you for your review, guys! Here are some comments about the issues you guys reported:

- The issues 1, 2, and 3 reported by @romartin  are done.
- Issue 4 requires more time and discussion. I think there are other priorities in auto layout, but we can do it in the future.

**Done:**
[X] If you specify non-existing state as a start node, the start is shown in top left corner.
[X] If end node is not specified, or if it is specified, but it's not boolean, or it is false, it is shown in top left corner.

**Those issues are doable but require a bit more time:**
[ ] All states without specified transition or end are shown in top left corner.

**Need clarification**
I don't understand those issues, but by description, I **think it can be fixed with the previous issue and the last one in this list**:
[ ] The states with the transition to the first level of states are shown next to the Start node and the Start node is not in the center.
[ ] If there is odd number of states below Start node, the Start state is not in the center.


**This issue is not related to this PR:**
[-] If there are two states with the same name, the diagram gets broken

**Not done:**
This issue requires a lot of time and effort because we need to implement another step in the auto layout algorithm specific to handle edges:
[-] The connections intersect the other nodes, or the connections overlap each other.

I tried to fix this issue with a "easy" fix but the result wasn't good. There is a solution: we handle each isolated group of nodes as a single diagram, apply the layout, and put the results on the screen. This solution can also fix the issue "All states without specified transition or end are shown in top left corner."
[-] The states connected with transition are sometimes shown in the same level, but they shouldn't

@romartin @LuboTerifaj

romartin:
Hello  @LuboTerifaj   @danielzhe 

Thanks for your work and feedback!

Given the [above status summary](https://github.com/kiegroup/kie-tools/pull/1146#issuecomment-1267760649) I suggest to take these actions, and just go ahead with this PR in its actual state:

**1. Postpone the reported improvements into another ticket / PR**

The idea would be to report the following points as improvements to do for a next iteration, in a separate ticket (https://issues.redhat.com/browse/KOGITO-8100):

- [KOGITO-7197](https://issues.redhat.com/browse/KOGITO-7197) - Connectors endings
- All states without specified transition or end are shown in top left corner.
- The states with the transition to the first level of states are shown next to the Start node and the Start node is not in the center.
- If there is odd number of states below Start node, the Start state is not in the center.
- The connections intersect the other nodes, or the connections overlap each other.
- The states connected with transition are sometimes shown in the same level, but they shouldn't

This way, we can go ahead with this PR, although with a few improvements pending, but it will help on reducing the amount of codebase, dropping the use of ELK and improving the actual performance as well. Also once merged, we'll be able to gather more feedback from third parties and so also continue with the improvements.

**2.- Reported new issue** -> https://issues.redhat.com/browse/KOGITO-8101 (Not related to this PR)

This way guys, as mentioned above, I propose to go ahead with this PR as it is, and move to the work on the improvements in a separate PR. @LuboTerifaj @ederign WDYT on this strategy? I am just trying to split the work into iterations and deliver this incrementally. I see this feature as more long term task, by doing iterations about gathering feedback and fixing properly the different use cases. Also drop as soon as possible the use of ELK. 

Thanks!

ederign:
+1 for this strategy @romartin . Delivering iterative is always better and gives us smaller PR's. Hopefully, we can have all those points fixed before the next release!

romartin:
Hey @danielzhe 

Thanks for the updates! I have been testing the vsix form the generated artifacts in this PR, and the start / end node are now not showing, if not present, but I still see **issues with magnets**.

See both cases:

**Issue1 - Line back to previous layer**

Screenshot from 2022-10-13 23-35-18

The magnet for for "Wait for completion" is correct, the below one. But the magnet used in the "DetermineCompletion" is still the bottom one, and it should be the top one. This way the line will not overlap the shape itself.

**Issue2 - Line back in same layer**

Screenshot from 2022-10-13 23-37-52

The magnet for for "State2" is correct, the left one. But the magnet used in the "State1" is still the bottom one, and it should be the right one. This way the line will not overlap the shape itself.

Rather than this, I have properly updated the improvements identified for a further task, see [KOGITO-8100](https://issues.redhat.com/browse/KOGITO-8100)

Can you please fix both issues? Then I'll be ok to approve..

@LuboTerifaj WDYT?

Thanks guys!

danielzhe:
@romartin Changes applied!

romartin (approved):
Thanks @danielzhe for the latest fixes! I works nicely now! :D 
Approving, all good to me. 
Let's go for as next step for topics at [KOGITO-8100](https://issues.redhat.com/browse/KOGITO-8100) :+1:

romartin:
@LuboTerifaj review on your side now! :) Thanks!

LuboTerifaj (approved):
Hello @danielzhe and @romartin ,

I have retested the PR. Some of the issues have been fixed or I couldn't have reproduced any more, or are present also on main.

Here are the issues which are still present:

- All states without specified transition or end are shown in top left corner.
- The states with the transition to the first level of states are shown next to the Start node and the Start node is not in the center.
- The connections intersect the other nodes, or the connections overlap each other.
- The states connected with transition are sometimes shown in the same level, but they shouldn't.

I agree with fixing them in separate jira, specifically in https://issues.redhat.com/browse/KOGITO-8100

I am about to attach sw.json reproducers to the jira.

Approving the PR.

Thank you!

romartin:
Thanks @LuboTerifaj !

Please @kiegroup/gatekeepers **DO NOT merge** yet, although approving. Waiting to merge just after the release. Next monday will be fine to merge.

PR #1151, opened 2022-08-05
Improve `EnvelopeMapping`

caponetto:
There is confusion when setting `resourcesPathPrefix` and `envelopePath` for `EnvelopeMapping`s on each `extension.ts` (VS Code channel) file. Although this confusion is "undone" in `VsCodeKieEditorControllerFactory` and no issues are seen, this is still incorrect.

To avoid such confusion, I'm adding an `args` object to the `EnvelopeMapping` constructor so that auto-completion can guide with named parameters when creating `EnvelopeMapping`s.

I also fixed the `@kie-tools-examples/webapp` during the validation of these changes.
We should probably revisit [1] because this is another potential point for confusion.

[1] https://github.com/kiegroup/kie-tools/blob/main/packages/editor/src/embedded/embedded/EmbeddedEditor.tsx#L89


tiagobento (approved):
@caponetto Great refactor! 😍

tiagobento:
cc @yesamer

ederign (approved):
LGTM

yesamer (approved):
@caponetto Thank you!

PR #1152, opened 2022-08-05
Finish removing Lerna and allow sparse checkouts

tiagobento:
Since we migrated to `pnpm`, we stopped using Lerna for our day to day development, but for releases, Lerna was still necessary for updating versions and checking tags against the monorepo version.

Now, the `update_version.js` script will depend on pnpm's internal packages to list the workspace packages and, more than that, by using the `workspace:` protocol with `workspace:*` versions, we don't need to update references inside the monorepo anymore.

To make this work when publishing to NPM, we need to use `pnpm publish` instead of `npm publish`.

Other scripts were depending on `@lerna/project` to list the workspace packages as well, so they were updated too.

---

In additional to that, a new script called `list_packages_dependencies.js` was created. This script operates together with `generate_packages_graph.js` to read a serialized dependency graph in a traverse-able format.

All this enables us to do sparse checkouts of the monorepo. For example:

```bash
ORG="tiagobento"
BRANCH="update-version-without-lerna"
PKGS="@kie-tools-core/envelope @kie-tools/kn-plugin-workflow"
bash <(curl -s https://raw.githubusercontent.com/$ORG/kie-tools/$BRANCH/scripts/sparse_checkout.sh) $ORG $BRANCH $PKGS
```

This is an experiment and can or cannot go somewhere :)

---

> P.S.: TODOs will be addressed on my next PR.


caponetto (approved):

thiagoelg (approved):
Sparse checkouts sound awesome, great work!

PR #1154, opened 2022-08-05
KOGITO-7719: Dashbuilder: Bring Victory Charts as supported external component

jesuino:


ederign:
@jesuino thanks for the quick PR. I'll look at this tomorrow!

package.json
```diff
@@ -79,6 +79,7 @@
     "ts-loader": "^9.1.2",
     "typescript": "^4.4.2",
     "url-loader": "^4.1.1",
+    "victory-zoom-container": "^35.11.4",
```
tiagobento: Please add this dependency to the package where it needs to be. We're not adding package-specific dependencies to the root package.json anymore.

packages/dashbuilder-component-victory-charts/src/VictoryChartComponent.tsx
```diff
@@ -42,6 +42,8 @@ export function VictoryChartComponent(props: Props) {
       setVictoryChartProps((props: VictoryChartProps) => {
         return {
           ...props,
+          width: +params.get("width") || 600,
+          height: +params.get("height") || 400,
```
tiagobento: What does `+` do here?

repo/graph.dot
```diff
@@ -95,6 +95,7 @@ digraph G {
   "vscode-extension-backend" [ color = "black", fontcolor = "black", style = "dashed, rounded" ];
   "vscode-extension-bpmn-editor" [ color = "blue", fontcolor = "blue", style = "rounded" ];
   "@kie-tools/vscode-java-code-completion-extension-plugin" [ color = "blue", fontcolor = "blue", style = "rounded" ];
+  "vscode-extension-dashbuilder" [ color = "blue", fontcolor = "blue", style = "rounded" ];
```
tiagobento: Does this package exist?

tiagobento (changes requested):
Just left some comments inline. @jesuino The best way for you to fix the graph.json and graph.dot files are simply by running a `pnpm bootstrap`.

packages/dashbuilder-component-victory-charts/src/VictoryChartComponent.tsx
jesuino: it casts the result to a number

tiagobento: What happens if there is no string to cast? `+undefined` becomes `NaN` and `+""` becomes `0`. I think a better way to do this would be `+(params.get("width") || 600)`, WDYT?

tiagobento: Also note that `""`, `0`, `NaN` are all falsy values, so in this case, if you set `width` to 0, it will actually be 600.

repo/graph.dot
jesuino: thanks! It was a leftover, it is removed now

packages/dashbuilder-component-victory-charts/src/VictoryChartComponent.tsx
jesuino: great suggestion, thanks! Just made the changes accordingly

tiagobento (approved):

ederign (approved):

paulovmr (approved):

PR #1156, opened 2022-08-08
KOGITO-7699: [SWF Editor] LS auto-completion for functionRef arguments does not work

fantonangeli:
Fixes: https://issues.redhat.com/browse/KOGITO-7699

Some changes to the Language Service broke auto-completion for functionRef arguments.

This test [1] is passing when running against `0.21.1` tag (our latest release atm) but fails on `main` branch.

**Cause**: After `0.21.1` the use of `jsonc.getLocation()` has been removed in favour of `matchNodeWithLocation()` that uses `findNodesAtLocation()`, and it's based on the AST. 
`jsonc.getLocation()` scans the file's text, not the AST, but is not compatible with YAML.  

**Solution**: Changed the trigger for the autocomplete from `"arguments":🎯` to `"arguments":{🎯}` 
In the first case, the AST library (jsonc) can't select the value of `"arguments"` because it is not defined by the user and can be string, object or array. In the second case, I can select the value that is an object.
The same limitation is with the YAML format as well.

KOGITO-7699 JSON


packages/serverless-workflow-language-service/src/channel/SwfLanguageService.ts
```diff
@@ -81,6 +82,7 @@ export class SwfLanguageService {
     cursorPosition: Position;
     cursorWordRange: Range;
     rootNode: SwfLsNode | undefined;
+    completionTranslator: (completion: any) => string;
```
tiagobento: I think this may be worth an interface and a full object. The `any` type of the first parameter together with the boolean type of the second make for a very 'open' function, which I tend to not find very friendly for new people reading the code.

domhanak:
@tomasdavidorg fyi

caponetto:
Re-triggered CI to include https://github.com/kiegroup/kie-tools/pull/1159.

packages/serverless-workflow-language-service/src/channel/SwfLanguageService.ts
fantonangeli: Improved in commit:
https://github.com/kiegroup/kie-tools/pull/1156/commits/2c05832f0a49c72ad5fd4ec3a57ca9cdfe401332

tiagobento (approved):

PR #1157, opened 2022-08-08
Use Maven CI Friendly Versions

tiagobento:
Maven packages used to have their versions hard-coded on multiple pom.xml files, leading to special treatment when updating their versions for a release.

As part of an effort to make each package more independent and reduce the need for top-level scripts with individual package references, Maven packages now only declare their version exclusively at the `package.json` and use that version through the `${revision}` property.

This strategy is better described at the Maven CI Friendly Versions article [1].

An implementation detail of this PR is the new `maven-config-setup-helper` package, with a simple script that will append to an existing `.mvn/maven.config` file. This is important because on our CI we configure some extra Maven args and we don't want to override them during the build. More on `maven-config-setup-helper`'s README.

---

This PR also aims to provide an insight on how Maven packages can be internally referenced inside the Monorepo. By declaring a dependency on `package.json`, a Maven package A willing to consume another Maven package B can inherit its `env` and use B's version on its `install.js` script. This will then parameterize B's version inside A, making A always use whatever version B declares on its `package.json`.

---

[1] https://maven.apache.org/maven-ci-friendly.html


caponetto (approved):
🚀

PR #1159, opened 2022-08-08
Bring back JSON Schema to SWF Editor VS Code Extension

tiagobento:
For the `Serverless Workflow (JSON)` and `Serverless Workflow (YAML)` languages, SWF LS JSON Schema diagnostics should be enabled, but for the `JSON` and `YAML` languages, we use VS Code's built-in JSON Schema capabilities (though the YAML extension for the `YAML` language), it shouldn't, or otherwise it would result in duplication of validation messages.

`JSON` and `YAML` continue to be the preferred languages, but the `Serverless Workflow (JSON)` and `Serverless Workflow (YAML)` languages keep us aware of the capabilities that are exclusive to the SWF LS.

After #1102, `getLsForDiagramEditor` won't be necessary anymore, as we're gonna have a better separation between Text and Diagram editors for SWF.

Also note that although the VS Code Extension contribution point is called `jsonValidation`, it enables full JSON Schema capabilities on the built-in VS Code text editor, such as auto-complete, outlines, etc etc, that's why it's important that we keep it.

### DIFF
--- a/packages/serverless-logic-sandbox/src/editor/api/SandboxSwfJsonLanguageService.ts
+++ b/packages/serverless-logic-sandbox/src/editor/api/SandboxSwfJsonLanguageService.ts
@@ -35,6 +35,7 @@ export class SandboxSwfJsonLanguageService extends SwfJsonLanguageService {
       },
       config: {
         shouldDisplayServiceRegistriesIntegration: async () => false,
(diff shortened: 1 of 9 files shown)


caponetto (approved):
Thank you, @tiagobento!

ederign (approved):
/cc @fantonangeli

PR #1161, opened 2022-08-08
Rename from Sandbox to Serverless Logic Web Tools

ederign:
Other projects use the Sandbox name at Red Hat, so I've renamed our tools to 'Serverless Logic Web Tools' to avoid confusion.

Later, we should rename packages and, more importantly, the extended services binary.


ederign:
@thiagoelg FYI for the docs.

caponetto (approved):

tiagobento (approved):

PR #1174, opened 2022-08-10
Improve required preinstalled CLI commands checks and add `build-env` reports

tiagobento:
Now, using the `pnpm` filters we already have for `bootstrap`, we're able to check which CLI commands the filtered packages require to be ran. I decided not do break the `bootstrap` command if one of those commands is not present yet, but that might be a good alternative in the future.

Also, during `bootstrap`, now there's a new step that prints the current environment. It also works with pnpm filters. This is useful information for us to debug in case something doesn't work as expected.

Also in this PR:
- New checks that will fail any `build-env` evaluation if there are duplicated env vars or root properties.
- Bootstrap now always passes the `-F .` filter to `pnpm`, allowing users not to write it.


domhanak:
@tiagobento Is there a JIRA please?

caponetto (approved):

PR #1177, opened 2022-08-11
KOGITO-7731:  [SWF Editor] LS auto-completion for full functionRef object does not work

fantonangeli:
Fixes: https://issues.redhat.com/browse/KOGITO-7731

Please note: In the jira I wrote a comment with an issue/limit we have to keep in mind with YAML, for the next tasks.

KOGITO-7731


caponetto:
I went through all supported auto-completions and found 2 issues:

`"operation":🎯`
Expected: operations being auto-completed
Actual: the entire function is being auto-completed
Note: `"operation":"🎯"` works fine.

`"refName":🎯`
Expected: refNames being auto-completed
Actual: No suggestions
Note: `"refName":"🎯"` works fine.

These issues are not observed in the latest release (`0.21.1`).
I'll take a look at these issues and update this PR accordingly.

packages/serverless-workflow-language-service/src/channel/findNodesAtLocation.ts
```diff
@@ -49,8 +55,9 @@ export function findNodesAtLocation(root: jsonc.Node | undefined, path: any): js
         }
 
         for (const prop of n.children) {
-          if (Array.isArray(prop.children) && prop.children[0].value === segment && prop.children.length === 2) {
-            return [prop.children[1]];
+          if (Array.isArray(prop.children) && prop.children[0].value === segment) {
+            // if prop.children[1] doesn't exist, return prop.children[0].parent to have the same value of findNodeAtOffset()
+            return [prop.children[1] || prop.children[0].parent];
```
tiagobento, line 60: This method was heavily inspired by `jsonc`'s implementation. I don't know if we should deviate too much from it... please take a look at how jsonc does it so we can make sure that we're not creating a method that is specific to our usecases alone.

packages/serverless-workflow-language-service/src/channel/matchNodeWithLocation.ts
```diff
@@ -35,13 +36,14 @@ export function matchNodeWithLocation(
   }
 
   const nodesAtLocation = findNodesAtLocation(root, path);
+  const nodeToMatch = nodeUpUntilType(node, "object");
+  const starSelector = path[path.length - 1] === "*";
```
tiagobento, line 40: What happens if we have a `*` in the middle of the JSON Path?

fantonangeli: That case is already managed in line 43 and tested in `packages/serverless-workflow-language-service/tests/matchNodeWithLocation.test.ts`

packages/serverless-workflow-language-service/src/channel/findNodesAtLocation.ts
fantonangeli: This is to manage the case that we have a property without a value because the user is writing the json. 
This is not the usual case for `jsonc` that always parses a complete and valid json.

tiagobento: I think `jsonc` is one of the most flexible parsers there is.. it can go very far without breaking even with missing brackets and commas. I'm not saying I'm against this change, it's just that I suspect there are some use-cases for this to behave the way it does, and if we start bending and stretching it, we might be caught off guard in the future.

caponetto:
Okay, so I've found a couple more issues during my debug session mainly caused by not using `jsonc` functions anymore. I transformed these scenarios into unit tests for LS JSON to improve our coverage and aid in future refactorings.

Since autocompletion for YAML is not enabled yet and needs more improvements, I put back the usage of `jsonc` for LS JSON as it was before so that the upcoming release is not compromised. LS YAML still uses the new implementation though.

tiagobento (approved):

PR #1186, opened 2022-08-17
KOGITO-7774: Stunner editor packages depend on old hibernate-validator

domhanak:
Upgrades to latest final version - 7.0.4.Final


domhanak:
Also tried `6.0.19.Final` and there are failures so not sure if we should proceed with this, I noticed the snyk-report and tried a quick fix PR, but it seems we need to do some work for this upgrade.

Upgrade to `6.0.19` looks fixable thought.

domhanak:
Closing for now. The current result is that upgrade can only happen if GWT stops using `<version.javax.validation>1.0.0.GA</version.javax.validation>`

PR #1205, opened 2022-08-26
Knative Workflow - Implements Plugin interface

ljmotta:
## On this PR
Implements the Knative [Plugin](https://github.com/knative/client/blob/main/pkg/kn/plugin/manager.go) interface to be able to use the [plugin inlining](https://github.com/knative/client/tree/main/docs/plugins#plugin-inlining) approach.


ederign (approved):

packages/kn-plugin-workflow/go.mod
```diff
@@ -7,27 +7,29 @@ require (
 	github.com/jstemmer/go-junit-report/v2 v2.0.0
 	github.com/ory/viper v1.7.5
 	github.com/spf13/cobra v1.4.0
+	knative.dev/client v0.34.0
```
dsimansk: This version should be `v0.31.z`. We're aligning to Knative version in Serverless releases, that is currently around latest-3 behind. It's mainly due to Serving and Eventing deps, that might pull unwanted changes, but it's not the case for workflow plugin (yet). 

Otherwise it seems OK.

ljmotta:
@dsimansk Thanks for reviwing! I've made the change.

@cimbalek Could you please review?

paulovmr (approved):

ederign (approved):

PR #1226, opened 2022-09-12
KOGITO-7568: [SWF Editor] Use the YAML API in SWF Text editor and VSCode ext

fantonangeli:
Fixes: https://issues.redhat.com/browse/KOGITO-7568

Please note:

- In `serverless-workflow-language-service/src/channel/SwfYamlLanguageService.ts` I created the method `SwfYamlLanguageService.isNodeUncompleted()`, used by `SwfYamlLanguageService.getCompletionItems()`, to manages the case:
```
 functionRef:
    refName: 🎯
```
Otherwise, `findNodeAtOffset`, because of the space character after `refName`, returns `functionRef` node (note that `refName` value is not defined).
It is a kind of workaround, so please let me know if you agree with this solution. Tests are already implemented.

- There are 3 skipped tests in `packages/serverless-workflow-language-service/tests/SwfYamlLanguageService.test.ts` to tests the case of JSON format in YAML. eg:
`functionRef: {🎯}`
I planned to add this support in this issue: https://issues.redhat.com/browse/KOGITO-7895

- In `/packages/serverless-workflow-language-service/tests/matchNodeWithLocation.test.ts` there is one skipped test regarding the case:

	```
	functions:
	- 🎯
	```

	Here the bug is that in the AST, `findNodeAtOffset()` returns the value (array) of the `functions` property, and not the first element.
	The `functions` value has an array with a `null` element because the object is still to be defined by the user.
	Anyway this bug is not affecting our YAML code completion, because the selection match the `functions *` selector.
	If makes sense, I skip testing this case with the `function` selector, that erroneusly returns true.

- Completions for `operation: '🎯'` are triggered and our code is returning the right result, but VSCode doesn't show the completions in the menu.
  Instead `operation: 🎯''` is working. 

  Is it to be considered as a bug? Anyone has some suggestion?

[KOGITO-7568.webm](https://user-images.githubusercontent.com/17780574/189680223-288bd9f3-3f4b-448d-b871-3a5c6a5da76b.webm)


packages/serverless-workflow-language-service/src/channel/SwfJsonLanguageService.ts
```diff
@@ -71,6 +71,17 @@ class JsonCodeCompletionStrategy implements CodeCompletionStrategy {
...
 
+  public formatLabel(label: string, completionItemKind: CompletionItemKind): string {
+    if (
+      completionItemKind === CompletionItemKind.Value ||
+      completionItemKind === CompletionItemKind.Function ||
+      completionItemKind === CompletionItemKind.Folder
+    ) {
+      return `"${label}"`;
+    }
+    return label;
```
caponetto:
```ts
return [CompletionItemKind.Value, CompletionItemKind.Function, CompletionItemKind.Folder].includes(completionItemKind) ? `"${label}"` : label;
```

I would do something like this but this is just my personal preference.
The same goes for the YAML variant.

packages/vscode-extension-serverless-workflow-editor/package.json
```diff
@@ -358,6 +358,9 @@
     "onLanguage:serverless-workflow-yaml",
     "onCustomEditor:kieKogitoWebviewEditorsServerlessWorkflow"
   ],
+  "extensionDependencies": [
+    "redhat.vscode-yaml"
```
caponetto, line 362: Do we really need to enforce this dependency?

packages/serverless-workflow-language-service/src/channel/SwfLanguageService.ts
```diff
@@ -205,6 +210,7 @@ export class SwfLanguageService {
     }
 
     const document = TextDocument.create(args.uri, this.args.lang.fileLanguage, 0, args.content);
+    const isYaml = getFileLanguage(args.uri) === FileLanguage.YAML;
```
caponetto: IMO `SwfLanguageService` must not know anything about YAML (or JSON, etc).
If there is specific code for file languages, then this code should be provided through `codeCompletionStrategy`, thus making it transparent for `SwfLanguageService`.

packages/serverless-workflow-language-service/src/channel/SwfLanguageService.ts
```diff
@@ -551,14 +570,17 @@ const completions = new Map<
...
-            kind: CompletionItemKind.Module,
-            label: `${swfFunctionRef.refName}`,
-            sortText: `${swfFunctionRef.refName}`,
+            kind,
+            label,
+            sortText: label,
             detail: `${swfServiceCatalogFunc.operation}`,
             textEdit: {
-              newText: codeCompletionStrategy.translate(swfFunctionRef),
+              newText: codeCompletionStrategy.translate(swfFunctionRef, CompletionItemKind.Module),
```
caponetto:
```suggestion
              newText: codeCompletionStrategy.translate(swfFunctionRef, kind),
```
You can use `kind` here, right? Also applicable for similar cases below.

packages/serverless-workflow-language-service/src/channel/findNodesAtLocation.ts
```diff
@@ -22,9 +22,14 @@ import { SwfLsNode } from "./types";
  *
  * @param root root node
  * @param path the location of the node to search
+ * @param includeUncompleteProps true to include uncomplete properties. eg: { parent: { child: }}
  * @returns an array of nodes matching the path, empty array if no matches
  */
-export function findNodesAtLocation(root: SwfLsNode | undefined, path: any): SwfLsNode[] {
+export function findNodesAtLocation(
```
caponetto: Can we use an `args` object for the parameters here?
Also, would it be possible to set a more restrictive type for `path`, rather than `any`?

packages/serverless-workflow-language-service/src/channel/indentText.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @returns the indented string
+ */
+export function indentText(text: string, indentNum = 0, indentChar = " ", skipFirstLine = true): string {
+  if (!text) {
+    return "";
+  }
+
+  const indentation = new Array(indentNum).fill(indentChar).join("");
+
+  const indentedText = text.replace(/^/gm, indentation);
```
caponetto, line 34: Despite being a very simple regex (`/^/gm`), can you put a comment explaining what it does? 
Or extract it into a const with a self-explanatory name.

caponetto (commented):
Hey @fantonangeli, thanks for the PR! :)
I took a look over the code and left some comments inline.

Also, two questions:
- Why is "Add function..." shown below "functions" for YAML and above for JSON?
- Why are values for `operation` and `name` autocompleted inside single quotes? `refName`, on the other hand, is not autocompleted inside single quotes.

I'll give the code a try as well and get back to you once I do.

fantonangeli:
@caponetto 

>     * Why is "Add function..." shown below "functions" for YAML and above for JSON?

It's because the CodeLens is placed before the value of `function:`, that is the `array`. In JSON it starts with the `[` character, instead in YAML starts with `-` at the first element of the array, which is where an array is defined. 
In fact in JSON you can put a newline after `function:` and you will have the same result as YAML

image

Possible solution: place the CodeLenses before the `function:` label, not the value.
Small downside: with 2 newlines after `function:` we will have:
```
Setup Service Registry... | +Add function.... 
function:

[{...
```
WDYT?






>     * Why are values for `operation` and `name` autocompleted inside single quotes? `refName`, on the other hand, is not autocompleted inside single quotes.

I use the [`dump()`](https://github.com/redhat-developer/yaml-ast-parser/blob/4cac55876ef365d452a9a9e817b0f74e010d4fdd/src/dumper.ts#L820) function from `yaml-ast-parser` to convert a JS object to YAML string. 
Unluckily it doesn't have an option to force quotes. 
`yaml-ast-parser` has been already installed previously in the `serverless-workflow-language-service` and I used the parser from that without adding more dependencies, but if it's important I can try with a different converter.

packages/vscode-extension-serverless-workflow-editor/package.json
fantonangeli: It adds the YAML support, that is needed to show CodeLenses and Completions, without it the `it-tests` fails.

packages/serverless-workflow-language-service/src/channel/SwfLanguageService.ts
fantonangeli: oops! I extracted the `kind` constant for this purpose and then I forgot to change `CompletionItemKind...`->`kind` :smile:

fantonangeli:
Jenkins retest please

zdrapela:
Hi @fantonangeli, could you please add a broader description to the JIRA for better tracking? Otherwise, it looks good to me 👍

fantonangeli:
> Hi @fantonangeli, could you please add a broader description to the JIRA for better tracking? Otherwise, it looks good to me +1

Jira updated @zdrapela

zdrapela (approved):
Thanks for the Jira. Approved!

caponetto:
@fantonangeli  Thanks for the details. As we talked offline, let's just make sure the autocomplete for `function` and `functionRef` behaves similarly in terms of the quotes (or at least understand the reason they are behaving differently).

fantonangeli:
@caponetto I investigated the `dump()` function in `yaml-language-server-parser`. It automatically puts the single quotes if there is a special character in the string, like "#" in our case.
But after looking better at the code I confirm I don't have an option to force the quotes.

Below are the points in the source where it the quotes are added to the string and where it's checked for "#" chars.
https://github.com/redhat-developer/yaml-ast-parser/blob/4cac55876ef365d452a9a9e817b0f74e010d4fdd/src/dumper.ts#L364
https://github.com/redhat-developer/yaml-ast-parser/blob/4cac55876ef365d452a9a9e817b0f74e010d4fdd/src/dumper.ts#L493

caponetto (approved):
LGTM

caponetto:
@paulovmr @tiagobento can you guys take a look at this PR when possible, please? :)

paulovmr (approved):
LGTM, thanks @fantonangeli !

PR #1229, opened 2022-09-13
KOGITO-6402: Unskip test

jomarko:
https://issues.redhat.com/browse/KOGITO-6402 was marke as resolved, thus it makes sense to unskip related test


yesamer (approved):
Thank you, @jomarko

paulovmr (approved):

PR #1258, opened 2022-10-05
KOGITO-8079: Add customization options to the SWF combined editor

paulovmr:


packages/serverless-workflow-combined-editor/src/editor/hooks/useSwfTextEditorChannelApi.tsx
```diff
@@ -37,7 +37,7 @@ export function useSwfTextEditorChannelApi(args: {
   const [serviceRegistriesSettings] = useSharedValue(
     args.channelApi?.shared.kogitoSwfServiceCatalog_serviceRegistriesSettings
   );
-  const stateControl = useMemo(() => new StateControl(), [args.embeddedEditorFile?.getFileContents]);
```
caponetto, line 40: This dependency came from [1], which was introduced in [2].
Can you please just double-check that this is indeed not necessary?

[1] https://github.com/kiegroup/kie-tools/blob/main/packages/editor/src/embedded/embedded/EmbeddedEditor.tsx#L85
[2] https://github.com/kiegroup/kie-tools/pull/273

packages/serverless-workflow-combined-editor/src/editor/hooks/useSwfDiagramEditorChannelApi.tsx
```diff
@@ -29,7 +29,7 @@ export function useSwfDiagramEditorChannelApi(args: {
   onEditorReady: () => void;
   swfTextEditorEnvelopeApi?: MessageBusClientApi<ServerlessWorkflowTextEditorEnvelopeApi>;
 }) {
-  const stateControl = useMemo(() => new StateControl(), [args.embeddedEditorFile?.getFileContents]);
```
caponetto, line 32: Same here.

packages/serverless-workflow-combined-editor/src/editor/hooks/useSwfTextEditorChannelApi.tsx
paulovmr: @caponetto I had removed it because of the React warning but you are right, I've put it back now with a comment to keep it.

caponetto (approved):

PR #1275, opened 2022-10-13
KOGITO-7098: SWF Viewer - refactor shapes from SVG to pure lienzo

hasys:
[KOGITO-7098](https://issues.redhat.com/browse/KOGITO-7098)

Hello @romartin, @domhanak, there is a PR for KOGITO-7098. It fixes:
* Module org.kie.kogito.stunner.serverless.editor:kie-wb-common-stunner-svg-client removed
* Module org.kie.kogito.stunner.serverless.editor:kie-wb-common-stunner-svg-gen removed
* Visually shapes and transitions should be changed in a way:
    * Default: #FFFFFF fill with #CCCCCC outline
    * Hover: #FFFFFF fill with #4F5255 outline
    * Selected: #E7F1FA fill with #0066CC outline with shadow
* On mouse over the borders are highlighted
* On selection node is moved to selected state and over is not applying anymore

image

For tests you can use: [VSIX for tests](https://drive.google.com/file/d/1ajuxF4SdvXFo76mZn9Ew9aR1_PoqHPaS/view?usp=share_link).

Thank you!


hasys:
Hello @romartin, @domhanak,

I finished sync up with UX team and also found a workaround for Rounded corners. It's not pretty so it's better to fix Lienzo bug in the future, but for now it works.

PR should be fully ready for the review.

Thank you!

romartin:
Hey @hasys 

Thanks for the summary of fixes, looks great. I'll take a look asap.

Also, from the comment above, you mean are also fixing the [KOGITO-8157](https://issues.redhat.com/browse/KOGITO-8157) or this will be part of a future PR?

hasys:
> Hey @hasys
> 
> Thanks for the summary of fixes, looks great. I'll take a look asap.
> 
> Also, from the comment above, you mean are also fixing the [KOGITO-8157](https://issues.redhat.com/browse/KOGITO-8157) or this will be part of a future PR?

Hi @romartin,

[KOGITO-8157](https://issues.redhat.com/browse/KOGITO-8157) is not fixed on lienzo side so the issue is still there. But for now I found a workaround for this behavior: The main layer border is shown only after we start interact with the shape itself, and the first interaction is Mouse Hover. So now on each mouse hover for the state I just changing the border radius and it works just fine. It helps me to instantiate/create magnets/connect to another shape initially and than change corners to rounded version on the first (and all ongoing on hover) interactions.

hasys:
I rebased to the master branch and added new JsInterop API moving back to the Draft state to check the API for now.

hasys:
Hello @romartin, @domhanak,

I rebased to the main branch and fixed all conflicts. Also I moved workaround for [KOGITO-8157](https://issues.redhat.com/browse/KOGITO-8157) from SW Viewer level to Lienzo level. It means that there is no more Null Pointer exception on magnets creation and also Center, North, West, East and South magnets works fine. Since SW Viewer so far using only those 5 magnets we are fine now and don't need any workaround in SW Viewer anymore. But in the future we, probably, will need to revisit NW, NE, SW, SE magnets.

Also fixed small issues like selection color for Start/End node and colors for Edge labels and mouse over Start/End titles changed Start/End Hover state to None.

Edit mode now also possible to enable (there were two issues, in my PR and on the main branch as well, now both fixed).

PR is ready for review again. 

Thank you!

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/types/PathPartList.java
```diff
@@ -207,13 +207,15 @@ public final PathPartList circle(final double r) {
 
         final double c = r * 2;
 
-        A(x + r, y, x + r, y + r, r);
+        M(x + r, y);
```
romartin, line 210: hey @hasys 
Are those updates (other changes below in this `circle` method as well) into the circle path generation expected? I am just asking because I was not aware of this logic being wrong. Was it wrong? What do these updates are intended for?

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/util/Geometry.java
```diff
@@ -1014,8 +1005,11 @@ public static final Point2DArray getCanvasArcToPoints(final Point2D p0, final Po
     }
 
     public static final Point2DArray getCardinalIntersects(final AbstractMultiPathPartShape<?> shape, Direction[] requestedCardinals) {
+        double corner = shape.getCornerRadius();
+        shape.setCornerRadius(0);
```
romartin: I see, here it is the workaround for arcs and magnets calculations (wires), well done :)

Anyway, can you please add a comment quickly explaining the reason for this code? Also can you please ensure the unit tests are covering this?

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/EndShape.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.kie.workbench.common.stunner.core.client.shape.ShapeState;
+import org.kie.workbench.common.stunner.core.client.shape.impl.AbstractShape;
+import org.kie.workbench.common.stunner.core.client.shape.impl.NodeShapeImpl;
+import org.kie.workbench.common.stunner.core.client.shape.impl.ShapeStateHandler;
+
+public class EndShape extends NodeShapeImpl {
+
+    public EndShape() {
+        super(new AbstractShape<EndShapeView>() {
```
romartin: Hey @hasys ,

Just proposing a quick improvement in order to reuse code. 

I see that on every `Shape` class (Start,End,State), you're creating an abstraction (`AbstractShape<>`) that IMO can be reused, instead of duplicating code on each shape class.

WDYT?

romartin (commented):
Hey @hasys 

Great work, great refactor!! Well done :clap: 

I just added a few review comments on code side, for now, Please let me know what do you think.

Thanks!

romartin:
@domhanak As far as I know this is ready for QE review as well. Thanks!

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/EndShape.java
hasys: Great catch, fixed. Much better now!

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/util/Geometry.java
hasys: So, tests won't be an easy task due to native method calls, but during tries to find a way how to write a test, I found the root cause and now the issue itself fixed, no more workarounds.

The issue was - rounded corners points (and bounding box) were calculated during the shape drawing so magnets were calculated before the points (and bounding box) were calculated.

The fix is calculate points (and bounding box) when it needed, not before the first drawing.

Values are also properly cached and cleared now, so no leaks and recalculations.

Thank you, I am closing  [KOGITO-8157](https://issues.redhat.com/browse/KOGITO-8157) as well since it is fixed now.

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/types/PathPartList.java
hasys: Hi @romartin,

see picture below

image

Picture 1 is how it was, as you see Y coordinates are prevented to be negative, but X coordinates weren't. And Picture 2 is how it is now: both X and Y are prevented to be negative.

It caused at least 3 issues:
1. Bounding box didn't calculated correctly with negative coordinates
2. Due to incorrect Bounding box calculations it breaks auto-alignment so we used addition `setX(radius)` to correct the bounding box calculations and as result to correct the auto-layout
3. It wasn't consistent with rectangle which adjusted not on Y but X as well so it was always as Picture 2 now for circles.

Please let me know if it does make sense for you.

hasys:
Hello, @romartin, @domhanak,

I fixed all suggestions from @romartin and answered questions. I rebuild the VSIX file, it shouldn't be different from the previous version from the user point of view, but just in case I rebased my changes to the `main` branch and rebuild it to be sure all works as expected.

Also I closed [KOGITO-8157](https://issues.redhat.com/browse/KOGITO-8157) since it is properly fixed now.

PR  is ready for review again. Thank you!

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/types/PathPartList.java
romartin: Hey @hasys 
Thanks for the updates. I have no complains about the approach on picture 2, but not sure if this change can have more implications that we don't have in mind yet, we should test this properly. For instance, did you tested how the rotate works now? Does it rotates properly the circle?

hasys: Hi @romartin,

I double checked the useage of `MultiPath.circle(radius)` method and it looks safe to me. Keep in mind that I didn't change the Circle class. For magnets and other elements used `Circle` class not `Multipath.circle` method and `Circle` class still drawing the circle around it's center at [0,0] coordinate (different from both Picture 1 and Picture 2).

So, since `Multipath.circle` is used just one in the SW Viewer (for StartShapeView) and couple examples from lienzo-webapp this change is safe.

Thank you for the great question!

romartin: that's true @hasys , ok makes sense to me! Thanks!

romartin (approved):

hasys:
Hi @domhanak,

all is ready for your review, thank you!

domhanak (approved):
LGTM

hasys:
Thank you @romartin, @domhanak!

@kiegroup/gatekeepers this PR is ready for merge, thank you!

PR #1285, opened 2022-10-17
KOGITO-8098: Incorporate the shared worker on Serverless Logic Web Tools

caponetto:
JIRA: https://issues.redhat.com/browse/KOGITO-8098

This PR incorporates the work done in https://github.com/kiegroup/kie-tools/pull/1244 into the Serverless Logic Web Tools.

Besides improving the Git and Storage experience with the worker mechanism and all other improvements provided in https://github.com/kiegroup/kie-tools/pull/1244, the goal with this PR is also converging as much as possible the workspace-related code between KIE Sandbox and Serverless Logic Web Tools to make it easier to extract common code into a new package. 

Although some adaptations needed to be done to align with Web Tools requirements, most of them are additions and can either be extracted to the new package as well or become part of Web Tools only. There are also some other differences in the code due to new changes introduced in https://github.com/kiegroup/kie-tools/pull/1269 but they can be incorporated in future PRs.

Since the Virtual Service Registry was heavily based on the descriptor+workspaces approach, I took the opportunity to make this module simpler by reusing generic stuff from LFS as much as possible but keeping the logic untouched. There is still room for improvement but it can be done in future PRs once new features for this area are worked on.

Deployment of this branch for tests: https://caponetto.dev/poc/webtools-worker2


packages/serverless-logic-sandbox/src/virtualServiceRegistry/VirtualServiceRegistryContextProvider.tsx
```diff
@@ -0,0 +1,251 @@
...
+  useCancelableEffect(
+    useCallback(
+      ({ canceled }) => {
+        const workspacesBroadcastChannel = new BroadcastChannel(WORKSPACES_BROADCAST_CHANNEL);
+
+        workspacesBroadcastChannel.onmessage = async ({ data }: MessageEvent<WorkspacesEvents>) => {
+          if (canceled.get()) {
+            return;
+          }
+          if (data.type === "ADD_WORKSPACE") {
```
thiagoelg, line 169: We should probably move this to a shared worker in the future, right? (to avoid multiple tabs conflicting)

caponetto: We can definitely improve it. Although the code looks a little bit different, the logic is kept the same way it is currently done on `main`.

thiagoelg (approved):
Awesome work, this looks great!

domhanak:
@tiagobento @thiagoelg @paulovmr Would it be possible to allow other browsers and keep the warning? It would be nice to help us identify issues preventing sandbox from working on firefox for example. Let me know

caponetto:
@domhanak 
The goal of this PR is to converge the workspace-related code between the sandboxes to make it easier to extract common code into a new package. It means that everything that was done for KIE Sandbox, was done too for Serverless Logic Web Tools in this PR.

However, right after this one, I'll send a new PR for [KOGITO-8156](https://issues.redhat.com/browse/KOGITO-8156), which at least enables Firefox for Serverless Logic Web Tools and keep the warning icon.
Regarding other browsers, an assessment will be done via [KOGITO-8166](https://issues.redhat.com/browse/KOGITO-8166) and enable them accordingly.

Regarding KIE Sandbox, this is up to @tiagobento.

paulovmr (approved):

PR #1290, opened 2022-10-20
KOGITO-8101: SWF Viewer - warn user about duplicated state name

hasys:
[KOGITO-8101](https://issues.redhat.com/browse/KOGITO-8101)

Hello @romartin, @domhanak,

notification added in case diagram has two states with the same name

image

[VSIX for tests](https://drive.google.com/file/d/1OEKTLe_rbluJsjDNhXShtwVNOGm3N9VM/view?usp=sharing)

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/BuilderContext.java
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/BuilderContext.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 
 import org.kie.workbench.common.stunner.core.api.DefinitionManager;
 import org.kie.workbench.common.stunner.core.api.FactoryManager;
+import org.kie.workbench.common.stunner.core.command.Command;
 import org.kie.workbench.common.stunner.core.command.CommandResult;
 import org.kie.workbench.common.stunner.core.command.impl.CompositeCommand;
 import org.kie.workbench.common.stunner.core.factory.graph.EdgeFactory;
@@ -56,6 +57,7 @@ import org.kie.workbench.common.stunner.sw.definition.OperationState;
 import org.kie.workbench.common.stunner.sw.definition.ParallelState;
 import org.kie.workbench.common.stunner.sw.definition.SleepState;
 import org.kie.workbench.common.stunner.sw.definition.Start;
+import org.kie.workbench.common.stunner.sw.definition.State;
 import org.kie.workbench.common.stunner.sw.definition.SwitchState;
 
 public class BuilderContext {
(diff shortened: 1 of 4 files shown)


domhanak (approved):
LGTM. Checked the use case and it works well.

Noting that I do not see any unit tests, but there are multiple ongoing efforts around it so I guess we can expect that here so there no are conflicts. Let me know.

Maybe `isDuplicateState()` would be better  name but just a nitpick, I ll leave that here as a suggestion :)

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/BuilderContext.java
```diff
@@ -111,6 +113,27 @@ public Node addNodeByUUID(String uuid,
         return node;
     }
 
+    public boolean isStateAlreadyExist(String name) {
```
romartin, line 116: Hey  @hasys 

This is working fine, although I am thinking that may exist a less intrusive way to check existing names in the workflow, rather than iterating, for every state, into all existing builder commands from the marshallers. See https://github.com/romartin/kie-tools/blob/main/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/BuilderContext.java#L72

As you can see, there actually exists a context during marshalling, where you have a map with the state names. So what about just checking if the state name is duplicated, by just checkinng if the key is present in the map? Does it makes sense? 

This way the amount of code would be much less, also it would perform faster as we're only searching a key in a map, not really iterating over commands etc.

WDYT?

Sorry for raising this up late, I just noticed about it today.

romartin (commented):
Hey  @hasys 

All looks good to me! Although please see my comment below, I think we can just do something easier, let me know what you think.

Thanks!

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/BuilderContext.java
hasys: Hey @romartin,

yes, I was looking at this, and what I found out this map is preserved between diagram reloads, so when you load the diagram for the first time, it will work just fine, but on reload/update/loading new diagram we will see "duplication" notifications for all state not changed from the last update. In theory we can create the another one map and store all names there as well, but I think it will complicate readability for the future and I don't think speed up will worth this complications.

Thank you for the good point and let me know if you agree with current solution.

romartin: ok it makes sense. Sorryt for bothering :)

romartin (approved):

PR #1291, opened 2022-10-20
KOGITO-8093: [SWF Editor] Improve auto-completion experience

fantonangeli:
Closes: https://issues.redhat.com/browse/KOGITO-8093

Goals:
    
    Improve auto-completion experience, to make it easier to invoke a service or orchestrate an event, i.e. to create specific states like operation and async

    improving the code completion to simplify adding full operation node referencing some function (imported from ServiceRegistry before)

New features:

* Add eventConditions transitions completion
* Add defaultCondition and dataConditions transitions completion
* Add transition and start completions
* Add eventRefs completions
* Add generic Functions node completions and unified similar tests
* Add Events node completions and unified similar tests
* Add most common states completion

Notes:
 
* for YAML files on `states` completion, i have to complete the property functionRef with empty brackets: `functionRef: {}`, The same happens for arrays.
  This because it's the only way in YAML to declare an empty object/array
* for YAML files there are completions in the menu that are not managed by our plugin. Eg for states completion. Those completions create nodes with very few properties, and with too many completion the user can feel "confused".  What should we do with the default completion?

KOGITO-8093


packages/serverless-workflow-language-service/src/assets/code-completions/states-completions.ts
```diff
@@ -0,0 +1,69 @@
...
+  ],
+  defaultCondition: {
+    transition: "${7:Default transition of the workflow}",
+  },
+};
+
+export const injectStateCompletion: OmitRecursively<Specification.Injectstate, "normalize"> = {
+  name: genericStateCompletion.name,
+  type: "inject",
+  data: [],
```
caponetto: According to the spec, `data` should be an object.

Ref: https://github.com/serverlessworkflow/specification/blob/main/specification.md#inject-state

caponetto (commented):
One more set of great improvements for this area. Thanks, @fantonangeli!
I have only one observation.

caponetto (approved):

packages/serverless-workflow-language-service/src/assets/code-completions/workflow-completions.ts
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Specification } from "@severlessworkflow/sdk-typescript";
+import { OmitRecursively } from "../../channel";
+
+export const workflowCompletion: OmitRecursively<Specification.Workflow, "normalize"> = {
+  id: "${1:Workflow unique identifier}",
+  version: "${2:0.1}",
+  specVersion: "${3:0.1}",
```
tomasdavidorg: I think we also use the `0.1` version in the autocompletion but I am just thinking if we should rather use here the latests version `0.8`. Wdyt?

tomasdavidorg (approved):
Looks good to me. Just 2 non-blocking notes:
1. The specVersion comment.
2. If a new event, function, state is added, it will break formatting in case of JSON file + if there is already an item, comma is missing in that case.
Example:
- before
Screenshot 2022-10-24 at 15 52 44
- after
Screenshot 2022-10-24 at 15 53 11

packages/serverless-workflow-language-service/src/assets/code-completions/workflow-completions.ts
fantonangeli: I agree to suggest the latest to the user and not an old one

fantonangeli:
> Looks good to me. Just 2 non-blocking notes:
> 
>     1. The specVersion comment.
> 
>     2. If a new event, function, state is added, it will break formatting in case of JSON file + if there is already an item, comma is missing in that case.

The last TAB key movement leaves the cursor at the position for the comma, so the user decides to put it or not without extra calculation for the library. If TAB is not enough I can implement it. WDYT?

tomasdavidorg:
Thats true, didn't notice that. Looks good to me. 🙂

caponetto (approved):

PR #1297, opened 2022-10-25
KOGITO-7833: [KIE Sandbox] Allow selecting the branch name to be used when cloning a Git repository

tiagobento:
This PR introduces the ability to choose which branch to use from a Git repository, when cloning. This is achieved by a new Modal (`AdvancedImportModalContent.tsx`) and some tweaks to the `Import > From URL` section on the Home page. This new Modal also appears as part of the `/import` route when the `?confirm=true` query parameter is enabled.

As consequence of this PR, the default branches are now fetched from the repositories and don't need to be guessed. (e.g. `kiegroup/kogito-examples`'s default branch is `stable`, and `kiegroup/kie-tools`'s is `main`. Also, validation and the UX of importing stuff is improved, so two additional Jiras are included here as well. See them listed below.

I also introduced a temporary concept of `Auth Sources`, which will then be later evolved to a much more robust Authentication subsystem. Right now it's only possible to select one GitHub account, or none.

Also on this PR:
- Added a new method on the Workspaces Shared Worker to fetch Git Server Refs from a URL. cc @caponetto 
- Fixed a bug on CORS Proxy, which will crash it if the Git URL did not exist (e.g.  `https://g`). Already sent a PR fixing it. See https://github.com/isomorphic-git/cors-proxy/pull/14). For this reason, the default setup now for local development includes running the CORS Proxy locally too. I've updated the instructions on this PR.
- Improved the `Import from URL` menu on the `New file` menu. Now it uses Auth Sources to import GitHub-specific private files as well.
- KOGITO-7814: Import -> From URL error handling improvements
- KOGITO-7815: New file -> From URL improvements

---
- The new card for importing at the Home page. `More options...` will open the new Modal. Also, notice that the Import button is now disabled when there's errors on the URL.
image

- `Change...` also opens the new Modal.
image

- The new Modal in the Home page
image

- The new Modal in the `/import` route. Try it on https://localhost:9001/#/import?confirm=true&url=https://github.com/kiegroup/kogito-examples&branch=stable&authSource=github.com
image


yesamer (approved):
LGTM

PR #1300, opened 2022-10-27
KOGITO-7963: Configure integration tests of cypress packages to run tests twice in case of fails

zdrapela:
**Jira:** https://issues.redhat.com/browse/KOGITO-7963

Retries in case of failure are set only for `run mode` in packages using Cypress (open mode is not used in automated tests, as far as I know).

Note that there are also packages, that have the commands for running Cypress set in `package.json`, but they don't contain any tests.

### DIFF
--- a/packages/boxed-expression-component/it-tests/cypress.config.ts
+++ b/packages/boxed-expression-component/it-tests/cypress.config.ts
@@ -21,4 +21,8 @@ export default defineConfig({
     specPattern: "e2e/**/*.cy.{js,jsx,ts,tsx}",
     supportFile: "support/e2e.ts",
   },
+  retries: {
+    runMode: 1,
+    openMode: 0,
+  },
 });
--- a/packages/kie-editors-standalone/it-tests/cypress.config.ts
+++ b/packages/kie-editors-standalone/it-tests/cypress.config.ts
@@ -17,4 +17,8 @@ export default defineConfig({
   e2e: {
     specPattern: "./cypress/e2e/**/*.cy.{js,jsx,ts,tsx}",
   },
+  retries: {
+    runMode: 1,
+    openMode: 0,
+  },
 });
--- a/packages/online-editor/it-tests/cypress.config.ts
+++ b/packages/online-editor/it-tests/cypress.config.ts
@@ -20,4 +20,8 @@ export default defineConfig({
     specPattern: "e2e/**/*.cy.{js,jsx,ts,tsx}",
     supportFile: "support/e2e.ts",
   },
+  retries: {
+    runMode: 1,
+    openMode: 0,
+  },
 });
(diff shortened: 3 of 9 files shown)


tiagobento:
I don't think that's the way to go.... Our tests should not fail if there isn't a problem with them. Re-running only increases the chaos in determining what's stable and what isn't..

zdrapela:
@tiagobento We previously discussed this inside the team and for us, it seemed like a way to avoid random fails of tests, that are generally stable, but from time to time fail in case of slow machines or a bug (e.g. chromedriver), which do not happen so often. @domhanak ?

tomasdavidorg:
@tiagobento you have right that it can hide some fails but if the fail is not random, it should fail again.

I am for using the rerun of the ui (and only ui) tests because the github action is very unstable environment for these tests and UI tests. We are using the similar rerun also for the business central tests https://github.com/kiegroup/kie-wb-distributions/blob/main/business-central-tests/business-central-tests-gui/pom.xml#L194

Anyway if there is a fail we usually run the actions again

Or we can run the tests on the more stable jenkins machines separately. 

But I am open to discussion because no solution is perfect :-/

domhanak:
@tiagobento this helps us  mitigate reruns in case there is random fail. There really is no downside to this, if we have 2 fails, we will see this test failed twice. If it passed on second run then good, no need to rerun the pipeline.

@zdrapela what about the failed CI? I see issue with `chrome-extension` package did the test run twice?

domhanak:
@zdrapela btw. careful with the branches, the JIRA automation paired this PR with incorrect ticket

zdrapela:
> @zdrapela what about the failed CI? I see issue with `chrome-extension` package did the test run twice?

I see it failed with `chrome-extension-pack-kogito-kie-editors` package, which have Selenium tests, so no rerun here :-/

> @zdrapela btw. careful with the branches, the JIRA automation paired this PR with incorrect ticket

Really sorry for that. And renaming the branch doesn't work [as advertised](https://github.com/github/renaming/#renaming-existing-branches) (or works differently with branches from a fork) and makes even more mess :-/

tiagobento:
I understand this brings some benefits, and basically automates what we already do manually, so I'm ok with it. My concern is that the stability of tests is "resolved" by the random factor of reruns. **What I think is important here is to understand that reruns are a temporary solution until we find a way to not need it anymore.**

Reruns only *reduce* the probability of a test failing, so this is not a permanent solution. Please don't consider my comment as a blocker, I know this will improve our lives in the short-term, but also I think it's important to highlight that rerunning flaky tests cannot be accepted as a **solution** to their instability problem.

domhanak:
@tiagobento sounded like a blocker :D now worries, we understand this is not a solution. Still a nice to have thing to increase the % of green builds.

domhanak (approved):
LGTM with green builds.

tomasdavidorg (approved):
All green. Looks good to me. 🙂

paulovmr (approved):

tiagobento (approved):

PR #1303, opened 2022-10-27
KOGITO-8156: Extract common workspace-related code from sandboxes into a new package

caponetto:
https://issues.redhat.com/browse/KOGITO-8156

Introducing two new packages:
- `@kie-tools-core/react-hooks`: Useful react hooks extracted from the sandboxes that can be used by other packages.
- `@kie-tools-core/workspaces-git-fs`: Workspace-related stuff (Git and FS) extracted from the sandboxes.

I quickly scanned over https://github.com/kiegroup/kie-tools/pull/1297 and saw some potential conflicts, but they should be easy to incorporate into this PR once merged. 

@tiagobento can you please review this one? I would really appreciate your feedback. :)


tiagobento:
❤️ 
image

tiagobento:
@caponetto Great work, really important stuff! Also, nice idea to extract the `reactExt` hooks to a new package. My only concern is that this package becomes a "catch-all" stuff when "extending" React, but for now it's fine :)

Another point is that I don't understand why some Lfs-related files are kept. For example, `LfsWorkspaceFsService.ts`, can you please give me more details?

tiagobento:
@caponetto Also, thanks for mentioning https://github.com/kiegroup/kie-tools/pull/1297, I think this PR will be better if merged after it.

caponetto:
@tiagobento 
I agree about the catch-all concern and we can definitely improve this new package as it evolves.
I think this is a good start for us to exercise and think more about reusable stuff at the application level that can be extracted into new packages for benefit of others. I see the same potential for react components as well.

The Virtual Service Registry (VSR) from Serverless Logic Web Tools is heavily based on the workspaces+descriptor approach for cross-workspace file management. So I created a couple more LfsXXX classes for general-purpose use and simplified the VSR as much as I could to reuse these new classes but kept the logic untouched. This, at least, enables other applications to reuse the old workspace approach we had for other purposes as VSR does.

tomasdavidorg (approved):

tiagobento (approved):

PR #1304, opened 2022-10-27
KOGITO-7448: Manage SWF dependencies, avoiding circular dependencies

thiagoelg:
This is a small part of a more significant change to how the Virtual Service Registry works, to better manage and deploy Serverless Workflows that depend on other files.

In this PR:
- Added a warning on the "Try on Openshift" dropdown when a workflow depends on other workflows;
- Added a check to avoid circular dependencies (for now, if Workflow A depends on Workflow B, A can't be used as a dependency for C or B, this is done by removing the generated OpenAPI Spec for Workflow A);
- Added a check to avoid a workflow adding itself as a dependency;
- Added a check to only allow specs from the same workspace as a dependency for a workflow.
image


packages/serverless-logic-sandbox/src/virtualServiceRegistry/hooks/useVirtualServiceRegistryDependencies.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from "react";
```
caponetto, line 17: Please add the copyright header to this file.

packages/serverless-logic-sandbox/src/editor/api/SwfServiceCatalogStore.ts
```diff
@@ -151,11 +153,35 @@ export class SwfServiceCatalogStore {
...
+        this.currentFile &&
+        metadata.id ===
+          toVsrFunctionPathFromWorkspaceFilePath({
+            vsrWorkspaceId: this.currentFile?.workspaceId,
+            relativePath: this.currentFile?.relativePath,
+          })
+      );
+    };
+
+    const isForeignWorkflow = (metadata: SearchedArtifact) => {
```
caponetto: Can we replace "Workflow" with something else? Maybe `isForeignArtifact` or something like that.

packages/serverless-logic-sandbox/src/virtualServiceRegistry/models/VirtualServiceRegistryFunction.ts
```diff
@@ -28,14 +30,18 @@ export class VirtualServiceRegistryFunction {
   }
 
   public async getOpenApiSpec(): Promise<string> {
-    const content = await this.file.getFileContents();
+    // Don't generate spec for files that depend on other workflows
+    if (await hasVirtualServiceRegistryDependency(this.file)) {
+      return "";
```
caponetto: Can we change the return type of this function to `Promise<string | undefined>` and handle calls properly in order to avoid generating empty files?

packages/serverless-logic-sandbox/src/virtualServiceRegistry/models/VirtualServiceRegistryFunction.ts
```diff
@@ -47,3 +53,28 @@ export class VirtualServiceRegistryFunction {
...
+  try {
+    parsedContent = isJson(file.relativePath) ? JSON.parse(content) : yaml.parse(content);
+  } catch (e) {
+    // Invalid file.
+    return [];
+  }
+  const workflowFunctions = parsedContent["functions"] as Array<{ operation?: string }> | undefined;
+
+  const dependencies: Array<string> = [];
+  workflowFunctions?.forEach((workflowFunction) => {
```
caponetto: Can we use `filter` and `map` instead of `forEach`?

packages/serverless-logic-sandbox/src/editor/api/SwfServiceCatalogStore.ts
```diff
@@ -151,11 +153,35 @@ export class SwfServiceCatalogStore {
...
+      return !isSpec(metadata.id) && metadata.groupId !== this.currentFile?.workspaceId;
+    };
+
+    const isLocalSpec = (metadata: SearchedArtifact) => {
+      return isSpec(metadata.id) && metadata.groupId === this.currentFile?.workspaceId;
+    };
+
+    // The list should:
+    // - Show workflows from other workspaces
```
caponetto, line 176: Shouldn't we also show artifacts from the same workspace?

packages/serverless-logic-sandbox/src/virtualServiceRegistry/hooks/useVirtualServiceRegistryDependencies.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { isServerlessWorkflow } from "../../extension";
+import { ActiveWorkspace } from "../../workspace/model/ActiveWorkspace";
+import { useWorkspaces, WorkspaceFile } from "../../workspace/WorkspacesContext";
+import { getVirtualServiceRegistryDependencies } from "../models/VirtualServiceRegistryFunction";
+import { useVirtualServiceRegistry } from "../VirtualServiceRegistryContext";
+
+export type VirtualServiceRegistryDependency = ActiveWorkspace & { vsrFiles: WorkspaceFile[] };
+
+export function useVirtualServiceRegistryDependencies(props: {
+  workspace?: ActiveWorkspace;
```
caponetto: Does it need to be optional? I think we can make this prop required along the whole call chain.

packages/serverless-logic-sandbox/src/virtualServiceRegistry/hooks/useVirtualServiceRegistryDependencies.tsx
```diff
@@ -0,0 +1,58 @@
...
+  workspaceFile?: WorkspaceFile;
```
caponetto: If `props.workspace` is required, I think we don't need this prop.

caponetto (commented):
Hi @thiagoelg, nice addition! Please see my inline comments.

Also, there is an empty file `DeploymentsContextProvider.tsx` being added. I guess this is a leftover.

packages/serverless-logic-sandbox/src/editor/api/SwfServiceCatalogStore.ts
thiagoelg: If you mean artifacts like specs, then specs from the same workspace are shown, but from other workspaces are not. I've talked about this with @ederign, and adding specs from foreign workspaces would add a lot of complexity when dealing with dependencies, so we are avoiding that (at least for now).

But if you mean artifacts like other workflows, those shouldn't be added as a function but as a `subFlowRef`, according to the specification, this way they can be deployed together.

caponetto: Got it! I meant the second one. Thanks for the info!

Can you please add a new line of comment saying that?
Like 
`- Hide workflows from the same workspace since they need to be added as subFlowRef`
or something like this. Just to make the list complete.

packages/serverless-logic-sandbox/src/virtualServiceRegistry/hooks/useVirtualServiceRegistryDependencies.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as React from "react";
+import { useState, useEffect, useCallback, useMemo } from "react";
+import { isServerlessWorkflow } from "../../extension";
+import { ActiveWorkspace } from "../../workspace/model/ActiveWorkspace";
+import { useWorkspaces, WorkspaceFile } from "../../workspace/WorkspacesContext";
```
caponetto: Note to self: update these imports in https://github.com/kiegroup/kie-tools/pull/1303 after this PR is merged.

caponetto: Ok, https://github.com/kiegroup/kie-tools/pull/1303 got merged first. 
@thiagoelg can you please update these imports to reference the new `workspace-git-fs` package?

caponetto (approved):

paulovmr (approved):

PR #1312, opened 2022-11-07
kie-issues#2: Prepare for supporting GHES on KIE Sandbox

### LINKED ISSUE #2: Prepare for supporting GHES on KIE Sandbox
KIE Sandbox allows authenticating with GitHub using PATs (Personal Access Tokens), but when trying to do the same with a GHES instance, it fails. The reason is because we currently validate this token against `github.com`.

We should be able to authenticate with any number of GHES instances on KIE Sandbox, and making it possible for users to select which one they want to use, depending on the repository and the action they're doing.

tiagobento:
Closes https://github.com/kiegroup/kie-issues/issues/2

----

This PR creates a new subsystem on KIE Sandbox: Accounts. It's composed by two main new entities, AuthProviders and AuthSessions. The new Accounts modal allows users to create any number of AuthSessions using the available AuthProviders. This mechanism is going to be used to support custom GHES instances, GitLab, and Bitbucket, and their on-prem use as well. More than that, authenticating with other kinds of services is also possible, like with OpenShift clusters, for Development Deployments of DMNs, for example.

AuthSessions are persisted on IndexedDB.

Next steps are to make the AuthProviders configurable through `env.json`. See https://github.com/kiegroup/kie-issues/issues/8

----
The new Accounts modal is accessible using the little User icon on the top-right corner of the header.
image
Selecting GitHub (the only one available at this moment), you'll be asked to provide your Token, with instructions to create one.
image
If your Token is valid, you'll see a Success message.
image
From this point on, you'll be authenticated with GitHub, and you'll have a choice to use this authentication session in your Workspaces. In this image, I already had a connection with `kie-tools-bot` set up.
image
If you select "+ Add", you're presented with a choice of providers to authenticate with.
image
After you have your authentication session setup, you can use it on Workspaces. Each Workspace can be linked to one authentication session at the time.
image
If you remove this authentication session, you'll see this warning in the Workspaces that were linked to it.
image
You can choose a new authentication session by selecting one, or use the "+ Connect to an account..." button to create a new authentication session.
image
Authentication sessions are used in many places, so if you don't have it properly setup, you're going to see a message like this one were it is required to be authenticated.
image
If your token is revoked or expires, KIE Sandbox will let you know by settings this authentication session as Invalid, displaying a red exclamation mark.
image


packages/online-editor/src/accounts/ConnectToGitHubSection.tsx
```diff
@@ -0,0 +1,285 @@
...
+        if (
+          [...authSessions.values()]
+            .filter(({ type }) => type === "git")
+            .some(({ token }: GitAuthSession) => token === githubToken)
+        ) {
+          setNewAuthSession({ error: "You're already logged in with this Token." });
+          return;
+        }
+
+        delay(600)
```
thiagoelg, line 85: Is this delay here to debounce user input? If so, adding a `if (canceled.get()) return` before calling the fetch method would prevent an unnecessary request.

packages/online-editor/src/accounts/authProviders/AuthProvidersContext.tsx
```diff
@@ -0,0 +1,104 @@
+import * as React from "react";
```
thiagoelg, line 17: Missing copyright header

packages/online-editor/src/accounts/authProviders/AuthProvidersGallery.tsx
```diff
@@ -0,0 +1,74 @@
+import { Card, CardBody, CardHeader, CardHeaderMain, CardTitle } from "@patternfly/react-core/dist/js/components/Card";
```
thiagoelg, line 17: Missing copyright header

packages/online-editor/src/accounts/AccountsDispatchContext.tsx
```diff
@@ -0,0 +1,145 @@
...
+      backActionKind: AccountsDispatchActionKind.SELECT_AUTH_PROVDER | AccountsDispatchActionKind.GO_HOME;
+      onNewAuthSession?: (newAuthSession: AuthSession) => any;
+    };
+
+// Reducer
+
+export enum AccountsDispatchActionKind {
+  CLOSE = "CLOSE",
+  GO_HOME = "GO_HOME",
+  SELECT_AUTH_PROVDER = "SELECT_AUTH_PROVDER",
```
thiagoelg: Small typo

packages/online-editor/src/accounts/authProviders/AuthProvidersGallery.tsx
```diff
@@ -0,0 +1,74 @@
...
+          .map((authProvider) => (
+            <Card
+              key={authProvider.id}
+              isSelectable={authProvider.enabled}
+              isRounded={true}
+              style={{
+                opacity: authProvider.enabled ? 1 : 0.5,
+              }}
+              onClick={() => {
+                if (authProvider.enabled && authProvider.type === "github") {
```
thiagoelg, line 52: I think this could use a map of dispatch actions for each provider type in the future, but this is fine for now.

packages/online-editor/src/accounts/authSessions/AuthSessionApi.ts
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const AUTH_SESSION_NONE: AuthSession = {
+  id: "none",
+  name: "Unauthenticated",
+  type: "none",
+  login: "Unauthenticated",
+};
```
thiagoelg: Just an alternative, but both methods work.

```suggestion
export const AUTH_SESSION_NONE = {
  id: "none",
  name: "Unauthenticated",
  type: "none",
  login: "Unauthenticated",
} as const;

export type NoneAuthSession = typeof AUTH_SESSION_NONE;
```

packages/online-editor/src/accounts/authSessions/AuthSessionSelect.tsx
```diff
@@ -0,0 +1,298 @@
...
+) => {
+  groups: AuthSessionSelectGroup[];
+  items: Array<AuthSessionSelectItem>;
+};
+
+export function AuthSessionSelect(props: {
+  authSessionId: string | undefined;
+  setAuthSessionId: React.Dispatch<React.SetStateAction<string | undefined>>;
+  isPlain: boolean;
+  title: string;
```
thiagoelg, line 70: The title could be optional here, as it's not necessary in this case:
image

packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -743,34 +821,55 @@ If you are, it means that creating this Gist failed and it can safely be deleted
                 </DropdownItem>
               </Tooltip>
               {!canPushToGitRepository && (
```
thiagoelg, line 823: It would be cool to be able to change the selected git authentication if the repository/gist was not created yet (maybe by checking if `workspacePromise.data.descriptor.origin.url` is not set)

packages/online-editor/src/editor/EditorToolbar.tsx
```diff
@@ -472,9 +532,19 @@ If you are, it means that creating this Gist failed and it can safely be deleted
         throw new Error("Gist creation failed.");
       }
 
+      const gistDefaultBranch = (
```
thiagoelg, line 535: ❤️ 

(This fixes [KOGITO-6912](https://issues.redhat.com/browse/KOGITO-6912) right?)

thiagoelg (commented):
Great addition, nice job!

I've left some suggestions, let me know what you think.

packages/online-editor/src/accounts/ConnectToGitHubSection.tsx
tiagobento: No, the delay is something I added to make it look like something's happening. I didn't like the experience of failing too fast, it didn't feel like an authentication was taking place.

packages/online-editor/src/accounts/authProviders/AuthProvidersGallery.tsx
tiagobento: Absolutely. This is just the first step. A lot of this code will evolve in future PRs.

packages/online-editor/src/accounts/authSessions/AuthSessionApi.ts
tiagobento: Great suggestion!

packages/online-editor/src/accounts/authSessions/AuthSessionSelect.tsx
tiagobento: The title being mandatory is something I'm leaving for the future. We're going to have more types of authentication other than Git, so making this explicit will help avoid confusion in the future.

packages/online-editor/src/editor/EditorToolbar.tsx
tiagobento: It is possible to do that using the top-left AuthSessionSelect above the file name. But it only works if you have more than one file in your Workspace.

packages/online-editor/src/editor/EditorToolbar.tsx
tiagobento: Probably. I'll leave it open for other people to investigate if this issue is still present.

thiagoelg (approved):

PR #1317, opened 2022-11-09
NO ISSUE: Check for workflow file contents before searching for vsr functions

thiagoelg:


packages/serverless-logic-sandbox/src/virtualServiceRegistry/models/VirtualServiceRegistryFunction.ts
```diff
@@ -55,6 +55,10 @@ export class VirtualServiceRegistryFunction {
 
 export async function getVirtualServiceRegistryDependencies(file: WorkspaceFile) {
   const content = await file.getFileContentsAsString();
+  if (!content) {
```
caponetto, line 61: Should we also add this if block after line 38?

thiagoelg: In there it would end up in the `else` section, with the `console.debug("No workflow ID!")`, but adding this if block would make it more explicit.

caponetto (approved):
Thanks, @thiagoelg! 😃

paulovmr (approved):

PR #1322, opened 2022-11-10
KOGITO-7078: SWF Viewer - Connector labels

inodeman:
https://issues.redhat.com/browse/KOGITO-7078 (First comment)

Since we don't have orthogonal lines with the new autolayout impl, labels must be fixed as:
- Fix the position of the label for vertical (less that 45degreesl) connectors
- Also consider the possiblty to reduce the width of the label (in case of overlapping)
- Fix the tooltip, which does not appear on mouse hover the label


romartin (approved):

romartin:
Hey @domhanak 
Ready for QE testing, please! 
Thanks!

romartin:
Hey @domhanak
Ready for QE testing, please!
Thanks!

domhanak:
In general looks good, I need to know if current behavior in case the labels does not overlap is correct. I see them being truncated in any case.

romartin:
hey @domhanak 

Yes, all labels, if long, are truncated, even if not overlapping. We're discussing this with UX and in fact, it was their idea, otherwise having longer and shorter labels across the workflow may look weird.

This way, do you think we can approve this?

Thanks!

domhanak (approved):

PR #1327, opened 2022-11-14
KOGITO-8214: Improvements to the deployment flow on Serverless Logic Web Tools

caponetto:
https://issues.redhat.com/browse/KOGITO-8214

This PR includes code reorganization on the Serverless Logic Web Tools package to make the deployment flow extensible, thus enabling the incorporation of customizable deployment flows.
Now we have OpenShift pipelines and deployment strategies. An OpenShift pipeline executes a set of related operations on OpenShift, such as resources that need to be created during the deployment operation, and resources that need to be fetched to load the user's deployments. A deployment strategy, on the other hand, builds up what needs to be done for creating a particular deployment kind.


domhanak:
@caponetto Would it be feasible to introduce unit tests for these changes?

caponetto:
> @caponetto Would it be feasible to introduce unit tests for these changes?

Thanks for bringing this up! I've added a few unit tests that made sense. The other parts of the code would basically end up with walls of mocks, which I don't believe are valuable.

paulovmr (approved):

PR #1329, opened 2022-11-15
KOGITO-8227: [SWF Editor] Add comma automatically after completions

fantonangeli:
Closes: https://issues.redhat.com/browse/KOGITO-8227

Description
The idea is with every auto complete action, we end up with a valid SWF, adding commas after JSON completions.

Goal:
Add the comma after a JSON completion, if the element is not the last one in the object.

KOGITO-8227


fantonangeli:
@ederign, traversing nodes in the memory required too much logic but in the end, using regexes, I managed to achieve the task with a few logic. This way is lighter and less prone to bugs.

tomasdavidorg (approved):
Works great. Thanks 🙂

caponetto (approved):
Great work, @fantonangeli!

PR #1331, opened 2022-11-16
KOGITO-8104: SWF Viewer - Ability to customize the states icon (json)

hasys:
Hello @romartin, @domhanak, @LuboTerifaj,

there is the PR for [KOGITO-8104](https://issues.redhat.com/browse/KOGITO-8104). All info is in the Jira issue, the only addition note, when you set `metadata.icon` property and refresh the process, you will see that icon is not loaded instantly and loaded with small delay. It is happening because of the Scale. I don't know the size of the image until page loaded it, after it I scale and place Image icon to it's place that's why there is a small lag. Not sure if it is fixable, looks like we need to choose between small blink and image scale. @romartin do you know other solutions for it?


packages/serverless-workflow-diagram-editor/lienzo-webapp/src/main/java/org/kie/lienzo/client/AutomaticLayoutExample.java
```diff
@@ -192,6 +192,7 @@ private void performAutomaticLayout() {
         }
 
         layer.refresh();
+        layer.draw();
```
romartin: Hey  @hasys , let you know calling `layer.draw` causes more computation that calling `layer.batch`, this way just asking if it should be possible to change this call and just call `batch` instead. I may not work, as draw parses the objects as well, but if it works, i'll help to improve performance. Can you test it, if makes sense to you?

packages/serverless-workflow-diagram-editor/lienzo-webapp/src/main/webapp/LienzoShowcase.html
```diff
@@ -24,6 +24,19 @@
     <meta content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no" name="viewport" />
     <link rel="stylesheet" href="css/app.css" />
     <script language="javascript" src="org.kie.lienzo.LienzoShowcase/org.kie.lienzo.LienzoShowcase.nocache.js"></script>
+    <style>
```
romartin, line 27: which is the `menu` element that this stylesheet applies for? Is this really necessary?

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/ShapeFactory.java
```diff
@@ -67,12 +74,58 @@ public Shape newShape(Object instance) {
             return new EndShape();
         } else if (instance instanceof State) {
             State state = (State) instance;
-            return StateShape.create(state.getName()).setType(state.getType());
+            StateShape shape = new StateShape(state.getName());
+            if (state.metadata == null) {
```
romartin: Hey  @hasys 

I think all this initialization stuff about metadata and icon, also the resourceAPI calls, should be moved to the StateShape class. This is specific logic for the state shape, which IMO can be managed by the shape itself, rather than the factory class (which I would prefer to keep it really simple, with no specific shape logic). This way you can pass the necessary arguments to the StateShape class' constructor, for instance, the resourceContentService, and so the shape class itself will manage this logic.

Also this is because the factory is called only once, one creating a shape. On the other hand, the code in the shape class itself (applyProperties, etc) gets executed when the bean changes. I know this works because we're recreating shape in every single change in the json , but conceptually is not completely correct IMO. In case in future we avoid reloading the whole workflow on changes, this won't work.

WDYT?

romartin (commented):
Hey   @hasys 

Thanks for this work, it is really good! 

Before doing the UI tests, I just did a quick code review and added a few comments. Can you please take a look and let me know your opinion?

Thanks!

packages/serverless-workflow-diagram-editor/lienzo-webapp/src/main/webapp/LienzoShowcase.html
hasys: It is the change in the lienzo-webapp and doesn't change the designer. This is just making showcase application menu a bit more fancy:
image

romartin: oh cool! Thanks!

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/ShapeFactory.java
hasys: Hi @romartin,

I tried to keep Controller free of Kogito specific things, but yea, make sense that shape factory and controller are at the same package anyway.

Fixed.

packages/serverless-workflow-diagram-editor/lienzo-webapp/src/main/java/org/kie/lienzo/client/AutomaticLayoutExample.java
hasys: Good point, thank you. Fixed.

hasys:
Hi @romartin,

great review as always, all issues resolved and rebased to the main branch as well. Thank you, ready for review again 👍

romartin (approved):
Thanks @hasys !

romartin:
Hey @LuboTerifaj @domhanak 
Early review appreciated please, as this ticket's deadline is end of this week. Can you please take a look?
Thanks!

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/Circle.java
```diff
@@ -80,7 +76,6 @@ protected boolean doStrokeExtraProperties() {
 
     /**
      * Sets this circle's radius.
-     *
```
LuboTerifaj, line 83: Why did you change formatting here? I would keep it as it is.

LuboTerifaj: Please check the change in all affected files in this PR.

LuboTerifaj (commented):
Hello @hasys , @romartin ,

The issues I found:

- When both type and icon are properly set, the type takes precedence, so the "type image" is rendered in the diagram. I think it would be better to set it in opposite way: the icon should always take the precedence if a valid image is used.
- If icon is set to empty string and type is set to non-existing type, the rendered image is empty circle, but when type is set to empty string and icon is set to non-existing image, or invalid base64 image, the rendered image is the default one. I think this should be unified. The default image could be rendered when non-existing type is set. What do you think about this?

Would it make sense to you to add some tests?

Otherwise it works good. I didn't find any blocker issues, so those mentioned above can be possibly addressed in a separate jira, if they are not quick fixes. Please let me know what do you think.

Thank you!

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/Circle.java
hasys: Hi @LuboTerifaj, this change is from IntelliJ IDEA auto formatter which we are using during development. I prefer to have all the code including comments formatted automatically rather than do it manually.

hasys:
Hi @LuboTerifaj,

Great job, thank you for the review!

About issues:
* The type property is actually not changing SW type of the State. It changes only icon to some specific predefined value, so it make sense to me to make priority to custom predefined icon value, rather then third party image. If user doesn't want to use custom predefined icon, the `type` parameter can be removed from `metadata` without any issues.
* Good catch. I suggest to move this issue with tests for it and surrounding code to another jira, since otherwise we won't be in time for the release. WDYT?

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/Circle.java
LuboTerifaj: It depends on your configuration. It is more readable with empty line between the description and the parameters from my pov, but feel free to let it as you have it.

LuboTerifaj:
Hello @hasys , 

Regarding the first issue, we can say the same thing about `icon` field. The scenario I am describing is setting the both values and choosing the one image. For me it makes more sense to prefer `icon` from `type` since as you said, the `type` property doesn't change the SW type, and it is only metadata, which could be potentially also used for other purpose than choosing an icon. The setting of `icon` field is more time consuming and I would say there is a higher probability a user might want to see rather "icon" image than `type` image if they have it in their project.

As I mentioned above, it is not necessary to address the issues in this PR.

Thanks!

LuboTerifaj (approved):
@hasys , approving for now, please let me know if you decide to make any changes within this PR. Thanks.

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/Circle.java
hasys: It is not my configuration it is KIE Java conventions, so I suggest to stick to it and don't introduce changes to this convention in this PR. But you are always fell free to suggest the change to the KIE Java conventions itself if you think it worth it. Thank you!

hasys:
Hello @kiegroup/gatekeepers this PR is ready for merge. Thank you!

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/Circle.java
LuboTerifaj: I use KIE Java conventions as well and it formats the code in a way I suggested.

hasys: That's weird, maybe you have EditorConfig setting enabled or other modifications? I can't reproduce what you are saying.

image

LuboTerifaj:
I reported the issues/enhancements from my comments above here:

- [KOGITO-8345](https://issues.redhat.com/browse/KOGITO-8345) _[SWF Viewer] Icon rendering improvements based on metadata fields_
- [KOGITO-8344](https://issues.redhat.com/browse/KOGITO-8344) Add missing tests for "Ability to customize the states' icon (json)"

LuboTerifaj:
I reported the issues/enhancements from my comments above here:

- [KOGITO-8345](https://issues.redhat.com/browse/KOGITO-8345) [SWF Viewer] Icon rendering improvements based on metadata fields
- [KOGITO-8344](https://issues.redhat.com/browse/KOGITO-8344) Add missing tests for "Ability to customize the states' icon (json)"

packages/serverless-workflow-diagram-editor/lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/Circle.java
LuboTerifaj: Yes, there must be misconfiguration on someone side.

LuboTerifaj: Yes, there must be misconfiguration on someone's side.

PR #1336, opened 2022-11-18
NO ISSUE: Improve Extended Services code

ljmotta:
## On this PR:
- Use envs to control build params.
- Remove config.yaml and install.js
- General code improvements:
  - Code reorganization;
  - Rename variables;


packages/extended-services/env/index.js
```diff
@@ -14,17 +14,34 @@
...
  */
 
-const { varsWithName, composeEnv } = require("@kie-tools-scripts/build-env");
+const { getOrDefault, varsWithName, composeEnv } = require("@kie-tools-scripts/build-env");
 
 module.exports = composeEnv([require("@kie-tools/root-env/env")], {
-  vars: varsWithName({}),
+  vars: varsWithName({
+    EXTENDED_SERVICES__modeler: {
+      default: "https://sandbox.kie.org/#/",
```
tiagobento: Default env vars should always target the development environment. I think `https://localhost:9001` would be the correct value here. Also, "modeler" is not a term we use. I think this should be `EXTENDED_SERVICES__kieSandboxUrl`. This would also require for this to be updated in other places that use the "modeler" term. WDYT?

packages/extended-services/pkg/server.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+}
+
+func createJitExecutor(jitexecutor []byte) string {
+	cacheDir, err := os.UserCacheDir()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	cachePath := filepath.Join(cacheDir, "org.kogito")
```
tiagobento: Probably `org.kie.kogito` would be better.

packages/extended-services/pkg/server.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+	}()
+
+	err = p.cmd.Start()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	router := mux.NewRouter()
+	router.PathPrefix("/devsandbox").HandlerFunc(p.devSandboxHandler())
```
tiagobento: This route was created to start as a CORS Proxy for the DMN Dev Sandbox feature, but I think we could revisit this, as this is a generic proxy. Do you think you could change this to "/cors-proxy"? Of course this would require changing other places as well. WDYT?

tiagobento (changes requested):
Made some comments inline! Overall, great stuff. Thanks for improving the Extended Services :D

ljmotta:
@tiagobento Thanks for your review. I've done your change requests.

tiagobento (approved):

PR #1341, opened 2022-11-23
kie-issues#34 Fix context entries rows generated names on paste

jomarko:
For more details see: 
- https://github.com/kiegroup/kie-issues/issues/34
- https://issues.redhat.com/browse/KOGITO-5920

This PR fixes just the basic reported scenario. The copy and paste into Context Expression will work correctly in a single scenario, and it is when ContextExpression contains single entry, defined as LiteralExpression. See the picture bellow

Screenshot from 2022-11-23 11-51-00

Other situations, when user tries to paste into ContextExpression with multiple context entries don't work even on the main branch.

To try the fix:
- Open some spreadhsheet tool like google spreadsheets
- Copy multiple cells of one column (Press 'Ctrl+C')
- Select the LiteralExpression of the ContextEntry-1
- paste value (Press 'Ctrl+V')

The existing issues, that are not addressed as part of this PR are:
- https://issues.redhat.com/browse/KOGITO-6436
- https://issues.redhat.com/browse/KOGITO-6437
- https://issues.redhat.com/browse/KOGITO-6438

### DIFF
--- a/packages/boxed-expression-component/src/components/ContextExpression/ContextExpression.tsx
+++ b/packages/boxed-expression-component/src/components/ContextExpression/ContextExpression.tsx
@@ -50,6 +50,7 @@ const DEFAULT_CONTEXT_ENTRY_NAME = "ContextEntry-1";
 const DEFAULT_CONTEXT_ENTRY_DATA_TYPE = DataType.Undefined;
 
 export const ContextExpression: React.FunctionComponent<ContextProps> = (contextExpression: ContextProps) => {
+  const existingEntryInfos: EntryInfo[] = [];
   const { i18n } = useBoxedExpressionEditorI18n();
   const { setSupervisorHash, boxedExpressionEditorGWTService, decisionNodeId } = useBoxedExpression();
 
@@ -179,13 +180,12 @@ export const ContextExpression: React.FunctionComponent<ContextProps> = (context
   );
 
   const onRowAdding = useCallback(
-    (existingRowsCount: number) => {
-      const generatedName = generateNextAvailableEntryName(
-        _.map(rows, (row: ContextEntryRecord) => row.entryInfo) as EntryInfo[],
-        "ContextEntry",
-        existingRowsCount + 1
-      );
-      return {
+    (allRows: DataRecord[]) => {
+      if (!existingEntryInfos.length) {
+        existingEntryInfos.push(...(_.map(allRows, (row: ContextEntryRecord) => row.entryInfo) as EntryInfo[]));
+      }
+      const generatedName = generateNextAvailableEntryName(existingEntryInfos, "ContextEntry");
+      const row = {
         entryInfo: {
           id: generateUuid(),
           name: generatedName,
@@ -198,8 +198,12 @@ export const ContextExpression: React.FunctionComponent<ContextProps> = (context
         editInfoPopoverLabel: i18n.editContextEntry,
         nameAndDataTypeSynchronized: true,
       };
+
+      existingEntryInfos.push(row.entryInfo);
+
+      return row;
     },
-    [i18n.editContextEntry, rows]
+    [i18n.editContextEntry]
   );
 
   const onRowsUpdate = useCallback(
--- a/packages/boxed-expression-component/src/components/ContextExpression/ContextExpression.tsx
+++ b/packages/boxed-expression-component/src/components/ContextExpression/ContextExpression.tsx
@@ -178,25 +178,29 @@ export const ContextExpression: React.FunctionComponent<ContextProps> = (context
     [contextExpression, spreadContextExpressionDefinition]
   );
 
-  const onRowAdding = useCallback(() => {
-    const generatedName = generateNextAvailableEntryName(
-      _.map(rows, (row: ContextEntryRecord) => row.entryInfo) as EntryInfo[],
-      "ContextEntry"
-    );
-    return {
-      entryInfo: {
-        id: generateUuid(),
-        name: generatedName,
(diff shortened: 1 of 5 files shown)


jomarko:
Reported by @fantonangeli , we have duplicated name in this scenario:
Peek 2022-11-23 12-28

packages/boxed-expression-component/src/components/ContextExpression/ContextExpression.tsx
```diff
@@ -50,6 +50,7 @@ const DEFAULT_CONTEXT_ENTRY_NAME = "ContextEntry-1";
 const DEFAULT_CONTEXT_ENTRY_DATA_TYPE = DataType.Undefined;
 
 export const ContextExpression: React.FunctionComponent<ContextProps> = (contextExpression: ContextProps) => {
+  const existingEntryInfos: EntryInfo[] = [];
```
jomarko: I see this line as the most problematic. When debugging, it is clear between two `onRowAdding` invocations, `existingEntryInfos` is never cleared, its content is just getting new and new elements.

While, everytime `onRowsUpdate` is invoked, `existingEntryInfos` is always empty.

Screenshot from 2022-11-24 09-36-51
Screenshot from 2022-11-24 09-36-20

fantonangeli (approved):

packages/boxed-expression-component/src/components/ContextExpression/ContextExpression.tsx
```diff
@@ -72,6 +72,8 @@ export const ContextExpression: React.FunctionComponent<ContextProps> = (context
     [contextExpression.contextEntries]
   );
 
+  const existingEntryInfos = useMemo<EntryInfo[]>(() => [], [rows]);
```
yesamer: @jomarko Can you please check this warning 

`React Hook useMemo has an unnecessary dependency: 'rows'. Either exclude it or remove the dependency array`

fantonangeli: Without that dependency, `existingEntryInfos` would not be re-initialized on `rows` changes.
I saw there we have also other components giving the same warning. 
Would you know a way to avoid this warning otherwise we can use the `// eslint-disable-next-line react-hooks/exhaustive-deps` comment?

tiagobento: When an unnecessary dependency is needed in a Hook, that's a very strong indicator that the problem is bigger. Ignoring the warning is ok, but does not fix the root cause. We need to dog deeper to find where's the wrong abstraction so that we don't need a Hook with an extra dependency.

fantonangeli: True, I used dependencies in the wrong way. 
Re-writing better the useMemo and dependencies the resulting code is much better and we don't need to receive `allRows` in `onRowAdding`.
Now I opened a PR on @jomarko branch and we both test it

jomarko: lint warning are fixed now, thank you @fantonangeli 

@yesamer @tiagobento feel free to re-review

yesamer: @jomarko @fantonangeli Much butter, thank you both. Reviewing it.

yesamer (approved):
LGTM

packages/boxed-expression-component/src/components/ContextExpression/ContextExpression.tsx
```diff
@@ -196,7 +198,11 @@ export const ContextExpression: React.FunctionComponent<ContextProps> = (context
       editInfoPopoverLabel: i18n.editContextEntry,
       nameAndDataTypeSynchronized: true,
     };
-  }, [i18n.editContextEntry, rows]);
+
+    existingEntryInfos.push(row.entryInfo);
```
tiagobento, line 202: Mutanting an array that is created on a `useMemo` is not something we should be doing. If this was necessary, it probably means that there's a bigger problem somewhere else. Most of the time we need to rethink our strategy to adapt to React's declarative model.

tiagobento (changes requested):
@jomarko Please take a look at my comment below. Thanks!

packages/boxed-expression-component/src/components/ContextExpression/ContextExpression.tsx
fantonangeli: The problem here is that `onRowAdding()` is called for each row before it is added to the `rows` variable.
The callstack is: `pasteOnTable()`--for each row--> `Table.onRowAddingCallback()`---> `onRowAdding()`.

As a solution I wanted to use a temporary array to store the rows info before they are added to `rows`, reinitializing it when `rows` is changed.
Using a normal constant for the array, without `useMemo`, the linter gives a warning.

Do we have a better solution, avoiding a refactoring that can impact all the components with the Copy&Paste functionality?

tiagobento: I'm not familiar with the boxed expression component (yet), so I won't be able to advise. What I can say is that there's a lot of things in the whole component architecture that does not conform with how React wants us to use its features. Scenarios like this one you're describing are a symptom of a much larger structural problem with it. I'm starting to look at the component as a whole trying to understand how we're going to change it to be easier to evolve and maintain, but right now, I'm not able to help much more than just pointing out these kind of things...

jomarko:
Waiting for @tiagobento's refactoring of the component

PR #1345, opened 2022-11-24
KOGITO-8280: preserve JSON properties that aren't defined as target for marshalling

treblereel:
@romartin possible fix

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/Marshaller.java
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/marshall/Marshaller.java
@@ -133,6 +133,7 @@ public class Marshaller {
     public Promise<ParseResult> unmarshallGraph(String raw) {
         try {
             workflow = parser.parse(mapper.fromJSON(raw));
+            MarshallerUtils.onPostDeserialize(raw, workflow);
         } catch (Exception e) {
             return promises.create(new Promise.PromiseExecutorCallbackFn<ParseResult>() {
                 @Override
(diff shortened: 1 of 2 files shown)


romartin:
@handreyrc can you please review this? Thx in advance!

handreyrc:
Hi @treblereel @romartin,

I tested the changes and It works, however, I have some concerns about having a post-deserialiser/serializer to merge the parsed object retrieved from the marshallers with the "original" content. 

- The merge process does not take into account the declaration order of properties. After reloading the JSON, undefined properties are forced to the bottom resulting in a different JSON even if no changes are made. It may be an inconvenience for the user otherwise somehow we may have to deal with it before presenting the reloaded JSON to the user again.
- Lets suppose that in the future, for some reason, we have some implementation that removes or replaces some specific property. These properties would come back to the final JSON after the merging process? 

IMO we should consider dealing with the undefined properties in the parse process itself instead of using a post processing approach. This way the workflow object is always the latest, no need to merge properties.

WDYT?

treblereel:
> Hi @treblereel @romartin,
> 
> I tested the changes and It works, however, I have some concerns about having a post-deserialiser/serializer to merge the parsed object retrieved from the marshallers with the "original" content.
> 
> * The merge process does not take into account the declaration order of properties. After reloading the JSON, undefined properties are forced to the bottom resulting in a different JSON even if no changes are made. It may be an inconvenience for the user otherwise somehow we may have to deal with it before presenting the reloaded JSON to the user again.
> * Lets suppose that in the future, for some reason, we have some implementation that removes or replaces some specific property. These properties would come back to the final JSON after the merging process?
> 
> IMO we should consider dealing with the undefined properties in the parse process itself instead of using a post processing approach. This way the workflow object is always the latest, no need to merge properties.
> 
> WDYT?

Thanks for your review, i have never saw such requirement, but if it's exist, we should definitely file an issue.

handreyrc:
Hey @treblereel,

I undertand your point. Anyway we can do those improvements in a new JIRA.
So I'm approving this PR.

Thanks for your work here!

handreyrc (approved):
Approved!!

Thanks

treblereel:
> Approved!!
> 
> Thanks

From my point of view, each property must be defined in the corresponding definition, but as I understand, specs arent finished yet.

romartin (approved):
Approving, although reported a new JIRA for tracking the missing issues: 

https://issues.redhat.com/browse/KOGITO-8299

Thanks!

PR #1346, opened 2022-11-24
Embed locally-built native image JIT Executor binary on Extended Services

gitgabrio:
@tiagobento @tiagodolphine @ljmotta  

Implement environment variable to embed locally-built jitexecutor-runner


packages/extended-services/env/index.js
```diff
@@ -34,6 +34,10 @@ module.exports = composeEnv([require("@kie-tools/root-env/env")], {
       default: "21345",
       description: "Extended Services port",
     },
+    EXTENDED_SERVICES__nativeBuildPath_macOS: {
```
tiagobento: I think `EXTENDED_SERVICES__jitExecutorNativeBinaryPath_macOS` would be a better name.

packages/extended-services/package.json
```diff
@@ -26,7 +26,7 @@
     "build:prod": "rimraf dist jitexecutor && pnpm copy-jitexecutor && pnpm build && pnpm pack-app",
     "build:win32": "make win32",
     "copy-jitexecutor": "run-script-os",
-    "copy-jitexecutor:darwin": "cp ./node_modules/@kie-tools/jitexecutor-native/dist/darwin/jitexecutor jitexecutor && chmod +x jitexecutor",
+    "copy-jitexecutor:darwin": "cp $(build-env extendedServices.jitexecutor.nativeBuildPath.macOS) jitexecutor && chmod +x jitexecutor",
     "copy-jitexecutor:linux": "cp ./node_modules/@kie-tools/jitexecutor-native/dist/linux/jitexecutor jitexecutor && chmod +x jitexecutor",
     "copy-jitexecutor:win32": "COPY /B /Y .\\node_modules\\@kie-tools\\jitexecutor-native\\dist\\win32\\jitexecutor.exe jitexecutor",
```
tiagobento: Can you please apply the same strategy for Windows and Linux?

packages/extended-services/env/index.js
```diff
@@ -42,6 +46,11 @@ module.exports = composeEnv([require("@kie-tools/root-env/env")], {
         kieSandboxUrl: getOrDefault(this.vars.EXTENDED_SERVICES__kieSandboxUrl),
         ip: getOrDefault(this.vars.EXTENDED_SERVICES__ip),
         port: getOrDefault(this.vars.EXTENDED_SERVICES__port),
+        jitexecutor: {
+          nativeBuildPath: {
```
tiagobento: I think `nativeBinaryPath` would be a better name..

tiagobento (changes requested):

packages/extended-services/package.json
gitgabrio: done

packages/extended-services/env/index.js
gitgabrio: done

packages/extended-services/env/index.js
gitgabrio: done

yesamer (approved):

yesamer:
@gitgabrio 

The system cannot find the file specified.
 ELIFECYCLE  Command failed with exit code 1.
undefined
D:\a\kie-tools\kie-tools\packages\extended-services:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @kie-tools/extended-services@0.0.0 build:prod: `rimraf dist jitexecutor && pnpm copy-jitexecutor && pnpm build && pnpm pack-app`
Exit status 1
Error: Process completed with exit code 1.

tiagobento:
Yeah, for Windows, the sub-command expression won't work with some extra tweaks. package.json scripts are ran on CMD by default, which doesn't have this syntax available. The solution is to invoke powershell there and change the copy command for the Power Shell equivalent.

gitgabrio:
> Yeah, for Windows, the sub-command expression won't work with some extra tweaks. package.json scripts are ran on CMD by default, which doesn't have this syntax available. The solution is to invoke powershell there and change the copy command for the Power Shell equivalent.

So ?

yesamer:
@gitgabrio not totally sure, but I think you should follow the same approach used in other modules (eg. stunner-editors package.json file)


Screenshot 2022-11-25 at 09 29 12

` change the copy command for the Power Shell equivalent.`
... and understand how to apply this :)

tiagobento:
> > Yeah, for Windows, the sub-command expression won't work with some extra tweaks. package.json scripts are ran on CMD by default, which doesn't have this syntax available. The solution is to invoke powershell there and change the copy command for the Power Shell equivalent.
> 
> 
> 
> So ?
> 
> 

What do you mean "So ?"?

gitgabrio:
> > > Yeah, for Windows, the sub-command expression won't work with some extra tweaks. package.json scripts are ran on CMD by default, which doesn't have this syntax available. The solution is to invoke powershell there and change the copy command for the Power Shell equivalent.
> > 
> > 
> > So ?
> 
> What do you mean "So ?"?

So: what should be done? I've no way to test on windows, and I'm not even experienced with power shell (almost always used linux on my career) - so : what should be done with it ?

tiagobento:
I trust you can find a way to execute a copy command on Power Shell without being experienced with it! 😎

As for testing, you can install a Virtual Machine, or even blindly push a command and the CI will run it for you. You can also do a custom GitHub Action on a repository in your fork that only runs a command on Windows using Power Shell.

gitgabrio:
> @gitgabrio not totally sure, but I think you should follow the same approach used in other modules (eg. stunner-editors package.json file)
> 
> Screenshot 2022-11-25 at 09 29 12
> 
> ` change the copy command for the Power Shell equivalent.` ... and understand how to apply this :)

NOt sure how can I "understand" without trying it. Beside, what I can't follow is that that PR simply replace a value with a variable, and everything is managed inside th ejs/go environment (no pnpm actually involved)
So, replacing the hardcoded path here

`COPY /B /Y .\node_modules\@kie-tools\jitexecutor-native\dist\win32\jitexecutor.exe`
 with a js. variable should not involve anything specific - maybe just the syntax should be tweaked a little bit, but can't test it

tiagobento:
NOTE: You can also use a platform-agnostic tool to do it as well, like 'cpr' or 'copyfiles'. We have plenty of examples of that in the repository.

gitgabrio:
@tiagobento could you pls approve/merge this ? Thanks!

ljmotta (approved):

tiagobento (approved):

PR #1350, opened 2022-11-29
Bump kie version to 8.31.0

yesamer:


yesamer:
Closing - I think it's better to keep Kogito and this kie version synchronized. 
kie-tools is currently using the 1.29.0 Kogito version, which depends on the 8.29.0 kie assets version,

PR #1352, opened 2022-11-29
kie-issues#46: Disable OK button properly in DMN Data Type Constraint modal

### LINKED ISSUE #46: DMN Editor: It is possible to save empty range constraint
The form for data types constraints of range type presents two fields as required, "start value" and "end value".

The "OK" button is disabled until both are filled out by the user. Anyway it's still clickable (or actionable via keyboard), letting the user save a constraint without values or with just one of the two.

When editing a saved constraint which only has one of the two limits, the "start value" and "end value" fields are presented empty.

Tested on Chrome 86 on Linux.

jomarko:
Closes https://github.com/kiegroup/kie-issues/issues/46


yesamer (approved):
LGTM

jomarko:
@tiagobento may I ask you for a merge?

PR #1353, opened 2022-11-30
KOGITO-8258: [SWF Editor] Implement click action over an error

fantonangeli:
Closes: https://issues.redhat.com/browse/KOGITO-8258

### Description

Please implement the click action over the error in the "Problems" tab in Webtool.

### Goal
By clicking over the error, the app should scroll the document and move the cursor to the line with the error.

### Note:
This doesn't affect the VSCode Extension that has its built-in errors list.

KOGITO-8258


packages/serverless-workflow-text-editor/src/envelope/ServerlessWorkflowTextEditorEnvelopeApiImpl.ts
```diff
@@ -42,11 +43,19 @@ export class ServerlessWorkflowTextEditorEnvelopeApiImpl
     super(serverlessWorkflowArgs, editorFactory);
   }
 
-  public kogitoSwfTextEditor__moveCursorToNode(args: { nodeName: string; documentUri: string }): void {
+  private getEditorOrThrowError() {
```
fantonangeli: @caponetto 
`getEditorOrThrowError()` is implemented also in "serverless-workflow-diagram-editor-envelope/src/envelope/ServerlessWorkflowDiagramEditorEnvelopeApiImpl.ts", does it make sense to move it to `KogitoEditorEnvelopeApiImpl`?

caponetto: Yes, I think it does.

fantonangeli: Thank you, I will modify it

packages/serverless-logic-sandbox/src/editor/EditorPage.tsx
```diff
@@ -352,6 +361,26 @@ export function EditorPage(props: Props) {
       .catch((e) => console.error(e));
   }, [workspaceFilePromise.data, editor, swfLanguageService, editorPageDock, i18n.terms.validation]);
 
+  const editorChannelApi = useMemo(
+    () =>
+      editor?.getEnvelopeServer()
+        .envelopeApi as unknown as MessageBusClientApi<ServerlessWorkflowCombinedEditorChannelApi>,
```
caponetto: This channel API shouldn't be created if the file is not SWF, right?

fantonangeli: You're right, thanks

caponetto (approved):
Great job, @fantonangeli!

tomasdavidorg (approved):
Looks good to me. 🙂

packages/serverless-workflow-standalone-editor/src/swf/channel/StandaloneServerlessWorkflowCombinedEditorChannelApi.ts
```diff
@@ -35,6 +35,7 @@ import {
 } from "@kie-tools/serverless-workflow-service-catalog/dist/api";
```
fantonangeli, line 35: @caponetto 
Because it's a channel implementation, I expected this file path to be something like this:
`impl/StandaloneServerlessWorkflowCombinedEditorChannelApiImpl.ts`
and not:
`channel/StandaloneServerlessWorkflowCombinedEditorChannelApi.ts`

Am I right?

caponetto: AFAIK we don't have a hard req about this kind of naming convention. 
So I think both are valid (though I am inclined to choose the `impl` one.

fantonangeli: Fine. 
I'm available to rename it or we do it during another activity

caponetto: Thanks! Let's keep it that way for now :)

PR #1357, opened 2022-12-02
KOGITO-7736: Stunner Editors - Performance improvements 

yesamer:
Part of https://github.com/kiegroup/kie-issues/issues/47 initiative

Scope of this PR is to apply the same changes introduced in SWF Editor into BPMN/DMN/SCESIM stream. So, a copy of https://github.com/kiegroup/kie-tools/pull/1248 (not a 1:1 copy)

WORK IN PROGRESS!


handreyrc:
Hi @yesamer, 

The commits below are not applicable to BPMN/DMN:
- 7 canvas for each lienzo panel - reduce number of instances in use:  We had 8 canvases in swf because we added an extra for gray background so this one was removed. However, about the 7 others remaining, they are supporting multiple distinct functionalities, from drag and drop to preview. One alternative would be to merge those layers but I personaly don't see benefit in having a super complex layer without warranties it would improve the performance. Currently SWF has the same 7 layers we have in BPMN/DMN.
- Replace ImageLoader for ImageBitmap: In SWF all the SVG processing was removed now images are loaded directly into canvas. In bpmn images are not loaded directly into diagram (except toolbar image strips). They are SVGs processed and coverted to multipath. So no immediate benefit here.

The remaining changes are focused in reducing the refresh calls overhead, actually it saves quite a few time alone.

The performance results of the original PR in SWF are the sum of all these changes. 
In my analysis I don't see much of loading time improvement when looking at the applicable changes to BPMN/DMN domain.

WDYT?

yesamer:
@handreyrc Thank you for your analysis and support. During my testing phase, I didn't notice significant performance improvements, which confirms your analysis. Based on these facts, it doesn't worth cherrypicking these changes in BMPN/DMN/SCESIM stream. Closing.

PR #1361, opened 2022-12-08
Remove dead code of linked grid columns on Stunner Editors

jomarko:
Uberfire grids framework contains possibility of displaying multiple grids in single canvas while columns of these grids could be linked, in other words, there was a line rendered between column pairs.

This feature has historical origin in guided decision tables graphs. The current kie tooling doesn't engage similar feature thus maintaining such 'dead code' is not an effective.


yesamer (approved):
@jomarko Thank you

PR #1365, opened 2022-12-09
KOGITO-8317: Show condition names, instead of expressions, when possible

hasys:
[KOGITO-8317](https://issues.redhat.com/browse/KOGITO-8317)

Hello @romartin, @LuboTerifaj,

There is a PR for KOGITO-8317 it gives priority to show condition name for data switch over condition value itself, since values mostly too long to see it property, but names are easily configurable to have better reading options.

Also default branch color made more contrast to distinguish it easier at first glance at the diagram.

To improve visibility differences for connections, I also did them a bit thicker and decreased the tip size.

See screenshots: 

image

image

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/TransitionShape.java
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/TransitionShape.java
@@ -19,6 +19,7 @@ package org.kie.workbench.common.stunner.sw.client.shapes;
 import org.kie.workbench.common.stunner.client.lienzo.shape.impl.ShapeStateDefaultHandler;
 import org.kie.workbench.common.stunner.core.client.shape.common.DashArray;
 import org.kie.workbench.common.stunner.core.client.shape.impl.ConnectorShape;
+import org.kie.workbench.common.stunner.core.util.StringUtils;
 import org.kie.workbench.common.stunner.sw.definition.ActionTransition;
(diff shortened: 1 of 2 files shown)


romartin (approved):

LuboTerifaj (commented):
Hello @hasys , @romartin ,

Why we are changing the colour of default condition and the width of transitions? From my pov it doesn't look good since the arrows are now wider than state edges, the colour is too light and intensive. See also my [comment](https://issues.redhat.com/browse/KOGITO-8317?focusedCommentId=21395337&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21395337).

The found issue: Name doesn't work in eventConditions.

I also found the issues regarding label hovering, but I reported it in separate jira as it is not related to the PR's changes:
- [KOGITO-8346](https://issues.redhat.com/browse/KOGITO-8346) _[SWF Diagram Editor] Displaying the string on hover over label is not formatted properly_

If we don't want to introduce the label name change based on `name` field in `eventConditions` and you are okay with the width and colour transition changes I mentioned above, feel free to merge the PR.

Thank you!

hasys:
Hi @LuboTerifaj,

great review, thank you. 

We changed colors and thickness because it was hard to distinguish connections one from another, now it is much easier. Connections now aren't wider than shape border, it's the same value.

About eventConditions it is correct for now, since event names are much readable than data conditions. And addition visualization about other properties will be done in [KOGITO-7072](https://issues.redhat.com/browse/KOGITO-7072).

Thank you!

romartin:
**Merging** as from Lubo also proposed, as this looks better than before, even there might be some more improvements

romartin:
Hey @hasys 

Can you please sync with @LuboTerifaj and report a new ticket, in case of any improvement pending around labels.

Thanks!

PR #1370, opened 2022-12-14
KOGITO-8277: FunctionRef.arguments could be String or JsObject

treblereel:


romartin (commented):
Code looks good! Thanks @treblereel 

Now going to test a little bit the UI.

@LuboTerifaj can you please prioritize this blocker issue? 

Thanks in advance!

romartin (approved):
Examples tested. All good, thanks @treblereel !

treblereel:
> Examples tested. All good, thanks @treblereel !

And would like to repeat: there would be no issue like this if we had the correct specs.

romartin:
@treblereel I don't think this is due to not having the whole spec declared in the model. The `arguments` have to be an `Object` anyway, in  order to support either string or a another object as well, but is a fix specific to this field. Isn't it?

LuboTerifaj (commented):
Hello @treblereel ,

Could you please add tests? There are no any in this PR.

The issue present also in previous release:
- There is a warning  message when using string or array as parameter: `Incorrect type. expected "object".`
It should be addressed in the follow-up jira.

Otherwise, I didn't find any issues, respectively regressions, so only tests are missing.

Thanks!

romartin:
Hey @LuboTerifaj !

Yes, we have plans to incorportate all those examples from the ticket as integration tests, but I was suggesting @treblereel to do it i n a separate PR. This way we can just cherry-pick this fix to the release branch (deadline is tomorrow), and tests will be added during the next following days, into the main branch.

@LuboTerifaj Saying this, can you please approve it, so we can merge?

BTW @treblereel can you pleaes cherry-pick thiis to this branch: https://github.com/kiegroup/kie-tools/tree/0.26.0-prerelease ? We need this merged on the branch for tomorrow.

Thanks both guys!!

LuboTerifaj (approved):
Hello @romartin , @treblereel ,

The PR is missing not only the integration tests, but also unit tests.

However, I am okay with implementing them in the following PR since this is the blocker issue.

Approving.

PR #1390, opened 2022-12-27
KOGITO-7533: Keyboard navigation though workflow states

Josephblt:
**JIRA:** [KOGITO-7533 - Keyboard navigation though workflow states](https://issues.redhat.com/browse/KOGITO-7533)

Node Navigation using arrows is now possible. Use the up, down, left, and right keys to navigate. Left and Right keys navigate through sibling nodes. Up navigates to source nodes and down navigates to target nodes. When using up or down it tries to go through the path it came from in cases where there are multiple targets or sources. In VSCode the text editor still selects the line corresponding to the selected node, but it does not steal the focus from the canvas.

**VS Code Extension:** [Serverless Workflow](https://drive.google.com/file/d/1G2jAAgrmIzfaopnbEvz6YHpU-up3afYW/view?usp=share_link)

### DIFF
--- a/packages/serverless-workflow-diagram-editor/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/select/AbstractSelectionControl.java
+++ b/packages/serverless-workflow-diagram-editor/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-commons/kie-wb-common-stunner-client-common/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/controls/select/AbstractSelectionControl.java
@@ -16,8 +16,10 @@
 
 package org.kie.workbench.common.stunner.core.client.canvas.controls.select;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


Josephblt:
Hello @romartin. All set here. Ready for review. Thanks.

LuboTerifaj (changes requested):
Hello @Josephblt , @romartin 

Here are the issues I found:
	
- If there are more incoming/outgoing connections, it is not possible to choose a path, so you cannot navigate through nodes easily.
  - Consider the following workflow:
  <details>
  <summary>key-navigation.sw.json</summary>

  ```json
  {
      "id": "key-navigation",
      "version": "1.0",
      "specVersion": "0.8",
      "name": "Key navigation",
      "description": "Key navigation",
      "start": "State1",
      "states":[
        {
           "name":"State1",
           "type":"inject",
           "data": {
              "result": "Hello World!"
           },
           "end": true
        },
        {
         "name":"State3",
         "type":"inject",
         "data": {
            "result": "Hello World"
         },
         "end": true
      },
      {
          "name":"State2",
          "type":"inject",
          "data": {
             "result": "Hello World!"
          },
          "transition": "State3"
       },
       {
         "name": "ConfirmBasedOnFunds",
         "type": "switch",
         "dataConditions": [
           {
             "condition": "${ .funds | .available == \"true\" }",
             "transition": "SendPaymentSuccess"
           },
           {
             "condition": "${ .funds | .available == \"false\" }",
             "transition": "SendInsufficientResults"
           },
           {
             "condition": "${ .funds | .available == \"false\" }",
             "transition": "OtherState"
           }
         ], 
         "defaultCondition": {
           "transition": "SendPaymentSuccess"
         }
       },
       {
         "name":"SendPaymentSuccess",
         "type":"inject",
         "data": {
            "result": "Hello World!"
         },
         "transition": "State5"
      },
      {
         "name":"SendInsufficientResults",
         "type":"inject",
         "data": {
            "result": "Hello World!"
         },
         "transition": "State5"
      },
      {
         "name":"OtherState",
         "type":"inject",
         "data": {
            "result": "Hello World!"
         },
         "end":true
      },
      {
         "name":"State5",
         "type":"inject",
         "data": {
            "result": "Hello World!"
         },
         "transition": "State5"
      }
      ]
      }
  ```
  </details>

The loops are being made when navigating through nodes without one direction path.
- For example monitor-job.sw.json, car-vitals-checks-parent.sw.json, continuing-as-a-new-execution.sw.json from specification, or example above.

Not always the correct node is selected when pressing the right or left arrow key.
- Select JobSucceeded or JobFailed state and press keys in monitor-job.sw.json from specification
	
Arrow key sometimes doesn't work as expected.
- Select SendPaymentSuccess state and press right arrow key in reusing-function-and-event-definitions.sw.json from specification

Possible enhancements:
- It is not possible to navigate through states that are not connected. It would be great if it was possible to navigate between states/nodes on the same level
  - It is also not possible to navigate from start/end at the same level.
  - Probably out of scope of this PR
- If you press and hold the key, the selection is changed only once.
  - This could be added in this PR, although not necessary.

Thanks!

LuboTerifaj:
Hey @Josephblt , @romartin , please see also my [jira-comment](https://issues.redhat.com/browse/KOGITO-7533?focusedCommentId=21542476&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21542476). Thanks!

ederign:
@romartin one thing that crossed my mind, is that we should figure out a way to allow users to override the shortcuts (probably on settings) in case that those conflict with other shortcuts. (I'm worried especially for the tab)

Maybe we can do this in a FUP task , wdyt?

ederign:
Also, is there a way to switch from the text editor to the canvas selected?

romartin:
Hey @ederign 

This task assumes that it possible to swtich the focus between the canvas and the text editor by using a key combination _Shift+Tab_

Anyway, IMO makes sense to be able to customize the shortcuts. Should I create a ticket for it, or something already exists, do you know? (I understand this is not just at stunner editor level, it is also about the whole tooling impl)

PR #1392, opened 2022-12-28
kie-issues#77: DMN Editor: It is not possible to add constraints to nested data types

### LINKED ISSUE #77: DMN Editor: It is not possible to add constraints to nested data types
Steps to reproduce:

1 - Click on the tab Data Types
2 - Create a new data type
3 - Add a nested String data type to it
4 - Click on the edit icon then on add constraints icon  
5 - Select  Enumeration as the constraint type

The following errors are being presented in the VSCode console, and it is not possible to define enumeration items:
image-2022-12-19-15-14-38-040

image-2022-12-19-15-15-11-647

yesamer:
Closes https://github.com/kiegroup/kie-issues/issues/77

This regression is a consequence of GWT 2.10 update: https://github.com/kiegroup/kie-tools/pull/1330.
As I mentioned:
This commit https://github.com/gwtproject/gwt/commit/657ae98c5d8bb20b67639d3175cf843e86ce3cda in GWT repo, fixed the String#substring emulation behavior.
Therefore, `substring` usage can potentially throw an exception in case of bad conditions.
In this case, the `substring` method throws an exception when the given string is empty. 

All `substring` usage can lead to this behavior, thus we need to pay great attention to that.

### DIFF
--- a/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/string/StringSelector.java
+++ b/packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/constraint/common/typed/string/StringSelector.java
@@ -33,7 +33,7 @@ public class StringSelector extends BaseSelector {
 
     @Override
     public void setValue(final String value) {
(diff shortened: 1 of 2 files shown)


jomarko (approved):
Thank you @yesamer . The regression seems to be fixed to me. I have manually checked string custom type as:
- top level custom type, that is enhanced string with constraint, not a structure
- second level custom type, that is a string field with constraint of a more complex structure

PR #1400, opened 2023-01-04
Fix build:dev to generate arm artifacts

ederign:
Fix build:dev to generate arm and amd platforms depending on OS.

This also fixed debug in ARM + VS Code

### DIFF
--- a/packages/kn-plugin-workflow/Makefile
+++ b/packages/kn-plugin-workflow/Makefile
@@ -33,16 +33,16 @@ LDFLAGS                       := "-X $(SET_QUARKUS_PLATFORM_GROUP_ID) -X $(SET_Q
 ARCH := $(shell uname -m)
 ifeq ($(ARCH),arm64)
 	GOARCH = arm64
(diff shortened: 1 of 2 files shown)


packages/kn-plugin-workflow/Makefile
```diff
@@ -33,14 +33,16 @@ LDFLAGS                       := "-X $(SET_QUARKUS_PLATFORM_GROUP_ID) -X $(SET_Q
 ARCH := $(shell uname -m)
 ifeq ($(ARCH),arm64)
 	GOARCH = arm64
+	BIN_BUILD-DARWIN=$(BIN_DARWIN_ARM64)
```
ljmotta: Use _ instead of -

ljmotta (changes requested):
Hi @ederign Thanks for your PR. I've made just a comment inline.

packages/kn-plugin-workflow/Makefile
ederign: Fixed! Thanks!

ljmotta (approved):

PR #1410, opened 2023-01-13
kie-issues#31: DMN/SceSim - Docks and tabs are visible in case of an error on `setContent`

### LINKED ISSUE #31: DMN/SceSim - Docks and tabs are visible in case of an error on `setContent`

yesamer:
Closes https://github.com/kiegroup/kie-issues/issues/31

Before:
Screenshot from 2021-04-06 08-08-40

After:
Screenshot 2023-01-16 at 20 43 22

**All PR changes**
- In case of unexpected error during the file unmarshalling phase, due to an invalid file or not supported version, an error page is shown. Before the changes, this error page contained the Tab ("Editor"), the search button, and the docks on the right. With provided changes, now these three widgets are removed, so the user can see an "empty" and not actionable page. To achieve that, I basically introduced a set of new APIs to interact with the Tab bar, the search button, and the Docks
- The same changes apply to Test Scenario. Bear in mind that editors share only a subset of the architecture, so the points where I applied the changes are not the same.
- The error page is a bit different. An error icon is present and no more a triangle warning.
- I added an additional message containing the reason for the issue, so the user has more clue about the error, as requested  [here](https://stackoverflow.com/questions/74882299/how-to-find-the-vscode-error-when-reading-dmn-file)
- I moved the `DiagramNavigationExplorer` dock to the EAST side in ALL channels. Before the change, the dock was on the EAST side in VSCODE, KIE_SANDBOX, and WEST in the STANDALONE channel. I did this to unify the editor UI in all channels and to simply code.
- During Docks management, I discovered that one of the docks (the Preview dock) was implemented in `kie-wb-common-dmn-webapp-common` module. That module contains that class **ONLY**. It doesn't make sense to have a module for a single class, so I moved that into `kie-wb-common-dmn-client` and removed the useless module.
- Removed code smells (redundant `eq()` mostly)


yesamer:
@jomarko Can you please review it?

jomarko:
### Manual checks
#### Online editor (DMN)
- [x] Open in a row valid, invalid, valid files
- [x] Open, close and reopen invalid file
- [x] Open and close invalid file, open valid file
- [x] Resize screen for invalid file

#### VS Code (DMN, SceSim)
- [x] Open in a row valid, invalid, valid files
- [x] Open, close and reopen invalid file
- [x] Open and close invalid file, open valid file
- [x] Resize screen for invalid file

packages/stunner-editors/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationViewImpl.java
```diff
@@ -45,6 +49,29 @@ public void setContentWidget(final Widget widget) {
...
+     */
+    @Override
+    public void setScenarioTabBarVisibility(boolean visible) {
+        /* The element className is bound with "TabPanelWithDropdowns.ui.xml" */
+        HTMLCollection<Element> elements = DomGlobal.document.getElementsByClassName("uf-tabbar-panel-nav-tabs");
+        if (elements.length == 1) {
+            HTMLDivElement element = (HTMLDivElement) elements.getAt(0);
+            element.style.display = visible ? "" : "none";
+        }
+    }
```
jomarko, line 73: Just a question, shouldn't we implement similar logic rather in `GwtEditorWrapper.tsx` ? Asking just because there it seems to me we manipulate GWTEditors panels due to some specifics in kogito channels.

jomarko (commented):
Thank you a lot @yesamer . I always like similar efforts of unification of things and removal of obsolete things.

### Manual review
I have single manual review observation. There is still that icon that opens keyboard shortcuts cheetsheat. It is not blocking frommy approval. Just bringing it to your attention to be sure you are ware of it.
Screenshot_20230125_121358


### Code review
I think we can go further with the 'west' panel placement removal as: https://github.com/yesamer/kie-tools/pull/4

packages/stunner-editors/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationViewImpl.java
yesamer: I see your point. Just consider that block of code is for SCESIM only, the APIs I wrote are working well for DMN. Considering our future plans for SCESIM, probably it doesn't worth it.  In addition, that solution increases the coupling between GWT and JS Modules, so I would personally avoid that solution, but I'm open to discussions :)

yesamer:
> I have single manual review observation. There is still that icon that opens keyboard shortcuts cheetsheat. It is not blocking frommy approval. Just bringing it to your attention to be sure you are ware of it. Screenshot_20230125_121358

Thank you for pointing that out. IIRC that component is not defined in the GWT layer, thus we should manage that in the JS layer, in this case

yesamer:
@jomarko Can you please review this PR? Thx

jomarko (approved):
Thank you @yesamer

PR #1413, opened 2023-01-18
KOGITO-8404: Dashbuilder VSCode extension: Auto Complete

AjayJagan:
adding autocomplete support to vscode extension of dashbuilder
JIRA: https://issues.redhat.com/browse/KOGITO-8404

### DIFF
--- a/packages/dashbuilder-editor/src/monaco/augmentation/language/schemas/dashbuilderSchema.ts
+++ b/packages/dashbuilder-editor/src/monaco/augmentation/language/schemas/dashbuilderSchema.ts
@@ -299,8 +299,8 @@ export const DASHBUILDER_SCHEMA = {
         pattern: {
           type: "string",
         },
-        required: ["id", "name", "expression", "pattern"],
       },
+      required: ["id", "name", "expression", "pattern"],
       title: "SettingsColumn",
     },
     SettingsTable: {
--- a/packages/dashbuilder-editor/src/monaco/augmentation/language/schemas/dashbuilderSchema.ts
+++ b/packages/dashbuilder-editor/src/monaco/augmentation/language/schemas/dashbuilderSchema.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2023 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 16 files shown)


AjayJagan:
need to wait for this minor fix: https://github.com/jesuino/dashbuilder-website/pull/10 .. before we can start testing :)

jesuino:
Hello Ajay,

Thank you so much for this PR! Your work is looking great and I really appreciate your hard work on it!

I can see auto complete working very well, just a few thoughts:

1) I see that some minor problems are displayer as errors - would it be possible to set as WARNINGS instead? Here's for example, this works as String or number, so not an issue if this is a String

image


2) In properties it says that the property should be String, which is not true - it can be any primitive type:

image


3) This also could be a warning because lower or upper case would work

image


4) In `settings`, lookup is indeed obligatory, but it may be defined globally, inside `global`, so it would be great to have it as a warning as well instead error

5) In the case showed on the image below baseUrl is not required and width/height for external are not numbers:

image


6) groupFunctions also work as "functions" - I would prefer to validate to functions because it is what we are going to use

image


7) In this case I don't know why it complains about the fields under "echarts" - this is a free object where users can add any property to the component - so we should not have validation under the component configuration

image


8) the fields html and javascript under displayer/html are indeed String - I am not sure why it complained:

image

That's all for now. 

Thanks!

jesuino (changes requested):
Please see my comment!

fantonangeli (approved):
Thank you @AjayJagan. 
From the code viewpoint, this work looks great!

jesuino:
Hello @AjayJagan ,

Thanks for your changes, it is working great now!

I will approve this PR and suggest some changes to the schema:

1) For `axis -> x` or `axis -> y `any of the properties are required:

image

I know that you did this because the auto complete when adding the property is really great

2) For sort we made a change to sortOrder to "order"

image

3) On settings.chart.zoom it is actually a boolean or string 

image

4) On settings.columns only ID is required

image

5) For settings.filter only enabled is required

image

6) With lookup uuid is indeed is required - but sometimes we define it globally in global.settings - but I don't know if this is possible to solve on the schema - if does it is not possible to solve don't worry

image

7) Finally I still see message as ERRORS, but it would be good it it was as WARNING - if it is not possible just let me know 

We can approve and merge this PR and do these improvements to the schema in parallel

Thanks!

jesuino (approved):
LGTM!

paulovmr (approved):

PR #1421, opened 2023-01-25
Bump `tycho` to `2.7.5` version in VSCode Java Code Completion Extension Plugin 

yesamer:
Updating `tycho` to `2.7.5` version, which is the last supported version for JDK 11. Newer versions (3.0.0+) require JDK 17.
The reason for that is just to keep the component more stable, relying on an updated `tycho` version. 

Updated `jacoco` as well, just to keep it aligned with `stunner-editors`

This change impacts the Java Code Completion Extension component only.


jomarko (approved):
Tried to finish tutorial on https://blog.kie.org/2022/05/dmn-types-from-java-classes.html using build artifacts of this PR. No troubles spotted.

PR #1426, opened 2023-01-26
kie-issues#26: DMN Runner enhancements and bug fixes

### LINKED ISSUE #26: DMN Runner enhancements and bug fixes
This issue groups DMN Runner enhancements and bug fix.

### General

- [x] Enable "Delete inputs" button.
- [x] Modal doesn't open when click on “connected icon”

### Persistence

- [x] KOGITO-7044 DMN Runner persistence doesn't work correctly between files
- [x] KOGITO-7328 DMN Runner form input is lost if form field is in focus when creating a new file
- [x] KOGITO-7515 DMN Runner inputs invalidate the workspace if they're wrongly formatted

## **Enhancements**

- [ ] Make DMN Runner panels open and close using a query param (e.g. ?runner={table|form}&row=1)
- [x] Create DMN Language Service

ljmotta:
Closes https://github.com/kiegroup/kie-issues/issues/26

## On this PR
This PR solves some bugs and introduce some enhancements for the DMN Runner.

### general fixes
 - Renders the DMN Runner form and table after the Editor is ready

https://user-images.githubusercontent.com/24302289/217062796-4ba298a6-86c9-49b9-9123-794bc1ced5ef.mov

 - [FORM ONLY]  Loading and deleting inputs will refresh the form;

https://user-images.githubusercontent.com/24302289/217061606-15274ac9-84ff-44ac-9316-3bb3b6b11035.mov

### imported models
 - Only the imported models of the current DMN are sent to the JIT executor.
 - Problems dock was enhanced so it could show errors from imported models.

https://user-images.githubusercontent.com/24302289/217061552-34a4bd46-1d40-42cd-b9b1-fa350b0d74ae.mov

### extended-services
 - Removed "Connected" modal;
 - Opens the Extended Services settings after clicking the "connected" icon;

https://user-images.githubusercontent.com/24302289/217061447-59b1a15c-218c-42a6-a68a-b5f488929189.mov

 - Stopping the Extended Services closes the DMN Runner table and disables the Run button;

https://user-images.githubusercontent.com/24302289/217061809-6e2b213a-7bc1-4ac0-a98a-d07ed47b676e.mov

## [EDIT] Removed query params
The query params are extensively used in KIE Sandbox, adding the DMN Runner query params is a challenge as we have many `useEffect` hooks that rely on them. All `useEffect` hooks are triggered simultaneously with the current value, so changing a query param in one of these `useEffect` will be overridden by another `useEffect` that was already scheduled. This process makes adding a new query param not trivial.


packages/dmn-language-service/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ResourceContent } from "@kie-tools-core/workspace/dist/api";
+import { WorkspacesContextType } from "@kie-tools-core/workspaces-git-fs/dist/context/WorkspacesContext";
+
+const DMN_IMPORT = "dmn:import";
+const DMN_INPUT_DATA = "dmn:inputData";
```
porcelli: the use of "dmn:" hard coded xml namespace prefix can be fragile as you could have a valid DMN file without namespace prefix.

ljmotta: I've changed to check any `import` tag that matches a regexp: `([a-z]*:)?import`. This way we can cover the cases the DMN file uses another prefix or even don't use any prefix.

jomarko (commented):
@ljmotta please let us know, once this is ready for review. I have briefly checked tabular form of the DMN runner.
- It has two header rows while first is completley empty
- There is no horizontal scrollbar
- I was not able to see execution problems
Screenshot_20230130_021711
Screenshot_20230130_021700

ljmotta:
@jomarko Hi, thanks for testing.
> It has two header rows while first is completley empty

The DMN Runner Table shows the first row to render to complete the table

> There is no horizontal scrollbar
> I was not able to see execution problems

As far I could see this DMN model is not supported in our table component. But I've created an issue (https://github.com/kiegroup/kie-issues/issues/99) so we could make a fix for it.

ljmotta:
@jomarko Hi. Thanks for testing!

> It has two header rows while first is completley empty

This happens so we can allign it with the Input table.

> There is no horizontal scrollbar

I've already made a fix for it, this happened because the Input table was broken.

> I was not able to see execution problems

This DMN is not been fully supported in the latest release, so I've created a new issue so we can keep track of it. https://github.com/kiegroup/kie-issues/issues/99

packages/dmn-language-service/src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  content: string;
+  path: string;
+}
+
+export class DmnLanguageService {
+  private readonly parser = new DOMParser();
+  private readonly importTagRegExp = new RegExp(`([a-z]*:)?(${IMPORT})`);
+  private readonly inputDataRegEx = new RegExp(`([a-z]*:)?(${INPUT_DATA})`);
+
+  constructor(private readonly readDmn: (importedModel: string) => Promise<DmnLanguageServiceImportedModel>) {}
```
tiagobento: This should be an object parameter so we know what is this function we're giving to the constructor.

tiagobento: Also, `importedModel` is not a good name... especially because this is a `string`. What kind of thing should I expect from this string? a path? a uuid?

packages/form/src/FormHook.tsx
```diff
@@ -150,7 +150,7 @@ export function useForm<Input extends Record<string, any>, Schema extends Record
   useEffect(() => {
     const newFormModel = cloneDeep(formInputs);
     setFormModel(newFormModel);
-  }, [name]);
+  }, [formInputs]);
```
tiagobento, line 153: 🙏

packages/online-editor/src/editor/DmnRunner/DmnRunnerContext.tsx
```diff
@@ -24,23 +24,19 @@ export interface DmnRunnerContextType {
...
-  didUpdateOutputRows: boolean;
   status: DmnRunnerStatus;
 }
 
 export interface DmnRunnerCallbacksContextType {
   preparePayload: (formData?: InputRow) => Promise<KieSandboxExtendedServicesModelPayload>;
-  setExpanded: React.Dispatch<React.SetStateAction<boolean>>;
   setError: React.Dispatch<React.SetStateAction<boolean>>;
-  setCurrentInputRowIndex: React.Dispatch<React.SetStateAction<number>>;
+  setExpandedLocation: (isExpanded: boolean) => void;
```
tiagobento: I don't understand how this receives a boolean.. I don't think this name matches its function.

packages/online-editor/src/editor/DmnRunner/DmnRunnerContext.tsx
```diff
@@ -24,23 +24,19 @@ export interface DmnRunnerContextType {
...
   setInputRows: React.Dispatch<React.SetStateAction<Array<InputRow>>>;
-  setDidUpdateInputRows: React.Dispatch<React.SetStateAction<boolean>>;
-  setDidUpdateOutputRows: React.Dispatch<React.SetStateAction<boolean>>;
-  setMode: React.Dispatch<React.SetStateAction<DmnRunnerMode>>;
+  setParams: (mode: DmnRunnerMode, expand?: true, row?: number) => void;
```
tiagobento: `setParams` is too broad. There are some types of params when we're dealing with URLs. `setQueryParams` would be better. However, in this case, I think this could be the former `setMode`. Consumers of DMN Runner components don't need to know how it works (in this case, with queryParams), they only care about the functionality.

packages/online-editor/src/editor/DmnRunner/DmnRunnerDrawerPanelContent.tsx
```diff
@@ -322,16 +318,15 @@ export function DmnRunnerDrawerPanelContent(props: Props) {
   const onAddNewRow = useCallback(() => {
     dmnRunnerDispatch.setInputRows((previousData: Array<InputRow>) => {
       const newData = [...previousData, {}];
-      dmnRunnerDispatch.setCurrentInputRowIndex(newData.length - 1);
+      dmnRunnerDispatch.setParams(DmnRunnerMode.FORM, true, newData.length - 1);
```
tiagobento: Floating boolean argument does not make code very readable. Please consider giving it an object parameter. Like this:

```suggestion
      dmnRunnerDispatch.setParams({ mode: DmnRunnerMode.FORM, expand: true, row: newData.length - 1 });
```

packages/online-editor/src/editor/DmnRunner/DmnRunnerDrawerPanelContent.tsx
```diff
@@ -475,7 +466,7 @@ export function DmnRunnerDrawerPanelContent(props: Props) {
                       <Text component={"h3"}>{i18n.terms.outputs}</Text>
                     </TextContent>
                     {dmnRunnerStylesConfig.buttonPosition === ButtonPosition.OUTPUT && (
-                      <DrawerCloseButton onClick={(e: any) => dmnRunnerDispatch.setExpanded(false)} />
+                      <DrawerCloseButton onClick={() => dmnRunnerDispatch.setExpandedLocation(false)} />
```
tiagobento: Yeah.. I don't really understand what `setExpandedLocation(false)` means..

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
```diff
@@ -109,13 +125,40 @@ export function DmnRunnerProvider(props: PropsWithChildren<Props>) {
...
+    }
+
+    if (queryParams.has(QueryParams.DMN_RUNNER_MODE)) {
+      const mode = queryParams.getString(QueryParams.DMN_RUNNER_MODE);
+      if (mode === "form") {
+        setMode(DmnRunnerMode.FORM);
+      }
+      if (mode === "table") {
+        setMode(DmnRunnerMode.TABLE);
+      }
```
tiagobento: Could you be using `DmnRunnerMode` for comparison instead of plain strings?

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
```diff
@@ -109,13 +125,40 @@ export function DmnRunnerProvider(props: PropsWithChildren<Props>) {
...
+    }
+
+    if (queryParams.has(QueryParams.DMN_RUNNER_ROW)) {
+      const row = queryParams.getNumber(QueryParams.DMN_RUNNER_ROW);
+      if (row !== undefined) {
+        setCurrentInputRowIndex(row);
+      }
```
tiagobento: What if it IS `undefined`?

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
```diff
@@ -109,13 +125,40 @@ export function DmnRunnerProvider(props: PropsWithChildren<Props>) {
   }, [extendedServices.status, extendedServices.client, props.workspaceFile.extension, preparePayload]);
 
   useEffect(() => {
+    if (queryParams.has(QueryParams.DMN_RUNNER_IS_EXPANDED)) {
+      const isExpanded = queryParams.getBoolean(QueryParams.DMN_RUNNER_IS_EXPANDED);
+      if (isExpanded !== undefined) {
+        setExpanded(isExpanded);
+      }
```
tiagobento: What if it IS `undefined`?

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
```diff
@@ -45,46 +46,61 @@ export function DmnRunnerProvider(props: PropsWithChildren<Props>) {
...
   const [error, setError] = useState(false);
   const [jsonSchema, setJsonSchema] = useState<DmnSchema | undefined>(undefined);
   const [isExpanded, setExpanded] = useState(false);
   const [mode, setMode] = useState(DmnRunnerMode.FORM);
   const [currentInputRowIndex, setCurrentInputRowIndex] = useState<number>(0);
+  const dmnLanguageService = useMemo(() => {
+    return new DmnLanguageService(async (importedModel: string) => {
+      const resourceContent = await workspaces.resourceContentGet({
+        workspaceId: props.workspaceFile.workspaceId,
+        relativePath: importedModel,
```
tiagobento: Ok, it's a path.. so in this case, `importedModel` -> `importedModelRelativePath`.

packages/online-editor/src/editor/EditorPageDockDrawer.tsx
```diff
@@ -94,14 +98,25 @@ export const EditorPageDockDrawer = React.forwardRef<
...
+        query = query.without(QueryParams.PROBLEMS_IS_EXPANDED).without(QueryParams.DMN_RUNNER_IS_EXPANDED);
       }
-      return PanelId.NONE;
-    });
-  }, []);
+
+      history.replace({ search: query.toString() });
+    },
+    [panel, history, queryParams]
+  );
```
tiagobento: This code is really hard to read, maybe it would benefit from some comments.

packages/online-editor/src/editor/EditorPageDockDrawer.tsx
```diff
@@ -151,8 +166,24 @@ export const EditorPageDockDrawer = React.forwardRef<
...
+        setPanel(PanelId.NOTIFICATIONS_PANEL);
+        return;
+      }
+      const isRunnerExpanded = queryParams.getBoolean(QueryParams.DMN_RUNNER_IS_EXPANDED);
+      const mode = queryParams.getString(QueryParams.DMN_RUNNER_MODE);
+      if (mode === "table" && isRunnerExpanded === true) {
+        setPanel(PanelId.DMN_RUNNER_TABLE);
+        return;
+      }
+    }
```
tiagobento: Same here..

tiagobento (changes requested):
Please consider my comments inline, and some other questions...


```
runnerIsExpanded=[true|false]
runnerMode=[form|table]
```

Why not make `runnerMode: "form" | "table" | undefined`? This way we avoid one extra parameter.


----

`problemsIsExpaded=[true|false]`

There's a typo here, and I don't understand how that relates to the DMN Runner.. If `runnerMode` is set, why do we have this param as well? 


---

Also, `runner` is not very descriptive IMHO... URL queryParams are in the context of KIE Sandbox as a whole, so we need to be more specific. `dmnRunnerMode` and `dmnRunnerRow` sound better to me.

ljmotta:
@tiagobento Thanks for your review

> Why not make runnerMode: "form" | "table" | undefined? This way we avoid one extra parameter.

As the user has the option to keep a mode of visualization of the runner even if it's closed, I thought it was better to add the "runnerIsExpanded".

An example: I've changed to the "table" view, and after that I've closed the DMN Runner. It keeps the state "table", but now the it's not expanded.

> There's a typo here, and I don't understand how that relates to the DMN Runner.. If runnerMode is set, why do we have this param as well?

Sorry, the typo is only present in the description, it is "problemsIsExpanded" 🙈 . I've thought that was a good idea to control the "Problems" dock with a query param as well, but I can remove it.

> Also, runner is not very descriptive IMHO... URL queryParams are in the context of KIE Sandbox as a whole, so we need to be more specific. dmnRunnerMode and dmnRunnerRow sound better to me.

Sure, I'll update it.

packages/online-editor/src/editor/DmnRunner/DmnRunnerContext.tsx
ljmotta: I should have changed the name to "setIsExpandedQueryParams".. At first I was working on changing the URL location so it was the first name I gave, but I'm going to update it. 🙈

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
ljmotta: If it's an `undefined` value, it's not a boolean, and it's not handled. So if the user passes "runnerIsExpanded=something", nothing should happen.

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
ljmotta: If it's an `undefined` value, it's not a number, and it's not handled. So if the user passes "runnerRow=something", nothing should happen.

tiagobento: I think we should remove the queryParam if there's an invalid value for it.

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
tiagobento: I think we should remove the queryParam if there's an invalid value for it.

tiagobento:
Yeah, I think it's not very good to keep this "preference" saved on the URL. It looks to me like we need a "Settings" entry for this, so users can store locally in their browser what's their preferred way of running DMN models. It looks odd to me that a URL with `dmnRunnerMode=table` does not show the table, you know what I mean? I guess we can first save this preference in-memory in a "lastDmnRunnerMode" state or something like that, and we create a new issue to create a Settings entry for this. I assume we're going to have other DMN Runner-related preferences in the future anyway. WDYT?

tiagobento:
I would remove the problems tab queryParam for now too. It's too tightly coupled with the DMN Runner in a way that I think we're adding more complexity to it for little gain.

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
```diff
@@ -240,11 +242,11 @@ export function DmnRunnerProvider(props: PropsWithChildren<Props>) {
     () => ({
       preparePayload,
       setError,
-      setExpandedLocation,
+      setIsExpandedQueryParams,
       setInputRows,
-      setParams,
+      setMode: setQueryParams,
```
tiagobento: Still referencing queryParams. Again, I don't think consumers of this context should know about HOW it's doing behind the scenes, they should be able to just express what they need from it. `setIsExpandedQueryParams` can have a better name.. Why not simply `setExpanded`?

packages/dmn-language-service/src/index.ts
```diff
@@ -29,9 +29,13 @@ export class DmnLanguageService {
   private readonly importTagRegExp = new RegExp(`([a-z]*:)?(${IMPORT})`);
   private readonly inputDataRegEx = new RegExp(`([a-z]*:)?(${INPUT_DATA})`);
 
-  constructor(private readonly readDmn: (importedModel: string) => Promise<DmnLanguageServiceImportedModel>) {}
+  constructor(
+    private readonly helperFunction: {
+      getDmnImportedModel: (importedModelRelativePath: string) => Promise<DmnLanguageServiceImportedModel>;
+    }
```
tiagobento: Why is this called `helperFunction`? This is not a function, it's an object with a property called `getDmnImportedModel`. We usually call constructor object parameters simply `args`.

packages/dmn-language-service/src/index.ts
```diff
@@ -66,18 +70,18 @@ export class DmnLanguageService {
...
-            return this.readDmn(importedModel);
+          importedModelPaths.map((importedModelPath) => {
+            return this.helperFunction.getDmnImportedModel(importedModelPath);
           })
         )
       ).filter((e) => e !== undefined) as DmnLanguageServiceImportedModel[];
 
       const contents = importedModelsContent.map((resources) => resources.content ?? "");
-      const importedFiles = this.getImportedModels(contents);
+      const importedFiles = this.getImportedModelPaths(contents);
```
tiagobento: On KIE Sandbox, I try not to name things simply `path`, as there's different types of path, mainly relative (and its reference point, e.g. current file), and absolute. What kind of path are those? Are they relative to the `models`' paths? When we can't use the type system to really express what's happening, we need to rely on variable names to do it.

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
tiagobento: Also, why is there a separate method to expand it? As far as I understood, you can do it using `setMode` too, right?

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
```diff
@@ -62,12 +62,14 @@ export function DmnRunnerProvider(props: PropsWithChildren<Props>) {
...
-      });
-      return { content: resourceContent?.content ?? "", path: resourceContent?.path ?? "" };
+    return new DmnLanguageService({
+      getDmnImportedModel: async (importedModelRelativePath: string) => {
+        const resourceContent = await workspaces.resourceContentGet({
+          workspaceId: props.workspaceFile.workspaceId,
+          relativePath: importedModelRelativePath,
+        });
+        return { content: resourceContent?.content ?? "", path: resourceContent?.path ?? "" };
+      },
```
tiagobento: I don't think `resourceContentGet` is the best option for this.. Maybe `getFileContent`?

tiagobento: Also, I think there's a hint here due to the fact that you're creating two separate instances of `DmnLanguageService`. As of today, I think we should only have one instance of `DmnLanguageService` instance per Workspace, as it CAN be that in the future we're going to have shared state inside it, for performance reasons.

tiagobento (changes requested):
Thanks for the changes! Left another round of review.

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
ljmotta: the `setExpanded` have a different behavior as its used to tweak between open/close, and the "setMode" only opens the DMN runner.

ljmotta: maybe a `toggleExpanded` would be a better name 😅

packages/dmn-language-service/src/index.ts
ljmotta: I'm not sure if the DMN model uses only relative paths, as the property name is called "locationURI". The spec doens't say anything regarding that too. But our editor always uses relative paths, so I guess it wouldn't be a problem, or should I add a path validator to check if it's a relative path or not?

ljmotta:
@tiagobento The preference already exists on the DMN Runner, as the user can choose the visualization between "as Form" and "as Table", so I tried to reflect this in the URL. But I'll remove it, and save it in a cookie for now.

packages/online-editor/src/editor/DmnRunner/DmnRunnerProvider.tsx
ljmotta: `resourceContentGet` returns the content as a `string` and the `path`. If I use the `getFileContent` method I'll need to decode the `Uint8Array` and them manually map with the `path`.

tiagobento (approved):

PR #1431, opened 2023-01-30
KOGITO-8426:  Implement a quickstart guide component in the settings section

fantonangeli:
Jira: https://issues.redhat.com/browse/KOGITO-8426

**Description**
Implement a "quickstart guide" component according to the wireframe.

**Notes**
- this feature uses [patternfly-quickstarts](/https://github.com/patternfly/patternfly-quickstarts) and can be reused for other purposes outside of the `Settings` section.
- after some research, I used HTML content, for the quickstart content, because the result is the closest possible to [`kie-docs`](https://kiegroup.github.io/kogito-docs). But we can evaluate to find a way to use the Antora Asciidoc format directly.

[KOGITO-8426.webm](https://user-images.githubusercontent.com/17780574/215455776-0a4b3269-6c08-4c29-84c0-18adff7b7edd.webm)


paulovmr (approved):

caponetto (approved):
Great work @fantonangeli! 
My only concern is the fact that the HTML code is inside a string.
Not sure if there is a better way, but let's revisit that later.

fantonangeli:
> Great work @fantonangeli! My only concern is the fact that the HTML code is inside a string. Not sure if there is a better way, but let's revisit that later.

@caponetto I agree, I tried converting the KIE-docs pages to Markdown, wich is supported by Quickstart, but it was difficult to be really close to the `KIE-docs`look.

PR #1439, opened 2023-02-02
KOGITO-8580: Disable JSON support for now

Rikkola:
https://issues.redhat.com/browse/KOGITO-8580

Removing JSON support for now. YAML is the main focus.

Also removing the need to call the file .yard.yml or .yard.yaml. Unless we need this to differentiate the files from other YAML files in VSCode?


paulovmr (commented):
@Rikkola We do need the `yard` at the suffix because of the file associations in the VS Code extension, so it only opens the editor for `yard` files and not all yaml files: https://github.com/kiegroup/kie-tools/blob/main/packages/vscode-extension-yard-editor/package.json#L132

Rikkola:
@paulovmr ok suspected that. I will change that back for now. The upload widget could be less picky even if the produced file is yard.yml.

packages/yard-editor/dev-webapp/App.tsx
```diff
@@ -76,7 +76,7 @@ export const App = () => {
   }, [editor]);
 
   const onSetContent = useCallback((path: string, content: string) => {
-    const match = /\.sw\.(json|yml|yaml)$/.exec(path.toLowerCase());
+    const match = /\.sw\.(yml|yaml)$/.exec(path.toLowerCase());
```
caponetto: Not related to your changes but this should be `yard` not `sw`, right?
```suggestion
    const match = /\.yard\.(yml|yaml)$/.exec(path.toLowerCase());
```

Rikkola: @caponetto Good point, I think `sw` is a bug in the old code that was copied from serverless workflow.

caponetto: Makes sense! Thanks for the fix :)

caponetto:
@Rikkola There are also some `yard.json` entries in [1]. Should those be removed as well?

[1] https://github.com/kiegroup/kie-tools/blob/main/packages/vscode-extension-yard-editor/package.json

Rikkola:
@caponetto I added that to my workspace and fixed it, thank you.

caponetto:
@Rikkola I still see some `yard.json` entries in these files. Can you please check?
- https://github.com/Rikkola/kie-tools/blob/KOGITO-8580-disable-json/packages/vscode-extension-yard-editor/package.json
- https://github.com/Rikkola/kie-tools/blob/KOGITO-8580-disable-json/packages/vscode-extension-yard-editor/README.md

Rikkola:
@caponetto Indeed there was a lot more.

caponetto (approved):
Thanks @Rikkola! LGTM

paulovmr (approved):

PR #1443, opened 2023-02-06
kie-issues#15: Create an 'About' modal on KIE Sandbox

### LINKED ISSUE #15: Create an 'About' modal on KIE Sandbox
- Add a new `i` icon at the top-right corner that opens a modal with the meta information about KIE Sandbox
- This will require the “Webpack replace” mechanism to consume information from outside source files (e.g. build-env). There are other places where we already use this strategy. Search for `WEBPACK_REPLACE__buildInfo` for an example.
- Some possible information on the ‘About’ modal
  - Version from build (build time info)
  - Quarkus version  (build time info)
  - Kogito runtimes version (build time info)
  - DMN Runner image name (build time info)
  - Extended Services version (info via http call)

klbowers:
Closes https://github.com/kiegroup/kie-issues/issues/15


packages/online-editor/webpack.config.ts
```diff
@@ -94,6 +101,7 @@ export default async (env: any, argv: any) => {
             },
           ]),
           new EnvironmentPlugin({
+            COMMIT_HASH: JSON.stringify(lastCommitHash),
```
tiagobento: Can we keep the same naming strategy? I know this is not good, as we're not using Webpack replacement anymore, but might be confusing when we actually fix this, to have only one variable that's not the same.

packages/online-editor/webpack.config.ts
```diff
@@ -52,7 +52,14 @@ export default async (env: any, argv: any) => {
     devDeployments_onlineEditorUrl,
   ] = getDevDeploymentsArgs();
   const gtmResource = getGtmResource();
-
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const childProcess = require("child_process");
```
tiagobento: You can move this to the top of the file, no? This will get rid of the warning..

packages/online-editor/src/aboutModal/AboutButton.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { Text, TextContent, TextVariants } from "@patternfly/react-core/dist/js/components/Text";
+import { Flex, FlexItem } from "@patternfly/react-core/dist/js/layouts/Flex";
+import { Bullseye } from "@patternfly/react-core/dist/js/layouts/Bullseye";
+import {
+  DescriptionList,
+  DescriptionListTerm,
+  DescriptionListGroup,
+  DescriptionListDescription,
+  Divider,
+} from "@patternfly/react-core";
```
tiagobento: Please don't use the top-level import for `@patternfly/react-core`, as it won't be tree-shaken. Use the specific ones like you see above.

packages/online-editor/src/aboutModal/AboutButton.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+export function OnlineEditorPage(props: { children?: React.ReactNode }) {
+  const history = useHistory();
+  const routes = useRoutes();
+}
```
tiagobento: ? :D

packages/online-editor/src/aboutModal/AboutButton.tsx
```diff
@@ -0,0 +1,136 @@
...
+  }, []);
+  const dmnImage = useMemo(() => {
+    return process.env["WEBPACK_REPLACE__dmnDevDeployment_baseImageFullUrl"];
+  }, []);
+  const extendedVersion = useMemo(() => {
+    return process.env["WEBPACK_REPLACE__kieSandboxExtendedServicesCompatibleVersion"];
+  }, []);
+  const commitSha = useMemo(() => {
+    return process.env.COMMIT_HASH;
+  }, []);
```
tiagobento: TBH I don't see any reason for these to be `useMemos`, you can just inline everything.

packages/online-editor/src/aboutModal/AboutButton.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const history = useHistory();
+  const routes = useRoutes();
+}
+
+export const AboutButton: React.FunctionComponent = () => {
+  const [isModalOpen, setIsModalOpen] = React.useState(false);
+  const history = useHistory();
+  const routes = useRoutes();
+  const handleModalToggle = () => {
+    setIsModalOpen(!isModalOpen);
```
tiagobento: Always prefer using the previous value when doing this.
```suggestion
    setIsModalOpen(prev => !prev);
```

Also, it's a good practice to memoize those, so this can be wrapped inside a `useCallback`.

tiagobento (changes requested):
Why is this on `Draft`? Looks much more ready than draft to me :P

Jokes aside, great work @kbowers-ibm! I'm missing a nice Gif on the description of the PR, though :(

Left some minor comments inline, mostly related to code style and conventions.

packages/online-editor/webpack.config.ts
```diff
@@ -53,6 +54,13 @@ export default async (env: any, argv: any) => {
   ] = getDevDeploymentsArgs();
   const gtmResource = getGtmResource();
 
+  let lastCommitHash = "";
+  try {
+    lastCommitHash = childProcess.execSync("git rev-parse --short HEAD").toString().trim();
+  } catch (e) {
+    console.error(e);
+  }
```
tiagobento: Hmm... here I think you should throw an exception if you can't get this information. We shouldn't be able to build if not inside a versioned directory... Also, you might want to JSON.stringify() inside the try/catch block.

tiagobento (changes requested):
Almost perfect! Please keep in mind that you have some unused imports on `AboutButton`. A couple of extra new lines around `handleModalToggle` wouldn't hurt too 😝.

tiagobento (approved):

PR #1447, opened 2023-02-07
KOGITO-8599: Fix Table External Component paging style in Dark Mode

kumaradityaraj:
The pagination section used to disappear when the screen size was decreased below 768px but now it comes down after the filter tab. This issue still needs to be addressed for light mode.
Screenshot from 2023-02-08 10-14-31

The pagination number which is in red used to be black in dark mode due to which it was invisible; in this PR it has been corrected and now its white in dark mode.
Screenshot from 2023-02-08 10-12-31

The updated code result is as in the below pic:-
Screenshot from 2023-02-08 10-20-37

### DIFF
--- a/packages/dashbuilder-component-table/src/FilteredTable.tsx
+++ b/packages/dashbuilder-component-table/src/FilteredTable.tsx
@@ -164,7 +164,7 @@ export const FilteredTable = (props: Props) => {
               onClear={() => onSearch("")}
             />
           </FlexItem>
-          <FlexItem align={{ default: "alignRight" }}>
+          <FlexItem align={{ default: "alignLeft" }}>
             <Pagination
               itemCount={rows.length}
               perPage={perPage}
--- a/packages/dashbuilder-component-table/src/FilteredTable.tsx
+++ b/packages/dashbuilder-component-table/src/FilteredTable.tsx
@@ -164,7 +164,7 @@ export const FilteredTable = (props: Props) => {
               onClear={() => onSearch("")}
             />
           </FlexItem>
-          <FlexItem align={{ default: "alignLeft" }}>
+          <FlexItem align={{ default: "alignRight" }}>
             <Pagination
               itemCount={rows.length}
               perPage={perPage}
--- a/packages/dashbuilder-component-table/static/index.css
+++ b/packages/dashbuilder-component-table/static/index.css
@@ -42,9 +42,36 @@
   --pf-global--BackgroundColor--100: #100c2a !important;
 }
 
-[mode="dark"] .pf-c-options-menu__menu {
+[mode="dark"] .pf-c-options-menu__menu,
+[mode="dark"] .pf-c-pagination__nav,
+[mode="dark"] .pf-c-pagination__total-items {
   background-color: #100c2a !important;
   color: #eef1fa;
+}
+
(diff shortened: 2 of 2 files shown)


paulovmr:
Hi @kumaradityaraj , could you please add two images (or video, but I think image is good in this case) to the PR description showing the before and after the change? Thanks.

kumaradityaraj:
> Hi @kumaradityaraj , could you please add two images (or video, but I think image is good in this case) to the PR description showing the before and after the change? Thanks.

The image has been updated. Thanks.

paulovmr (approved):

packages/dashbuilder-component-table/src/FilteredTable.tsx
```diff
@@ -164,7 +164,7 @@ export const FilteredTable = (props: Props) => {
               onClear={() => onSearch("")}
             />
           </FlexItem>
-          <FlexItem align={{ default: "alignRight" }}>
+          <FlexItem align={{ default: "alignLeft" }}>
```
jesuino: Could you please clarify why the position was changed?

packages/dashbuilder-component-table/static/index.css
```diff
@@ -42,11 +42,38 @@
...
+  padding-left: 20px;
+  padding-top: 10px;
+  padding-bottom: 10px;
+}
+
+.pf-l-flex {
+  justify-content: center;
+}
+
+.pf-m-align-left {
```
jesuino: same question, why changing the padding and alignment?

jesuino (changes requested):
Changes looks good to me, thanks! I am just wondering about some additional CSS - could you please clarify it?

Thanks!

packages/dashbuilder-component-table/src/FilteredTable.tsx
kumaradityaraj: If we align it right then on changing the screen size less than 768px that particular div used to disappear; I tried alligning it left and then it worked.

packages/dashbuilder-component-table/static/index.css
kumaradityaraj: In the small screen size, the UI was looking bad so I shifted the pagination part after the filter div and applied padding and aligned for smooth UI.

jesuino:
Hello @kumaradityaraj 

Thanks for the clarification

Could we please keep only the color change for now?

On a later ticket we can work on the elements position. 

Basically it comes from patternfly[1] and we may have a patternfly way to fix sizing issues. We had to modify for dark mode because it is not supported for this component and it had to look like it is integrated into other components.

Thanks!

[1] https://www.patternfly.org/v4/components/table

kumaradityaraj:
> Hello @kumaradityaraj
> 
> Thanks for the clarification
> 
> Could we please keep only the color change for now?
> 
> On a later ticket we can work on the elements position.
> 
> Basically it comes from patternfly[1] and we may have a patternfly way to fix sizing issues. We had to modify for dark mode because it is not supported for this component and it had to look like it is integrated into other components.
> 
> Thanks!
> 
> [1] https://www.patternfly.org/v4/components/table

The changes has been done. 
Thanks @jesuino

jesuino (approved):
Great work, thanks!

PR #1462, opened 2023-02-14
KOGITO-8639: SWF Viewer - display actions icon

hasys:
Hello @romartin, @LuboTerifaj, there is the PR about action icons. This PR include:

1. **action** and **actions** icon will have the same symbol
2. **action** and **actions** are required parameters so icons should be always visible. On mouse hover it should display name or functionRef or eventRef or subFlowRef (in this order) parameter for action and parameters for actions or notify user about missing value.

    - if there is no action/actions, notify user
    - if there is action `name` set, use name
    - if there is FunctionRef (string) defined, use `functionRef` value
    - if there is FunctionRef (object) defined, use `functionRef.refName` required parameter, even if null
    - if there is EventRef defined, use `eventRef.consumeEventRef` required parameter, even if null
    - if there is SubFlowRef (string) defined, use `subFlowRef`
    - if there is SubFlowRef (object) defined, use `subFlowRef.workflowId` required parameter, even if null

3. Supported parameters:

    - Callback State - **action**
    - ForEach, Operation States - **actions** (simple array of action)

4. On mouse click the state should be activated.
image


packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/HasAction.java
```diff
@@ -0,0 +1,79 @@
...
+package org.kie.workbench.common.stunner.sw.client.shapes;
+
+import org.kie.workbench.common.stunner.core.util.StringUtils;
+import org.kie.workbench.common.stunner.sw.definition.ActionNode;
+import org.kie.workbench.common.stunner.sw.definition.FunctionRef;
+import org.kie.workbench.common.stunner.sw.definition.SubFlowRef;
+
+public interface HasAction {
+
+    String ACTION_IS_NULL = "Action is not defined";
```
romartin, line 25: Could all these Strings below be i18n?

romartin (commented):
Hey @hasys 

Thanks for this! The code review looks good! Just missing to take care about i18n for labels being displayed to the end users.  Although just populating the `en` language, could it be possible to prepare the Strings for i18n?

Thx!

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/HasAction.java
hasys: Good point, thank you Roger. I created ticket for this issue, since it should be done not only for string constants for this PR but for all String Constants in the editor.

https://issues.redhat.com/browse/KOGITO-8677

romartin: ok, makes sense to me, let's address all i18n at once in [KOGITO-8677](https://issues.redhat.com/browse/KOGITO-8677)
Thanks!

romartin (approved):
Hey @hasys 

I have tested the UI and it works well, thanks! The only thing I am not completely sure is about the icon, I see it contains several "gears" and some other thing on the bottom. IMO I would just use a single gear icon, anyway it's my own opinion. Let's wait for feedback from others and we can change it later, if necessary, just easy :)

My idea now would be to merge this, and then start gathering feedback about the concrete label texts, from others. This way I don't discard further small changes on how to populate the labels, but anyway actually it looks good to me :)

I also noticed that if the duration is not set now, instead of showing all possible entries, you're just showing `undefined`, which makes sense to me.

Thanks!

hasys:
Good points, thank you @romartin,

about icon the idea was to use Patternfly icon for Services. Gears in Patternfly stands for "configurable settings". See Patternfly [icons sections](https://www.patternfly.org/v4/guidelines/icons/) for more details.

Anyway good idea about gathering more feedback. If Gears will raise better associations with actions among users let's change it!

LuboTerifaj (approved):
Hello @hasys ,

I have identified only two issues.

Actions are not shown in diagram when they are inside `onEvents` block. 
- See [event-based-greeting.sw.json](https://github.com/LuboTerifaj/specification/blob/examples-in-separate-files/examples/resources/event-based-greeting.sw.json), [new-patient-onboarding.sw.json](https://github.com/LuboTerifaj/specification/blob/examples-in-separate-files/examples/resources/new-patient-onboarding.sw.json) and other examples from specification.
- I understand this may be expected, so I would report a separate jira for this in case we want to change the actual behaviour.
	
In `eventRef`, there may be two references: `consumeEventRef` and `produceEventRef`. From the diagram, it is not clear which value is referenced to.
- We might want to adjust shown text in the diagram to better distinguish the references.
- I would report it in a separate jira.

Please let me know what do you think about the issues mentioned above.

Other than that, the feature works as expected.

Thank you!

romartin:
thanks @LuboTerifaj !

Please @hasys take a look at Lubo's comments and if so, let's report a new issue.

Merging. Thanks guys!

PR #1474, opened 2023-02-21
KOGITO-8714: SWF Viewer - display stateDataFilter parameter with icon

hasys:

there is the new parameter visualized:
* stateDataFilter
* States which supports this parameter: Event, Operation, Switch, Parallel, Inject, ForEach, Callback.
* To test add following parameter to supported State:
```json
   "stateDataFilter": {
    "input": "${ {fruits: .fruits} }",
    "output": "${ .provisionedOrders }"
   },
```

image


LuboTerifaj (changes requested):
Hello @hasys ,

I have found two issues:
- Empty string value is not trimmed in the diagram. It can cause showing a box with very wide width excessing the diagram.

- The parameters that are not present in source file should not be listed the diagram.
  - I understand that the similar issue is present also for other icons, so it is not necessary to change it in this PR, but it is definitely something to discuss. 

Potential improvement:
- If there are no other icons in state, currently shown icon could be aligned to the top right corner.
  - I reported the issue here: https://issues.redhat.com/browse/KOGITO-8726

Thanks!

hasys:
Hello @LuboTerifaj, great feedback thank you!

Just not sure the point about:

* Empty string value is not trimmed in the diagram. It can cause showing a box with very wide width excessing the diagram.

Can I ask you to provide JSON and screenshot to illustrate the issue there?

LuboTerifaj (approved):
Hey @hasys 

It looks good to me now.

I reported the issue with showing undefined values here:
https://issues.redhat.com/browse/KOGITO-8762

Thanks!

romartin (approved):

PR #1481, opened 2023-02-23
kie-issues#85: Add more Git operations on KIE Sandbox

### LINKED ISSUE #85: Add more Git operations on KIE Sandbox
KIE Sandbox needs to allow users to have more control over their changes before they push. Implementing the operations below will improve the experience considerably.
 
- Reverting changes of a single file (like `git checkout [file-path]`) (Probably at the Kebab menu? This needs a confirmation alert/modal)
- Reverting changes of the entire Workspace (like `git checkout .`), going back to the state of last commit. (Probably at the Kebab menu as well? This needs a confirmation alert/modal).
- Show which files are modified on the File Switcher (for current Workspace only). An icon indicating that a file has changes is enough.
- Add an option on the File Switcher to show only files that contain changes.

EDIT:
- The reverting operation needs to handle also files being deleted and need to be restored.

jstastny-cz:
Resolves kiegroup/kie-issues#85 .

New:
* FileSwitcher menuItems contain git status indicator (Modified, Deleted, Added) and a hoverable icon for reverting the change. On hover a popover appears with inline alert contents.
* New Component `GitStatusIndicatorActions.tsx` implements two variants of how actions are presented: popover vs. dropdown based design, details hidden under the component itself. It's passed to GitStatusIndicator.tsx as children.
* WorkspaceStatusIndicator contains statuses for all files in workspace, e.g. `M | A | D` in case there are files in such states.
* WorkspaceStatusIndicator contains a dropdown variant of the git actions.
* Implemented revert operation as checkout of files from HEAD using isomorphic-git in `GitService.ts`. Invoked through the GitStatusIndicatorActions.tsx passing
  * single file in FileSwitcher menu item
  * all modified files in WorkspaceStatusIndicator dropdown
* Introduced event type `WS_CHECKOUT_HEAD_FILES` to signal changes to the workspace itself. The message contains relativePaths to all the files that were processed - allows for checking if update of current Editor contents should be refreshed.

Noteworthy refactoring:
* FileSwitcher now shows also the active file, to be able to display git operations related just to the current file.
* FilesMenuItems component reused in HomePage.tsx - reduction of duplicated code and unification - it was not possible to show the files as SVG cards on HomePage before.
* `WorkspaceHooks.tsx` signature changed to `useWorkspaceGitStatusPromise(workspace: WorkspaceDescriptor | undefined)` to enable reusing in HomePage.tsx
* Workspace toolbar is shown also for Local WorkspaceKind now, it enables the additional git actions also for Local workspaces.

### DIFF
--- a/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
+++ b/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
@@ -1012,7 +1012,7 @@ const FilesMenuItemCarouselCard = (props: {
       isRounded={true}
       isCompact={true}
       isFullHeight={true}
-      onSelect={props.onSelectFile}
+      onClick={props.displayMode === FileListItemDisplayMode.enabled ? props.onSelectFile : undefined}
       className={switchExpression(props.displayMode, {
         enabled: "kie-tools--file-list-item-enabled",
         deleted: "kie-tools--file-list-item-deleted",
--- a/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
+++ b/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
@@ -393,7 +393,6 @@ export function FileSwitcher(props: {
                   <MenuGroup
                     style={{
                       maxHeight: `calc(${MENU_HEIGHT_MAX_LIMIT_CSS} - ${DRILLDOWN_NAVIGATION_MENU_ITEM_HEIGHT_IN_PX})` /* height of menu minus the height of Current item*/,
-                      overflowY: "auto",
                     }}
                   >
                     {workspacesMenuItems}
@@ -694,8 +693,8 @@ export function FilesMenuItems(props: {
           }}
           placeholder={`In '${props.workspace.descriptor.name}'`}
           style={{ fontSize: "small" }}
-          onClick={(ev) => {
-            ev.stopPropagation();
+          onClick={(e) => {
+            e.stopPropagation();
           }}
         />
       </MenuInput>
@@ -753,37 +752,36 @@ export function FilesMenuItems(props: {
 
   return (
     <>
-      <Grid style={{ alignItems: "center" }}>
-        <GridItem span={12}>
-          <MenuItem direction={"up"} itemId={`${props.workspace.descriptor.workspaceId}-breadcrumb`}>
-            All
-          </MenuItem>
-        </GridItem>
-        <GridItem span={12}>{searchInput}</GridItem>
-        <GridItem span={6}></GridItem>
-        <GridItem span={4}>
+      <MenuItem direction={"up"} itemId={`${props.workspace.descriptor.workspaceId}-breadcrumb`}>
+        All
+      </MenuItem>
+      {searchInput}
+      <Flex
+        justifyContent={{ default: "justifyContentFlexEnd" }}
+        alignItems={{ default: "alignItemsCenter" }}
+        style={{ paddingRight: "16px" }}
+      >
+        <FlexItem>
           {props.currentWorkspaceGitStatusProps !== undefined && (
-            <MenuInput>
-              <Tooltip content={"Filter only modified files"} position={"bottom"}>
-                <Checkbox
-                  label={"Only changed"}
-                  id={"filter-git-status-changed-locally"}
-                  aria-label={"Select to display only modified files"}
-                  isChecked={filteredGitSyncStatus === WorkspaceGitLocalChangesStatus.pending}
-                  onChange={(checked) => {
-                    setFilteredGitSyncStatus(checked ? WorkspaceGitLocalChangesStatus.pending : undefined);
-                  }}
-                />
-              </Tooltip>
-            </MenuInput>
+            <Tooltip content={"Select to display only modified, added, or deleted files"} position={"bottom"}>
+              <Checkbox
+                label={"Only modified"}
+                id={"filter-git-status-modified-locally"}
+                aria-label={"Select to display only modified, added, or deleted files"}
+                isChecked={filteredGitSyncStatus === WorkspaceGitLocalChangesStatus.pending}
+                onChange={(checked) => {
+                  setFilteredGitSyncStatus(checked ? WorkspaceGitLocalChangesStatus.pending : undefined);
+                }}
+              />
+            </Tooltip>
           )}
-        </GridItem>
-        <GridItem span={2}>
+        </FlexItem>
+        <FlexItem>
           <Flex justifyContent={{ default: "justifyContentCenter" }}>
             <FilesMenuModeIcons filesMenuMode={props.filesMenuMode} setFilesMenuMode={props.setFilesMenuMode} />
           </Flex>
-        </GridItem>
-      </Grid>
+        </FlexItem>
+      </Flex>
       <Divider component={"li"} />
 
       {props.filesMenuMode === FilesMenuMode.LIST && (
@@ -932,7 +930,7 @@ export function FilesMenuItems(props: {
                     setTopSectionHeightInPx((prev) => prev + MENU_SHOW_CHANGED_CHECKBOX_HEIGHT_IN_PX);
                   }}
                 >
-                  Back To Models
+                  Back
                 </MenuItem>
                 {searchInput}
                 <Divider component={"li"} />
--- a/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
+++ b/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
@@ -74,7 +74,6 @@ import {
 import { WorkspaceListItem } from "../../workspace/components/WorkspaceListItem";
 import { usePreviewSvg } from "../../previewSvgs/PreviewSvgHooks";
 import { WorkspaceLoadingMenuItem } from "../../workspace/components/WorkspaceLoadingCard";
-import { Grid, GridItem } from "@patternfly/react-core/dist/js/layouts/Grid";
 import {
   listDeletedFiles,
   resolveGitLocalChangesStatus,
@@ -99,7 +98,7 @@ const FILE_SWITCHER_INITIAL_HEIGHT_OFFSET_IN_PX = 204;
 const MENU_HEIGHT_MAX_LIMIT_CSS = `calc(100vh - ${FILE_SWITCHER_INITIAL_HEIGHT_OFFSET_IN_PX}px)` as const;
 const DRILLDOWN_NAVIGATION_MENU_ITEM_HEIGHT_IN_PX = 40;
 const MENU_SEARCH_HEIGHT_IN_PX = 64;
-const MENU_SHOW_CHANGED_CHECKBOX_HEIGHT_IN_PX = 53;
+const MENU_SHOW_CHANGED_CHECKBOX_HEIGHT_IN_PX = 36;
 const MENU_DIVIDER_HEIGHT_IN_PX = 17;
 
 export function FileSwitcher(props: {
@@ -392,7 +391,8 @@ export function FileSwitcher(props: {
                   </MenuItem>
                   <MenuGroup
                     style={{
-                      maxHeight: `calc(${MENU_HEIGHT_MAX_LIMIT_CSS} - ${DRILLDOWN_NAVIGATION_MENU_ITEM_HEIGHT_IN_PX})` /* height of menu minus the height of Current item*/,
+                      maxHeight: `calc(${MENU_HEIGHT_MAX_LIMIT_CSS} - ${DRILLDOWN_NAVIGATION_MENU_ITEM_HEIGHT_IN_PX}px)` /* height of menu minus the height of Current item*/,
+                      overflowY: "auto",
                     }}
                   >
                     {workspacesMenuItems}
@@ -703,11 +703,9 @@ export function FilesMenuItems(props: {
 
   const fileListMenuItemsHeightMaxLimitCss = useMemo(() => {
     return `calc(${MENU_HEIGHT_MAX_LIMIT_CSS} - ${
-      (!props.currentWorkspaceGitStatusProps
-        ? topSectionHeightInPx - MENU_SHOW_CHANGED_CHECKBOX_HEIGHT_IN_PX
-        : topSectionHeightInPx) + (otherFiles.length ? bottomSectionHeightInPx : 0)
+      topSectionHeightInPx + (otherFiles.length ? bottomSectionHeightInPx : 0)
     }px)`;
-  }, [bottomSectionHeightInPx, otherFiles.length, props.currentWorkspaceGitStatusProps, topSectionHeightInPx]);
+  }, [bottomSectionHeightInPx, otherFiles.length, topSectionHeightInPx]);
 
   const isCurrentWorkspaceFile = useCallback(
     (file: WorkspaceFile) => {
--- a/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
+++ b/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
@@ -16,7 +16,6 @@
 
 import { ActiveWorkspace } from "@kie-tools-core/workspaces-git-fs/dist/model/ActiveWorkspace";
 import { useWorkspaces, WorkspaceFile } from "@kie-tools-core/workspaces-git-fs/dist/context/WorkspacesContext";
-import { useRoutes } from "../../navigation/Hooks";
 import * as React from "react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { join } from "path";
@@ -46,19 +45,14 @@ import { ThLargeIcon } from "@patternfly/react-icons/dist/js/icons/th-large-icon
 import { ListIcon } from "@patternfly/react-icons/dist/js/icons/list-icon";
 import { useWorkspaceDescriptorsPromise } from "@kie-tools-core/workspaces-git-fs/dist/hooks/WorkspacesHooks";
 import { PromiseStateWrapper, useCombinedPromiseState } from "@kie-tools-core/react-hooks/dist/PromiseState";
-import { Split, SplitItem } from "@patternfly/react-core/dist/js/layouts/Split";
 import { Button } from "@patternfly/react-core/dist/js/components/Button";
-import { WorkspaceDescriptor } from "@kie-tools-core/workspaces-git-fs/dist/worker/api/WorkspaceDescriptor";
 import { useWorkspacesFilesPromise } from "@kie-tools-core/workspaces-git-fs/dist/hooks/WorkspacesFiles";
 import { Skeleton } from "@patternfly/react-core/dist/js/components/Skeleton";
 import { Card, CardBody, CardHeader, CardHeaderMain } from "@patternfly/react-core/dist/js/components/Card";
 import { Gallery } from "@patternfly/react-core/dist/js/layouts/Gallery";
-import { useHistory } from "react-router";
 import { Bullseye } from "@patternfly/react-core/dist/js/layouts/Bullseye";
 import { EmptyState, EmptyStateIcon } from "@patternfly/react-core/dist/js/components/EmptyState";
 import { CubesIcon } from "@patternfly/react-icons/dist/js/icons/cubes-icon";
-import { ArrowRightIcon } from "@patternfly/react-icons/dist/js/icons/arrow-right-icon";
-import { ArrowLeftIcon } from "@patternfly/react-icons/dist/js/icons/arrow-left-icon";
 import { useEditorEnvelopeLocator } from "../../envelopeLocator/hooks/EditorEnvelopeLocatorContext";
 import { VariableSizeList } from "react-window";
 import AutoSizer from "react-virtualized-auto-sizer";
@@ -66,6 +60,7 @@ import {
   FileDataList,
   FileLink,
   FileListItem,
+  FileListItemDisplayMode,
   getFileDataListHeight,
   SingleFileWorkspaceDataList,
 } from "../../filesList/FileDataList";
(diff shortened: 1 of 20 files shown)


paulovmr:
@jstastny-cz This PR should fix the CI error: https://github.com/kiegroup/kie-tools/pull/1484

jstastny-cz:
@caponetto or @paulovmr  if you could please review too, there are changes in workspaces-git-fs affecting serverless-logic-web-tools too:
1. WorkspaceHooks.tsx signature changed to `useWorkspaceGitStatusPromise(workspace: WorkspaceDescriptor | undefined)`
2. WorkspaceDescriptorService.ts removing broadcast in `bumpLastUpdatedDate` function which was causing duplicate git status query on each change in the editor. Originally this was added as part of the fix for refreshing workspaces order on the HomePage when a workspace is updated async, like in a different browser window. This will be reported as an issue and needs a different fix.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -478,26 +483,9 @@ export function NewModelCard(props: { title: string; extension: SupportedFileExt
...
-        .sort((a, b) => a.relativePath.localeCompare(b.relativePath))
-        .filter((file) => editorEnvelopeLocator.hasMappingFor(file.relativePath)),
-    [editorEnvelopeLocator, workspacePromise.data?.files]
-  );
-
-  const arrayWithModelsThenOtherFiles = useMemo(() => {
-    return [...models, ...otherFiles];
-  }, [models, otherFiles]);
+  const workspaceGitStatusPromise = useWorkspaceGitStatusPromise(workspacePromise.data?.descriptor);
+  const [filesDropdownMode, setFilesDropdownMode] = useState(FilesDropdownMode.LIST_MODELS_AND_OTHERS);
```
jstastny-cz: Current implementation of the dropdown mode is causing issues in home page with width, investigating if View others/View models can be implemented as a drilldown instead of Split as it is now.

packages/online-editor/src/editor/FileSwitcher.tsx
```diff
@@ -586,14 +645,19 @@ export function SearchableFilesMenuGroup(props: {
...
+            <TextInput
+              ref={searchInputRef}
+              value={props.search}
+              aria-label={"Other files menu items"}
+              iconVariant={"search"}
+              type={"search"}
+              onChange={(value) => props.setSearch(value)}
+            />
+          </FlexItem>
+          <FlexItem>{filterByGitSyncStatusSwitch()}</FlexItem>
```
tiagobento: Can you inline this? Avoid indirection...

packages/online-editor/src/workspace/components/GitStatusIndicatorActions.tsx
```diff
@@ -0,0 +1,513 @@
...
+  CaretDownIcon,
+  CaretLeftIcon,
+  CaretRightIcon,
+  CaretUpIcon,
+  CheckCircleIcon,
+  ExclamationCircleIcon,
+  InfoCircleIcon,
+  UndoAltIcon,
+  WarningTriangleIcon,
+} from "@patternfly/react-icons";
```
tiagobento: Wrong imports... please always prefer the most specific ones. Also, double-check other places on the PR.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -558,24 +545,17 @@ export function WorkspacesListDrawerPanelContent(props: { workspaceId: string |
...
-              )}
-            </AutoSizer>
+            <FilesMenuItems
+              shouldFocusOnSearch={false}
+              filesDropdownMode={filesDropdownMode}
+              setFilesDropdownMode={setFilesDropdownMode}
+              workspace={workspace}
+              onSelectFile={() => {}}
+              workspaceGitStatusPromise={workspaceGitStatusPromise}
+              workspaces={workspaces}
```
tiagobento: This is not good.. you don't need to pass `workspaces` as props.. you should be able to `useWorkspaces` where you need it.

tiagobento: There's a lot of other places with this same behavior too.

packages/online-editor/src/i18n/locales/en.ts
```diff
@@ -626,4 +626,16 @@ export const en: OnlineI18n = {
...
+      title: "Revert Changes",
+      description: ["This action is permanent. Following file will be reverted to the last commit:", wrapped("paths")],
+      confirmButtonText: "Revert Permanently",
+    },
+    revertAll: {
+      title: "Revert All Changes",
+      description: ["This action is permanent. Following files will be reverted to the last commit:", wrapped("paths")],
+      confirmButtonText: "Revert Permanently",
+    },
+  },
```
tiagobento: At some point, non-programmers will need to write i18n files. Our setup is already more complicated than .properties files, so we should also avoid complex constructs here.. IDK if there's any other place using an array, but if possible, we should avoid doing so...

packages/online-editor/src/workspace/components/GitStatusIndicatorActions.tsx
```diff
@@ -0,0 +1,513 @@
...
+        filepaths,
+      });
+      if (
+        props.currentWorkspaceFile &&
+        !(await props.workspaces.existsFile({
+          workspaceId: props.workspaceDescriptor.workspaceId,
+          relativePath: props.currentWorkspaceFile.relativePath,
+        }))
+      ) {
+        props.onDeletedWorkspaceFile && props.onDeletedWorkspaceFile();
```
tiagobento:
```suggestion
        props.onDeletedWorkspaceFile?.();
```

packages/online-editor/src/workspace/components/GitStatusIndicatorActions.tsx
```diff
@@ -0,0 +1,513 @@
...
+        ref: props.workspaceDescriptor.origin.branch,
+        filepaths,
+      });
+      if (
+        props.currentWorkspaceFile &&
+        !(await props.workspaces.existsFile({
+          workspaceId: props.workspaceDescriptor.workspaceId,
+          relativePath: props.currentWorkspaceFile.relativePath,
+        }))
+      ) {
```
tiagobento: There are many places passing `workspaces` down as a prop just for this moment here to happen.. That is not very good IMHO.

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -28,61 +28,214 @@ import { SecurityIcon } from "@patternfly/react-icons/dist/js/icons/security-ico
...
+
+  const deletedFilesForWorkspace = React.useMemo(() => {
+    return listDeletedFiles({
+      workspaceDescriptor: props.workspaceDescriptor,
+      workspaceGitStatus: props.workspaceGitStatusPromise?.data,
+    });
+  }, [props.workspaceDescriptor, props.workspaceGitStatusPromise]);
+
+  const isThereDeletedFile = React.useMemo(() => {
+    const maybeWorkspaceFile = props.workspaceFile;
```
tiagobento: No reason for this variable to exist, honestly..

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -28,61 +28,214 @@ import { SecurityIcon } from "@patternfly/react-icons/dist/js/icons/security-ico
...
+    return (
+      (!maybeWorkspaceFile && deletedFilesForWorkspace.length) ||
+      (maybeWorkspaceFile &&
+        deletedFilesForWorkspace.some((deleted) => deleted.relativePath === maybeWorkspaceFile.relativePath))
```
tiagobento: Why the two possibilities? They seem equivalent to me.

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -28,61 +28,214 @@ import { SecurityIcon } from "@patternfly/react-icons/dist/js/icons/security-ico
...
+    });
+    if (props.onStatusChanged) {
+      props.onStatusChanged(
+        switchExpression(status, {
+          pending: isThereDeletedFile ? "deleted" : "modified",
+          synced: "synced",
+          default: undefined,
+        })
+      );
+    }
```
tiagobento:
```suggestion
      props.onStatusChanged?.(
        switchExpression(status, {
          pending: isThereDeletedFile ? "deleted" : "modified",
          synced: "synced",
          default: undefined,
        })
      );
```

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -28,61 +28,214 @@ import { SecurityIcon } from "@patternfly/react-icons/dist/js/icons/security-ico
...
+        modified: modifiedTooltip,
+        deleted: deletedTooltip,
+        default: <></>,
+      });
+    const maybeWorkspaceFile = props.workspaceFile;
+    if (maybeWorkspaceFile) {
+      tooltips.push(
+        tooltipForStageStatus(
+          props.workspaceGitStatusPromise?.data.fileStageStatuses.find(
+            (it) => it.path === maybeWorkspaceFile.relativePath
```
tiagobento: If you don't know what to call a variable, you can resort to object destruction... 
```suggestion
            ({path}) => path === maybeWorkspaceFile.relativePath
```

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -28,61 +28,214 @@ import { SecurityIcon } from "@patternfly/react-icons/dist/js/icons/security-ico
...
+          )?.status
+        )
+      );
+    } else {
+      new Set(props.workspaceGitStatusPromise?.data.fileStageStatuses.map((it) => it.status)).forEach((it) => {
+        tooltips.push(tooltipForStageStatus(it));
+      });
```
tiagobento: Same here.

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -28,61 +28,214 @@ import { SecurityIcon } from "@patternfly/react-icons/dist/js/icons/security-ico
...
+    );
+    const tooltipForStageStatus = (stageStatus?: FileStageStatus) =>
+      switchExpression(stageStatus, {
+        added: addedTooltip,
+        modified: modifiedTooltip,
+        deleted: deletedTooltip,
+        default: <></>,
+      });
+    const maybeWorkspaceFile = props.workspaceFile;
+    if (maybeWorkspaceFile) {
```
tiagobento: Really not necessary for this variable to exist.

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -28,61 +28,214 @@ import { SecurityIcon } from "@patternfly/react-icons/dist/js/icons/security-ico
...
     </Flex>
   );
 }
 
-export function WorkspaceStatusIndicator(props: { workspace: ActiveWorkspace }) {
+export function WorkspaceStatusIndicator(props: {
+  workspaceDescriptor: WorkspaceDescriptor;
+  workspaces: WorkspacesContextType;
+  workspaceGitStatusPromise: PromiseState<WorkspaceGitStatusType>;
+  currentWorskpaceFile: WorkspaceFile;
```
tiagobento: Typo.

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -28,61 +28,214 @@ import { SecurityIcon } from "@patternfly/react-icons/dist/js/icons/security-ico
...
+        default: <></>,
+      })}
     </Flex>
   );
 }
 
-export function WorkspaceStatusIndicator(props: { workspace: ActiveWorkspace }) {
+export function WorkspaceStatusIndicator(props: {
+  workspaceDescriptor: WorkspaceDescriptor;
+  workspaces: WorkspacesContextType;
```
tiagobento: Not good to have this as a prop, as this is a context.

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -96,46 +249,92 @@ export function WorkspaceStatusIndicator(props: { workspace: ActiveWorkspace })
...
+    >
+      <GitStatusIndicatorActions
+        variant={GitStatusIndicatorActionVariant.dropdown}
+        isOpen={isActionsDropdownOpen}
+        setOpen={setActionsDropdownOpen}
+        workspaceDescriptor={props.workspaceDescriptor}
+        workspaces={props.workspaces}
+        currentWorkspaceFile={props.currentWorskpaceFile}
+        onDeletedWorkspaceFile={props.onDeletedWorkspaceFile}
+        workspaceGitStatus={props.workspaceGitStatusPromise}
```
tiagobento: Please keep prop names consistent.. `workspaceGitStatus` -> `workspaceGitStatusPromise`

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
```diff
@@ -96,46 +249,92 @@ export function WorkspaceStatusIndicator(props: { workspace: ActiveWorkspace })
...
+}
+export const resolveGitLocalChangesStatus = (args: {
+  workspaceGitStatus?: WorkspaceGitStatusType;
+  file?: WorkspaceFile;
+}) => {
+  if (args.workspaceGitStatus === undefined) {
+    return WorkspaceGitLocalChangesStatus.unknown;
+  }
+  const maybeFile = args.file;
+  if (maybeFile) {
```
tiagobento: :D?

tiagobento (changes requested):
Great job! Left some comments inline, and as a general comment, I'd say let's try to keep new concepts cohesive. I know naming things is really hard, but it helps if the names are consistent. For example:

WS_CHECKOUT_HEAD_FILES
checkoutFilesFromLocalHead
checkoutLocalHead

They are part of the same concept, but they have slight variations on their names, making it hard to know for sure the they are really related or not.

I'll keep reviewing and will test it locally, but this first round is ready.

packages/online-editor/src/workspace/components/GitStatusIndicatorActions.tsx
```diff
@@ -0,0 +1,513 @@
...
+
+type GitStatusIndicatorActionsPopover = GitStatusIndicatorActionsBase & {
+  variant: GitStatusIndicatorActionVariant.popover;
+  isOpen: boolean;
+  setOpen: (isOpen: boolean) => void;
+};
+
+export type GitStatusIndicatorActionsType = GitStatusIndicatorActionsDropdown | GitStatusIndicatorActionsPopover;
+
+export const GitStatusIndicatorActions = (props: GitStatusIndicatorActionsType) => {
```
tiagobento: It's weird that a props type is not called `Props` at the end...

packages/online-editor/src/i18n/locales/en.ts
jstastny-cz: That's for the I18nWrapper component that is used also elsewhere. Just grep `wrapped` in this file. Does that mean we should not use that construct?

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
jstastny-cz: So we have an optional workspaceFile property passed to the indicator - if that's set it means the indicator applies just to that single file (in file menu).
When it's not set, it means the indicator is for whole workspace.

So when we don't have a specific file it is enough to check length of the deletedFilesForWorkspace array. But if we have a specific WorkspaceFile set, we need to check specifically, if deletedFilesForWorkspace contain exactly same relativePath.

packages/online-editor/src/home/HomePage.tsx
jstastny-cz: Right, this is what I was asking earlier - if we should be passing rather than using the use* hooks or not. Is there a rule of thumb which are better to pass and which are safe to invoke?
The concern was not to re-evaluate those like workspaceGitStatus, which might be expensive.

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
jstastny-cz: if I put `if(props.workspaceFile)` then it does not serve as type guard when accessing in the find function below. Am I missing something, or is it fine to put `props.workspaceFile?.relativePath` below, even though I have checked it's value in this if statement?

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
jstastny-cz: Anyway this method is not needed, removed now.

packages/online-editor/src/home/HomePage.tsx
tiagobento: Hmmmm, yeah, I can see how that's confusing.. Usually, as hooks should be done in a way that it's safe to call them multiple times, but the git status hook is an exception I guess... we could've materialized it to a Context rather then a simple hook.. this would make it safe for any component to call it without performance implications.. Let's say that for all hooks it's okay to call them any amount of times, except for this one for now.. :)

packages/online-editor/src/i18n/locales/en.ts
tiagobento: Oh, alright. If that was already in place, then it's okay. I don't exactly remember when we did this, but if the array is necessary, then let's leave it be. We'll have to revisit it at some point when we go back to working on i18n again.

packages/online-editor/src/workspace/components/WorkspaceStatusIndicator.tsx
tiagobento: Hmmm.. interesting.. Looks like a limitation from the TypeScript type inference.. I'd rather use an extra `?` then naming an unnecessary temporary variable.. WDYT?

thiagoelg: I've had this problem before, not sure why it happens 😕 
Destructuring the props outside the `useMemo` and then checking it might be a solution... not sure if you guys would like it though.

```js
const { workspaceFile } = props;

...

const indicatorTooltips = React.useMemo(() => {
  ...
  if (workspaceFile) {
     ...
   }
   ...
}, [workspaceFile]);
```

packages/online-editor/src/i18n/locales/en.ts
jstastny-cz: I've changed to plain string instead, there was no actual value brought by this construct.

jstastny-cz:
Latest changes include not only resolution of previous comments, but also further refactoring of FileSwitcher.tsx. The problem there was that the initial refactoring didn't account for displaying otherFiles inside a Split, which doubled width of the file menu. That was alright in EditorPage, but was causing issues in HomePage where I reuse the same component from FileSwitcher. To achieve consistent rendering in both places, I've switched to display otherFiles as another level of drilldownMenu, which keeps the width same all the time.
Next issue on HomePage were nested scrolls when there were too many files, while in FileSwitcher we need to put limits on max height, in HomePage this resulted in nested scrollbars, making interaction problematic. So I've introduced a conditional `isNoHeightMaxLimit` which conditionally removes the restrictions for file listings. One of the implications is that `View other files` menu item in HomePage is displayed after all the models (all the way down), while in EditorPage the button stays accessible all the time - because of the restricted height of preceding files listing.

packages/workspaces-git-fs/src/context/WorkspacesContext.tsx
```diff
@@ -148,6 +151,8 @@ export interface WorkspacesContextType {
 
   hasLocalChanges(args: { workspaceId: string }): Promise<boolean>;
 
+  getUnstagedModifiedFileRelativePaths(args: { workspaceId: string }): Promise<FileStageStatusType[]>;
```
caponetto: There are places where the returned list of this method is stored in a const named `fileStageStatuses` but I would expect a list of paths from a method called `getUnstagedModifiedFileRelativePaths`. Maybe call it `getUnstagedModifiedFileStageStatuses` or something like that?

packages/workspaces-git-fs/src/services/GitService.tsx
```diff
@@ -68,6 +68,16 @@ export interface RemoteRefArgs {
   };
 }
 
+export enum FileStageStatus {
+  added = "added",
+  modified = "modified",
+  deleted = "deleted",
+}
+export type FileStageStatusType = {
```
caponetto: Can the name `FileStageStatusType` be improved? IMO it may be confusing to have an enum called `FileStageStatus` that has status and a type with the same name (+`Type`) that has a path inside.

packages/workspaces-git-fs/src/services/GitService.tsx
```diff
@@ -158,6 +168,22 @@ export class GitService {
     });
   }
 
+  public async checkoutFilesFromLocalHead(args: {
```
caponetto, line 170: I am wondering if we can expand the args of the `checkout` method above to include `filepaths`, `noUpdateHead` and `force` because I don't see the point of having 2 checkout methods at this level (`GitService`). These different implementations could be done on `WorkspacesWorkerApiImpl`.

More on this: `kieSandboxWorkspacesGit_checkout` can call `gitService.checkout` in a way and `kieSandboxWorkspacesGit_checkoutFilesFromLocalHead` can call `gitService.checkout` in a different way.

WDYT?

caponetto (commented):
Overall looks good to me. Thanks, @jstastny-cz!
I've left some minor comments inline, please check if they make sense. :)

packages/workspaces-git-fs/src/services/GitService.tsx
jstastny-cz: @tiagobento what do you think? We have discusses this before and thought separate methods would be better - though I am not sure now if it meant already on GitService or only at the WorkspacesWorkerApi level.

packages/workspaces-git-fs/src/services/GitService.tsx
jstastny-cz: I agree, I've switched now to following:
```
export enum FileModificationStatus {
  added = "added",
  modified = "modified",
  deleted = "deleted",
}
export type UnstagedModifiedFilesStatusEntryType = {
  path: string;
  status: FileModificationStatus;
};
```
If you have a better proposal, please, help.

packages/workspaces-git-fs/src/context/WorkspacesContext.tsx
jstastny-cz: Renamed api methods to:
* `unstagedModifiedFilesStatus`
* `kieSandboxWorkspacesGit_getUnstagedModifiedFilesStatus`

Variables holding the value are now named `unstagedModifiedFilesStatus`. And while iterating through contents, the local var is named `stageStatusEntry`.

jstastny-cz:
I will shortly push new changes. Due to limited space on HomePage, it was not convenient to reuse same menu as in FileSwitcher, thus reverting those modifications.

packages/workspaces-git-fs/src/services/GitService.tsx
tiagobento: I think we could rename the first one to `checkoutAll` or `checkoutEverything`.. I don't think we should be exposing the full power of the `checkout` method from `isomorphic-git`.

packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
```diff
@@ -371,20 +377,26 @@ export function FileSwitcher(props: { workspace: ActiveWorkspace; workspaceFile:
...
+                          currentWorkspaceGitStatusProps={props.gitStatusProps}
+                          onDeletedWorkspaceFile={props.onDeletedWorkspaceFile}
                         />
                       </DrilldownMenu>
                     }
                   >
                     Current
                   </MenuItem>
-                  {workspacesMenuItems}
+                  <MenuGroup
```
jstastny-cz, line 392: @tiagobento The purpose of the MenuGroup maxHeight here was only to enable scrolling inside MenuContent while keeping the `Current >` menu item all the time at the top of the menu. In the last commit the overflow was moved to MenuContent itself thus scrolling all items including the `Current >`, I wanted to check if that was on purpose, if so, I will remove the maxHeight specification here.

jstastny-cz:
Scope was limited to the Editor page only, home page changes were reverted.

tiagobento (approved):

packages/online-editor/src/filesList/FileDataList.tsx
```diff
@@ -42,31 +46,74 @@ export function getFileDataListHeight(file: WorkspaceFile) {
...
+            <EyeIcon title="Currently selected file." />
+          </FlexItem>
+        )}
         <FlexItem style={{ minWidth: 0 /* This is to make the flex parent not overflow horizontally */ }}>
-          <Tooltip distance={5} position={"top-start"} content={fileName}>
+          <Tooltip
+            distance={5}
+            position={"top-start"}
+            content={
+              props.file.nameWithoutExtension +
```
thiagoelg, line 87: Shouldn't we follow the same rule as before here?
`const fileName = props.isEditable ? props.file.nameWithoutExtension : props.file.name;`

Showing the file extension if it's not editable.

PR #1482, opened 2023-02-23
kie-issues#125: Remove unnecessary FieldName type from PMML Marshaller

### LINKED ISSUE #125: Remove unnecessary `FieldName` type from PMML Marshaller
The types defined in [pmml4_4.ts](https://github.com/kiegroup/kie-tools/blob/main/packages/pmml-editor-marshaller/src/marshaller/model/pmml4_4.ts) includes an unnecessary `FieldName` class that just wraps a string value. This class isn't really used neither in the Marshaler (that generate the objects with the names as strings) neither in the Editor that is basically using the names as string (not as a `FieldName` instance). 

This is introducing a lot of unnecessary casts around the code that could be skipped if we were using a string directly. This situation is blocking the React 18 migration ([kie-issues#118](https://github.com/kiegroup/kie-issues/issues/118))

Note: The [pmml4_4.ts](https://github.com/kiegroup/kie-tools/blob/main/packages/pmml-editor-marshaller/src/marshaller/model/pmml4_4.ts) is autogenerated from the [pmml-4-4.xsd](https://github.com/manstis/pmml-codegen/blob/main/pmml-codegen/src/main/resources/pmml-4-4.xsd), if we check the `xsd` there are a bunch of `simpleTypes` like `FIELD-NAME` but all of them end up being a basic type in the models... only `FIELD-NAME` results in a class. That makes me think this could be a bug in the codegen tool.

pefernan:
Closes https://github.com/kiegroup/kie-issues/issues/125

Remove unnecessary `FieldName` class from the  `pmml-editor-marshaller`.

### DIFF
--- a/packages/pmml-editor-marshaller/src/marshaller/Marshaller.ts
+++ b/packages/pmml-editor-marshaller/src/marshaller/Marshaller.ts
@@ -38,7 +38,7 @@ import { textModelFactory } from "./jsonata/json2ui/TextModel";
 import { timeSeriesModelFactory } from "./jsonata/json2ui/TimeSeriesModel";
 import { treeModelFactory } from "./jsonata/json2ui/TreeModel";
 import { UI2JSON_TRANSFORMATION as ui2json } from "./jsonata/UI2JSON";
-import { CompoundPredicate, False, FieldName, PMML, SimplePredicate, True } from "./model/pmml4_4";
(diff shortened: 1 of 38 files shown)


yesamer (approved):
LGTM and Tested

Screenshot 2023-02-23 at 17 56 24

PR #1484, opened 2023-02-24
KOGITO-8729: Out of memory error while building dashbuilder-component-swf-editor in development

paulovmr:
To reproduce the error, run these commands without this change:
```
export NODE_OPTIONS="--max-old-space-size=2000"
pnpm bootstrap && pnpm -F @kie-tools/dashbuilder-component-swf-editor... build:dev
```
After the change, the build should be successful.


caponetto (approved):

jesuino (approved):
Editor is opening when running `dashbuilder-component-swf-editor` webapp and the build is successful

PR #1496, opened 2023-03-01
KOGITO-8731: Make filter enabled by default when a "filter" is added to the YAML

kumaradityaraj:
The filter attribute once added in YAML then it's not necessary to add the parameter enabled: 'true' ;  the dashbuilder will enale it itself.

Screenshot from 2023-03-01 11-08-02


jesuino (approved):
good work!

jesuino:
looks like the CI error is unrelated to the PR - Ubuntu build is green

paulovmr (approved):

PR #1501, opened 2023-03-02
KOGITO-8677: SWF Diagram Editor - i18n

hasys:

there are discussed changes about i18n of all string literals. Visually should be nothing changed, but now all literals can be changed in `SWConstants.properties` files and i18n if needed.


packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/HasTranslation.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.stunner.sw.client.shapes;
+
+public interface HasTranslation {
```
romartin, line 19: Hey @hasys 
Do you think we really need this interface? I don't see any usage. I think it does not provides any additional benefit, I would just drop this, trying to make the code smaller and cleaner (not so many interfaces). WDYT?

romartin (commented):
Hey @hasys 

Code looks good, just added a single suggestion on code. Let me know what you think.

Thanks!

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/HasTranslation.java
hasys: Hi @romartin, yes, good point. It was an initial idea, but the issue I met is we need this method in several interfaces (default methods implementations) and in implementations (classes) as well. So there will be many `getTranslations` and `translationService` should be send from implementation to interface which complicates design and interface even further.

Making getTranslation as separate interface looks like the simplest solution. Since we implement it once in root `ServerlessWorkflowShape` and than just attach the HasTranslation to all needed interfaces and we can use `getTranslation` everywhere for free in all interfaces and in all Shapes without passing method parameters from anywhere to anywhere, just calling `getTranslation` and it's there.

We can have a call if you have another idea.

Thank you!

romartin: ok @hasys , looks good to me

romartin (approved):

LuboTerifaj (approved):
It looks good to me.

hasys:
Hello @kiegroup/gatekeepers this PR is ready for merge. Thank you!

PR #1502, opened 2023-03-02
kie-issues#42: DMN Editor: Repetitive opening of properties panel for List expression

### LINKED ISSUE #42: DMN Editor: Repeatitive opening of properties panel for List expression
If User tries to open properties panel for multiple List Expression entries, an error is thrown - stack size exceeded.

For more details see the issue and attached video there https://issues.redhat.com/browse/KOGITO-2961

Or alternatively:
- Create new diagram
- Add a Decision node
- Open the decision node
- Set its expression type as List expression
- Open and Close (~5times) properties panel for the first list entry
- Add one more list entry
- Again Open and Close (~5times) properties panel for the second list entry
- When properties panel is closed/collapsed, select first list entry
- Try to open properties panel, an exception will be thrown

Screenshot_20230111_110927

danielzhe:
Closes https://github.com/kiegroup/kie-issues/issues/42

An easier way to reproduce the issue, as analyzed by @yesamer in https://github.com/kiegroup/kie-issues/issues/42:
1. Add a Decision
2. Open de Decision node
3. Add a List
4. Add a new list item
5. Select the first item

**Analysing the issue**
As @yesamer mentioned in his first analysis, the error was caused because of a recursive reference in 'DMNModelInstrumentedBase:getDefaultNamespace()'

Every row in a 'List' is a 'HasExpression' encapsulated in 'WrappedHasExpression'.
The _parent_ of 'WrappedHasExpression' is the list itself and the _expression_ is the real expression (a 'LiteralExpression' in this case).
The issue was happening because in the 'AddListRowCommand' the _parent_ of the 'WrappedHasExpression' (which is the 'List') was being set to the 'List' but it should be the _expression_ inside 'WrappedHasExpression' that needs to have its _parent_ set to 'List'.

Maybe the method 'asDMNModelinstrumentedBase()' is incorrect because its returning the _parent_ instead of the _expression_ wrapped in _WrappedHasExpression_, but change this method will require to check it in all other places where it is called....


yesamer (commented):
Thank you @danielzhe I agree with your analysis, based on your findings, LGTM.
Can you please double-check the places where that `WrappedHasExpression.asDMNModelInstrumentedBase()` is actually used? Maybe the same bug could be present in that points as well. 
Thx!

danielzhe:
@yesamer In my opinion it is not feasible to track all places where `WrappedHasExpression.asDMNModelInstrumentedBase()` is used, because `asDMNModelInstrumentedBase()` is an implementation from `HasExpression`. So we gonna need to track all places where the instance of `HasExpression` is `WrappedHasExpression` and analyze if the logic is correct.

yesamer:
@danielzhe This is the PR that introduced it: https://github.com/kiegroup/kie-wb-common/pull/3220. 
Looking in the test It seems that returning the parent was intended https://github.com/kiegroup/kogito-editors-java/commit/24a424c0eb4a6c1e64c18cb1f638bcc978048c11#diff-0ee778ee563e991f475a6fdaa1079e9533357bb53f877521ec120344accab7c7R55.

yesamer (approved):
LGTM and Tested, thank you @danielzhe

PR #1366, opened 2022-12-09
KOGITO-7999: Stunner - Editor API Canvas/Session/Domain Definitions exposed in JS

handreyrc:
Hey @romartin, could you please review this PR?  

**JIRA**: [KOGITO-7999](https://issues.redhat.com/browse/KOGITO-7999)

Thanks!

PR #1505, opened 2023-03-03
KOGITO-8798: Skip VSCode Select states test because of bug in Canvas API

LuboTerifaj:
https://issues.redhat.com/browse/KOGITO-8798

### DIFF
--- a/packages/vscode-extension-serverless-workflow-editor/it-tests/serverless-workflow-editor-extension-diagram-navigation.test.ts
+++ b/packages/vscode-extension-serverless-workflow-editor/it-tests/serverless-workflow-editor-extension-diagram-navigation.test.ts
@@ -43,7 +43,8 @@ describe("Serverless workflow editor - Diagram navigation tests", () => {
     await testHelper.closeAllNotifications();
   });
 
-  it("Select states", async function () {
+  // The following test is skipped because there is a bug in Canvas API: https://issues.redhat.com/browse/KOGITO-8793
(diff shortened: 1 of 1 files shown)


LuboTerifaj:
Hello @handreyrc , @domhanak , could you please have a look at this PR? Thank you!

domhanak (approved):
What's the reason behind this PR? Was the test failing somewhere? You reference a bug with coordinates but I do not see them used now which is confusing me. Can't we just wait until #1366  and do a proper fix after it is merged? 

Here is my approval although I would prefer fixing the test directly in #1366

LuboTerifaj:
Hello @domhanak ,

the coordinates and dimensions are not used because the part of the test is commented out. It was commented out in https://github.com/kiegroup/kie-tools/pull/1498.

As a result, the test is passing without testing what it is supposed to test. The PR changes the passed test, which is the state we want to avoid because the issue mentioned in my previous point, to the skipped one with the reference to the issue why it is skipped.

Thanks!

handreyrc (approved):
Hey @LuboTerifaj,

I agree that conceptualy it is a better approach to skip those tests other than commenting them out. This way we keep a better track of the tests executed and test summary consistent with what was indeed tested and what wasn't.
That being said I'm approving this PR.

Thanks

LuboTerifaj:
Hello @paulovmr , this is ready to merge. Thanks!

caponetto (approved):

PR #1515, opened 2023-03-08
KOGITO-8793: BoudingBox values of states with tooltip boxes are wrong

hasys:
Hello @LuboTerifaj, @handreyrc,

there is a fix, now tooltip is created on mouse hover and removed on mouse out, so BoundingBox not expanding by the tooltip until it is not created.

Thank you!

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/icons/BottomDepiction.java
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/icons/BottomDepiction.java
@@ -31,7 +31,7 @@ public class BottomDepiction extends Group {
     private final HandlerRegistration mouseClickHandler;
 
     public BottomDepiction(String icon) {
-        setLocation(new Point2D(STATE_SHAPE_WIDTH / 2 - 8, STATE_SHAPE_HEIGHT - 18));
+        setLocation(new Point2D(STATE_SHAPE_WIDTH / 2 - 8, STATE_SHAPE_HEIGHT - 20));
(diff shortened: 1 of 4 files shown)


hasys:
Hi @LuboTerifaj,

we checked with @handreyrc  the it tests and they are passing locally and failing on the CI, looks like we have random fails back on the menu. Is it known issue?

I restarted the build, let's see if it will pass now. Thank you!

LuboTerifaj (changes requested):
Hello @hasys ,

I found issue with node dimensions. The function returns [250, 92] instead of [250, 90] for states with icons.

Could you please also add the tests for checking the expected values?

Regarding the test stability, yes there may be a different environment configuration using CI in comparison with local env. We should definitely stabilize it once this issue is resolved.

I can have a look at the issue afterwards.

Thank you!

handreyrc (changes requested):
Hey @hasys, 

Great job here the code looks good and works fine to me!

In my opinion it would be good to add some regular unit tests around the creation and destruction of the the tooltip box making sure those methods are being called in the process. 
It should also be possible to add some IT tests in the scope of the sw-editor to assert the dimensions of the group. There are some helper classes implemented to call js methods.

WDYT?

hasys:
Hello @handreyrc, @LuboTerifaj,

I added integration test for Dimension. Thank you for the quick feedback!

handreyrc (approved):
Hey @hasys,

Thanks for adding IT tests!
Everything looks good to me.

 Thanks a lot for your work here.

LuboTerifaj (approved):
Hey @hasys ,

it looks good to me now.

Thank you!

hasys:
Thank you @LuboTerifaj, @handreyrc!

Hello @kiegroup/gatekeepers, this PR is ready for merge. Thank you!

romartin (approved):

PR #1517, opened 2023-03-08
KOGITO-8824: [Dashbuilder] Support "accumulate" flag in data sets

jesuino:
This is a new feature for Dashbuilder datasets where it is possible to configure a dataset to keep the old values in memory.

To test: 

* Build until dashbuilder-editor and run dashbuilder-editor as a webapplication.
* Use the following YAML - make sure that the file data exists on the dataset url

```
datasets:
  - uuid: acc
    url: http://127.0.0.1:3001/data
    accumulate: true
    cacheMaxRows: 5    
pages:
  - components:  
      - settings:                    
          component: table
          refresh:
              interval: 1
          lookup:
            uuid: acc
```
* Create some dataset, here's a sample:

```
[
    ["A", 3000]
]
```
* Notice that dashbuilder constantly pulls data from the dataset, so if you modify the file you will see that the old values are kept

image

You can see also in this video

https://drive.google.com/file/d/1ZwDq0jgnMeYvdS5GFuGRTrJO7T2VXxc8/view?usp=share_link


jesuino:
Hello @barboras7 could you please review this? Thanks!

packages/dashbuilder-component-timeseries/src/TimeSeriesComponent.tsx
```diff
@@ -54,13 +54,8 @@ const INIT_OPTION = {
...
 const validate = (columns: Column[]): string | undefined => {
-  if (
-    columns.length < 3 ||
-    columns[0].type !== "LABEL" || // series
-    columns[1].type !== "NUMBER" || // time
-    columns[2].type !== "NUMBER" // value
-  ) {
-    return "Data Set is Invalid! You must provide at least 3 columns container the series (LABEL), timestamp (NUMBER) and the value (NUMBER)";
+  if (columns.length < 3) {
+    return "Data Set is Invalid! You must provide at least 3 columns containing the series, timestamp and the value";
```
barboras7:
```suggestion
    return "Data Set is invalid! You must provide at least 3 columns containing the series, timestamp and the value";
```

barboras7:
Hi @jesuino,

for interval, the number is underlined even though the value is correct (Incorrect type, expected string). This could be misleading for users, see the screenshot attached. 
Screenshot from 2023-03-17 15-24-15

jesuino:
Hello @barboras7 

Thanks for the feedback.

We are working to improve our schema and avoid these misleading messages, please see [1]

So this problem will be addressed soon!


[1] https://issues.redhat.com/browse/KOGITO-8898

barboras7:
Then everything looks good! Just the minor comment I left with the "typo" (but that's mostly optional, not a bug).

Thanks! :tada:

barboras7 (approved):

caponetto (approved):

PR #1525, opened 2023-03-15
kie-issues#91: An overhaul of how we handle extensions both in KIE Sandbox and Serverless Logic Web Tools 

### LINKED ISSUE #91: An overhaul of how we handle extensions both in KIE Sandbox and Serverless Logic Web Tools
Steps to reproduce:
- Import this URL on KIE Sandbox https://gist.github.com/tiagobento/6cd55debe616913f986dc9e661a68b07#file-test-bpmn-bpmn
- Observe that it opens as a BPMN, but the File Label is not shown (????)
- File name seems correct, but on the URL we see `test.bpmn.bpmn`, so we should see `test.bpmn` at the file name on the Editor Page.
 
> NOTE: There's important information and comments at the original Jira too. See https://issues.redhat.com/browse/KOGITO-4708

Expected behaviour:
- The UI should be able to display the File Label correctly and the name displayed on screen should be `test.bpmn`, as the original file name is `test.bpmn.bpmn`.

klbowers:
Closes https://github.com/kiegroup/kie-issues/issues/91


packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const REGEX = {
+  supported:
```
tiagobento: Is this used somewhere? I couldn't find usages for it..

packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
```diff
@@ -0,0 +1,129 @@
...
+  DASH_YML = "dash.yml",
+}
+
+export enum FileTypesSandbox {
+  DMN = "dmn",
+  BPMN = "bpmn",
+  BPMN2 = "bpmn2",
+  SCESIM = "scesim",
+  PMML = "pmml",
+}
```
tiagobento: Why not make these a single enum?

tiagobento (changes requested):
Very good! Left some minor comments inline.

packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
klbowers: It was causing an error at \serverless-logic-web-tools\src\workspace\components\FileLabel.tsx with the const labelColors, which wanted the sandbox file types to be defined in there.

packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
klbowers: No usages anywhere. I thought maybe it was there as a sort of "comment" to have all supported filetypes in one type

klbowers:
~~Do not merge! Investigating a bug where a file with file extension e.g. .dmn, will error if .dmn is anywhere in the name of the file except at the end.~~ Fixed!

packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
tiagobento: Alright, so if there are no usages, let's delete it :D

packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
tiagobento: Can you give more details? I'm not sure I understand the problem..

tiagobento (changes requested):
Alright, looking good! Let's solve the two remaining comments before proceeding.. I need to test it locally too..

@caponetto Can you help us verify this on Serverless Logic Web Tools?

packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export function isDashbuilderYml(path: string): boolean {
+  return REGEX.dashYml.test(path);
+}
+
+export function isDashbuilderYaml(path: string): boolean {
+  return REGEX.dashYaml.test(path);
+}
+
+export function isDecision(path: string): boolean {
```
caponetto: I know these methods (`isXYX`) were only moved but I think we can replace them with some TS tricks, like:

```ts
type FileRegexKind = keyof typeof REGEX;
const matchers: Record<FileRegexKind, (path: string) => boolean> = {} as any;
for (const key in REGEX) {
  const kind = key as FileRegexKind;
  matchers[kind] = (path: string): boolean => REGEX[kind].test(path);
}

function isOfKind(kind: FileRegexKind, path: string) {
  return matchers[kind](path);
}

// usage examples
isOfKind("swYaml", "foo/bar/model.sw.json") // ----> false 
isOfKind("swJson", "foo/bar/model.sw.json") // ----> true
isOfKind("bpmn", "model.bpmn") // ----> true
isOfKind("bpmn", "model.dmn") // ----> false
isOfKind("bpmn", "model.bpmn.bpmn") // ----> true
```

Anyway, just a suggestion.

packages/workspaces-git-fs/src/relativePath/WorkspaceFileRelativePathParser.ts
```diff
@@ -28,18 +42,35 @@ export function parseWorkspaceFileRelativePath(relativePath: string) {
...
-  const matchDots = fileName.match(/\./g);
-  if (matchDots && matchDots.length > 1) {
-    return fileName
-      .split(/\.(.*)/s)
-      .slice(1)
-      .join("");
+  const fileName = basename(relativePath).toLowerCase();
+  if (fileName.includes(".")) {
+    switch (true) {
+      case isServerlessWorkflowJson(fileName):
```
caponetto: Why use a switch-case statement here?
I think if-return would suit better.

caponetto (commented):
Thanks guys! I will test locally the serverless logic web tools tomorrow.

packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
klbowers: Screenshot (4)
There's a serverless-logic const called 'labelColors' which is a record with 'FileTypes' as a property, so when the sandbox FileTypes and serverless FileTypes are a single enum, labelColors requires dmn,bpmn,etc. to be defined in there. I've changed it to have string as a property instead of FileTypes (which is what online-editor does)

packages/workspaces-git-fs/src/constants/ExtensionHelper.ts
klbowers: Feels a lot cleaner, thanks for that!

packages/workspaces-git-fs/src/relativePath/WorkspaceFileRelativePathParser.ts
```diff
@@ -13,13 +13,13 @@
...
  */
-
 import { basename, extname, parse } from "path";
+import { isOfKind, FileTypes } from "../constants/ExtensionHelper";
 
 export function parseWorkspaceFileRelativePath(relativePath: string) {
   const extension = extractExtension(relativePath);
   return {
-    relativePathWithoutExtension: relativePath.replace(`.${extension}`, ""),
+    relativePathWithoutExtension: relativePath.substring(0, relativePath.lastIndexOf("." + extension)),
```
caponetto: The only issue I found on Serverless Logic Web Tools is that I cannot open files that do not have extensions anymore. 

Probably because of this change.
For example, if I have `relativePath=test`, `relativePathWithoutExtension` returns `""` but it should be `test`.

klbowers: Fixed, sorry about that! Didn't realize that SLWT could open files without extensions

caponetto: Thanks!!

caponetto (approved):
LGTM! Thanks @kbowers-ibm!
I suggest changing the title of this PR to something that better describes the issue.

tiagobento (approved):

PR #1527, opened 2023-03-17
kie-issues#85: fix SearchInput focus and FileListItem display regressions

jstastny-cz:
Problem was duplicate usage of a single ref object in two places (I though that the fact that containing component was a Memo value would be enough). 

Switching to callback and conditional assignment of the ref object into a specific occurrence of the SearchInput does resolve. Together with that I renamed `isScrollUsingRefEnabled` state property to `isModelsMenuActive` and reusing for 2 purposes now:
* scroll enabled only when isModelsMenuActive===true
* search ref object is assigned conditionally based on isModelsMenuActive value. (true -> search in models, false -> search in other files).

### DIFF
--- a/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
+++ b/packages/online-editor/src/editor/Toolbar/FileSwitcher.tsx
@@ -376,6 +376,7 @@ export function FileSwitcher(props: {
                     drilldownMenu={
                       <DrilldownMenu id={`dd${props.workspace.descriptor.workspaceId}`}>
                         <FilesMenuItems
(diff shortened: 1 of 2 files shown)


jstastny-cz:
One pending fix will be pushed shortly.

thiagoelg:
On `packages/online-editor/src/filesList/FileDataList.tsx` we should still be able to see the file extension if it's not editable, right?

See this comment:
https://github.com/kiegroup/kie-tools/pull/1481/files#r1140276318

jstastny-cz:
@thiagoelg thank you, I will adjust the rendering as it used to be.

tiagobento (approved):

PR #1541, opened 2023-03-28
KOGITO-8316: SWF Viewer - Parallel state requires more details

hasys:
Hello @romartin, @LuboTerifaj,

there is a PR we discussed yesterday. For more info see [KOGITO-8316](https://issues.redhat.com/browse/KOGITO-8316).

Thank you!


fjtirado (commented):
@hasys Thanks a lof for the quick reaction. Feature looks pretty cool!!!

fjtirado (approved):

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/definition/custom/NumCompletedJsonTypeDeserializer.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    public Object deserialize(JsonValue value, DeserializationContext ctx) {
+        switch (value.getValueType()) {
+            case STRING:
+                return new StringJsonDeserializer().deserialize(value, ctx);
+            case NUMBER:
+                return new BaseNumberJsonDeserializer.IntegerJsonDeserializer().deserialize(value, ctx);
+            default:
+                return null;
```
fjtirado, line 34: Null or exception? ;)

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/HasBranches.java
```diff
@@ -0,0 +1,42 @@
...
+        if (branches == null || branches.length == 0) {
+            return getTranslation(BRANCHES_NOT_DEFINED);
+        }
+
+        StringBuilder result = new StringBuilder();
+        for (ParallelStateBranch branch : branches) {
+            result.append(getTranslation(BRANCH))
+                    .append(": ")
+                    .append(branch.getName())
+                    .append("\r\n");
```
fjtirado, line 38: Hmmm it wont be better to use System.getProperty("line.separator") rather than \r\n (which is windows)?
Or is this intentional because the tooling requires it?

hasys: Yes, it is Lienzo specific new line. We converting java to javascript and executing in the browser, so tooltip new line is purely Lienzo thing here.

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/definition/custom/NumCompletedJsonTypeDeserializer.java
hasys: It's Null, for almost all situations it means parameter is not present in the JSON which is correct it can be omitted by the spec. For the rest situations like user putted object instead of String or Number I guess it's better in this case to excuse the user and just pretend it never happened and just load the rest of the diagram.

If this parameter is actually required but not provided or provided incorrect value in this case on the diagram user will se "atLeast: undefined" in the bottom of the corresponding Parallel State instead of crashed loading of the whole diagram. I guess it will be pretty easy to identify and fix the issue for the user. Also user will be notified on the Text Editor side about invalid value (at least if our VS Code plugin is used). So I guess it's pretty safe to just ignore invalid value here. WDYT?

fjtirado: I thini is ok to return null then, thanks for the detailed explanation.

romartin (approved):
Great stuff @hasys ! Thanks

LuboTerifaj (changes requested):
Hello @hasys ,

I have found four issues. I have reported three of them here as they are the corner cases and not critical ones:

- https://issues.redhat.com/browse/KOGITO-8939 _SWF Viewer - Parallel state completionType info and toolbox text exceed the boundaries_

The one that needs to be fixed:

- If there is no action specified in one of the branches, there is a message "No actions provided" shown without a new line separator. The next branch follows immediately without any separation.

Thanks!

hasys:
Hi @LuboTerifaj,

great catches! Fixed.

LuboTerifaj (approved):
Thank you for the fix @hasys . It looks good now.

PR #1556, opened 2023-04-06
kie-issues#157:  Connection to Shared Worker is lost when KIE Sandbox tab is frozen by chrome

### LINKED ISSUE #82: Having multiple tabs of different versions KIE Sandbox open may cause issues with the Shared Worker
Since a lot of our processing in the frontend is done via Shared Worker most releases will need to update it.
That can be an issue when the user has multiple tabs of the KIE Sandbox open and some of them are on older versions, making the Shared Worker unable to reset after an update. This can break the experience on the tabs that were updated, forcing them to work with the outdated Shared Worker.

A good workaround/fix for this situation is to concatenate the KIE Sandbox version to the generated Shared Worker, this way each version can have its own worker registered on the browser.

### LINKED ISSUE #157: Connection to Shared Worker is lost when KIE Sandbox tab is frozen by Chrome
Whenever Chrome freezes a tab (normally to save CPU cycles) the Shared Worker connection is lost and not initiated again, leaving all functionality that depends on it unavailable, without warning the user or recovering itself until the tab is reloaded manually.

Check this video on how to emulate a Chrome tab being frozen:

https://user-images.githubusercontent.com/10515135/229836496-58052257-eeeb-4001-bbfa-c8ced682f702.mov

Obs.: You can check the tab status at [`chrome://discards`](chrome://discards)

The Shared Worker connection is lost after a few seconds, to debug it go to [`chrome://inspect/#workers`](chrome://inspect/#workers) and click on `Inspect` on the `workspaces-shared-worker` Shared Worker.

thiagoelg:
Closes https://github.com/kiegroup/kie-issues/issues/157
Closes https://github.com/kiegroup/kie-issues/issues/82

Fixes the connection to the Shared Worker whenever the browser freezes and the resumes a tab.
This happens because while frozen the `ping` function from the Shared Worker will timeout and eventually close the connection.

This PR adds an event listener to the `resume` event and recreates the Shared Worker connection.

This PR also fixes the problem of having multiple tabs of different versions of KIE Sandbox, by prefixing the Shared Worker name with the current version.

### DIFF
--- a/packages/serverless-logic-web-tools/src/App.tsx
+++ b/packages/serverless-logic-web-tools/src/App.tsx
@@ -38,7 +38,11 @@ export const App = () => (
       [SettingsContextProvider, {}],
       [
         WorkspacesContextProvider,
-        { workspacesSharedWorkerScriptUrl: "workspace/worker/sharedWorker.js", shouldRequireCommitMessage: false },
+        {
+          workspacesSharedWorkerScriptUrl: "workspace/worker/sharedWorker.js",
+          shouldRequireCommitMessage: false,
+          workerNamePrexi: `serverless-logic-web-tools-${process.env.WEBPACK_REPLACE__version}`,
+        },
       ],
       [OpenShiftContextProvider, {}],
       [VirtualServiceRegistryContextProvider, {}],
--- a/packages/serverless-logic-web-tools/src/App.tsx
+++ b/packages/serverless-logic-web-tools/src/App.tsx
@@ -41,7 +41,7 @@ export const App = () => (
         {
           workspacesSharedWorkerScriptUrl: "workspace/worker/sharedWorker.js",
           shouldRequireCommitMessage: false,
-          workerNamePrexi: `serverless-logic-web-tools-${process.env.WEBPACK_REPLACE__version}`,
+          workerNamePrefix: `serverless-logic-web-tools-${process.env.WEBPACK_REPLACE__version}`,
         },
       ],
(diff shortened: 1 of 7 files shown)


packages/serverless-logic-web-tools/src/App.tsx
```diff
@@ -38,7 +38,11 @@ export const App = () => (
       [SettingsContextProvider, {}],
       [
         WorkspacesContextProvider,
-        { workspacesSharedWorkerScriptUrl: "workspace/worker/sharedWorker.js", shouldRequireCommitMessage: false },
+        {
+          workspacesSharedWorkerScriptUrl: "workspace/worker/sharedWorker.js",
+          shouldRequireCommitMessage: false,
+          workerNamePrexi: `serverless-logic-web-tools-${process.env.WEBPACK_REPLACE__version}`,
```
caponetto:
```suggestion
          workerNamePrefix: `serverless-logic-web-tools-${process.env.WEBPACK_REPLACE__version}`,
```
🙈

thiagoelg: ooops

thiagoelg: That made me realize that the `nest` helper function doesn't check for types 👀 

Not sure on how to implement that though...

caponetto (approved):

tiagobento (approved):

PR #1571, opened 2023-04-18
kie-issues#138: Create new feature on KIE Sandbox to allow the enabling/disabling of the PMML editors as a supported graphical editor

### LINKED ISSUE #138: Create new feature on KIE Sandbox to allow the enabling/disabling of the PMML editors as a supported graphical editor
Some distributions of KIE Sandbox might not be interested in allowing users to manually tweak PMML files on KIE Sandbox. We need a new environment variable to enable/disable the PMML Editor on KIE Sandbox. This env var must be usable on builds directly, or on `kie-sandbox-image`.

klbowers:
Closes https://github.com/kiegroup/kie-issues/issues/138

### DIFF
--- a/packages/kie-sandbox-image/README.md
+++ b/packages/kie-sandbox-image/README.md
@@ -237,3 +237,46 @@ Here's an example of what it should look like:
     otherFilesDestinationFolder: "src/main/resources/others",
 }
 ```
+
+### Editors
+
+By default all editors will be enabled. To disable an editor simply delete/comment out the respective json.
+
+- **type**: The type of file that you want to edit.
+- **filePathGlob**: The glob pattern of the file you want to edit.
+- **resourcesPathPrefix**: The path to the editor you want to get to.
+- **path**: Populates the envelopeContent.
+- **title**: The title that will show up on the home page.
+- **extension**: The extension of the file type that you want to edit.
+- **description**: An optional parameter that will display a short description under the title on the home page.
+  Here's an example of what it should look like:
+
+```js
+    {
+      type: FileTypes.BPMN,
+      filePathGlob: GLOB_PATTERN.bpmn,
+      resourcesPathPrefix: "gwt-editors/bpmn",
+      path: "bpmn-envelope.html",
+      title: "Workflow",
+      extension: "bpmn",
+      description: "BPMN files are used to generate business workflows.",
+    },
+    {
+      type: FileTypes.DMN,
+      filePathGlob: GLOB_PATTERN.dmn,
+      resourcesPathPrefix: "gwt-editors/dmn",
+      path: "dmn-envelope.html",
+      title: "Decision",
+      extension: "dmn",
+      description: "DMN files are used to generate decision models",
+    },
+    {
+      type: FileTypes.PMML,
+      filePathGlob: GLOB_PATTERN.pmml,
+      resourcesPathPrefix: "",
+      path: "pmml-envelope.html",
+      title: "Scorecard",
+      extension: "pmml",
+      description: "PMML files are used to generate scorecards",
+    },
+```
--- a/packages/kie-sandbox-image/README.md
+++ b/packages/kie-sandbox-image/README.md
@@ -58,7 +58,7 @@ This package contains the `Containerfile/Dockerfile` and scripts to build a cont
    | :---------------------------------------------------------: | :---------------------------------------------------------------------------------------------------------------------------: | :-----------------------------------------------------------------: |
    |             `KIE_SANDBOX_EXTENDED_SERVICES_URL`             |                                   The URL that points to the KIE Sandbox Extended Services.                                   | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
    |              `KIE_SANDBOX_GIT_CORS_PROXY_URL`               |                         The URL that points to the Git CORS proxy for interacting with Git providers.                         | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
-   |        `KIE_SANDBOX_REQUIRE_CUSTOM_COMMIT_MESSAGES`         |                           Require users to type a custom commit message when creating a new commit.                           | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
+   |         `KIE_SANDBOX_REQUIRE_CUSTOM_COMMIT_MESSAGE`         |                           Require users to type a custom commit message when creating a new commit.                           | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
    | `KIE_SANDBOX_CUSTOM_COMMIT_MESSAGES_VALIDATION_SERVICE_URL` |                                           Service URL to validate commit messages.                                            | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
    |                `KIE_SANDBOX_AUTH_PROVIDERS`                 |     Authentication providers configuration. Used to enable integration with GitHub Enterprise Server instances and more.      | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
    |                 `KIE_SANDBOX_ACCELERATORS`                  |  Accelerators configuration. Used to add a template to a set of Decisions and Workflows, making it buildable and deployable.  | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
@@ -246,10 +246,10 @@ By default all three standard editors will be enabled (BPMN, DMN, PMML). To disa
 
 - **extension**: The extension of the file that you want to edit.
 - **filePathGlob**: The glob pattern of the file you want to edit.
-- **resourcesPathPrefix**: The path to the gwt-editor.
-- **path**: The path of the editor envelope.html.
-- **title**: The title of the editor that will be displayed on the home page.
-- **description**: Displays a short description of the editor under the title on the home page.
+- **editor.resourcesPathPrefix**: The path to the gwt-editor.
+- **editor.path**: The path of the editor envelope.html.
+- **card.title**: The title of the editor that will be displayed on the home page.
+- **card.description**: Displays a short description of the editor under the title on the home page.
 
 Here's an example of what it should look like:
 
--- a/packages/kie-sandbox-image/README.md
+++ b/packages/kie-sandbox-image/README.md
@@ -240,15 +240,14 @@ Here's an example of what it should look like:
 
 ### Editors
 
-By default all editors will be enabled. To disable an editor simply delete/comment out the respective json.
+By default all three standard editors will be enabled (BPMN, DMN, PMML). To disable an editor simply delete/comment out the respective json.
 
 - **type**: The type of file that you want to edit.
 - **filePathGlob**: The glob pattern of the file you want to edit.
 - **resourcesPathPrefix**: The path to the editor you want to get to.
-- **path**: Populates the envelopeContent.
-- **title**: The title that will show up on the home page.
-- **extension**: The extension of the file type that you want to edit.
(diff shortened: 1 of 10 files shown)


packages/kie-sandbox-image/README.md
tiagobento, line 1: We need to add this new env var to the table here as well. Please follow the other variables and do a similar thing to this new one.

packages/online-editor/src/envelopeLocator/hooks/EditorEnvelopeLocatorContext.tsx
```diff
@@ -38,3 +42,11 @@ export function EditorEnvelopeLocatorContextProvider(props: { children: React.Re
 export function useEditorEnvelopeLocator() {
   return useContext(EditorEnvelopeLocatorContext);
 }
+
+export function getEditorConfig() {
```
tiagobento: This is a custom hook, right? I mean, you're "abstracting" a `useEnv` and a `useMemo` here. I think this needs to be called `useEditorEnvelopeConfig`.

packages/online-editor/src/workspace/worker/sharedWorker.ts
```diff
@@ -30,16 +31,24 @@ async function gitCorsProxyUrl(): Promise<string> {
...
-// eslint-disable-next-line prefer-const
 onconnect = async (e: MessageEvent) => {
-  console.log("Connected to Workspaces Shared Worker");
+  const editorEnvelopeConfigs = await fetchEditorEnvelopeConfig();
+  const editorEnvelopeLocator = new EditorEnvelopeLocatorFactory().create({
+    targetOrigin: "",
+    editorEnvelopeConfigs,
+  });
 
+  // eslint-disable-next-line prefer-const
```
tiagobento: This line was before `onconnect = ...` before, right? We need to keep it in the same place.

packages/kie-sandbox-image/README.md
```diff
@@ -237,3 +237,43 @@ Here's an example of what it should look like:
...
+    },
+    {
+      type: FileTypes.PMML,
+      filePathGlob: GLOB_PATTERN.pmml,
+      resourcesPathPrefix: "",
+      path: "pmml-envelope.html",
+      title: "Scorecard",
+      description: "PMML files are used to generate scorecards",
+    },
+```
```
tiagobento: We're not in a JS context here. This needs to be plain JSON, so people configuring their env vars using a terminal will be able to use this one as reference.

tiagobento: I.e., GLOB_PATTERN and FileTypes do not exist in your terminal.

packages/online-editor/src/envelopeLocator/EditorEnvelopeLocatorApi.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+export type EditorEnvelopeConfig = {
+  type: string;
+  filePathGlob: string;
+  resourcesPathPrefix: string;
+  path: string;
+  title: string;
+  description: string;
+};
```
tiagobento:
```suggestion
export type EditorEnvelopeConfig = {
  extension: string;
  filePathGlob: string;
  editor: {
    resourcesPathPrefix: string;
    path: string;
  },
  card: {
    title: string;
    description: string;
  }
};
```

Please change this type to the structure above. It's easier to group related properties inside a nested object. Also change the necessary places after this change.

packages/online-editor/src/home/HomePage.tsx
```diff
@@ -144,24 +146,21 @@ export function HomePage() {
...
-                  title={"Scorecard"}
-                  extension={"pmml"}
-                  description={"PMML files are used to generate scorecards"}
-                />
+                {editorConfig.map((config, index) => {
+                  return (
+                    <NewModelCard
+                      key={index}
+                      title={config.title}
+                      extension={config.type}
```
tiagobento: This is a discrepancy in naming. `extension` receiving a `type`. See my suggestion below about changing the EditorEnvelopeConfig type.

tiagobento (changes requested):
Thanks @kbowers-ibm ! Left some comments inline.

As a general comment, when introducing a new type or concept, we need to be careful to address it the same way everywhere. I see in this PR you're some times calling it `editorConfig`, and some times `editorEnvelopeConfig`. We need to choose one and make sure every place uses the same term.

More than that, maybe internally "Editor envelope config" makes sense to us, but for the outside (thinking about the env var here), we should not say "envelope". What do you think about simply `KIE_SANDBOX_EDITORS`? I mean, the fact that it is a env var already says it's a "config"... Then internally we can call it `EditorConfig`, following the same we did for `AuthProvider`.

Please make sure to change all the places with the new name we choose.

packages/online-editor/src/home/HomePage.tsx
klbowers: Fixed! The only thing now is that in the EditorEnvelopeLocatorFactory class we're now returning a new EnvelopeMapping with type: extension. Is it worth refactoring EnvelopeMapping from type to extension?
`return new EnvelopeMapping({
          type: config.extension,`

tiagobento: If everywhere else was referring to this as `type`, then I think just the prop on `NewModelCard` that should be renamed to `type`, instead of renaming everything else to `extension`..

klbowers: Explored renaming, but we decided that the amount of refactoring needed wasn't worth it

packages/kie-sandbox-image/README.md
```diff
@@ -237,3 +239,58 @@ Here's an example of what it should look like:
...
+### Editors
+
+By default all three standard editors will be enabled (BPMN, DMN, PMML). To disable an editor simply delete/comment out the respective json.
+
+- **extension**: The extension of the file that you want to edit.
+- **filePathGlob**: The glob pattern of the file you want to edit.
+- **resourcesPathPrefix**: The path to the gwt-editor.
+- **path**: The path of the editor envelope.html.
+- **title**: The title of the editor that will be displayed on the home page.
+- **description**: Displays a short description of the editor under the title on the home page.
```
tiagobento: This is not aligned with the recent changes introduced to the `EditorConfig` type.

packages/kie-sandbox-image/README.md
```diff
@@ -54,14 +54,15 @@ This package contains the `Containerfile/Dockerfile` and scripts to build a cont
...
+   | :---------------------------------------------------------: | :---------------------------------------------------------------------------------------------------------------------------: | :-----------------------------------------------------------------: |
+   |             `KIE_SANDBOX_EXTENDED_SERVICES_URL`             |                                   The URL that points to the KIE Sandbox Extended Services.                                   | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
+   |              `KIE_SANDBOX_GIT_CORS_PROXY_URL`               |                         The URL that points to the Git CORS proxy for interacting with Git providers.                         | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
+   |        `KIE_SANDBOX_REQUIRE_CUSTOM_COMMIT_MESSAGES`         |                           Require users to type a custom commit message when creating a new commit.                           | See [ defaultEnvJson.ts ](../online-editor/build/defaultEnvJson.ts) |
```
tiagobento: Can you please rename this to `KIE_SANDBOX_REQUIRE_CUSTOM_COMMIT_MESSAGE` without the `S` at the end? Me and Tim found this error this week.. Also add to the PR description that this unrelated change is present.

packages/online-editor/src/env/EnvJson.ts
```diff
@@ -16,6 +16,7 @@
 
 import { AuthProvider } from "../authProviders/AuthProvidersApi";
 import { AcceleratorConfig } from "../accelerators/AcceleratorsApi";
+import { EditorConfig as EditorConfig } from "../envelopeLocator/EditorEnvelopeLocatorApi";
```
tiagobento:
```suggestion
import { EditorConfig } from "../envelopeLocator/EditorEnvelopeLocatorApi";
```

That's VS Code's fault, I know :P

packages/online-editor/src/envelopeLocator/hooks/EditorEnvelopeLocatorContext.tsx
```diff
@@ -18,16 +18,20 @@ import * as React from "react";
...
 
 export function EditorEnvelopeLocatorContextProvider(props: { children: React.ReactNode }) {
+  const editorConfig = useEditorConfig();
   const value = useMemo(
     () =>
       new EditorEnvelopeLocatorFactory().create({
         targetOrigin: window.location.origin,
+        editorConfig: editorConfig,
       }),
     []
```
tiagobento: Please always use the correct dependencies on hooks. Trust the linter! :)

You can also use a shorthand property here like 

```typescript
      new EditorEnvelopeLocatorFactory().create({
        targetOrigin: window.location.origin,
        editorConfig,
      })
```

packages/online-editor/src/envelopeLocator/hooks/EditorEnvelopeLocatorContext.tsx
```diff
@@ -18,16 +18,20 @@ import * as React from "react";
...
 import { EditorEnvelopeLocatorFactory } from "../EditorEnvelopeLocatorFactory";
+import { useEnv } from "../../env/hooks/EnvContext";
+import { EditorConfig } from "../EditorEnvelopeLocatorApi";
 
 export type SupportedFileExtensions = "bpmn" | "bpmn2" | "BPMN" | "BPMN2" | "dmn" | "DMN" | "pmml" | "PMML";
 
 export const EditorEnvelopeLocatorContext = React.createContext<EditorEnvelopeLocator>({} as any);
 
 export function EditorEnvelopeLocatorContextProvider(props: { children: React.ReactNode }) {
+  const editorConfig = useEditorConfig();
```
tiagobento: Hmmm.. I think `editorsConfig` would be better, as this returns an array. Same for the name of the hook.

packages/online-editor/src/workspace/worker/sharedWorker.ts
```diff
@@ -30,16 +31,24 @@ async function gitCorsProxyUrl(): Promise<string> {
...
+const workspaceServices = createWorkspaceServices({ gitCorsProxyUrl: gitCorsProxyUrl() });
 
 declare let onconnect: any;
 // eslint-disable-next-line prefer-const
 onconnect = async (e: MessageEvent) => {
-  console.log("Connected to Workspaces Shared Worker");
+  const editorConfig = await fetchEditorConfig();
+  const editorEnvelopeLocator = new EditorEnvelopeLocatorFactory().create({
+    targetOrigin: "",
+    editorConfig: editorConfig,
```
tiagobento: Same here.

packages/online-editor/src/workspace/worker/sharedWorker.ts
```diff
@@ -30,16 +31,24 @@ async function gitCorsProxyUrl(): Promise<string> {
   return env.KIE_SANDBOX_GIT_CORS_PROXY_URL;
 }
 
-const editorEnvelopeLocator = new EditorEnvelopeLocatorFactory().create({ targetOrigin: "" });
-const workspaceServices = createWorkspaceServices({ gitCorsProxyUrl: gitCorsProxyUrl() });
+async function fetchEditorConfig(): Promise<EditorConfig[]> {
```
tiagobento:
```suggestion
async function fetchEditorsConfig(): Promise<EditorConfig[]> {
```

tiagobento (changes requested):
Few minor issues commented inline. Ready to go when these are fixed and the PR is green. Thanks @kbowers-ibm !

tiagobento (approved):

PR #1579, opened 2023-04-25
NO-ISSUE: Add option to set step value in the `uniforms-patternfly` NumField component

tplevko:
Fixes same issue as in https://github.com/vazco/uniforms/issues/855

Also added test for set step prop

### DIFF
--- a/packages/uniforms-patternfly/src/NumField.tsx
+++ b/packages/uniforms-patternfly/src/NumField.tsx
@@ -49,7 +49,7 @@ function NumField(props: NumFieldProps) {
       onChange={onChange}
       placeholder={props.placeholder}
       ref={props.inputRef}
-      step={props.decimal ? 0.01 : 1}
+      step={props.step ?? (props.decimal ? 0.01 : 1)}
       type="number"
       value={props.value ?? ""}
       validated={props.error ? "error" : "default"}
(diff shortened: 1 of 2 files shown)


packages/uniforms-patternfly/src/NumField.tsx
```diff
@@ -49,7 +49,7 @@ function NumField(props: NumFieldProps) {
       onChange={onChange}
       placeholder={props.placeholder}
       ref={props.inputRef}
-      step={props.decimal ? 0.01 : 1}
+      step={props.step || (props.decimal ? 0.01 : 1)}
```
ljmotta: Please use `??` instead of `||`

```suggestion
      step={props.step ?? (props.decimal ? 0.01 : 1)}
```

packages/uniforms-patternfly/src/__tests__/NumField.test.tsx
```diff
@@ -88,6 +88,13 @@ test("<NumField> - renders an input with correct step (integer)", () => {
   expect(screen.getByTestId("num-field").getAttribute("step")).toBe("1");
 });
 
+test("<NumField> - renders an InputNumber with correct step (set)", () => {
```
ljmotta: Please use the same pattern of other tests
```suggestion
test("<NumField> - renders an input with correct step (set)", () => {
```

ljmotta (changes requested):
@tplevko Hi there! Thanks a lot for your contribution. I've made some change requests inline.

packages/uniforms-patternfly/src/NumField.tsx
tplevko: thanks - updated

packages/uniforms-patternfly/src/__tests__/NumField.test.tsx
tplevko: sure - updated

ljmotta (approved):
@tplevko Approved! Thanks again.

PR #1582, opened 2023-04-26
NO-ISSUE: Fix `supportedDoubleExtensions` regex

caponetto:
YAML files (not sw.yaml) are being treated as a doubled extension file due to the lack of parenthesis in the regex.


tiagobento (approved):

caponetto:
Error 500 when setting up default reviewers 🙈

PR #1597, opened 2023-05-02
KOGITO-9053: [Diagram Viewer] stateDataFilter doesn't show details in diagram properly

treblereel:
JIRA: https://issues.redhat.com/browse/KOGITO-9053

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/test/resources/org/kie/workbench/common/stunner/sw/client/selenium/yaml/FoodOrderWorkflow.sw.yaml
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/test/resources/org/kie/workbench/common/stunner/sw/client/selenium/yaml/FoodOrderWorkflow.sw.yaml
@@ -1,0 +1,40 @@
+id: foodorderworkflow
+version: 1.0.0
+specVersion: 0.8
+name: Food Order Workflow
+start: Place Order
+states:
+  - name: Place Order
+    type: operation
+    actions:
+      - subFlowRef: placeorderworkflow
+    transition: Wait for ETA Deadline
+  - name: Wait for ETA Deadline
+    type: event
+    onEvents:
+      - eventDataFilter:
+          data: "${ .results.status }"
+          toStateData: "${ .status }"
+        eventRefs:
+          - ETA Deadline Event
+    transition: Deliver Order
+  - name: Deliver Order
+    type: operation
+    actions:
+      - subFlowRef: deliverorderworkflow
+    transition: Charge For Order
+  - name: Charge For Order
+    type: operation
+    actions:
+      - functionRef:
+          refName: Charge For Order Function
+          arguments:
+            order: "${ .order.id }"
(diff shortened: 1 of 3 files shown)


romartin (commented):
Does it makes sense to you @treblereel to add some unit test for this regexp logic only?

LuboTerifaj (changes requested):
Hello @treblereel ,

the PR is missing tests. Could  you please add some?

Thanks!

treblereel:
@romartin I planned to add tests to all yaml  examples I saw ( well, yaml from test.html, vscode tests and Lubos examples) as separate PR. But if we want to have a test for this regex fix, yeap, why not, it's easy.

romartin (approved):
great, thanks @treblereel

packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-kogito-app/src/test/resources/org/kie/workbench/common/stunner/sw/client/selenium/yaml/FoodOrderWorkflow.sw.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - functionRef:
+          refName: Charge For Order Function
+          arguments:
+            order: "${ .order.id }"
+        actionDataFilter:
+          results: "${ .outcome.status }"
+          toStateData: "${ .status }"
+    end: true
+    stateDataFilter:
+      output: '.status} | .orderstatus += ["Order Completed"] }'
```
LuboTerifaj, line 38: It looks like the output value is incomplete. Could you please update it?

LuboTerifaj (changes requested):
Hello @treblereel ,

I have an inline comment regarding test.

There is also an issue with opening [finalize-college-application.sw.yaml](https://github.com/LuboTerifaj/specification/blob/examples-in-separate-files/examples/resources/finalize-college-application.sw.yaml) workflow. Loading takes very long time in comparison with main. Could you please have a look at it?

Thanks!

treblereel:
@LuboTerifaj let's create separate a Jira task for the performance issue. And which browser is slow?

LuboTerifaj:
Hello @treblereel,

 I tested it using VSCode extension. I think we should not introduce such regression. Anyway, I reported the jira in case we decide to fix it separately:
- [KOGITO-9222](https://issues.redhat.com/browse/KOGITO-9222) _SWF - Workflow yaml file takes too much time to load_.

cc @romartin 

Could you please update the suggested change from my previous comment?

Thanks!

romartin:
Thanks @LuboTerifaj  for the feedback.

We have some important release in a few weeks, so I think we should address also https://issues.redhat.com/browse/KOGITO-9222 as part of this work, otherwise, if merging it implies the issues, we should wait for the release.

So just let us know @treblereel if it is related, also if it is feasible to fix, before second week of June.

Thanks!

LuboTerifaj:
The issue will be fixed in https://github.com/kiegroup/kie-tools/pull/1680. @treblereel , please close this pull-request. Thanks!

PR #1603, opened 2023-05-03
kie-issues#246: allow rootless container build

### LINKED ISSUE #246: Allow rootless container build of dmn-dev-deployment-base-image
Currently the existing Containerfile fails to build when running without root.

It explicitly sets chown to root gid and executes chmod among the others.

jstastny-cz:
Closes kiegroup/kie-issues#246

Adjusting Containerfile so that when being run rootless, it would not fail due to permission issues.

@thiagoelg @caponetto please check if the changes make sense to you.

### DIFF
--- a/packages/dmn-dev-deployment-base-image/Containerfile
+++ b/packages/dmn-dev-deployment-base-image/Containerfile
@@ -18,15 +18,15 @@ ARG QUARKUS_PLATFORM_VERSION
 ARG KOGITO_RUNTIME_VERSION
 ARG ROOT_PATH
 
+USER 1000
+
 RUN mkdir /tmp/kogito/
 
-COPY --chown=185:root dist-dev/dmn-dev-deployment-quarkus-app /tmp/kogito/dmn-dev-deployment-quarkus-app/
-COPY --chown=185:root dist-dev/dmn-dev-deployment-form-webapp/ /tmp/kogito/dmn-dev-deployment-quarkus-app/src/main/resources/META-INF/resources/
+COPY --chown=1000:1000 --chmod=775 dist-dev/dmn-dev-deployment-quarkus-app /tmp/kogito/dmn-dev-deployment-quarkus-app/
+COPY --chown=1000:1000 --chmod=775 dist-dev/dmn-dev-deployment-form-webapp/ /tmp/kogito/dmn-dev-deployment-quarkus-app/src/main/resources/META-INF/resources/
 
 WORKDIR /tmp/kogito/dmn-dev-deployment-quarkus-app/
 
-RUN ./mvnw clean package -B -ntp -Dmaven.test.skip -DQUARKUS_PLATFORM_VERSION=${QUARKUS_PLATFORM_VERSION} -DKOGITO_RUNTIME_VERSION=${KOGITO_RUNTIME_VERSION} -Dquarkus.http.root-path=${ROOT_PATH} \
-    && mv ~/.m2 /tmp/kogito \
-    && chmod -R 775 /tmp/kogito
-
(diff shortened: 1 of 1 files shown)


thiagoelg (approved):

caponetto:
Hi @jstastny-cz,
I see no problems with the changes if you managed to get this working on OpenShift.

jstastny-cz:
@caponetto I'll double-check with the latest changes to be sure.

jstastny-cz:
@caponetto I needed to change a bit, what helped was only the --chmod=775 specifier to COPY command. I have to admit I am not entirely clear on why the lack of it would endup in lack of permissions in the final chmod -R command.

I've also removed the copying of maven repository directory and instead using the same location for both builds using `-Dmaven.repo.local`.

I've built locally with both docker and podman.

One catch is that with older docker versions, one needs to enable docker buildkit. Which can be done using `export DOCKER_BUILDKIT=1`. As for podman support for `--chmod` I have not been able to find any information about it, but as I said it worked in my case.

packages/dmn-dev-deployment-base-image/Containerfile
```diff
@@ -18,15 +18,15 @@ ARG QUARKUS_PLATFORM_VERSION
...
+COPY --chown=1000:1000 --chmod=775 dist-dev/dmn-dev-deployment-form-webapp/ /tmp/kogito/dmn-dev-deployment-quarkus-app/src/main/resources/META-INF/resources/
 
 WORKDIR /tmp/kogito/dmn-dev-deployment-quarkus-app/
 
-RUN ./mvnw clean package -B -ntp -Dmaven.test.skip -DQUARKUS_PLATFORM_VERSION=${QUARKUS_PLATFORM_VERSION} -DKOGITO_RUNTIME_VERSION=${KOGITO_RUNTIME_VERSION} -Dquarkus.http.root-path=${ROOT_PATH} \
-    && mv ~/.m2 /tmp/kogito \
-    && chmod -R 775 /tmp/kogito
-
+RUN ./mvnw clean package -B -ntp -Dmaven.test.skip -Dmaven.repo.local=/tmp/kogito/.m2/repository -DQUARKUS_PLATFORM_VERSION=${QUARKUS_PLATFORM_VERSION} -DKOGITO_RUNTIME_VERSION=${KOGITO_RUNTIME_VERSION} -Dquarkus.http.root-path=${ROOT_PATH}
+RUN chmod -R 775 /tmp/kogito/
```
caponetto: I would only merge these 2 RUN into a single RUN.

caponetto (commented):
Interesting @jstastny-cz!
I remember that I did those things because I faced permission issues only on OpenShift Sandbox, although it was working locally. So, I think it is important to validate on OpenShift too, especially the OpenShift Sandbox.

packages/dmn-dev-deployment-base-image/Containerfile
jstastny-cz: I thought this would be more readable, is there a significant difference between this and multi-line command to a single RUN commands?

jstastny-cz:
@caponetto I've run this on our OpenShift instance on IBM Cloud. And I believe I've seen the problems with permissions you mention, the current version works fine in that respect.

packages/dmn-dev-deployment-base-image/Containerfile
caponetto: Yes, each RUN will create a new docker layer, so better merge them together in a single one, when possible.

caponetto: You can break the lines to improve readability.

jstastny-cz: Did as instructed, please check.

caponetto (approved):

thiagoelg (approved):

jstastny-cz:
@kiegroup/gatekeepers could you please merge when possible?

PR #1612, opened 2023-05-05
kie-issues#228: On the DMN runner table, the SelectField will reset its value after "Add above" or "Duplicate" operations

### LINKED ISSUE #228: On the DMN runner table, the SelectField will reset its value after "Add above" or "Duplicate" operations
Performing a "Add above" or "duplicate" table operation will reset the SelectField value to its previous one. The value will not be reset if the user closes and opens the DMN Runner table after a change. e.g. Change the SelectField value, close the table, and open it again.

### Steps to reproduce
 - Open the DMN sample;
 - Open the DMN Runner table;
 - Change a SelectField value using your mouse (e.g. Martial Status);
 - With the field still selected, make a table operation ("add above" or "duplicate");
 - The SelectField will reset to its previous value;

### Example
The example below show the cases where the SelectField value will reset after a "Add above" or "Duplicate" operation.

https://user-images.githubusercontent.com/24302289/234093176-9f6bc861-47a3-419b-b936-671bae7fde27.mp4

### Code
The bug cause will be in one (or two) of the two places:
 1. `package/online-editor/src/dmnRunner/DmnRunnerContextProvider.tsx`
 2. `packages/unitables/src/bee/UnitablesBeeTables.tsx`

On 1. the cause can be in the table operation methods `onRowAdded` or `onRowDuplicated`. 
On 2. the cause could be part of the `UnitablesBeeTableCell` component.

```[tasklist]
### Acceptance criteria
- [x] "Add above" should not change the SelectField value;
- [x] "Duplicate" should not change the SelectField value;
```

yesamer:
Closes https://github.com/kiegroup/kie-issues/issues/228

The described issue reproduces in two scenarios:
- Updating a boolean value (Checkbox)
- Updating an Enum value (Button + Dropdown HTML elements)

The root cause for both cases is basically the same: the `onBlur` event (which is responsible to submit the change) it is not triggered in some cases of the above scenarios, and in detail:
- **Case 1:** Updating a boolean value (Checkbox):  Sometimes when clicking on the checkbox the focus on that input element is not triggered. To solve that, I added the code in `onClick` method (503:511) which forces the focus on the input element. Please note, the same approach is used on `onKeyDown` management when pressing ENTER (line 401)
- **Case 2:** Updating an Enum value (Button + Dropdown HTML elements):  In this case, we need to submit the change when closing the dropdown. To solve that, I added the code in `onClick` method (489:501) submit the change if the dropdown is open. Please note, the same approach is used on `onKeyDown` management when pressing ENTER (line 389)


packages/unitables/src/bee/UnitablesBeeTable.tsx
```diff
@@ -465,28 +465,53 @@ function UnitablesBeeTableCell({
...
+        setEditingCell(false);
       }
       if (
         e.target.tagName.toLowerCase() === "button" ||
         (e.relatedTarget as HTMLElement)?.tagName.toLowerCase() === "button"
       ) {
         // if the select field is open and it blurs to another cell, close it;
         const selectOptions = document.getElementsByName(fieldName)?.[0]?.getElementsByTagName("button");
-        if ((selectOptions?.length ?? 0) > 0 && (e.relatedTarget as HTMLElement).tagName.toLowerCase() === "td") {
+        if ((selectOptions?.length ?? 0) > 0 && (e.relatedTarget as HTMLElement)?.tagName?.toLowerCase() === "td") {
```
yesamer, line 477: That thrown a NPE when clicking OUTSIDE the table area

yesamer:
Hey @ljmotta, I tried to solve this issue, not sure if that is the correct strategy or if I missing something, can you please check? 
Please note, I basically replicated the same logic used to manage the ENTER key, during my tests it seems working well.
I just notice this Warning:

Screenshot 2023-05-08 at 12 53 41

that is thrown when calling this `setIsSelectFieldOpen`. As I said, ENTER key logic uses the same approach, therefore is affected by the same warning (try navigating with the keyboard to an Enumerated field and press enter=

WDYT? Thank you :)

ljmotta:
@yesamer Thanks for your PR. I test it locally and it looks great.

Regarding the warning, I investigated it, and indeed, the problem is the set state inside the `setIsSelectFieldOpen` callback in the "enter" logic.

```tsx
setIsSelectFieldOpen((prev) => {
  if (prev === true) {
    submitRow();
    setEditingCell(false);
  } else {
    setEditingCell(true);
  }
  return !prev;
});
```

The value of `isEditing` is changed with the `setEditingCell` and this causes a re-calculation of the `onKeyDown` as the `isEditing` is a dependency of the callback. Now, you PR is re-using the same logic, and `onClick` callback has the `isEditing` as dependency, so it will trigger the same warning.

I think this will do the trick:

```tsx
setIsSelectFieldOpen((prev) => {
  if (prev === true) {
    submitRow();
  }
  return !prev;
});
setEditingCell(!isEditing);
```

Removing the `setEditingCell` from inside the `setIsSelectFieldOpen` will make React batch the `setState`, so all states will update at the same time.

yesamer:
@ljmotta Thank you!!! Changes applied! It works nicely!

ljmotta (approved):

PR #1614, opened 2023-05-05
KOGITO-8988: SWF Viewer - add icons for ForEch state

hasys:

I added two icons: for mode I reused actionMode since it is the same enum. And for other 4 parameters: inputCollection, outputColloection, itemIteration, batchSize I added single icon since all of them are related to each other.

I putted the icon to the right since it is data or parameters icon so putted it near the filter icon.

The icon is always shown, since inputCollection is required parameter so it should be always defined.

image

Thank you!


packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-client/src/main/java/org/kie/workbench/common/stunner/sw/client/shapes/HasCollections.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.kie.workbench.common.stunner.sw.resources.i18n.SWConstants.FOR_EACH_BATCH_SIZE;
+import static org.kie.workbench.common.stunner.sw.resources.i18n.SWConstants.INPUT_COLLECTION;
+import static org.kie.workbench.common.stunner.sw.resources.i18n.SWConstants.ITERATION_PARAMETER;
+import static org.kie.workbench.common.stunner.sw.resources.i18n.SWConstants.OUTPUT_COLLECTION;
+
+public interface HasCollections extends HasDataFilter {
+
+    default String getCollections(ForEachState state) {
+        StringBuilder result = new StringBuilder();
+        result.append(getTranslation(INPUT_COLLECTION) + ": " + truncate(state.inputCollection, 30) + "\r\n");
```
domhanak: just a nitpick, but feels like it could improve readability -> rename 30 to a constant `MAX_VISIBILE_CHARS = 30` 

2nd, just an observation maybe we could add dots after long ones indicating visually it is longer like this -> `${ .messages.parents.conten...` dots when the truncated string is >30 chars.

domhanak (approved):
@hasys LGTM, added comment but not mandatory, the JIRA's goals are working as expected imho.

LuboTerifaj (commented):
Hello @hasys ,

I found the following issues:
- `iterationParam` value is not truncated in the diagram.
- If you input a large integer to `batchSize` (e.g.: 2147483648),  2147483647 is shown, which is maximum positive value for a 32-bit signed binary integer.
  - Since the real value is different. It should be handled in a different way. For example by showing the message the unsupported value is used.
	
- The icons may cover the state's name.
Screenshot from 2023-05-10 13-55-39
I reported the issue here: https://issues.redhat.com/browse/KOGITO-9135

Thanks!

hasys:
Hello @domhanak, @LuboTerifaj,

great points, thank you!

I rebased to master branch and also fixed:
* Created default constant value for truncate method
* On the end of the string there is ... as expected
* `iterationParam` value now also truncated
* Default number parser is changed to correct Long type from Integer. So it is aligned with JavaScript number type now.

Thank you guys!

romartin (approved):

LuboTerifaj (commented):
Hello @hasys ,

unfortunately the issues I mentioned are not fixed:

- iterationParam value is still not truncated in the diagram
- If you input a large number to batchSize a wrong value is shown

Thanks!

LuboTerifaj:
Hello @hasys , regarding the iterationParam issue, it seems it is more generic one. I have reported it here:

- [KOGITO-9155](https://issues.redhat.com/browse/KOGITO-9155) _SWF Viewer - Tooltip is shown outside the viewable area_

The only one left issue is regarding showing the wrong value.

Thanks!

LuboTerifaj (approved):
Hello @hasys , @romartin ,

I have reported also the remaining issue since it may affect also other parts of editor.

- [KOGITO-9156](https://issues.redhat.com/browse/KOGITO-9156) _SWF Viewer - The diagram shows only maximum of 64-bit integer_

I didn't find any more issues.

Thanks!

PR #1629, opened 2023-05-11
KOGITO-9100: [SWF Editor] Add AsyncAPI autocomplete support for events array

Sara4994:
Hi,

Please review this PR, which include auto complete feature on events array of swf specifications.

This works only with AsyncAPI specs for now.

Related JIRAs:
https://issues.redhat.com/browse/SRVLOGIC-148
https://issues.redhat.com/browse/KOGITO-8915 (EPIC)

### DIFF
--- a/packages/serverless-workflow-service-catalog/src/api/types.ts
+++ b/packages/serverless-workflow-service-catalog/src/api/types.ts
@@ -65,10 +65,7 @@ export enum SwfServiceCatalogFunctionType {
 }
 
 export enum SwfServiceCatalogEventType {
-  rest = "rest",
-  graphql = "graphql",
   asyncapi = "asyncapi",
-  custom = "custom",
 }
 
 export enum SupportArtifactTypes {
--- a/packages/serverless-workflow-service-catalog/src/api/types.ts
+++ b/packages/serverless-workflow-service-catalog/src/api/types.ts
@@ -40,8 +40,8 @@ export interface SwfServiceCatalogService {
   name: string;
   type: SwfServiceCatalogServiceType;
   source: SwfServiceCatalogServiceSource;
-
   functions: SwfServiceCatalogFunction[];
+  events?: SwfServiceCatalogEvent[];
   rawContent: string;
 }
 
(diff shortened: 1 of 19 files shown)


Sara4994:
https://github.com/kiegroup/kie-tools/assets/40579476/756ed162-3f75-455f-a583-ba36c0bd8360

paulovmr (changes requested):
Hi @Sara4994 , looking at the video, I see that a few things we discussed still have to be changed:
* The source and type of an event are totally different fields from the source and type from functions. In case of events, they are both string and for now can be empty by default when the autocompletion is used.
* The kind, metadata and reference fields are not being written in the event when the autocompletion is used.

Sara4994:
short clip after making minor enhancements to the autocomplete

https://github.com/kiegroup/kie-tools/assets/40579476/5f4038a9-8c0c-4fab-8a76-713c95fee420

packages/serverless-workflow-service-catalog/src/api/types.ts
```diff
@@ -64,6 +64,13 @@ export enum SwfServiceCatalogFunctionType {
   custom = "custom",
 }
 
+export enum SwfServiceCatalogEventType {
```
paulovmr, line 67: Currently, `asyncapi` should be the only type for events

paulovmr (changes requested):
Hi @Sara4994 , looks great, just one observation on the code, and one request: could you please add the "Refresh Service Registries..." action to the events array as well, similarly to the functions array? Thanks!

tomasdavidorg (approved):
Looks good to me. Thanks @Sara4994 🙂

packages/serverless-workflow-service-catalog/src/api/types.ts
```diff
@@ -95,9 +107,29 @@ export type SwfServiceCatalogFunctionSource =
...
   name: string;
   arguments: Record<string, SwfServiceCatalogFunctionArgumentType>;
   type: SwfServiceCatalogFunctionType;
 }
+
+export interface SwfServiceCatalogEvent {
+  source: SwfServiceCatalogEventSource;
+  name: string;
+  kind: SwfServiceCatalogEventKind;
+  type: SwfServiceCatalogEventType.asyncapi;
```
paulovmr: I think it is better if the enum just have the asyncapi item instead of having all functions types and the asyncapi being used here.

paulovmr (changes requested):

paulovmr (approved):

PR #1631, opened 2023-05-12
kie-issues#255: A Not-collect Hit Policy should clear aggregation type 

### LINKED ISSUE #255: A Not-collect Hit Policy should clear aggregation type
Step to reproduce:

- Open any Decision Table with HIT POLICY != "COLLECT"
- Change the HIT POLICY to Collect, and select ANY aggregator.
- Change again the HIT POLICY to any other type (eg. "UNIQUE")
- Save
- Validate the DMN file. The following error is thrown:

image-2023-05-08-14-27-20-306

The root cause is that the Aggregation field should be removed from the XML for any HIT POLICY different than "COLLECT"

https://issues.redhat.com/browse/KOGITO-9116

yesamer:
Closes https://github.com/kiegroup/kie-issues/issues/255

Just setting aggregation with NULL in case of not "COLLECT" Hit Policy type solves the issue :)


jomarko (approved):
Code looks reasonable! Thank you for spotting @yesamer

PR #1633, opened 2023-05-12
kie-issues#256: kie-sandbox-image crashes on MacOS on the entrypoint step

### LINKED ISSUE #256: kie-sandbox-image crashes on MacOS on the entrypoint step
When running the currently published `kie-sandbox-image` on MacOS the image crashes on the ENTRYPOINT step, without providing any logs.

After some investigation, this seems to be related to the updated ubi9 base image being used. Reverting to the previous ubi8 base image should fix the issue.

thiagoelg:
Closes https://github.com/kiegroup/kie-issues/issues/256

The currently published kie-sandbox-image based on ubi9 doesn't seem to run well on different platforms (on MacOS it crashes after setting the env vars). Reverting to the previous ubi8 image and keeping the linux platform fixes this issue.

### DIFF
--- a/packages/kie-sandbox-image/package.json
+++ b/packages/kie-sandbox-image/package.json
@@ -26,9 +26,11 @@
     "image:docker:build": "run-script-if --bool $([ $(command -v docker) ] && echo true || echo false) --then \"docker build $(echo $(build-env kieSandbox.image.buildTags) | xargs printf -- \"-t $(build-env kieSandbox.image.registry)/$(build-env kieSandbox.image.account)/$(build-env kieSandbox.image.name):%s\n\" | xargs echo) . -f Containerfile\" --else \"echo Docker not found, skipping image build.\"",
     "image:podman:build": "run-script-if --bool $([ $(command -v podman) ] && echo true || echo false) --then \"podman build $(echo $(build-env kieSandbox.image.buildTags) | xargs printf -- \"-t $(build-env kieSandbox.image.registry)/$(build-env kieSandbox.image.account)/$(build-env kieSandbox.image.name):%s\n\" | xargs echo) -f Containerfile\" --else \"echo Podman not found, skipping image build.\""
   },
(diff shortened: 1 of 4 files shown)


packages/kie-sandbox-image/package.json
```diff
@@ -26,9 +26,11 @@
     "image:docker:build": "run-script-if --bool $([ $(command -v docker) ] && echo true || echo false) --then \"docker build $(echo $(build-env kieSandbox.image.buildTags) | xargs printf -- \"-t $(build-env kieSandbox.image.registry)/$(build-env kieSandbox.image.account)/$(build-env kieSandbox.image.name):%s\n\" | xargs echo) . -f Containerfile\" --else \"echo Docker not found, skipping image build.\"",
     "image:podman:build": "run-script-if --bool $([ $(command -v podman) ] && echo true || echo false) --then \"podman build $(echo $(build-env kieSandbox.image.buildTags) | xargs printf -- \"-t $(build-env kieSandbox.image.registry)/$(build-env kieSandbox.image.account)/$(build-env kieSandbox.image.name):%s\n\" | xargs echo) -f Containerfile\" --else \"echo Podman not found, skipping image build.\""
   },
-  "devDependencies": {
+  "dependencies": {
```
yesamer, line 29: @thiagoelg Can you please explain this change? I don't see this in the previous versions, so I guess this is not part of the "revert"

yesamer (commented):
@thiagoelg Just a question

packages/kie-sandbox-image/package.json
thiagoelg: This won't bring any practical change, mainly conceptual, since the `online-editor` and `image-to-env-json` packages are direct parts of this image I thought they should be dependencies instead of devDependencies.

yesamer (approved):
Thank you @thiagoelg

PR #1634, opened 2023-05-15
kie-issues#222: When editing a BeeTable cell with selected text, the selection is lost when mouse hovers in or out of it

### LINKED ISSUE #222: When editing a BeeTable cell with selected text, the selection is lost when mouse hovers in or out of it
I tracked it down to the presence or absence of the `<Resizer>` component depending on the hoverInfo of the `BeeTableTd` component. Commenting the code on this snippet solves the issue, but then the resizing capabilities on tds is lost.

https://github.com/tiagobento/kie-tools/blob/f0d12aba742e79035552afce848e7cf169c80e67/packages/boxed-expression-component/src/table/BeeTable/BeeTableTd.tsx#L208-L218

Steps to reproduce:
1. Launch the Boxed Expression Editor showcase webapp at `packages/boxed-expression-component` with `pnpm start --env live`
2. Create a new Decision Table
3. Start editing a cell  by double-clicking it and enter some text to it
4. Make a selection either with the mouse or the keyboard
5. Move the mouse out of it and see the selection resetting

https://user-images.githubusercontent.com/1584568/233214837-25945d4f-bfa0-4b88-ab8f-36b7fdc30a92.mov

```[tasklist]
### Acceptance criteria
- [ ] The selection of the text inside a cell is kept no matter where the mouse moves to
```

jomarko:
Closes kiegroup/kie-issues/issues/222

### Previously
We rendered `Resizer` for any hovered cell of the boxed expression table component (currently used in DMN editor and DMN Runner Tabular view). It causes issues because you can have at the same time One cell being edited and another cell just hovered. In other words during editing one cell, resizer were dynamically added and removed form the html DOM. Such dynamic removal of DOM elements causes window selection to be cleared, as described in https://github.com/kiegroup/kie-issues/issues/222#issuecomment-1543784277

### Currently
We decided to eliminate shown resizers. From now on it is not enough to hover cell with mouse to display the `Resizer`, but we also check if there is some selected text. If yes, we stop to display `Resizer`s until there is some selected text.

### DIFF
--- a/packages/boxed-expression-component/src/table/BeeTable/BeeTableTd.tsx
+++ b/packages/boxed-expression-component/src/table/BeeTable/BeeTableTd.tsx
@@ -134,9 +134,6 @@ export function BeeTableTd<R extends object>({
     }
 
     function onLeave() {
-      if (hasTextSelectedInBoxedExpressionEditor()) {
-        return;
-      }
       setHoverInfo((prev) => ({ isHovered: false }));
     }
     td?.addEventListener("mouseenter", onEnter);
--- a/packages/boxed-expression-component/src/table/BeeTable/BeeTableTd.tsx
+++ b/packages/boxed-expression-component/src/table/BeeTable/BeeTableTd.tsx
@@ -66,6 +66,7 @@ export function BeeTableTd<R extends object>({
 }: BeeTableTdProps<R>) {
   const [isResizing, setResizing] = useState(false);
   const [hoverInfo, setHoverInfo] = useState<HoverInfo>({ isHovered: false });
+  const [isFocused, setIsFocused] = useState(false);
 
   const tdRef = useRef<HTMLTableCellElement>(null);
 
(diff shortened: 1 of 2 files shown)


danielzhe:
Great work, @jomarko ! This is a very good solution for the issue. 
I just found a minor issue:
1. Select the text
2. Click in another cell
3. Two resizes are shown


https://github.com/kiegroup/kie-tools/assets/7305741/3e2688bd-bbf1-4627-9c7f-69bea564f26d

thiagoelg:
@danielzhe Good catch!

I think this can be fixed by calling `setHoveredInfo({ isHovered: false })` once the `isActive` changes from `true` to `false` since this would be automatically done before, but now the `hasTextSelectedInBoxedExpression` function blocks it from happening.

What do you think @jomarko?

jomarko:
@danielzhe @thiagoelg thank you for good points raised. I addressed them in a separate commit b262ca153a

danielzhe (approved):

thiagoelg (approved):

yesamer (approved):
@jomarko Thank you!

PR #1635, opened 2023-05-15
KOGITO-8153: Create an extensible Samples page (now Catalog)

fantonangeli:
https://issues.redhat.com/browse/KOGITO-8153

**Description**

Users should be able to see samples when accessing the Sample Catalog page

Please note that, after [KOGITO-8875](https://issues.redhat.com/browse/KOGITO-8875), samples are imported from https://github.com/kiegroup/kie-samples

    Samples should be paginated with a fixed number per page (maybe 10?)
    The search and filter must be kept
    The layout should be a grid of cards
    The card layout can be completely changed from what we have today
    Each card should include the information we already have (title, description, SVG and category)
    SVGs should fit their container (must never be cropped)
    The sample is opened when users click on the card (no need to have a button for that)
    Currently, all samples are cached in the first load. With pagination, we can change the strategy to cache only the SVGs that were loaded. (For example: the sample SVG that lives on the last page doesn't need to be loaded if the user never goes to the last page)

Nice to have but not required: Some way to see a bigger version of the SVG. One option is to have a magnifying glass button.

**Preview**
https://fantonangeli.github.io/kie-tools/kogito-8015

[KOGITO-8153.webm](https://github.com/kiegroup/kie-tools/assets/17780574/89e1376c-4ec6-4651-8ee7-798437e3c7e4)


packages/serverless-logic-web-tools/src/home/sample/Showcase.tsx
```diff
@@ -14,28 +14,30 @@
...
   DropdownToggle,
 } from "@patternfly/react-core/dist/js/components/Dropdown";
+import { EmptyState, EmptyStateIcon } from "@patternfly/react-core/dist/js/components/EmptyState";
+import { Page, PageSection } from "@patternfly/react-core/dist/js/components/Page";
+import { SearchInput } from "@patternfly/react-core/dist/js/components/SearchInput";
+import { Text, TextContent, TextVariants } from "@patternfly/react-core/dist/js/components/Text";
+import { Title } from "@patternfly/react-core/dist/js/components/Title";
+import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core/dist/js/components/Toolbar";
+import { Gallery } from "@patternfly/react-core/dist/js/layouts/Gallery";
+import "@patternfly/react-core/dist/styles/base.css";
```
caponetto: Do we need this import?

packages/serverless-logic-web-tools/src/home/sample/Showcase.tsx
```diff
@@ -50,18 +52,46 @@ const LABEL_MAP: Record<SampleCategory, JSX.Element> = {
 };
 
 const ALL_CATEGORIES_LABEL = "All categories";
-
+const perPage = 9;
```
caponetto:
```suggestion
const CARDS_PER_PAGE = 9;
```

packages/serverless-logic-web-tools/src/home/sample/Showcase.tsx
```diff
@@ -50,18 +52,46 @@ const LABEL_MAP: Record<SampleCategory, JSX.Element> = {
...
 const CATEGORY_ARRAY = Object.keys(SAMPLE_PRIORITY) as SampleCategory[];
 
+export const perPageOptions: PerPageOptions[] = [
```
caponetto:
```suggestion
export const SAMPLE_CARDS_PER_PAGE_OPTIONS: PerPageOptions[] = [
```

packages/serverless-logic-web-tools/src/home/sample/Showcase.tsx
```diff
@@ -50,18 +52,46 @@ const LABEL_MAP: Record<SampleCategory, JSX.Element> = {
...
+  {
+    title: "9",
+    value: 9,
```
caponetto: Can we reuse `CARDS_PER_PAGE` here instead of hardcoding the values?

packages/serverless-logic-web-tools/src/home/sample/sampleApi.ts
```diff
@@ -143,56 +167,56 @@ async function listSampleDefinitionFiles(args: {
...
+ * @returns The content of the SVG cover file for the sample, or undefined if it could not be fetched.
+ */
+export async function fetchSampleCover(args: { octokit: Octokit; sample: Sample }): Promise<string | undefined> {
+  const { sample } = args;
+
+  const svgContent = await fetchFileContent({
+    octokit: args.octokit,
+    fileInfo: {
+      ...KIE_SAMPLES_REPO,
+      path: join("samples", sample.sampleId, sample.definition.cover),
```
caponetto: Please create a const for `"samples"`.

packages/serverless-logic-web-tools/src/home/sample/Showcase.tsx
```diff
@@ -166,46 +189,71 @@ export function Showcase() {
...
+            </ToolbarItem>
+          </ToolbarContent>
+        </Toolbar>
+      </PageSection>
+
+      <PageSection isFilled>
+        {sampleLoadingError && <SamplesLoadError errors={[sampleLoadingError]} />}
+        {!sampleLoadingError && (
+          <>
+            {loading && <SampleCardSkeleton numberOfCards={4} />}
```
caponetto: I think we need to update the skeleton to match the new UX.

packages/serverless-logic-web-tools/src/home/sample/SampleCard.tsx
```diff
@@ -46,72 +46,114 @@ const tagMap: Record<SampleCategory, { label: string; icon: React.ComponentClass
...
+                      WebkitBoxOrient: "vertical",
+                      WebkitLineClamp: 5,
+                      overflow: "hidden",
+                      textOverflow: "ellipsis",
+                      whiteSpace: "pre-wrap",
+                    }}
+                  >
+                    {props.sample.definition.description}
+                  </Text>
+                  <Button
```
caponetto: I wonder if it would be better if we place this button on the bottom right of the SVG area.

caponetto (commented):
Great work, @fantonangeli!

Besides a few inline comments, I only have one additional point.

I can rapidly see this (highlighted in red) while doing a page refresh. Can we avoid this somehow?

image

caponetto (approved):

PR #1636, opened 2023-05-15
NO-ISSUE: Stabilize randomly failing GraphUtilsTest#testComputeGraphHashCode

jomarko:
The mentioned test `GraphUtilsTest#testComputeGraphHashCode` started to fail in CI while locally is working fine. 

During investigation it was found we use `junit` method for comparing `double`s but real compared values are `int`s

The CI error message is
```
Error: s/stunner-editors build:prod: [ERROR] GraphUtilsTest.testComputeGraphHashCode:248 Values should be different. Actual: -1.8597954E7
```

It can be seen for example here https://github.com/kiegroup/kie-tools/actions/runs/4978117965/jobs/8910744518?pr=1634


yesamer:
Closing,  fixed here https://github.com/kiegroup/kie-tools/pull/1639

PR #1637, opened 2023-05-16
KOGITO-9151: Remove old Dashbuilder dependencies leftovers

jesuino:
Removing old libraries that were not used and also adding a new component for markdown


packages/dashbuilder/deps.txt
```diff
@@ -0,0 +1,2745 @@
+[INFO] Scanning for projects...
```
caponetto: Is this file a leftover?

caponetto (approved):

PR #1638, opened 2023-05-16
kie-issues#258: Use data type current value when opening DMN constraint modal

### LINKED ISSUE #258: Unable to create demanded DMN Constraint for fresh data type entry
When creating a new data type (Enumeration and Range are greyed out). User needs to save the type first and edit again to convert it to enum or range. It would be nice to be 
able to create an enum or range right when creating the type. Mentioning it here so it could be considered as part of this ticket. 

image (4)

_Originally posted by @baldimir in https://github.com/kiegroup/kie-issues/issues/209#issuecomment-1547238983_

jomarko:
Closes https://github.com/kiegroup/kie-issues/issues/258

Currently when opening a DMN constraint modal, not the latest DMN entry data type is taken into account. 

After this fix, DMN constraint modal will allow to use `Enumeration`, `Range` and `Expression` constraint type according to the latest dmn entry data type, even not saved.


yesamer (approved):
Manually tested, code LGTM. Thank you @jomarko

baldimir (approved):

PR #1639, opened 2023-05-16
NO-ISSUE: Ignoring failing GraphUtilsTest

yesamer:
In the last few days, the `GraphUtilsTest.testComputeGraphHashCode()` randomly failed.
The root cause is still not determined, and very hard to find, because:
- There are no changes in that area that impacted the involved code;
- In local environments, the test never fails;

I tried to change the JVM version but with no luck. 

Considering that failure is impacting our CI/CD builds, I propose to Ignore that, guessing that may be related to CI/CD configuration setup - but no evidence for that.


yesamer:
Screenshot 2023-05-16 at 14 10 18

yesamer:
Hi @romartin, in the last few days we report that test failing in our CI ^
Do you have any idea of the reason? 
//cc @jomarko

romartin:
Hey @yesamer , honestly no idea before hand. It looks the hashCode for some bpmn model assertion is failing.

As far as I know we just fixed a single customer issue a few time ago, about bpmn: https://issues.redhat.com/browse/RHPAM-4604. On the other hand, we merged because of green checks. @hasys you think it can be related?

hasys:
Hi @romartin, @yesamer,

yes, when we merged my PR all was green, and also we merged it about 1 month ago. So it shouldn't be caused by my PR.

I think I saw something similar before and it was caused by Java version. But it was in different project, so not sure if it is the case here. I did a glance and it looks like there is some Java_Zulu_jdk 11 is used. Is it the same JDK as before? Didn't you or Github change something around JDK?

yesamer:
@romartin @hasys Thank you for looking into that. It seems that occurs in GitHub CI/CD only, in my local, I launched it dozen times and it never failed. So, yes, it can be related to some change on GitHub side

yesamer:
@jomarko @romartin @hasys I didn't find any consistent reason to explain this failure. So, I propose to ignore that test in both stunner-editor and swf editor code. If you agree, please approve.

.github/actions/setup-env/action.yml
```diff
@@ -30,7 +30,7 @@ runs:
     - name: "Setup JDK 11"
       uses: actions/setup-java@de1bb2b0c5634f0fc4438d7aa9944e68f9bf86cc
       with:
-        java-version: 11
+        java-version: 11.0.18
```
hasys: Hello @yesamer if unstable test is disabled, is there a need to freeze the java version?

hasys (approved):
I see java freeze already removed. PR looks good to me, thank you!

.github/actions/setup-env/action.yml
yesamer: @hasys It was a typo, I already removed that :)

caponetto (approved):
Thanks!

paulovmr (approved):

PR #1647, opened 2023-05-17
kie-issues#252: A Decision Table's single output column header should be merged

### LINKED ISSUE #252: A Decision Table's single output column header should be merged
**ISSUE 1**

In the case of a single output column in the decision header, the header contains 2 different rows, as shown

image

From DMN Specs:

`
A decision table consists of:

- A list of output clauses (one or more). Each output clause is made of a name and optional allowed values for the output entries that correspond to the clause. The output entries are contained in the rules, and the ith output entry corresponds to the ith output clause **A single output clause has no name**. 

Therefore, we should merge the 2 header rows, as the old editor did before.

Screenshot 2023-05-09 at 20 57 26

**ISSUE 2**

In this ticket, I would evaluate merging the "U" and "O" cells in the first header column.

```[tasklist]
### Acceptance Criteria
- [ ] Creating a Decision table based on an output simple type should have a single merged header row
- [ ] The output column should correctly be persisted (save /close / reopen )
- [ ] Merging the "U" and "O" on the first header row (HIT POLICY Cell)
```

danielzhe:
Closes: https://github.com/kiegroup/kie-issues/issues/252

### DIFF
--- a/packages/boxed-expression-component/src/table/BeeTable/BeeTableHeader.tsx
+++ b/packages/boxed-expression-component/src/table/BeeTable/BeeTableHeader.tsx
@@ -123,14 +123,16 @@ export function BeeTableHeader<R extends object>({
     [onColumnUpdates]
   );
 
-  const renderRowIndexColumn = useCallback<(column: ReactTable.ColumnInstance<R>, rowIndex: number) => JSX.Element>(
-    (column, rowIndex) => {
+  const renderRowIndexColumn = useCallback<
+    (column: ReactTable.ColumnInstance<R>, rowIndex: number, rowSpan: number) => JSX.Element
+  >(
+    (column, rowIndex, rowSpan) => {
       const columnKey = getColumnKey(column);
       const classNames = `${columnKey} fixed-column no-clickable-cell counter-header-cell`;
 
       return (
         <BeeTableTh
-          rowSpan={1}
+          rowSpan={rowSpan}
           key={columnKey}
           column={column}
           columnIndex={0}
@@ -171,7 +173,7 @@ export function BeeTableHeader<R extends object>({
 
       const ret = column.isRowIndexColumn ? (
         <React.Fragment key={"row-index-column"}>
-          {shouldRenderRowIndexColumn && renderRowIndexColumn(column, rowIndex)}
+          {shouldRenderRowIndexColumn && renderRowIndexColumn(column, rowIndex, rowSpan)}
         </React.Fragment>
       ) : (
         <React.Fragment key={getColumnKey(column)}>
@@ -293,7 +295,7 @@ export function BeeTableHeader<R extends object>({
       // rowIndex === -1 --> Last headerGroup
       // rowIndex === -2 --> Second to last headerGroup
       // ... and so on
-      const rowIndex = -(reactTableInstance.headerGroups.length - 1 - index + 1);
+      const rowIndex = -(reactTableInstance.headerGroups.length - index);
       let lastParentalHeaderCellIndex = 0;
 
       const { key, ...props } = { ...headerGroup.getHeaderGroupProps(), style: {} };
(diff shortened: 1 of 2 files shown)


jomarko (approved):
@danielzhe thank you. I did manual check using vsix from CI artifacts, There the behavior seems OK to me. Would be possible to double check DMN runner table view row index column effect of this PR before merge?

ljmotta:
@danielzhe Did you manage to take a look in the DMN Runner table? If you need any help let me know.

danielzhe:
> @danielzhe Did you manage to take a look in the DMN Runner table? If you need any help let me know.

No, I will contact you in private chate to help me with this test.

yesamer:
@danielzhe Clickable area to change the HIT Policy seems not covering the entire cell 

https://github.com/kiegroup/kie-tools/assets/16005046/2bf8d1e5-b233-4e0e-8033-aa64eb46d01f

yesamer (commented):
It seems that the behavior in the DMN Runner is correct 
Screenshot 2023-05-25 at 10 15 17

@ljmotta WDYT?

danielzhe:
> @danielzhe Clickable area to change the HIT Policy seems not covering the entire cell
> 
>  Screen.Recording.2023-05-25.at.10.08.26.mov

This is a known issue not introduced by this change.

yesamer:
> > @danielzhe Clickable area to change the HIT Policy seems not covering the entire cell
> > Screen.Recording.2023-05-25.at.10.08.26.mov
> 
> This is a known issue not introduced by this change.

Ok, Should I open a new ticket for that?

danielzhe:
> > > @danielzhe Clickable area to change the HIT Policy seems not covering the entire cell
> > > Screen.Recording.2023-05-25.at.10.08.26.mov
> > 
> > 
> > This is a known issue not introduced by this change.
> 
> Ok, Should I open a new ticket for that?

Yes, please.

packages/boxed-expression-component/src/table/BeeTable/BeeTableHeader.tsx
```diff
@@ -306,7 +308,24 @@ export function BeeTableHeader<R extends object>({
                 getColumnIndexOfHeader(reactTableInstance, placeholder) >= 0
                   ? getColumnIndexOfHeader(reactTableInstance, placeholder)
                   : lastParentalHeaderCellIndex++;
-              return renderColumn(rowIndex + depth - 1, placeholder, columnIndex, visitedColumns, depth);
+
+              if (placeholder.isRowIndexColumn) {
+                if (headerVisibility === BeeTableHeaderVisibility.AllLevels) {
+                  if (rowIndex == -reactTableInstance.headerGroups.length) {
```
yesamer: @danielzhe I guess these 2 `if` can be merged (Optional)

yesamer (approved):
LGTM @danielzhe, please wait for @ljmotta approval before merging it, just to doublecheck everything is fine on Runned side.

packages/boxed-expression-component/src/table/BeeTable/BeeTableHeader.tsx
ljmotta: Please consider using triple equals instead of double;

```suggestion
                  if (rowIndex === -reactTableInstance.headerGroups.length) {
```

ljmotta: I don't think he can merge those ifs, or it would go to the else clause even when the `headerVisibility === BeeTableHeaderVisibility.AllLevels` is `true`

ljmotta (commented):
I've made some comments inline.

packages/boxed-expression-component/src/table/BeeTable/BeeTableHeader.tsx
yesamer: @ljmotta You're right, I "linked" the `else`s with the wrong `if `s 🤦 . Thank you for spotting

danielzhe: Exactly, @ljmotta . I want to call `return renderColumn(rowIndex + depth - 1, placeholder, columnIndex, visitedColumns, depth);` only when `headerVisibility === BeeTableHeaderVisibility.AllLevels` are `false`.

If its true, I want to check if we are in the first row index (`rowIndex == -reactTableInstance.headerGroups.length`)

If is true, we render. Otherwise, we just ignore and do not render.

ljmotta:
@yesamer As this PR introduces a merge behavior on rows with one input/output, I think the Sample model doesn't capture this new feature. A more appropriate test would be to create a single Input node and a single Decision node as the following one:

image

Then attribute this structure Data Type to the Input Node:

image

And the this Expression to the Decision Node:

image

As a result we should have something like this in the DMN Runner table:

image

If everything is working as shown, and adding output columns in the Decision Table inside the Context Expression correctly creates new columns, so everything is good. I would play a little with the DMN by adding a few more input nodes, and deleting them to check the table behavior.

packages/boxed-expression-component/src/table/BeeTable/BeeTableHeader.tsx
danielzhe: For people reading this and not understanding how can the minus length of groups be the first row index: Notice that the `first row index` in the headers are negative, so the top row in header is ´-reactTableInstance.headerGroups.length´ and the bottom is 0, because we start to count the rows in table's content for 1 to _n_.

yesamer: Yeah, that was clear. Thank you for underlining that.

yesamer:
@ljmotta Thank you, I tested the DMN Runner, and I didn't notice any unexpected behavior. Merging.

PR #1656, opened 2023-05-22
KOGITO-8966: Ability to completely erase stored data on browser

fantonangeli:
Jira: https://issues.redhat.com/browse/KOGITO-8966

**Description**
Users should be able to completely erase data saved on their browser through Web Tools settings.

-  We should have a new menu below "Feature Preview" called "Storage"
-  The "Storage" page should have a button to clean up everything (indexed DBs) with a nice explanation
-  A confirmation modal should be opened when the user clicks on the button
-  After the clean-up action, users should be redirected to the Overview page
-  The confirmation modal should also have a checkbox to clean up cookies and local storage
-  Users should know that if they erase cookies and local storage, all settings will be gone

    
**Preview**
https://fantonangeli.github.io/kie-tools/kogito-8015/
    
[KOGITO-8966.webm](https://github.com/kiegroup/kie-tools/assets/17780574/8ad334e1-cb78-4666-8b95-84fbd430e9e1)


packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const dbNames = await Dexie.getDatabaseNames();
+
+  // is not possible to use await here: Dexie doesn't reject or resolve the promise if the db is locked
+  dbNames.forEach(Dexie.delete);
+};
+
+/**
+ * Delete all cookies
+ */
+const deleteAllCookies = () => {
```
caponetto: Maybe move this function to `packages/serverless-logic-web-tools/src/cookies/index.ts`?

packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
```diff
@@ -0,0 +1,173 @@
...
+import { ConfirmDeleteModal } from "../../table/ConfirmDeleteModal";
+import { SETTINGS_PAGE_SECTION_TITLE } from "../SettingsContext";
+import Dexie from "dexie";
+
+const PAGE_TITLE = "Storage";
+
+/**
+ * Delete all indexed DBs using Dexie
+ */
+const deleteAllIndexedDBs = async () => {
```
thiagoelg: I believe this can be achieved without Dexie. We used to have a startupBlocker that cleaned the indexedDB storage if its structure was changed (this blocker has been removed).

You can see how it was done here: https://github.com/kiegroup/kie-tools/pull/1450/files#diff-23b3c862425324f1b16f2c400f1e9b95eca5af77f75f1e8e739f7f70806ae2a7L56

packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const toggleConfirmModal = useCallback(() => {
+    setIsConfirmDeleteModalOpen((isOpen) => !isOpen);
+  }, []);
+
+  const deleteSuccessAlert = useAlert(
+    alerts,
+    useCallback(({ close }) => {
+      setTimeout(() => {
+        window.location.href = routes.home.path({});
```
caponetto: On second thought, I think redirecting is unnecessary now that the settings are not a modal anymore. WDYT?

packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
caponetto: Hi @thiagoelg, unfortunately `indexedDB.databases()` does not work on Firefox.

packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
fantonangeli: @caponetto I preferred redirecting to ensure every configuration is completely reloaded and the "Delete Storage" operation is something the user should not frequently do, so a few seconds more should be ok.
If you think it's nicer to use `history.push()`, I'm available to change it

caponetto: I think users might get confused because there is no warning saying that the page will be redirected.
An option would be to add something like "Redirecting to Overview page in X seconds..." when you show the "Data deleted successfully" alert.

packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
thiagoelg: Oh, wow, you're right!
Really weird for something that is part of the spec to not be implemented in Firefox.

thiagoelg: But how does Dexie do it?
Checking this link (https://dexie.org/docs/Dexie/Dexie.getDatabaseNames()) it seems that it uses a `__dbnames` database to store the database names and then list them from there (as a fallback, if the `.databases()` method is not supported).
But since our databases are not created via Dexie, how is this `__dbnames` populated?

This is just me being curious right now 😄

caponetto: Wow, I always thought that `__dbnames` was something maintained by Firefox so that there is a way to list the databases due to the lack of that API implementation, but this is in fact done by Dexie.
I will create a JIRA to improve this but at the moment we will have to stick to that warning we have on Firefox.
Thanks for raising this up, @thiagoelg!

caponetto: @fantonangeli Let's hide this tab for Firefox only and use the standard API for db listing/deleting.

packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
```diff
@@ -0,0 +1,197 @@
...
+      }
+      deleteSuccessAlert.show();
+    } catch (e) {
+      deleteErrorAlert.show();
+    }
+  }, [toggleConfirmModal, deleteSuccessAlert, isDeleteLocalStorageChecked, deleteErrorAlert, isDeleteCookiesChecked]);
+
+  useEffect(() => {
+    if (!isBrowserChromiumBased()) {
+      history.push(routes.settings.home.path({}));
```
caponetto:
```suggestion
      history.replace(routes.settings.home.path({}));
```

packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  const toggleConfirmModal = useCallback(() => {
+    setIsConfirmDeleteModalOpen((isOpen) => !isOpen);
+  }, []);
+
+  const deleteSuccessAlert = useAlert(
+    alerts,
+    useCallback(({ close }) => {
+      setTimeout(() => {
+        window.location.href = routes.home.path({});
```
caponetto:
```suggestion
        history.push({ pathname: routes.home.path({}) });
```

packages/serverless-logic-web-tools/src/settings/storage/StorageSettings.tsx
```diff
@@ -0,0 +1,197 @@
...
+      }, 5000);
+      return (
+        <Alert
+          variant="success"
+          title={
+            <>
+              Data deleted successfully. <br />
+              You will be redirected to the home page in <Timer delay={5} /> seconds
```
caponetto: Please create a const for this time value and reuse it here and in line 80.

caponetto (approved):

PR #1657, opened 2023-05-22
kie-issues#245: Add default .env and update README

### LINKED ISSUE #245: Create a Docker Compose file to run KIE Sandbox and its related services
**Description**

This task will help us to quickly get the KIE Sandbox running on local. Currently, the only way to run KIE sanbox is to manually install all tools on the infrastructure like Node 18, Java 11 etc. In addition to this, KIE Sandbox also needs other services to run, For example, KIE Sandbox (at kie-sandbox-image), Extended Services (at extended-services-image), and the Git CORS Proxy (at git-cors-proxy-image). This will speed up and streamline the process for users to run their instance of KIE Sandbox on local. 

Docker Compose is a technology that can enable us to run these multi-container services with a single command ([read more](https://docs.docker.com/compose/)). 

**Acceptance Criteria**

- A docker compose file should be created containing configurations to run KIE Sandbox and all related services.
- On running command `docker compose up`, All services related to KIE Sandbox should run in their specified port.
- We should be able to access KIE Sandbox at the port specified in docker-compose.yml file.

thiagoelg:
Adds a default environment vars file: `.env.default`.

Docker compose now can be run with `docker compose --env-file .env.default up` or with the `pnpm docker:start` command.

Closes https://github.com/kiegroup/kie-issues/issues/245


tiagobento:
@thiagoelg Shouldn't this .env.default be generated? I mean, we're duplicating a lot of values there. I think maintaining `env/index.js` is already hard enough 😅

thiagoelg:
@tiagobento the `.env` file is generated, but having a default means that we can just zip the package and release it and users can run the docker compose without ever needing to bootstrap our repo. Here's an example of where it could be useful: https://github.com/IBM/kie-roadmap/issues/96

tiagobento:
@thiagoelg We can generate and commit it at the same time, just like we do with `repo/graph.dot` and `repo/graph.json`. We also do that on `scripts/run-script-if/README.md`.

The build itself generates the contents of the file, and the CI will catch if someone fails to commit it, as we run a `git diff` as part of our PR checks.

tiagobento (approved):

PR #1665, opened 2023-05-24
kie-issues#129: chrome-extension: Open in KIE Sandbox button doesn't work in a private repository

### LINKED ISSUE #129: chrome-extension: "Open in KIE Sandbox" button doesn't work in a private repository
After configuring the token in the KIE Sandbox and clicking in the button, the following error appears:
cant import

Even though the possibility to import the repository using the import URL is working:
url import is working

## Backend part
Was addressed by https://github.com/apache/incubator-kie-tools/pull/2087

## Frontend part analysis
We have this kind of views of github repositories

### Project root
- Github url: https://github.com/apache/incubator-kie-kogito-examples 🔴 
- Sandbox import url: https://localhost:9001/#/import?url=https://github.com/apache/incubator-kie-kogito-examples 🟢 

For this view we miss the button in UI, but the backend is able to import using this kind of URL.

### Project folder
- Github url: https://github.com/apache/incubator-kie-kogito-examples/tree/stable/kogito-quarkus-examples/dmn-quarkus-example 🟠  
- Sandbox import url: https://localhost:9001/#/import?url=https://github.com/apache/incubator-kie-kogito-examples/tree/stable/kogito-quarkus-examples/dmn-quarkus-example 🔴  

There is bug in UI that the button is not showed for the folder view if it is opened for the first time, for second opening the button is present. For the backend importing this URL is not supported at all. User receives an error about unsupported URL type.

### Project file
- Github url: https://github.com/apache/incubator-kie-kogito-examples/blob/stable/kogito-quarkus-examples/dmn-drools-quarkus-metrics/src/main/resources/org/kie/kogito/examples/LoanEligibility.dmn 🔴 
- Sandbox import url: https://localhost:9001/#/import?url=https://github.com/apache/incubator-kie-kogito-examples/blob/stable/kogito-quarkus-examples/dmn-drools-quarkus-metrics/src/main/resources/org/kie/kogito/examples/LoanEligibility.dmn 🟢 

When single file is opened, we do not have there 'Open in KIE Sandbox' button.

klbowers:
Closes https://github.com/kiegroup/kie-issues/issues/129
When clicking the Chrome Extensions's "Open in KIE Sandbox" button on a private repo, with no connected account, on KIE Sandbox, a user will now be directed to an error page that has a link to authenticate and reimport.
When clicking the Chrome Extensions's "Open in KIE Sandbox" button on a private repo, with a relevant connected account, it will now be imported in the same way as public repos currently are.

Video demonstrating changes:
https://github.com/kiegroup/kie-tools/assets/92726146/bc52cb97-3cca-4b63-9094-364e1d2982b2


tiagobento:
@kbowers-ibm The video looks great, I think it's very nice the way it is!

Regarding the code, I think there's a better way to do that, instead of letting an exception occur and have the option to fix the error be shown in EditorPageErrorPage..

From what I understood, if there are no compatible authSessions with the URL being imported, a "cannot read 'id' from undefined" will happen, which will result in an Error being thrown, showing the Error page.

I think we could do better, and instead of relying on the EditorPageErrorPage, we could create an internal state directly on `NewWorkspaceFromUrlPage` that controls whether or not the user needs to authenticate to proceed. WDYT?

klbowers:
Yeah that sounds cleaner, good idea!

klbowers:
Updated functionality from an error message with a clickable button to instead automatically open the import modal if it can't find a branch name - which should never happen with public repos and will always happen with a private repo (if no valid auth token available). If a valid auth token exists it will find the branch and therefore import automatically without opening the modal.

https://github.com/kiegroup/kie-tools/assets/92726146/86a1297d-3ddf-43e6-baad-1b605fdf65b3

tiagobento:
@kbowers-ibm Aren't we still erroring out when there's no `compatible` authSession? I mean, I think the rule should be a combination of things, not only based of the branch name not being present.. I'm not saying it doesn't work like it is, just saying that the code can be written in a way that expressed the intended behavior better.

tiagobento:
These two tests might be actually failing @kbowers-ibm... I ran three times now, and they failed consistently.. can you please take a look?

klbowers:
@tiagobento Passing now! Changed code to only execute when the URL being imported isCertainlyGit. Made an oversight as I didn't realize that the code would be run when e.g. clicking on 'Try sample,' which is what was causing the tests to fail.

packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
```diff
@@ -142,35 +140,39 @@ export function NewWorkspaceFromUrlPage() {
...
+    }
     history.replace({
       pathname: routes.import.path({}),
       search: queryParams
+        .without(QueryParams.CONFIRM)
         .with(QueryParams.BRANCH, selectedGitRefName)
-        .with(QueryParams.AUTH_SESSION_ID, authSession?.id ?? compatible[0].id)
+        .with(QueryParams.AUTH_SESSION_ID, compatible[0].id)
         .toString(),
     });
```
tiagobento:
```suggestion
    setAuthSessionId(compatible[0].id);
    
    if (compatible[0].id === AUTH_SESSION_NONE.id && !selectedGitRefName) {
      history.replace({
        pathname: routes.import.path({}),
        search: queryParams.with(QueryParams.CONFIRM, "true").toString(),
      });
    } else {
      history.replace({
        pathname: routes.import.path({}),
        search: queryParams
          .without(QueryParams.CONFIRM)
          .with(QueryParams.BRANCH, selectedGitRefName)
          .with(QueryParams.AUTH_SESSION_ID, compatible[0].id)
          .toString(),
      });
    }
```

tiagobento: Adding a nice blank line before the `if` statement and grouping both cases in an `if-else` block, rather then and `if-return`. I think it looks better in this case, since we only have the two options.

packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
```diff
@@ -251,6 +253,10 @@ export function NewWorkspaceFromUrlPage() {
         return;
       }
 
+      if (!queryParamBranch && isCertainlyGit(importableUrl.type) && queryParamAuthSessionId !== AUTH_SESSION_NONE.id) {
+        return;
+      }
```
tiagobento, line 259: Can you please explain this change? I struggle to understand it.

packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
```diff
@@ -42,7 +42,7 @@ import { fetchSingleFileContent } from "./fetchSingleFileContent";
 import { useGitHubClient } from "../github/Hooks";
 import { AccountsDispatchActionKind, useAccountsDispatch } from "../accounts/AccountsContext";
 import { useAuthSession, useAuthSessions } from "../authSessions/AuthSessionsContext";
-import { useAuthProvider, useAuthProviders } from "../authProviders/AuthProvidersContext";
+import { useAuthProviders } from "../authProviders/AuthProvidersContext";
```
tiagobento, line 45: Unused?

tiagobento (changes requested):
Thanks @kbowers-ibm! We're almost there.

packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
klbowers: Yep, couldn't see any use of it anywhere, so got rid of that and a few others

packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
klbowers: So previously I had the state being set to false here. However when I removed it, I noticed that in the case of no auth token and repo is private we get the import modal as expected. However once the github token is added it briefly shows the error screen before carrying on through to import it properly. So this line stops that from happening.
https://github.com/kiegroup/kie-tools/assets/92726146/037a62d2-f0e9-4a9b-9393-548ede2b15d3

yesamer (commented):
Hi @kbowers-ibm, I manually tested your PR, please take a look at the below video

https://github.com/kiegroup/kie-tools/assets/16005046/fa94c617-40f3-4413-a8d5-3f1dcb7f13ed

Consider that I locally rebased your branch with `main`, so can you please rebase this PR and double-check this behavior?

yesamer:
@kbowers-ibm Not sure this behavior is related to your change, it seems it occurs with public repos too.

klbowers:
Hi Yeser,
Very confused how that's happening as I added that error message in my first commit, but removed it the second https://github.com/kiegroup/kie-tools/pull/1665/commits/41fc420687b381fcac0125d25471f1a5b1ab4281. Are you definitely testing the latest commit?

klbowers:
@yesamer Also judging by the error and that it's not working on public repos, can I confirm that you have git-cors-proxy-image package running?

yesamer:
@kbowers-ibm Good point, I forgot to launch it. Thank you.

yesamer:
Ah no, ok. It's correct. Nevermind

yesamer:
@kbowers-ibm Sorry, my bad. I had in my local an old version of your PR.

yesamer (approved):
Thank you @kbowers-ibm, it works well. Apologies for the confusion, my bad :/

klbowers:
Perfect! Thank you very much @yesamer, appreciate you taking the time :)

klbowers:
And no worries about the confusion! I forgot to launch git-cors-proxy myself a couple of times during this ticket as well and had a few moments where I thought I'd messed everything up!

PR #1667, opened 2023-05-25
kie-issues#158: Automatically focus and select file name when creating a new file on KIE Sandbox

klbowers:
Closes https://github.com/kiegroup/kie-issues/issues/158
Whenever a new file is created from the home page, or from the 'New File' dropdown in the Editor page, the name will be selected so user's can easily create a meaningful name


tiagobento:
@kbowers-ibm So I think there's a problem with this approach.... If I WANT my file to be called "Untitled", every time I open it, its name will be selected. I think we're going to need a queryParam here to make sure this behavior is only triggered when a "New file" operation happens.

klbowers:
Fair enough, will look into using a queryParam, thanks! :)

klbowers:
Video showing:
URL that has no queryparam, when clicking on add new file, it will add the queryparam, but not take you to a new file as expected. But when it does have the queryparam it will work and proceed to the new file 

https://github.com/kiegroup/kie-tools/assets/92726146/33f5489f-1f9d-4903-946c-889c0b64d672

danielzhe (changes requested):
I found an issue:
1. Create a new file and name it something like "TestName"
2. Try to create another new file

You will notice the the file isn't created instead the focus goes to the current "TestName", so user has to go to "New File" again.

It may be confuse to the users. I'm not sure if it's clear the description of the issue so there is a video.


https://github.com/kiegroup/kie-tools/assets/7305741/a40b7615-c454-43b6-b074-3ed96e1490ce

klbowers:
Hi @danielzhe,
Thanks for the review! Yeah apologies, this shouldn't really be a draft PR as there's still a few kinks to iron out. I have a comment in the [WorkspaceFileHooks.tsx](https://github.com/kiegroup/kie-tools/pull/1667/files#diff-4e147a84e76cf7745f983eca11b1f848a00c5905ec0fd11331c107ca6303ea83) that explains why the above is happening. The problem is the workspaceFilePromise is being cancelled before it's being set to the new value, and so reverts back to the old value. I'm just not quite sure how to stop it from being cancelled, e.g. putting something like don't cancel if only the queryparam in the URL has changed.

tiagobento:
Closing due to inactivity.

PR #1682, opened 2023-05-29
kie-issues#287: After deleting a Decision Table row, cells are wrongly selected

### LINKED ISSUE #287: Decision Table cells are wrongly selected after context menu operation is invoked
# Delete
As you can see from the below videos, after a row deletion, cells are wrongly selected

https://github.com/kiegroup/kie-issues/assets/16005046/a5ffb7a6-3d02-464b-9155-76314e9dbdd1

https://github.com/kiegroup/kie-issues/assets/16005046/e874a651-7a79-4545-8cd3-558a2c1480dfù

Previous behavior (1 week ago) 

https://github.com/kiegroup/kie-issues/assets/16005046/8df74e42-48cf-470a-982c-638025953107

# Insert left
After a column is inserted left from column header cell, the cells are wrongly selected in decision table
Screenshot 2023-05-30 135436

danielzhe:
Closes: https://github.com/kiegroup/kie-issues/issues/287

Before this change, the [registerSelectableCellRef](https://github.com/kiegroup/kie-tools/blob/f82f2244bc50fd47f46f6e08ed8c865b36b5af49/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx#L760) was being called with the old value of `selectionRef.current?`

The value of `selectionRef.current?` value changes after user delete a row/column and was being correctly set.

**So what was happening?**

Two `selectionRef.current?.active` cells was being rendered: the correct (set [here](https://github.com/kiegroup/kie-tools/blob/f82f2244bc50fd47f46f6e08ed8c865b36b5af49/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx#L524) by `adaptSelection`) and the old one 
from `registerSelectableCellRef`.

This was happening because [selectionRef](https://github.com/kiegroup/kie-tools/blob/f82f2244bc50fd47f46f6e08ed8c865b36b5af49/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx#L238) is only updated after the component is rendered and `useEffect` was happening before this updated since `useEffect` is asynchronous.

`useLayoutEffect` in other hand is synchronous and is called after [selectionRef](https://github.com/kiegroup/kie-tools/blob/f82f2244bc50fd47f46f6e08ed8c865b36b5af49/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx#L238) is updated.

Here is more info about this cycle: https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780

I didn't found any side-effect after this change but I can be wrong. Also, **if there is any issue that this change introduces and I'm not aware of, please, let me know.**

### DIFF
--- a/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx
+++ b/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx
@@ -909,7 +909,7 @@ export function useBeeTableSelectableCellRef(
 
   const [status, setStatus] = useState<BeeTableCellStatus>(NEUTRAL_CELL_STATUS);
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     const ref = registerSelectableCellRef?.(rowIndex, columnIndex, {
       setStatus,
       setValue,


danielzhe:
@tiagobento Please, take a look.

jomarko (approved):
Thank you @danielzhe . My local check passed for both `delete` and `insert` operation.

yesamer (approved):

PR #1693, opened 2023-06-02
kie-issues#174: Support constrained Data Types on inputs table of DMN Runner

### LINKED ISSUE #174: Support constrained Data Types on DMN Runner form
Adding a "Range" or "Expression" constrait for some Data Types add a "x-dmn-allowed-values" property to the model JSON Schema. This property is not used in the validation, enabling the user to add any values in the DMN Runner form without a warning. Adding a constraint to a "date", "time", "date-time", "days and time duratation", "years and months duration" will add the "x-dmn-allowed-values".

```[tasklist]
### Acceptance criteria
- [ ] Adding a "Range" constraint to a Data Type should reflect on DMN Runner form validation.
- [ ] Adding a "Expression" constraint to a Data Type should reflect on DMN Runner form validation.
```

jomarko:
Closes kiegroup/kie-issues#174

This PR:
- Set correct column minimal width for custom DMN data types that are date or time based, e.g. `myDateType: date`
- Check if there is `Range` constraint for dmn custom type and validates values put into DMN runner form [1]
- Check if there is `Enumeration` constraint for dmn custom type and validates values put into DMN runner form
- **does not** check free form **expression** constraints that can user put on DMN custom data types

[1]
Screenshot 2023-06-05 155431
Screenshot 2023-06-05 155422
Screenshot 2023-06-07 154347

### DIFF
--- a/packages/dmn-runner/src/ajv.ts
+++ b/packages/dmn-runner/src/ajv.ts
@@ -17,7 +17,12 @@
 import Ajv from "ajv";
 import * as metaSchemaDraft04 from "ajv/lib/refs/json-schema-draft-04.json";
 export { ValidateFunction } from "ajv";
+import { duration } from "moment";
 import {
+  DATE_AND_TIME_ENUM_REGEXP,
+  DATE_AND_TIME_FEEL_REGEXP,
+  DATE_ENUM_REGEXP,
+  DATE_FEEL_REGEXP,
   DAYS_AND_TIME_DURATION_FORMAT,
   DAYS_AND_TIME_DURATION_REGEXP,
   RECURSION_KEYWORD,
@@ -25,12 +30,107 @@ import {
   X_DMN_ALLOWED_VALUES_KEYWORD,
   X_DMN_DESCRIPTIONS_KEYWORD,
   X_DMN_TYPE_KEYWORD,
+  DURATION_ENUM_REGEXP,
+  DURATION_FEEL_REGEXP,
+  TIME_ENUM_REGEXP,
+  TIME_FEEL_REGEXP,
   YEARS_AND_MONTHS_DURATION_FORMAT,
   YEARS_AND_MONTHS_DURATION_REGEXP,
 } from "./constants";
 
+/**
+ * Please notice this is enum is slightly different from DMN built in types.
+ * These values are used in 'ajv' to represent corresponding DMN built-in types.
+ * This enum is primarily for 'ajv' validation code, where we want to validate DMN runner values according to X_DMN_ALLOWED_VALUES.
+ *
+ * There is this deviation for example:
+ * ajv: 'date-time'
+ * dmn: 'date and time'
+ */
+enum DmnAjvSchemaFormat {
+  DATE = "date",
+  TIME = "time",
+  DATE_TIME = "date-time",
+  DAYS_TIME_DURATION = "days and time duration",
+  YEARS_MONTHS_DURATION = "years and months duration",
+}
+
 export class DmnRunnerAjv {
   private ajv;
+
+  private parseRangeFromAllowedValues = (xDmnAllowedValues: string, type: DmnAjvSchemaFormat) => {
+    if (
+      (xDmnAllowedValues.startsWith("[") || xDmnAllowedValues.startsWith("(")) &&
+      (xDmnAllowedValues.endsWith("]") || xDmnAllowedValues.endsWith(")")) &&
+      xDmnAllowedValues.includes("..")
+    ) {
+      //It is a range
+      let feelFunction: string;
+      let regExp;
+      switch (type) {
+        case DmnAjvSchemaFormat.DATE:
+          feelFunction = "date";
+          regExp = DATE_FEEL_REGEXP;
+          break;
+        case DmnAjvSchemaFormat.TIME:
+          feelFunction = "time";
+          regExp = TIME_FEEL_REGEXP;
+          break;
+        case DmnAjvSchemaFormat.DATE_TIME:
+          feelFunction = "date and time";
+          regExp = DATE_AND_TIME_FEEL_REGEXP;
+          break;
+        case DmnAjvSchemaFormat.DAYS_TIME_DURATION:
+        case DmnAjvSchemaFormat.YEARS_MONTHS_DURATION:
+          feelFunction = "duration";
+          regExp = DURATION_FEEL_REGEXP;
+          break;
+      }
+      const matches = xDmnAllowedValues.match(regExp);
+      if (matches) {
+        const minAllowed = matches[0].replace(`${feelFunction}("`, "").replace('")', "");
+        const minAllowedIncluded = xDmnAllowedValues.startsWith("[");
+        const maxAllowed = matches[1].replace(`${feelFunction}("`, "").replace('")', "");
+        const maxAllowedIncluded = xDmnAllowedValues.endsWith("]");
+        return { minAllowed, minAllowedIncluded, maxAllowed, maxAllowedIncluded };
+      }
+    }
+    return {};
+  };
+
+  private parseEnumerationFromAllowedValues = (xDmnAllowedValues: string, type: DmnAjvSchemaFormat) => {
+    // try to check if it is enumeration
+    let feelFunction: string;
+    let regExp;
+    switch (type) {
+      case DmnAjvSchemaFormat.DATE:
+        feelFunction = "date";
+        regExp = DATE_ENUM_REGEXP;
+        break;
+      case DmnAjvSchemaFormat.TIME:
+        feelFunction = "time";
+        regExp = TIME_ENUM_REGEXP;
+        break;
+      case DmnAjvSchemaFormat.DATE_TIME:
+        feelFunction = "date and time";
+        regExp = DATE_AND_TIME_ENUM_REGEXP;
+        break;
+      case DmnAjvSchemaFormat.DAYS_TIME_DURATION:
+      case DmnAjvSchemaFormat.YEARS_MONTHS_DURATION:
+        feelFunction = "duration";
+        regExp = DURATION_ENUM_REGEXP;
+        break;
+    }
+    const matches = xDmnAllowedValues.match(regExp);
+    if (matches) {
+      return [
+        ...matches.map((matchedString) =>
+          matchedString.trim().replace(`${feelFunction}("`, "").replace('"),', "").replace('")', "")
+        ),
+      ];
+    }
+    return [];
+  };
 
   constructor() {
     this.ajv = new Ajv({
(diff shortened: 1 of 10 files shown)


ljmotta (commented):
@jomarko  Thanks for taking care of this issue.

We shouldn't add any DMN Runner specific checks or/and  DMN logic inside the `uniforms-patternfly` library. Its sole purpose is to generate a form based on a JSON Schema, and any other thing not related to the JSON Schema spec or the form generation shouldn't be in this library.

Regarding the data validation, we are just showing the inline errors in the Form view, as it will overlap the table rows.

Now, to address this issue, a more appropriate approach would verify if it's possible to add this check in our Ajv setup in the `packages/dmn-runner/ajv.ts`. Take a look in the Ajv docs on how to add [keywords](https://ajv.js.org/keywords.html).

If it isn't possible to add the validation in the Ajv setup, we should change our form and table `uniforms` Bridge.
 - form: `packages/form-dmn/src/uniforms/FormDmnJsonSchemaBridge.ts`
 - table: `packages/unitables-dmn/src/uniforms/DmnUnitablesJsonSchemaBridge.ts`

tiagobento:
@jomarko Formatting is broken for this PR... Can you please take a look?

packages/dmn-runner/src/ajv.ts
```diff
@@ -39,7 +132,87 @@ export class DmnRunnerAjv {
...
+              if (minAllowedDate && compareMinBoundary(actualDate, minAllowedDate)) {
+                return false;
+              }
+              const maxAllowedDate = new Date(maxAllowed);
+              if (maxAllowedDate && compareMaxBoundary(actualDate, maxAllowedDate)) {
+                return false;
+              }
+            }
+          }
+          if (enumeratedValues) {
```
yesamer: @jomarko A type can have an `allowedValues` only, I'd use a "else if" statement:

eg. 
`if (minAllowed && maxAllowed) {
 ....
} else if (enumeratedValues) {
 .....
`

yesamer (commented):
Thank you @jomarko, great feature. 

In my local, Range validation (for `date`) doesn't work as expected, please check the above video. It's weird because I guess you already tested that 

https://github.com/kiegroup/kie-tools/assets/16005046/08c181ad-4b46-4413-b440-3f08ae9228b8

packages/dmn-runner/src/ajv.ts
```diff
@@ -39,7 +132,87 @@ export class DmnRunnerAjv {
...
+
+        return (data: string) => {
+          if (data.includes(".") && data.endsWith("Z")) {
+            // adjusting from "2023-06-01T10:42:00.000Z" to "2023-06-01T10:42:00"
+            data = data.substring(0, data.lastIndexOf("."));
+          }
+          if (minAllowed && maxAllowed) {
+            // It is a Range constraint
+            if (parentSchema.format === "time") {
+              if (compareMinBoundary(data, minAllowed ?? "") || compareMaxBoundary(data, maxAllowed ?? "")) {
```
yesamer: `?? ""` are not necessary, based on the `if` in line 163

```suggestion
              if (compareMinBoundary(data, minAllowed) || compareMaxBoundary(data, maxAllowed)) {
```

packages/dmn-runner/src/ajv.ts
```diff
@@ -39,7 +132,87 @@ export class DmnRunnerAjv {
...
+          parentSchema.format !== "years and months duration"
+        ) {
+          return (data: string) => true;
+        }
+        const { minAllowed, minAllowedIncluded, maxAllowed, maxAllowedIncluded } = this.parseRangeFromAllowedValues(
+          schema ?? "",
+          parentSchema.format
+        );
+        const enumeratedValues = this.parseEnumerationFromAllowedValues(schema ?? "", parentSchema.format);
+        const compareMinBoundary: (a: Date | String | number, b: Date | String | number) => boolean = minAllowedIncluded
```
yesamer: I would rename that to `isUnderTheMinBoundary` to better clarify the scope.
I would rename `a` to `actual` or `value`, and `b` to `minBoundary`

packages/dmn-runner/src/ajv.ts
jomarko: will do, thank you, good suggestion

packages/dmn-runner/src/ajv.ts
jomarko: well spotted, removing

packages/dmn-runner/src/ajv.ts
jomarko: perfect suggestion for renaming, will do

packages/dmn-runner/src/ajv.ts
```diff
@@ -29,6 +38,90 @@ import {
 export class DmnRunnerAjv {
   private ajv;
 
+  private parseRangeFromAllowedValues = (
+    xDmnAllowedValues: string,
+    units: "date" | "time" | "date-time" | "days and time duration" | "years and months duration"
```
ljmotta: I don't think `units` correctly capture what this variable is. The method invocation passes  down a `type` in this parameter, so I guess `type` or `dmnType` would be better.

Also, you should create an enum in this case, to avoid repeated strings in your code.

```
enum DmnType {
  DATE = "date",
  TIME = "time",
  ...
}

File: packages/dmn-runner/src/ajv.ts [hunk 5]
```diff
@@ -29,6 +38,90 @@ import {
...
+  ) => {
+    if (xDmnAllowedValues) {
```
ljmotta: The `xDmnAllowedValues` is a string, it can't be `undefined` or `null`, so this if statement is not necessary, unless you're checking for an empty string. I suggest to add a short circuit "if" to avoid the entire code to be inside of one big if: 

```ts
if (xDmnAllowedValues === "") {
  return {}
}
```

```suggestion
```

packages/dmn-runner/src/ajv.ts
```diff
@@ -29,6 +38,90 @@ import {
...
+          const maxAllowed = matches[1].replace(`${feelFunction}("`, "").replace('")', "");
+          const maxAllowedIncluded = xDmnAllowedValues.endsWith("]");
+          return { minAllowed, minAllowedIncluded, maxAllowed, maxAllowedIncluded };
+        }
+      }
+    }
+    return {};
+  };
+
+  private parseEnumerationFromAllowedValues = (
```
ljmotta: Same comments as the above method.

packages/dmn-runner/src/ajv.ts
```diff
@@ -39,7 +132,87 @@ export class DmnRunnerAjv {
     });
     this.ajv.addMetaSchema(metaSchemaDraft04);
     this.ajv.addKeyword(X_DMN_TYPE, {});
-    this.ajv.addKeyword(X_DMN_ALLOWED_VALUES, {});
+    this.ajv.addKeyword(X_DMN_ALLOWED_VALUES, {
+      compile: (schema, parentSchema: { format?: string }, it) => {
```
ljmotta: With your enum you could add a more specific type here.

packages/unitables/src/uniforms/UnitablesJsonSchemaBridge.tsx
```diff
@@ -158,6 +158,14 @@ export class UnitablesJsonSchemaBridge extends JSONSchemaBridge {
             type: field.type,
           };
         }
+        switch (field["format"]) {
```
ljmotta: Please, use the `field.format` way. The `[" ... "]` should be used only when strictly necessary. Also, the format check could be moved to inside the line 72 if statement to reuse the switch case.

ljmotta (changes requested):
@jomarko Hi! The PR looks great. I've made some specific comments inline.

packages/dmn-runner/src/ajv.ts
jomarko: good point, thank you, removing

packages/dmn-runner/src/ajv.ts
jomarko: thank you, removing also here

packages/unitables/src/uniforms/UnitablesJsonSchemaBridge.tsx
jomarko: thank you @ljmotta I understand `"format"` vs `.format`, but I am not sure how should look code once I move it to line 72. could you please share some pseudocode at least?

packages/dmn-runner/src/ajv.ts
jomarko: good point

packages/dmn-runner/src/ajv.ts
jomarko: good point, introducing one enum

packages/unitables/src/uniforms/UnitablesJsonSchemaBridge.tsx
ljmotta: I didn't test it to check some corner cases. But what I meant was to make this change on line 72:

```ts
if (!xDmnType) {
  if (field.format) {
    type = field.format
  } else {
    type = field.type;
  }
} 
```

Then you will not need this new swtich case.

jomarko: Understand now, will do, thank you!

jomarko: @ljmotta I pushed the changes in a separate commit [8b49916](https://github.com/kiegroup/kie-tools/pull/1693/commits/8b49916c0022dd0db32dcb3bb8b791275964ba7f)

However I needed to put a code to slightly different place, because it is not true `!xDmnType` is not set. See screenshot from my debuging.

Screenshot 2023-06-13 115859

jomarko:
@yesamer I struggle to reproduce the issue from your video. I changed my browser locale and system setting to 'Slovak', to be totally different than english,

Snímka obrazovky 2023-06-13 140523


However #1693 seems to work properly on my machine. May I ask you to re-check once more on your machine?

@ljmotta all points should be addressed now.

yesamer:
@jomarko Now it works well, probably one of the changes you applied "fixed" that.

yesamer (approved):

ljmotta:
Hi @jomarko I was performing some manual checks, and I found a bug, please check the following scenario:

https://github.com/kiegroup/kie-tools/assets/24302289/40158252-6833-4532-9117-41b18a8570b4

I've already opened a PR to your branch solving the issue. https://github.com/jomarko/kie-tools/pull/8

Another thing that come to my attention is that the "validation" error and the calendar icon is overlaping, this is happening for you too or it's a problem in my side? I guess it's a Patternfly bug.

ljmotta (approved):

PR #1706, opened 2023-06-05
kie-issues#284: Make it possible to recover from errors of recursive Struct Data Types in the DMN Runner

### LINKED ISSUE #284: Make it possible to recover from errors of recursive Struct Data Types in the DMN Runner
Adding a recursive Data Type to a node will cause a "Maximum call stack size exceeded", it's not possible to recover from this error without refreshing the Editor page.

The user should be able to recover from any DMN Runner error by editing the DMN model;

```[tasklist]
### Acceptance criteria:
- [ ] Changing the Input or Decision node type to a recursive Struct should show an error message in the form or table;
- [ ] It should be possible to recover from this error by editing the node, the Data Type or the DMN model;
```

ljmotta:
## issue
Closes https://github.com/kiegroup/kie-issues/issues/284

## On this PR
 - Add a DMN Runner table and form placeholder component for recursive structure Data Types;
 - Add `json-refs` library to dereference the JSON schema, and check for circular dependencies;
 - Add `process` polyfill in webpack
 - Rename all form-dmn files to align with the package name

## Video
https://github.com/kiegroup/kie-tools/assets/24302289/8ada99ef-c191-4f59-ba53-d3cd6f559d4d


packages/form-dmn/src/uniforms/FormDmnJsonSchemaBridge.ts
```diff
@@ -24,14 +24,26 @@ export enum Duration {
...
     this.i18n = i18n;
   }
 
+  public getProps(name: string, props: Record<string, any>) {
+    const superProps = super.getProps(name, props);
+
+    if (!superProps.padding && !superProps.properties) {
+      superProps.style = { padding: "5px" };
+    } else if (!superProps.padding && superProps.properties) {
+      superProps.style = { margin: "5px" };
```
jomarko: maybe 5px also good candidate to store as constant somewhere, just bringing as point for discussion, I am fine if we keep as it is

packages/form-dmn/src/uniforms/FormDmnNotSupportedField.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <div
+          aria-label="field type not supported"
+          style={{
+            display: "flex",
+            justifyContent: "center",
+            alignItems: "center",
+            width: "100%",
+          }}
+        >
+          Recursive structures <br /> are not supported yet
```
jomarko, line 42: is it intentionally without some `i18n`?

pnpm-lock.yaml
```diff
@@ -5279,6 +5279,9 @@ importers:
       js-yaml:
         specifier: ^4.1.0
         version: 4.1.0
+      json-refs:
+        specifier: ^3.0.15
+        version: 3.0.15
```
jomarko, line 5284: `json-refs` last release seems to be 3years old, just asking if we can not use some alternative, not an expert, again, just opening discussion, found this:

https://www.npmjs.com/package/@apidevtools/json-schema-ref-parser

jomarko (approved):
@ljmotta my manual tests didn't reveal any issue. It works as a charm!

Just few questions/points to code, that probably do not block from merging.

pnpm-lock.yaml
ljmotta: I tryied to use a more recent library, but I extensily tested other libraries and this was the only one that correctly worked for our usecases with circular json schema and wasn't deprecated. 😞

packages/form-dmn/src/uniforms/FormDmnNotSupportedField.tsx
ljmotta: Good catch. I'll add it.

ljmotta: Actually, while I was performing this change, I remember why I didn't add 😅 . We need provide the i18n context to the form and to the table. This would require some component change and it would be better to address it in another issue. I'll create it.

ljmotta: https://github.com/kiegroup/kie-issues/issues/331

pnpm-lock.yaml
tiagobento: @jomarko Nice job checking the health of the dependency @ljmotta added. That's very valuable, nicely done!

From the chosen dependency's NPM page [1], I can see that it has a fairly high number of weekly downloads, and despite the last release being from 3 years ago, the maintainer seems to be actively replying to issues on the repo.

It's uncommon nowadays to see a project that has gotten so stable that it simply doesn't release new versions anymore, but I think that's the case of this one. Not that it doesn't have anything to change or improve, but maybe the maintainer just doesn't feel like evolving it anymore. It does what it does and that's it. 

Looking at issues and replies from the maintainers is also a good indication of activity and health of an open source project. The maintainer being active in the project means that it's very likely they will issue a patch release with a security fix if necessary.

I followed Luiz's research closely when he was looking for this library, therefore I know we can move forward with it, but I understand where you comment came from @jomarko, and please don't hesitate to keeping doing it! That's what code reviews are for :)

[1] https://www.npmjs.com/package/json-refs?activeTab=versions

tiagobento (approved):

PR #1722, opened 2023-06-07
NO-ISSUE: Remove Workspaces Shared Worker ping

ljmotta:
The keep alive interval in the shared worker tries to ping the channel every 1 minute. If it doesn't receive a response, it disconnects. This behavior was causing the shared worker to disconnect when the channel was busy. One way to make the channel busy is to make and edit, create a file or change between files. If the shared worker tries to ping the channel at the same time, it will disconnect, and the user will need to refresh the browser page to restart the shared worker.

Changing the ping to a small value make it easier to replicate, the following video changes the ping from 60000ms to 1000ms.

https://github.com/kiegroup/kie-tools/assets/24302289/cec57d36-3314-43e0-ab73-892f1e10aa81

### DIFF
--- a/packages/workspaces-git-fs/src/worker/setupWorkerConnection.ts
+++ b/packages/workspaces-git-fs/src/worker/setupWorkerConnection.ts
@@ -31,4 +31,8 @@ export function setupWorkerConnection(args: {
   args.port.addEventListener("message", (message) => bus.server.receive(message.data, args.apiImpl));
   args.port.start(); // Required when using addEventListener. Otherwise, called implicitly by onmessage setter.
   bus.clientApi.notifications.kieToolsWorkspacesWorker_ready.send();
+
+  args.fsFlushManager.subscribable.subscribe((flushes) => {
+    bus.shared.kieSandboxWorkspacesStorage_flushes.set(flushes);
+  });
 }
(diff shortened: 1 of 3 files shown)


thiagoelg (approved):

packages/workspaces-git-fs/src/worker/setupWorkerConnection.ts
```diff
@@ -31,24 +31,4 @@ export function setupWorkerConnection(args: {
   args.port.addEventListener("message", (message) => bus.server.receive(message.data, args.apiImpl));
   args.port.start(); // Required when using addEventListener. Otherwise, called implicitly by onmessage setter.
   bus.clientApi.notifications.kieToolsWorkspacesWorker_ready.send();
-
-  const flushManagerSubscription = args.fsFlushManager.subscribable.subscribe((flushes) => {
-    bus.shared.kieSandboxWorkspacesStorage_flushes.set(flushes);
-  });
```
tiagobento: This shouldn't be removed... it's not part of the `ping` mechanism.

tiagobento:
@caponetto Can you please take a look? As this impacts SLWT too.

caponetto (approved):

PR #1729, opened 2023-06-09
kie-issues#326: DMN Editor's Properties Panel changing element selection when blurring out of changed inputs

### LINKED ISSUE #326: DMN Editor's Properties Panel changing element selection when blurring out of changed inputs
1. When editing data on Properties Panel, moving over another TextBox (or TextArea) results to select the properties of another cell.

https://github.com/kiegroup/kie-issues/assets/1584568/dfcc568c-57c3-4fe5-b3d7-56dbfb35ec13

2. Entering the Decision node's expression, currently the first cell properties are shown, even if nothing has been selected. The cause of that is that the old BEE automatically selected the first cell of the table.

3. In some scenarios, Mouse Events that occurred in the Lienzo Canvas are propagated to the React-based BEE. This affects may affect several areas, and one of those is the Property Panel.

yesamer:
Closes https://github.com/kiegroup/kie-issues/issues/326 

Points 1 & 2 (described in the ticket) are fixed and simply removing the `selectExpressionEditorFirstCell();` call in `ExpressionContainerGrid.java`, which represents the old GWT-Based BEE, still present behind the curtain. 

To solve point 3, we agreed to unmount the new React component from GWT, in order to avoid any strange, unexpected and inexplicable interaction when it's not shown.


tiagobento:
@yesamer Won't we need to unmount the component when "undoing" a command that navigates back to the Diagram view too? I mean, I don't know if the click handler you changed it enough to cover all cases where we should be unmounting BEE.

yesamer:
> @yesamer Won't we need to unmount the component when "undoing" a command that navigates back to the Diagram view too? I mean, I don't know if the click handler you changed it enough to cover all cases where we should be unmounting BEE.

Very good point!

yesamer:
@tiagobento Thank you, changes applied!

tiagobento (approved):

PR #1733, opened 2023-06-09
KOGITO-9325: Support Join in Dashbuilder DataSets

jesuino:
Support a join dataset! Join datasets are a way to merge datasets into a single one and visualize or make lookup operations on top of it

 See the JIRA for more information.

image

### DIFF
--- a/packages/dashbuilder/kie-soup-dataset/kie-soup-dataset-api/src/main/java/org/dashbuilder/dataset/json/ExternalDefJSONMarshaller.java
+++ b/packages/dashbuilder/kie-soup-dataset/kie-soup-dataset-api/src/main/java/org/dashbuilder/dataset/json/ExternalDefJSONMarshaller.java
@@ -15,6 +15,7 @@
  */
 package org.dashbuilder.dataset.json;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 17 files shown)


packages/dashbuilder/kie-soup-dataset/kie-soup-dataset-api/src/main/java/org/dashbuilder/dataset/json/ExternalDefJSONMarshaller.java
```diff
@@ -46,9 +48,10 @@ public void fromJson(ExternalDataSetDef def, JsonObject json) {
         var headers = json.getObject(HEADERS);
         var accumulate = json.getBoolean(ACCUMULATE);
         var type = json.getString(TYPE);
+        var join = json.getArray(JOIN);
 
-        if (isBlank(url) && isBlank(content)) {
-            throw new IllegalArgumentException("External Data Sets must have \"url\" or \"content\" field");
+        if (isBlank(url) && isBlank(content) && (join == null || join.length() == 0)) {
```
caponetto, line 53:
```suggestion
        if (isBlank(url) && isBlank(content) && isBlank(join)) {
```
Can we use `isBlank` for `join` too?

jesuino: It will not work because `isBlank` is only for Strings and join is a JSON array

caponetto (approved):

PR #1744, opened 2023-06-14
kie-issues#342: specify target origin on postMessage

jstastny-cz:
Issue: kiegroup/kie-issues#342

Replacing asterisk in postMessage calls using component properties.


tiagobento (commented):
It's been a long time, so I can be wrong, but the asterisk was the way we found for our envelope to work cross-domain. Namely, the Chrome Extension. Please check if they still work after this change.

jstastny-cz:
@tiagobento based on tests results, the extension is the problem indeed.

jstastny-cz:
Closing as it is not possible at the moment to reduce target origins during build time and keep functionality intact, the reported vulnerabilities are at the moment inevitable to keep as known issues.

PR #1750, opened 2023-06-14
KOGITO-9301: Wrong number of pages and items after a search operation

fantonangeli:
Jira: https://issues.redhat.com/browse/KOGITO-9301

**Steps to reproduce:**
-  Go to Recent Models and set pagination to 5 per page
-  Create 6 workspaces
-  Search for a specific or an inexistent value
-  The number of pages and number of items in the table should change based on the number of search results

[KOGITO-9301.webm](https://github.com/kiegroup/kie-tools/assets/17780574/c550c5ce-08fb-4bf8-bc33-80c2e845f716)


caponetto (approved):

tomasdavidorg (changes requested):
Search does not work for single models but only for workspaces. If I create a single json file `aaa` and then I search for `a`, there is no result.
Screenshot from 2023-06-15 10-15-44

fantonangeli:
There is still a bug to be fixed when a workspace is deleted from a different browser tab

tomasdavidorg:
Hi @fantonangeli,
I have tested the changes and the searching works fine. I have found 3 possible issues:

1. In directories, searching also takes into account file extensions for the result (.sw.json, ...). If you hover over the file, you will see the whole file name + extension but in the recent files root, searching does not use the extension + the popup is not visible.

Screenshot from 2023-06-20 10-10-46

2. If you delete file from directory which has only 2 files, the directory changes into the single file but in the second opened tab it remains directory.

Screenshot from 2023-06-20 10-22-32

3. Uploading single file which is not editable will create directory. Uploading another non-editable file to the directory will cause that the files cannot be opened.

Screenshot from 2023-06-20 10-38-02

I can create jiras for the issues if they are not yet created.

I will approve the PR if none of these finding will be part of this PR.

fantonangeli:
Hi @tomasdavidorg ,

Thank you for your points. Below are my answers:
1. I suggest we can avoid blocking the PR as these 2 things are not related to the changes of the PR.
Otherwise, I can send a quick fix for both of these points, I think searching only on filenames without extensions as we maybe put the category filter in future.
2. Maybe not a bug as all workspaces and single files have the `<workspace-id>/files` route, because, in the logic, single files are workspaces with only one file. 
3.  I could not reproduce it. Please open a Jira. Maybe we can prevent the users to upload unsupported files from the Overview page.

Let me know WDYT

tomasdavidorg (approved):

tomasdavidorg:
Approved. For 1. and 3. I will report jiras.

tomasdavidorg:
- https://issues.redhat.com/browse/KOGITO-9422
- https://issues.redhat.com/browse/KOGITO-9423
- https://issues.redhat.com/browse/KOGITO-9424

packages/serverless-logic-web-tools/src/homepage/recentModels/hooks/useWorkspacesWithFilesPromise.tsx
```diff
@@ -34,16 +34,26 @@ type WorkspaceWithFilesResponse =
       errorMessage: string;
     });
 
-export function useWorkspacesWithFilesPromise(workspaceIds: WorkspaceDescriptor["workspaceId"][]) {
+export function useWorkspacesWithFilesPromise(workspaceIds?: WorkspaceDescriptor["workspaceId"][]) {
```
caponetto: Since no code is using this hook with `workspaceIds`, I think we can remove this arg (no need to keep "ids" as we will always fetch all workspaces).

Maybe rename this hook (and its file) to `useAllWorkspacesWithFilesPromise`.

packages/serverless-logic-web-tools/src/homepage/recentModels/hooks/useWorkspacesWithFilesPromise.tsx
```diff
@@ -34,16 +34,26 @@ type WorkspaceWithFilesResponse =
...
+  const [ids, setIds] = useState<string[]>([]);
 
   const refresh = useCallback(
     async (canceled: Holder<boolean>) => {
+      if (canceled.get()) {
+        return;
+      }
+
       try {
+        const idsToFetch: WorkspaceDescriptor["workspaceId"][] = workspaceIds
```
caponetto: `listAllWorkspaces` already returns all workspace descriptors, so there is no need to keep the first try-catch block below to get the descriptor again. If any of the `getWorkspace` calls fail, `listAllWorkspaces` will most likely fail first. Since now we need to use `listAllWorkspaces`, the first try-catch below is redundant.

packages/serverless-logic-web-tools/src/homepage/recentModels/hooks/useWorkspacesWithFilesPromise.tsx
```diff
@@ -59,18 +69,14 @@ export function useWorkspacesWithFilesPromise(workspaceIds: WorkspaceDescriptor[
             }
           })
         );
-        if (canceled.get()) {
-          return;
-        }
-
```
caponetto, line 65: Why move this if block?

fantonangeli: It was returning `true`, I think because of the previous call to `listAllWorkspaces`.
Moving the check to the beginning of the block, it returns `fail` as expected.

caponetto (approved):

PR #1753, opened 2023-06-14
KOGITO-9389: Support query params in dashbuilder datasets

jesuino:
Adding the support to safe query parameters in datasets.

Before this we had to encode and manually build the URL, with this new change parameters are now safely encoded and declared:

```
    - uuid: GPU by Day
      type: prometheus
      url: ${dataset_url}
      cacheEnabled: true
      query:
        query: >-
            sum(
              (increase(kepler_container_gpu_joules_total[1h])
                *${watt_per_second_to_kWh}
              ) *
              (count_over_time(kepler_container_gpu_joules_total[24h])/
                count_over_time(kepler_container_gpu_joules_total[1h])
              )
            )${time_window}
```


packages/dashbuilder/dashbuilder-runtime-parent/dashbuilder-runtime-client/src/main/java/org/dashbuilder/client/external/ExternalDataSetClientProvider.java
```diff
@@ -139,13 +140,25 @@ public void callback(DataSet dataSet) {
 
     private void fetch(ExternalDataSetDef def, DataSetReadyCallback callback) {
         var req = RequestInit.create();
+        URL url = null;
+
+        try {
```
ederign, line 141: Can we do this check without a try catch?

ederign (changes requested):

packages/dashbuilder/dashbuilder-runtime-parent/dashbuilder-runtime-client/src/main/java/org/dashbuilder/client/external/ExternalDataSetClientProvider.java
jesuino: We can do this, but I may forget some case and wrongly build URLs (e.g. forget to cover some protocol other than HTTP). Should I try or keep the try/catch, wdyt?

ederign: let's keep it for now!

ederign (approved):

caponetto (approved):

PR #1754, opened 2023-06-14
NO-ISSUE: kn-plugin-workflow: Temporarily apply/create warning (quick and dirty)

ederign:
@caponetto @zdrapela I've opened this PR more for discussion, because we don't have enough time for the final solution. For the real solution, we would need to check if "configmaps/service-props" are already created and then do a create or apply.

At this stage, I don't want to introduce an extra call and query for kubectl (that can fail). The real solution for this would be part of https://issues.redhat.com/browse/KOGITO-9381, using k8 client to do that, but meanwhile I'm trying to avoid this warning:

```
> ../cli deploy
🛠️️ Deploying a Kogito Serverless Workflow file on Kubernetes via the Kogito Serverless Workflow Operator...

🔎 Checking your deployment environment...
🔎 Checking if kubectl is available...
 - ✅ kubectl is available
🔎 Checking if kubectl has a context configured...
 - ✅ kubectl current context: minikube
🔎 Checking current namespace in kubectl...
 - ✅  kubectl current namespace: default
🔎 Checking if the Kogito Serverless Workflow Operator is correctly installed...
 - ✅ Kogito Serverless Workflow Operator is available

🛠️ Generating your deployment environment...
🔍 Looking for your Serverless Workflow File...
 - ✅ Serverless workflow file found: /Users/ederign/temp/cli/sample-project/workflow-service.sw.json
🔍 Looking for your configuration support files...
 - ✅ Properties file found: /Users/ederign/temp/cli/sample-project/application.properties
 - ✅ Support file found: /Users/ederign/temp/cli/sample-project/specs/workflow-service-openapi.json
 - ✅ Support file found: /Users/ederign/temp/cli/sample-project/specs/workflow-service-schema.json
🚚️ Generating your Kubernetes manifest files..
🛠 Deploying your Kogito Serverless project in namespace default
configmap/01-service-resources configured
 - ✅ Manifest 01-configmap_01-service-resources.yaml successfully deployed in namespace default
kogitoserverlessworkflow.sw.kogito.kie.org/service created
 - ✅ Manifest 01-kogitoserverlessworkflow_service.yaml successfully deployed in namespace default
Warning: resource configmaps/service-props is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
configmap/service-props configured
 - ✅ Manifest 02-configmap_service-props.yaml successfully deployed in namespace default
```

As far as I know, this warning message is generated when kubectl apply is used on an object that was not initially created using kubectl apply or kubectl create --save-config. But in our scenario, this doesn't change anything because k8 will apply the changes and add the missing annotation.

See: https://www.airplane.dev/blog/kubectl-apply-vs-create and https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/

I'm happy either way, so if you guys think we should go ahead and merge this until https://issues.redhat.com/browse/KOGITO-9381 is fixed, let's do this.

But if I went to far on hack mode, just block this and I'll happily close the PR.

### DIFF
--- a/packages/kn-plugin-workflow/pkg/common/operator.go
+++ b/packages/kn-plugin-workflow/pkg/common/operator.go
@@ -17,6 +17,8 @@
 package common
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
 	"github.com/kiegroup/kie-tools/packages/kn-plugin-workflow/pkg/metadata"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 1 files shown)


ederign:
@adrielparedes maybe you have an opinion about this one?

adrielparedes:
@ederign kubectl does not say anything like that, are you using kubectl directly or a different plugin? But again, if there is other way, and if we don't need a configmap (or the user) I would hide that warning since is not useful, right?

adrielparedes:
@ederign I saw the configmap you are trying to apply, and the problem is that it has an invalid annotation: `creationTimestamp: null`. There is no need to add this on the yaml files, it is just an internal annotation k8s adds once you apply that configmap (or any yaml file). That is why the warning message. You `don't need` to run `kubectl create`. Actually is not used by the community. `kubectl apply -f ...` should be enough. Here is a good explanation of the difference: https://www.airplane.dev/blog/kubectl-apply-vs-create.
If you are not creating those files, then it is a but from the app that generates those files. I think you can hide that warning if case you don't have time now to solve it.

zdrapela:
It looks fine to me to hide the warning for now 👍️

caponetto (approved):

PR #1769, opened 2023-06-16
NO-ISSUE: Switch cors-proxy dependency (maintain original isomorphic-git package-lock dependencies)

thiagoelg:
Previous version (`@pefernan/cors-proxy@0.0.3`) had deleted the entire `package-lock.json` file, causing all dependencies to be upgraded, which broke the app.

This new version (`@thiagoelg/cors-proxy@0.0.1`) only upgraded what had CVEs, so we're keeping the app stable and upgrade only what was necessary.


tiagobento (approved):

tiagobento:
The check failed at the "upload" phase. Maybe we should revisit what we keep during the build?

PR #1778, opened 2023-06-20
kie-issues#216: DMN's Editor Boxed Expression Editor: It's not possible to delete a table row in Safari

### LINKED ISSUE #216: DMN's Editor Boxed Expression Editor: It's not possible to delete a table row in Safari
The user experience in SAFARI suffers from some issues. This issue reports a couple of them, although the overall experience of the new component should be deeply checked.

Issues found so far:
- To keep the BEE menu open, you need to hold the mouse click. At the "mouse-up" event, the menu disappears.
- If you try to delete a row, nothing happens

https://user-images.githubusercontent.com/16005046/232757684-3e185300-9855-4c7a-b841-c67ea8c090a6.mov

```[tasklist]
### Acceptance criteria
- [x] BEE menu should keep open after releasing the mouse button in SAFARI
- [x] It should be possible to delete a row in SAFARI
- [x] Check that the overall UX works well in SAFARI
```

yesamer:
Closes https://github.com/kiegroup/kie-issues/issues/216

The problem:
To simulate the Right Click using the trackpad and the keyboard, the shortcut CTRL+CLICK is used.
In Chrome-based browsers, that combination results in a single `contextmenu` event. In SAFARI, that generates 2 mouse events: `contextmenu` and `click`.
So, the first `contextmenu` opens the popup, while the second `click` event suddenly closes the popup, because of this [handler registration](https://github.com/kiegroup/kie-tools/pull/1778/files#diff-7af3f1ca21b3ed58a1c4f9788ef2e5fae0b7a6c41a8a13d99a6de708f40a3451L112). 
To fix that, if the `click` event is triggered together with a CTRL, the event is ignored.

### DIFF
--- a/packages/boxed-expression-component/src/contextMenu/Hooks.ts
+++ b/packages/boxed-expression-component/src/contextMenu/Hooks.ts
@@ -38,6 +38,9 @@ export function useCustomContextMenuHandler(domEventTargetRef: React.RefObject<H
   const hide = useCallback(
     (e: MouseEvent) => {
       e.preventDefault();
+      /* In SAFARI only, CTRL + click shortcut used to open the Menu, results in two distinct mouse events: "contextmenu" and “click“, in this order.
+         Considering this hide() function is currently bound with both event handlers, the second event (click) will suddenly close the menu. 
+         To prevent this, if ctrlKey is actually pressed, the event is ignored. */
       if (!isOpen || e.ctrlKey) {
         return;
       }
--- a/packages/boxed-expression-component/src/contextMenu/Hooks.ts
+++ b/packages/boxed-expression-component/src/contextMenu/Hooks.ts
@@ -38,7 +38,7 @@ export function useCustomContextMenuHandler(domEventTargetRef: React.RefObject<H
   const hide = useCallback(
     (e: MouseEvent) => {
       e.preventDefault();
-      if (!isOpen) {
+      if (!isOpen || e.ctrlKey) {
         return;
       }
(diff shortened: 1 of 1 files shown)


packages/boxed-expression-component/src/contextMenu/Hooks.ts
```diff
@@ -38,7 +38,7 @@ export function useCustomContextMenuHandler(domEventTargetRef: React.RefObject<H
   const hide = useCallback(
     (e: MouseEvent) => {
       e.preventDefault();
-      if (!isOpen) {
+      if (!isOpen || e.ctrlKey) {
         return;
       }
```
tiagobento, line 46: This change actually is a very good candidate for a comment :D I think I would like to know what the CTRL key has to do with all that :) Can you please add a comment there?

yesamer: Fair enough, that's a classic example where it's easier to change the code than explain the change :)
Thank you @tiagobento, hope the comment is clear. If not or if an alternative solution could be evaluated, please let me know.

danielzhe: I understand now why I wasn't able to reproduce the issue: I'm using an external mouse. Great finding, @yesamer!

danielzhe (approved):
I can confirm that is working in MacBook. Thank you!

tiagobento (approved):

PR #1800, opened 2023-06-26
kie-issues#48: Implement DMN 1.3 new keywords support

### LINKED ISSUE #48: Implement DMN 1.3 new keywords support
We should be sure that 1.3 specific keywords and functions are supported by:
- autocompletion
- during save and reopen of diagram

See list of them
- https://issues.redhat.com/browse/DROOLS-5016
- https://blog.viadee.de/en/this-is-new-in-dmn-1_3

jomarko:
closes kiegroup/kie-issues#48

See commits:
- https://github.com/kiegroup/drools/commit/e991ef1ca9e6d7c63ae6b79dcab5dba13615136a
- https://github.com/kiegroup/drools/commit/f80acf947cb1c5f576ae859707b83ea890eba18d

### DIFF
--- a/packages/feel-input-component/src/FeelConfigs.ts
+++ b/packages/feel-input-component/src/FeelConfigs.ts
@@ -40,8 +40,10 @@ export const feelTheme = (): Monaco.editor.IStandaloneThemeData => {
 export const feelTokensConfig = (): Monaco.languages.IMonarchLanguage => {
   return {
     keywords: [
+      "@",
       "for",
       "in",
+      "instance of",
       "return",
       "if",
       "then",
@@ -94,6 +96,7 @@ export const feelTokensConfig = (): Monaco.languages.IMonarchLanguage => {
       "index of",
       "insert before",
       "invoke",
+      "is",
       "list contains",
       "log",
       "lower case",
@@ -191,7 +194,9 @@ export const feelDefaultSuggestions = (): Monaco.languages.CompletionItem[] => {
 
   const suggestionTypes = {
     Snippet: [
+      ["@", '@"23:59:00"', '@"P2DT20H14M"'],
       ["if", "if $1 then\n\t$0\nelse\n\t"],
+      ["instance of", "instance of number", "instance of string", "instance of boolean"],
       ["for", "for element in $1 return\n\t$0"],
     ],
     Function: [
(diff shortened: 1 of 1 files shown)


packages/feel-input-component/src/FeelConfigs.ts
```diff
@@ -191,7 +194,9 @@ export const feelDefaultSuggestions = (): Monaco.languages.CompletionItem[] => {
 
   const suggestionTypes = {
     Snippet: [
+      ["@", '@"23:59:00"', '@"P2DT20H14M"'],
       ["if", "if $1 then\n\t$0\nelse\n\t"],
+      ["instance of", "instance of number", "instance of string", "instance of boolean"],
```
jomarko: @yesamer can you share your opinion about these?
`@` and `instance of` may be followed by more different words, byt monaco supports probably just single suggestion:

should I do sth like
```
["@", '@"time or duration string"'],
["instance of", "[dmn built in type | dmn custom type | range]"]
```
Screenshot 2023-06-26 132213

packages/feel-input-component/src/FeelConfigs.ts
```diff
@@ -40,8 +40,10 @@ export const feelTheme = (): Monaco.editor.IStandaloneThemeData => {
 export const feelTokensConfig = (): Monaco.languages.IMonarchLanguage => {
   return {
     keywords: [
+      "@",
       "for",
       "in",
+      "instance of",
```
yesamer, line 45: @jomarko I agree to add these two entries as keyboard, but I would consider removing "instance" and "of" from the lists.

yesamer: Reference: 10.3.1.2 Grammar rules

packages/feel-input-component/src/FeelConfigs.ts
yesamer: @jomarko For `instance of` I'd use the same strategy we're using for `if` and `for`, using a parameter. Eg. `["instance of", "instance of $0"]` .
For `@`, I tested and it seems there are no effects. I guess a keyword with a single character is not managed. I suggest removing it, considering it is used in different contexts (eg. `"00:01:00@Etc/UTC"`)

jomarko:
thank you @yesamer . changes pushed

packages/feel-input-component/src/FeelConfigs.ts
```diff
@@ -40,17 +40,17 @@ export const feelTheme = (): Monaco.editor.IStandaloneThemeData => {
 export const feelTokensConfig = (): Monaco.languages.IMonarchLanguage => {
   return {
     keywords: [
+      "@",
```
yesamer: @jomarko please remove it accordingly :)

jomarko: @yesamer done

yesamer (approved):
Thank you @jomarko

PR #1810, opened 2023-06-27
kie-issues#180: On DMN Editor's Boxed Expression Editor, when resizing Context/Invocation entry info cells, minWidths are getting messy

### LINKED ISSUE #180: On DMN Editor's Boxed Expression Editor, when resizing Context/Invocation entry info cells, minWidths are getting messy
### Steps to Reproduce: 
- Create a Context expression with two entries, make those be Context expressions too. So we have 3 contexts in total. One parent and 2 children. [1]
- In [1] you can notice, we increased context entry info column width of the first `Context` child. It is  a blue rectangle in [1]
- As next step, we set context expression as literal for second 'Context' child. It is a blue rectangle in [2]
- Now, keeping focus to Literal/Blue rectangle from [2], we can resize the rectangle. Increasing width works fine. However lets try to decrease the width of the blue rectangle from [2]
- We will spot behavior like [3]. The blue rectangle width can be decreased, but the other siblings or parent do not reflect this width change. We should prevent from decreasing width of blue rectangle in situations like [3].

[1]
Screenshot 2023-06-23 055408

[2]
Screenshot 2023-06-23 055423

[3]
Screenshot 2023-06-23 055434

### Codebase
Problematic place seems to be in `Hooks.tsx`
Screenshot 2023-06-23 112842

Please notice `maxNestedExpressionMinWidth` that is always the same (default) value `210`. If we want to achieve behavior described above, we need to compute it dynamically.

### Observations
image

for ^ we execute this code snippet:
```js
  const { nestedExpressionContainerValue, onColumnResizingWidthChange: onColumnResizingWidthChange2 } =
    useNestedExpressionContainerWithNestedExpressions(
      useMemo(() => {
        contextExpression.contextEntries.forEach(e=> console.debug(e));
        return {
          nestedExpressions: [
            ...contextExpression.contextEntries.map((e) => e.entryExpression),
            contextExpression.result,
          ],
          fixedColumnActualWidth: entryInfoWidth,
          fixedColumnResizingWidth: entryInfoResizingWidth,
          fixedColumnMinWidth: CONTEXT_ENTRY_INFO_MIN_WIDTH,
          nestedExpressionMinWidth: CONTEXT_ENTRY_EXPRESSION_MIN_WIDTH,
          extraWidth: CONTEXT_EXPRESSION_EXTRA_WIDTH,
          expression: contextExpression,
          flexibleColumnIndex: 2,
        };
      }, [contextExpression, entryInfoResizingWidth, entryInfoWidth])
    );
```
Please notice there `nestedExpressionMinWidth: CONTEXT_ENTRY_EXPRESSION_MIN_WIDTH,` we set always to a constant value, what is semantically not correct, in the picture above, we shed set `nestedExpressionMinWidth` to a  sum of:
`sub entry 1 fixed column width` +  `sub sub entry 1 fixed column width` + `specific expression flexible column minimal allowed width`

in the picture above

jomarko:
Closes kiegroup/kie-issues#180

Screenshot 2023-07-11 154657

So we update the logic for minimal expression width, `getExpressionTotalMinimalWidth`, to return a value highlighted with red color in the attached picture. The computed value is a sum of:
- `A` `entry info` width 
- `CC` `entry info` width
- `BBB` `entry info` width
- `default minimal width` of literal expression, as literal expression is actual expression of `BBB`

### DIFF
--- a/packages/boxed-expression-component/src/expressions/ContextExpression/ContextExpression.tsx
+++ b/packages/boxed-expression-component/src/expressions/ContextExpression/ContextExpression.tsx
@@ -50,7 +50,7 @@ import { ContextEntryExpressionCell } from "./ContextEntryExpressionCell";
 import { ContextEntryInfoCell } from "./ContextEntryInfoCell";
 import "./ContextExpression.css";
 import { ContextResultExpressionCell } from "./ContextResultExpressionCell";
-import { getExpressionTotalMinimalWidth } from "../../resizing/WidthMaths";
+import { getExpressionTotalMinWidth } from "../../resizing/WidthMaths";
 
 const CONTEXT_ENTRY_DEFAULT_DATA_TYPE = DmnBuiltInDataType.Undefined;
 
@@ -95,11 +95,11 @@ export function ContextExpression(contextExpression: ContextExpressionDefinition
     useNestedExpressionContainerWithNestedExpressions(
       useMemo(() => {
         const entriesWidths = contextExpression.contextEntries.map((e) =>
-          getExpressionTotalMinimalWidth(0, e.entryExpression)
+          getExpressionTotalMinWidth(0, e.entryExpression)
         );
-        const resultWidth = getExpressionTotalMinimalWidth(0, contextExpression.result);
+        const resultWidth = getExpressionTotalMinWidth(0, contextExpression.result);
 
-        const biggestWidth = Math.max(...entriesWidths, resultWidth);
+        const maxNestedExpressionMinWidth = Math.max(...entriesWidths, resultWidth, CONTEXT_ENTRY_EXPRESSION_MIN_WIDTH);
 
         return {
           nestedExpressions: [
@@ -109,12 +109,12 @@ export function ContextExpression(contextExpression: ContextExpressionDefinition
           fixedColumnActualWidth: entryInfoWidth,
           fixedColumnResizingWidth: entryInfoResizingWidth,
           fixedColumnMinWidth: CONTEXT_ENTRY_INFO_MIN_WIDTH,
-          nestedExpressionMinWidth: biggestWidth > 0 ? biggestWidth : CONTEXT_ENTRY_EXPRESSION_MIN_WIDTH,
+          nestedExpressionMinWidth: maxNestedExpressionMinWidth,
           extraWidth: CONTEXT_EXPRESSION_EXTRA_WIDTH,
           expression: contextExpression,
           flexibleColumnIndex: 2,
         };
-      }, [contextExpression, entryInfoResizingWidth, entryInfoWidth, getExpressionTotalMinimalWidth])
+      }, [contextExpression, entryInfoResizingWidth, entryInfoWidth])
     );
 
   /// //////////////////////////////////////////////////////
--- a/packages/boxed-expression-component/src/expressions/ContextExpression/ContextExpression.tsx
+++ b/packages/boxed-expression-component/src/expressions/ContextExpression/ContextExpression.tsx
@@ -50,6 +50,7 @@ import { ContextEntryExpressionCell } from "./ContextEntryExpressionCell";
 import { ContextEntryInfoCell } from "./ContextEntryInfoCell";
 import "./ContextExpression.css";
 import { ContextResultExpressionCell } from "./ContextResultExpressionCell";
+import { getExpressionTotalMinimalWidth } from "../../resizing/WidthMaths";
 
 const CONTEXT_ENTRY_DEFAULT_DATA_TYPE = DmnBuiltInDataType.Undefined;
 
@@ -93,6 +94,13 @@ export function ContextExpression(contextExpression: ContextExpressionDefinition
   const { nestedExpressionContainerValue, onColumnResizingWidthChange: onColumnResizingWidthChange2 } =
     useNestedExpressionContainerWithNestedExpressions(
       useMemo(() => {
+        const entriesWidths = contextExpression.contextEntries.map((e) =>
+          getExpressionTotalMinimalWidth(0, e.entryExpression)
+        );
+        const resultWidth = getExpressionTotalMinimalWidth(0, contextExpression.result);
+
+        const biggestWidth = Math.max(...entriesWidths, resultWidth);
+
         return {
           nestedExpressions: [
             ...contextExpression.contextEntries.map((e) => e.entryExpression),
@@ -101,12 +109,12 @@ export function ContextExpression(contextExpression: ContextExpressionDefinition
           fixedColumnActualWidth: entryInfoWidth,
           fixedColumnResizingWidth: entryInfoResizingWidth,
           fixedColumnMinWidth: CONTEXT_ENTRY_INFO_MIN_WIDTH,
-          nestedExpressionMinWidth: CONTEXT_ENTRY_EXPRESSION_MIN_WIDTH,
+          nestedExpressionMinWidth: biggestWidth > 0 ? biggestWidth : CONTEXT_ENTRY_EXPRESSION_MIN_WIDTH,
           extraWidth: CONTEXT_EXPRESSION_EXTRA_WIDTH,
           expression: contextExpression,
           flexibleColumnIndex: 2,
         };
-      }, [contextExpression, entryInfoResizingWidth, entryInfoWidth])
+      }, [contextExpression, entryInfoResizingWidth, entryInfoWidth, getExpressionTotalMinimalWidth])
     );
 
   /// //////////////////////////////////////////////////////
--- a/packages/boxed-expression-component/src/expressions/InvocationExpression/InvocationExpression.tsx
+++ b/packages/boxed-expression-component/src/expressions/InvocationExpression/InvocationExpression.tsx
@@ -46,6 +46,7 @@ import { ArgumentEntryExpressionCell } from "./ArgumentEntryExpressionCell";
 import { ContextEntryInfoCell } from "../ContextExpression";
 import "./InvocationExpression.css";
 import { DEFAULT_EXPRESSION_NAME } from "../ExpressionDefinitionHeaderMenu";
+import { getExpressionTotalMinimalWidth } from "../../resizing/WidthMaths";
 
 type ROWTYPE = ContextExpressionDefinitionEntry;
 
@@ -81,17 +82,23 @@ export function InvocationExpression(invocationExpression: InvocationExpressionD
   const { nestedExpressionContainerValue, onColumnResizingWidthChange: onColumnResizingWidthChange2 } =
     useNestedExpressionContainerWithNestedExpressions(
       useMemo(() => {
+        const entriesWidths = invocationExpression.bindingEntries.map((e) =>
+          getExpressionTotalMinimalWidth(0, e.entryExpression)
+        );
+
+        const biggestWidth = Math.max(...entriesWidths);
+
         return {
           nestedExpressions: invocationExpression.bindingEntries?.map((e) => e.entryExpression) ?? [],
           fixedColumnActualWidth: parametersWidth,
           fixedColumnResizingWidth: parametersResizingWidth,
           fixedColumnMinWidth: INVOCATION_PARAMETER_MIN_WIDTH,
-          nestedExpressionMinWidth: INVOCATION_ARGUMENT_EXPRESSION_MIN_WIDTH,
+          nestedExpressionMinWidth: biggestWidth > 0 ? biggestWidth : INVOCATION_ARGUMENT_EXPRESSION_MIN_WIDTH,
           extraWidth: INVOCATION_EXTRA_WIDTH,
           expression: invocationExpression,
           flexibleColumnIndex: 2,
         };
-      }, [parametersWidth, parametersResizingWidth, invocationExpression])
+      }, [parametersWidth, parametersResizingWidth, invocationExpression, getExpressionTotalMinimalWidth])
     );
 
   /// //////////////////////////////////////////////////////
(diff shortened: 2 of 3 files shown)


packages/boxed-expression-component/src/expressions/InvocationExpression/InvocationExpression.tsx
```diff
@@ -79,20 +82,47 @@ export function InvocationExpression(invocationExpression: InvocationExpressionD
...
+        depths.push(depth);
+        i.bindingEntries.forEach((e) => computeNestedDepths(depths, depth, e.entryExpression));
+      } else {
+        // it is an expression without entryInfoWidth
+        const depth = currentDepth + getExpressionMinWidth(expression);
+        depths.push(depth);
+      }
+    },
+    []
+  );
```
tiagobento: 1. "depths" are not something that makes sense.. I think you meant "widths"?

2. Creating an empty array inside a `useMemo` and mutating it on a function that gets called inside another memo is not how React is intended to be used.

3. It seems to me that you don't need to pass `allDepths` as argument to the `computeNestedDepths` function, you can create a new array every time.

4. Also, it's not ok to cast an ExpressionDefinition to whatever specialized type you need without looking at its `logicType`. ExpressionDefinition is a union type that has many different specializations discriminated by the common `logicType` property.

5. This code it 100% duplicated on this PR too.

6. It looks like this will only work for some types of expressions, but remember that ALL expressions can be nested, so this solution is incomplete.

Please consider this points when reworking the proposed solution.

tiagobento (changes requested):
@jomarko Thanks for the PR!

I left a comment inline, please do read it carefully and come back if you have any questions. Please consider each bullet point I added in the comment.

Also, it might help to put your solution in words, on the PR description or in the code itself, as a comment. It will help not only you visualize exactly what you're doing, but also other that come later to understand what you did.

packages/boxed-expression-component/src/expressions/InvocationExpression/InvocationExpression.tsx
jomarko: thank you for a thorough review @tiagobento I will address the points the best as I can

jomarko: 1. `depth` vs `width`, sure no problem, I think it was just my non native speaker problem, when we solve a situation with nested expressions, that is why I used `depth`, I will use `width`

jomarko: 2., 3., 4., and 5. - I refactored code, I extracted `computeNestedDepths`, now it is `getExpressionTotalWidth` into `WidthMaths` util file.

[edited]
Also, the latest `getExpressionTotalWidth` is simplified, no more using arrays, but reducing context and invocation `entries` to maximal value using `reduce`.

jomarko: 6. I agree all expression can be nested, however it seems, if you check last `else` branch of `computeNestedWidths`, that existing `getExpressionMinWidth` works fine for expressions, that do not have two columns (`entry info` and `entry expression`)

packages/boxed-expression-component/src/expressions/InvocationExpression/InvocationExpression.tsx
```diff
@@ -81,17 +82,23 @@ export function InvocationExpression(invocationExpression: InvocationExpressionD
   const { nestedExpressionContainerValue, onColumnResizingWidthChange: onColumnResizingWidthChange2 } =
     useNestedExpressionContainerWithNestedExpressions(
       useMemo(() => {
+        const entriesWidths = invocationExpression.bindingEntries.map((e) =>
+          getExpressionTotalMinimalWidth(0, e.entryExpression)
+        );
+
+        const biggestWidth = Math.max(...entriesWidths);
+
```
tiagobento: You can add `INVOCATION_ARGUMENT_EXPRESSION_MIN_WIDTH` to the Math.max operation as well, as the argument expression for an invocation cannot, under any circumstance, be less than it.  Then you can remove the ternary on line 96 and use "biggestWidth" directly.

"biggestWidth" also doesn't seem like a good name to be used on a `nestedExpressionMinWidth` property :) I would go with `maxNestedExpressionWidth`

packages/boxed-expression-component/src/expressions/ContextExpression/ContextExpression.tsx
```diff
@@ -93,6 +94,13 @@ export function ContextExpression(contextExpression: ContextExpressionDefinition
   const { nestedExpressionContainerValue, onColumnResizingWidthChange: onColumnResizingWidthChange2 } =
     useNestedExpressionContainerWithNestedExpressions(
       useMemo(() => {
+        const entriesWidths = contextExpression.contextEntries.map((e) =>
+          getExpressionTotalMinimalWidth(0, e.entryExpression)
+        );
+        const resultWidth = getExpressionTotalMinimalWidth(0, contextExpression.result);
+
+        const biggestWidth = Math.max(...entriesWidths, resultWidth);
+
```
tiagobento: Same comment applies here.

packages/boxed-expression-component/src/resizing/WidthMaths.ts
```diff
@@ -108,6 +113,34 @@ export function getExpressionMinWidth(expression?: ExpressionDefinition): number
...
 }
 
+/**
+ * This function gous recursively trough all `expression` nested expressions and sum either `entryInfoWidth` or default minimal width, returned by `getExpressionMinWidth`, if it is last nested expression in the chain.
+ * This function returns maximal sum found bewteen all `expression` nested expressions
+ */
+export function getExpressionTotalMinimalWidth(currentWidth: number, expression: ExpressionDefinition): number {
+  if (expression.logicType === ExpressionDefinitionLogicType.Context) {
+    // it is a Context, we need to go trough entries and result row
+    const c = expression as ContextExpressionDefinition;
```
tiagobento: You don't need this line. When you check for the `logicType`, you TypeScript will automatically understand what type it has. You can use `expression` directly.

packages/boxed-expression-component/src/resizing/WidthMaths.ts
```diff
@@ -108,6 +113,34 @@ export function getExpressionMinWidth(expression?: ExpressionDefinition): number
...
+    const width = currentWidth + (c.entryInfoWidth ?? 0);
+    const contextEntriesMaxWidth = c.contextEntries.reduce((maxWidth, currentExpression) => {
+      return Math.max(maxWidth, getExpressionTotalMinimalWidth(width, currentExpression.entryExpression));
+    }, width);
+    const resultWidth = getExpressionTotalMinimalWidth(width, c.result);
+    return Math.max(contextEntriesMaxWidth, resultWidth);
+  } else if (expression.logicType === ExpressionDefinitionLogicType.Invocation) {
+    // it is an Invocation, we need to go trough entries
+    const i = expression as InvocationExpressionDefinition;
```
tiagobento: Same here.

packages/boxed-expression-component/src/resizing/WidthMaths.ts
```diff
@@ -108,6 +113,34 @@ export function getExpressionMinWidth(expression?: ExpressionDefinition): number
...
+    const width = currentWidth + (i.entryInfoWidth ?? 0);
+    return i.bindingEntries.reduce((maxWidth, currentExpression) => {
+      return Math.max(maxWidth, getExpressionTotalMinimalWidth(width, currentExpression.entryExpression));
+    }, width);
+  } else {
+    // it is an expression without entryInfoWidth
+    const width = currentWidth + getExpressionMinWidth(expression);
```
tiagobento: You can inline this variable.

packages/boxed-expression-component/src/expressions/InvocationExpression/InvocationExpression.tsx
```diff
@@ -81,17 +82,23 @@ export function InvocationExpression(invocationExpression: InvocationExpressionD
...
           fixedColumnResizingWidth: parametersResizingWidth,
           fixedColumnMinWidth: INVOCATION_PARAMETER_MIN_WIDTH,
-          nestedExpressionMinWidth: INVOCATION_ARGUMENT_EXPRESSION_MIN_WIDTH,
+          nestedExpressionMinWidth: biggestWidth > 0 ? biggestWidth : INVOCATION_ARGUMENT_EXPRESSION_MIN_WIDTH,
           extraWidth: INVOCATION_EXTRA_WIDTH,
           expression: invocationExpression,
           flexibleColumnIndex: 2,
         };
-      }, [parametersWidth, parametersResizingWidth, invocationExpression])
+      }, [parametersWidth, parametersResizingWidth, invocationExpression, getExpressionTotalMinimalWidth])
```
tiagobento: As `getExpressionTotalMinimalWidth` it not a `useCallback` anymore, you can safely remove it from the dependencies list. Global functions are immutable references.

packages/boxed-expression-component/src/expressions/ContextExpression/ContextExpression.tsx
```diff
@@ -101,12 +109,12 @@ export function ContextExpression(contextExpression: ContextExpressionDefinition
...
           fixedColumnResizingWidth: entryInfoResizingWidth,
           fixedColumnMinWidth: CONTEXT_ENTRY_INFO_MIN_WIDTH,
-          nestedExpressionMinWidth: CONTEXT_ENTRY_EXPRESSION_MIN_WIDTH,
+          nestedExpressionMinWidth: biggestWidth > 0 ? biggestWidth : CONTEXT_ENTRY_EXPRESSION_MIN_WIDTH,
           extraWidth: CONTEXT_EXPRESSION_EXTRA_WIDTH,
           expression: contextExpression,
           flexibleColumnIndex: 2,
         };
-      }, [contextExpression, entryInfoResizingWidth, entryInfoWidth])
+      }, [contextExpression, entryInfoResizingWidth, entryInfoWidth, getExpressionTotalMinimalWidth])
```
tiagobento: As getExpressionTotalMinimalWidth it not a useCallback anymore, you can safely remove it from the dependencies list. Global functions are immutable references.

packages/boxed-expression-component/src/resizing/WidthMaths.ts
```diff
@@ -108,6 +113,34 @@ export function getExpressionMinWidth(expression?: ExpressionDefinition): number
   }
 }
 
+/**
+ * This function gous recursively trough all `expression` nested expressions and sum either `entryInfoWidth` or default minimal width, returned by `getExpressionMinWidth`, if it is last nested expression in the chain.
+ * This function returns maximal sum found bewteen all `expression` nested expressions
+ */
+export function getExpressionTotalMinimalWidth(currentWidth: number, expression: ExpressionDefinition): number {
```
tiagobento: To keep consistency with the rest of the code, I would call this function `getExpressionTotalMinWidth`.

packages/boxed-expression-component/src/resizing/WidthMaths.ts
```diff
@@ -108,6 +113,34 @@ export function getExpressionMinWidth(expression?: ExpressionDefinition): number
   }
 }
 
+/**
+ * This function gous recursively trough all `expression` nested expressions and sum either `entryInfoWidth` or default minimal width, returned by `getExpressionMinWidth`, if it is last nested expression in the chain.
+ * This function returns maximal sum found bewteen all `expression` nested expressions
+ */
```
tiagobento:
```suggestion
/**
 * This function goes recursively through all `expression`'s nested expressions and sums either `entryInfoWidth` or default minimal width, returned by `getExpressionMinWidth`, if it is the last nested expression in the chain.
 *
 * This function returns maximal sum found in all `expression`'s nested expressions.
 */
```

tiagobento (changes requested):
Alright, we're getting there. I tested it locally and it works well, with the exception of one scenario. On an expression with other expressions (so not a single context expression), if we expand and shrink the same context entry info cell, over and over, we'll get an ever-expanding table. Take a look:

https://github.com/kiegroup/kie-tools/assets/1584568/dc473a5b-fd9a-4595-8880-7ca38d077da9

Also, please take a look at the comments I left inline.

Overall, I think the biggest issue is the terminology and excessive use of intermediate variables. I always say that if it's hard to name something, try to come up with a way to not have to name it at all :) Inlining variables helps a lot!

packages/boxed-expression-component/src/expressions/InvocationExpression/InvocationExpression.tsx
jomarko: good suggestions, thank you

packages/boxed-expression-component/src/expressions/ContextExpression/ContextExpression.tsx
jomarko: sure, will do

packages/boxed-expression-component/src/resizing/WidthMaths.ts
jomarko: oh, nice, thank you for suggestion

packages/boxed-expression-component/src/resizing/WidthMaths.ts
jomarko: perfect, will do a removal

packages/boxed-expression-component/src/resizing/WidthMaths.ts
jomarko: oki

packages/boxed-expression-component/src/resizing/WidthMaths.ts
jomarko: sure, no problem, will do

packages/boxed-expression-component/src/resizing/WidthMaths.ts
jomarko: good, thank you

tiagobento (approved):

PR #1813, opened 2023-06-27
kie-issues#368: git-cors-proxy-image and kie-sandbox-image don't work properly on macOS on ARM processors

### LINKED ISSUE #368: git-cors-proxy-image and kie-sandbox-image don't work properly on macOS on ARM processors
On MacOS devices running the ARM processors (M1 and M2), docker fails to virtualize some required parts to run these images.

Depending on the status of the `Use Rosetta for x86/amd64 emulation on Apple Silicon` setting these images will interchangeably work.

Image

If it's **ON**: `kie-sandbox-image` works with no problems. `git-cors-proxy-image` will start with no error messages, but the service won't be available (no response from the server).
If it's **OFF**: `kie-sandbox-image` container gets killed when it tries to run `httpd -d FOREGROUND` (may be related to some permissions issues on ubi9 based images, need to research). `git-cors-proxy-image` starts and runs with no problems.

thiagoelg:
Closes https://github.com/kiegroup/kie-issues/issues/368

- Changed how NodeJS is installed in the `git-cors-proxy-image` Containerfile, using rpm repositories from Nodesource (https://github.com/nodesource/distributions#installation-instructions-1);
- Added new `Mutex posixsem` config to `httpd` on `kie-sandbox-image` to fix a broken behavior when Docker is running without the `Use Rosetta for x86/amd64 emulation on Apple Silicon` config set (https://stackoverflow.com/questions/68365844/docker-apache-in-apple-silicon-m1 and https://httpd.apache.org/docs/2.4/mod/core.html#mutex);

Unfortunately `git-cors-proxy-image` won't work on Apple Silicon if the `Use Rosetta for x86/amd64 emulation on Apple Silicon` config is set to ON, as it's in beta and there are several bugs reported on this issue: https://github.com/docker/roadmap/issues/384#issuecomment-1455786117

It's recommended that this config is set to OFF whenever building kie-tools images, as all of them will work without it.

## Also in this PR
- Cross-build extended-service-headless on all platforms
- Added docker to macOS CI build
- Stop attempting to build container images on the Windows CI (it doesn't have support for it due to nested virtualization)
- Replaced appindicator lib on extended services (the old one was deprecated, now using libayatana-appindicator)
- Added cross-env to extended services Makefile (I know this is not ideal, but it was a better way to make sure Windows builds are setting env vars correctly)
- Updated extended-services systray lib version
- Start using docker by default on the following images: `kie-sandbox-image`, `kie-sandbox-extended-services-image`, and `git-cors-proxy-image`
- Created a new `image-builder` package to help build images on our repo
- Updated the `run-script-if` package to support multiple conditions (with `and` or `or`)
- Refactored `run-script-if` to use Typescript
- Added `kie-sandbox-distribution` automated test (simple, to check that all services and up and available)
- Small refactor to `build-env` to make some supporting functions available as a lib

___

Obs.: Since our images now build with docker by default, it's expected that the `push-to-registry` action can pull them from docker to podman so that it can push them to the registry (https://github.com/redhat-actions/push-to-registry/tree/main#note-about-images-built-with-docker)


tiagobento:
Nice! Would it be possible to add some sort of test to `kie-sandbox-distributions`? Spinning up all the containers and doing a fetch or something like that. Very simple...

caponetto (approved):

thiagoelg:
> Nice! Would it be possible to add some sort of test to `kie-sandbox-distributions`? Spinning up all the containers and doing a fetch or something like that. Very simple...

Working on it. Do we need it to run on the CI? If so, I think this should be another issue, because that would introduce some complications (i.e. building the images before testing).

thiagoelg:
@caponetto I haven't updated the SWF container images to use the new `image-builder` package because I didn't want to break any builds, but let me know if you would like to.

.github/actions/setup-env/action.yml
```diff
@@ -50,6 +50,11 @@ runs:
       with:
         maven-version: 3.8.6
 
+    - name: "Set up Docker"
+      uses: docker-practice/actions-setup-docker@v1
```
caponetto: Please consider using the commit id instead of the tag.

.github/workflows/daily_dev_publish.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
     env:
       KIE_SANDBOX_EXTENDED_SERVICES_VERSION: "0.0.0"
-      KIE_TOOLS_BUILD__buildContainerImages: "true"
+      KIE_TOOLS_BUILD__buildContainerImages: ${{ runner.os != 'Windows' }}
```
caponetto:
```suggestion
      KIE_TOOLS_BUILD__buildContainerImages: ${{ runner.os == 'Linux' }}
```

Since this workflow does not run tests, maybe build the container images only on linux?

caponetto:
> @caponetto I haven't updated the SWF container images to use the new `image-builder` package because I didn't want to break any builds, but let me know if you would like to.

Thanks for the heads up, @thiagoelg! I think for now we can keep those images as they are.

.github/workflows/daily_dev_publish.yml
thiagoelg: Good idea!

thiagoelg:
The CI is failing in different places on every run, at least on Ubuntu. On macOS it's failing while building `@kie-tools/serverless-logic-web-tools-swf-dev-mode-image`, apparently due to the machine running out of memory.
Investigating...

caponetto:
The last failed build will be probably fixed after https://github.com/kiegroup/kie-tools/pull/1837

.github/workflows/release_build.yml
```diff
@@ -49,7 +49,7 @@ env:
   KIE_TOOLS_BUILD__runLinters: "false"
   KIE_TOOLS_BUILD__runTests: "false"
   KIE_TOOLS_BUILD__runIntegrationTests: "false"
-  KIE_TOOLS_BUILD__buildContainerImages: "true"
+  KIE_TOOLS_BUILD__buildContainerImages: ${{ runner.os == 'Linux' }}
```
caponetto: Just noticed this https://github.com/kiegroup/kie-tools/actions/runs/5469520956.
At this point, we don't have this `runner` var.
I think we can keep "true" since the image builders run on ubuntu.

.github/workflows/daily_dev_publish.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
     env:
       KIE_SANDBOX_EXTENDED_SERVICES_VERSION: "0.0.0"
-      KIE_TOOLS_BUILD__buildContainerImages: "true"
+      KIE_TOOLS_BUILD__buildContainerImages: ${{ runner.os == 'Linux' }}
```
caponetto: https://github.com/kiegroup/kie-tools/actions/runs/5469520962
Looks like `runner` is not visible here?

.github/workflows/release_build.yml
thiagoelg: Nice catch!

caponetto:
Hopefully, the CI on Ubuntu will pass after https://github.com/kiegroup/kie-tools/pull/1842

PR #1820, opened 2023-06-30
NO-ISSUE: [kn-workflow-plugin] Use new client-pkg for plugin inline

dsimansk:
This PR add a dependency on `knative.dev/client-pkg` rather than directly client repository. The client-pkg was introduced to break potential cyclic deps etc. As an oversight on my end I haven't realized all other plugins are already on the package except `workflow` one.

Given that this's a monorepo of multiple tools. For the go depedency it's fine to consume it through commit `sha`. IMO there's no need to have a new patch tag. 

/cc @ederign


dsimansk:
Maybe we will need to go the other way around and revert other plugins. Please don't merge it yet.

dsimansk:
There's no immediate need for this PR to land wrt/ next Serverless release. We got more elegant solution in place for now. But IMO it should be revisited for the next cycle probably. We can have a bit of structured (what?why? etc.) conversation beforehand as well. 

I'll close the PR to avoid any confusion. :)

PR #1829, opened 2023-07-03
KOGITO-9423: Searching inside directories uses file extensions

fantonangeli:
https://issues.redhat.com/browse/KOGITO-9423

 **Steps to Reproduce:**
-  Create a directory with two SWF file: aaa.sw.json and bbb.sw.json.
-  Open directory in Recent Models.
-  Try to search "sw.json"
-  Current result: Both files are displayed.
-  Expected result: Search result should be empty.

**Description:**
If you search inside Recent models, only names of the files are used.
If you search inside a directory, file extensions are also used for the result.

**Note:**
Together with this fix, I changed the search methodology to use the same filtering library used in `Sample Catalog`, which is fusejs.
This applies to `Recent Models` and `Workspace Files`.

[KOGITO-9423.webm](https://github.com/kiegroup/kie-tools/assets/17780574/1ebbd999-8ff9-466b-913d-db76b68347a0)


packages/serverless-logic-web-tools/src/homepage/recentModels/RecentModels.tsx
```diff
@@ -32,7 +32,7 @@ import { ConfirmDeleteModal, defaultPerPageOptions, TablePagination, TableToolba
 import { WorkspacesTable, WorkspacesTableRowData } from "./WorkspacesTable";
 import { Link } from "react-router-dom";
 import { routes } from "../../navigation/Routes";
-import { escapeRegExp } from "../../regex";
```
caponetto, line 35: Since we don't need this regex file anymore, can you please remove it?

caponetto (approved):

tomasdavidorg (approved):
Looks good to me 🙂. Thanks.

PR #1830, opened 2023-07-04
KOGITO-9299: Cover kn-workflow-plugin CLI with integration tests II.

zdrapela:
**Jira:** https://issues.redhat.com/browse/KOGITO-9299

This PR incorporates new integration tests for `kn-workflow-plugin` and improvements for integration tests.

Tests for the following commands are contained:
- `run`
- `quarkus run` (this test is now disabled because of failures when using this command)
- `quarkus convert`

Tests now use `stretchr/testify/require` instead of `stretchr/testify/assert` for asserts. The difference is, that if the condition in require is not met, the single test run will fail immediately and will not continue with the test. According to this change, deferred clean-up is used to ensure clean-up if the test fails.

There are now two now env. variables (I am open to better name or other env. variabel related suggestions):
- `KN_PLUGIN_WORKFLOW__testPrintCmdOutput` - if set to `true`, output of all the commands executed during the test will by printed out to console.
- `KN_PLUGIN_WORKFLOW__suppressBrowserWindow` - if set to `true`, the `run` and `quarkus run` commands and the test that executes them won't open the project URL when ready. This is primarily used by the integration test to avoid popping out the browser window.

The timeout to run IT is increased to 20 min. to give all the test time to execute (typical integration test set run time is now around 10 min.).


ederign:
I'll take a look at this early next week.

ederign (approved):

ederign:
@zdrapela, when you are back from PTO, let's discuss the 'quarkus run (this test is now disabled because of failures when using this command)' issue! Probably this was related to an issue of the runtimes.

With KOGITO-9449  Add Kogito BOM automatically for generated and converted projects, we will have a more stable environment with fixed versions.

PR #1843, opened 2023-07-10
KOGITO-9248: Show create and import cards in RecentModels if there are no models

fantonangeli:
https://issues.redhat.com/browse/KOGITO-9248

**Description**
If you go to Recent Models and it is empty there is no way to create or redirect to a place where you can create model, import sample or use a model from the catalog.

KOGITO-9248

### DIFF
--- a/packages/serverless-logic-web-tools/src/homepage/overView/CreateOrImportModelGrid.tsx
+++ b/packages/serverless-logic-web-tools/src/homepage/overView/CreateOrImportModelGrid.tsx
@@ -16,53 +16,21 @@
 import React from "react";
 import { FileTypes } from "@kie-tools-core/workspaces-git-fs/dist/constants/ExtensionHelper";
 import { QuickStartContext, QuickStartContextValues } from "@patternfly/quickstarts";
-import { Card, CardBody } from "@patternfly/react-core/dist/js/components/Card";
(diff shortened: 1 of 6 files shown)


packages/serverless-logic-web-tools/src/homepage/overView/CreateOrImportModelGrid.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { QueryParams } from "../../navigation/Routes";
+import { useQueryParam, useQueryParams } from "../../queryParams/QueryParamsContext";
+import { ImportFromUrlCard } from "./ImportFromUrlCard";
+import { NewModelCard } from "./NewModelCard";
+import { UploadCard } from "./UploadCard";
+
+export function CreateOrImportModelGrid(props: { isNavOpen: boolean }) {
+  const routes = useRoutes();
+  const history = useHistory();
+  const expandedWorkspaceId = useQueryParam(QueryParams.EXPAND);
```
caponetto: I think we don't use this query param anymore, right?
Can you please clean up everything associated with this param?

caponetto:
We've got a conflict :)

fantonangeli:
I will reopen a new PR with the "New file" button.

PR #1850, opened 2023-07-12
kie-issues#382: [DMN Editor] Superfluous newline in multi-cell copy/paste

### LINKED ISSUE #382: [DMN Editor] Superfluous newline in multi-cell copy/paste
The new multi-cell copy/paste functionality is a great feature. However, when multiple cells in the same column are involved, a newline is added at the end of each cell (except for the cells in the first column and except for the last one in the column). I don't think this is semantically wrong, but the visual layout gets messed up. 
superfluous newline

It can be seen in the textual representation of the .dmn file as well:
image

### Investigation
During a copy operation, we store a string value into clipboard
Screenshot 2023-07-12 110904

During a paste operation is however the clipboard value changed somehow
image

jomarko:
Closes: https://github.com/kiegroup/kie-issues/issues/382

The web Clipboard API we use for copy and paste feature in boxed expression editor, doesn't allow to adjust formatting/encoding of strings, see:
- https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText
- https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

For some reason, even we write text into clipboard containing only `\n` new line characters, it is changed to `\r\n` during reading text from the clipboard.

### Writing to the clipboard / copy operation
Screenshot 2023-07-12 110904

### Reading from the clipboard / paste operation
Screenshot 2023-07-12 111950

Please compare `clipboardValue` in the screenshots above. You can see there the difference of `\n` vs `\r\n`

### DIFF
--- a/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx
+++ b/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx
@@ -122,7 +122,8 @@ export enum BeeTableSelectionPosition {
   Right = "right",
 }
 
-const CLIPBOARD_ROW_SEPARATOR = "\n";
+const TEXT_TO_CLIPBOARD_ROW_SEPARATOR = "\n";
+const CLIPBOARD_TO_TEXT_ROW_SEPARATOR = /\r?\n/;
 const CLIPBOARD_COLUMN_SEPARATOR = "\t";
 
 const NEUTRAL_SELECTION = {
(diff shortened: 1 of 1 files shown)


yesamer:
@jomarko This table should explain the reason, it's an OS-dependent behavior

Screenshot 2023-07-12 at 18 04 59

https://en.wikipedia.org/wiki/Newline

jomarko:
Thank you @yesamer . SO what is your opinion?

Should we detect the paltform and use approptiate line separator dynamically? Sth. like below in pseudocode?
```js
CLIPBOARD_TO_TEXT_ROW_SEPARATOR = Platform.isWin() ? \r\n : Platform.isLinux() ? \n : \r;
```

yesamer:
@jomarko Well, your current change should cover correctly the first 2 cases in the table (we can disregard the third, I guess nobody is using DMN Editor on a Commodore 64 :D :D ). In my opinion, we should just test the change in the different OS. Do you have Windows?

jomarko:
@yesamer yes, my current changes should cover first two cases from the table you shared.

I am using Windows, yes.

yesamer:
@jomarko ok, I'll test the change on my mac, that should be enough.

yesamer (approved):
It works. Thank you @jomarko

PR #1851, opened 2023-07-13
NO-ISSUE: Fix occasional no space left on device error during full CI build

paulovmr:


thiagoelg (approved):

caponetto (approved):

tiagobento:
@paulovmr Do we need to partition it in 5 pieces? I mean, wouldn't partitioning it in two be enough for now? :x

paulovmr:
@tiagobento These are not full partitions as we didn't cleared all the files created by each build, just the docker/podman resources. I prefered to do the 5 separate images (actually 4 more, because we already had one separate) to buy some time, but surely we need a more definite and structured solution in the near future.

tiagobento:
Ok, got it! Thanks for fixing it. Let's sync up regarding this topic later.

PR #1857, opened 2023-07-14
KOGITO-9248: Add "New Model" button to RecentModels if there are no models

fantonangeli:
https://issues.redhat.com/browse/KOGITO-9248

**Description**
If you go to Recent Models and it is empty there is no way to create or redirect to a place where you can create model, import sample or use a model from the catalog.

[KOGITO-9248.webm](https://github.com/kiegroup/kie-tools/assets/17780574/6f9d1c84-cc06-4254-aa7e-52e6f58f2ae1)


tomasdavidorg (approved):
Thanks @fantonangeli. Looks good to me 🙂.

caponetto (approved):

PR #1878, opened 2023-07-28
KOGITO-9595 Process SCSS file into CSS in editor package

bxf12315:


tiagobento:
'packages/editor' is a TypeScript library that we make available with tree shaking capabilities. Introducing Webpack to it breaks that principle, and doesn't make sense as tsc is still being called.

By design, the TypeScript libraries on KIE Tools don't use Webpack, and we would need a pretty substantial reason to break that principle.

Please consider an alternative solution or provide more details on why this is necessary.

tiagobento (changes requested):
See comment above.

caponetto:
Hi @tiagobento, great point! Thanks for raising this concern.
I dedicated some time to investigating the issue we were having and saw a problem in the folder structure of the SWF combined editor package. I will send a PR to fix it. 

Anyway, thanks for the investigation @bxf12315! I am closing this PR.

PR #1884, opened 2023-08-01
KOGITO-9399: SWF Combined Editor - Multiplying Architecture integration in Edit Mode

handreyrc:
Hey @romartin, could you please review this PR?

**JIRA**: [KOGITO-9399](https://issues.redhat.com/browse/KOGITO-9399)

Thanks!

### DIFF
--- a/packages/serverless-workflow-combined-editor/src/channel/ServerlessWorkflowDiagramEditorChannelApiImpl.ts
+++ b/packages/serverless-workflow-combined-editor/src/channel/ServerlessWorkflowDiagramEditorChannelApiImpl.ts
@@ -99,4 +99,12 @@ export class ServerlessWorkflowDiagramEditorChannelApiImpl implements Serverless
   public kogitoSwfDiagramEditor__setContentSuccess(): void {
     // Do nothing.
   }
+
+  public kogitoSwfDiagramEditor__onDiagramChanged(args: {
+    newContent: string;
+    nodeName: string;
+    documentUri?: string;
+  }): void {
+    return this.args.textEditorEnvelopeApi?.notifications.kogitoSwfTextEditor__updateContent.send(args);
+  }
 }
--- a/packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramExposedInteropApi.ts
+++ b/packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramExposedInteropApi.ts
@@ -30,4 +30,11 @@ export interface DiagramExposedInteropApi {
    * Notify that the diagram finished setting its content
    */
   setContentSuccess: () => void;
+
+  /**
+   * Notify that the diagram has been edited from Stunner
+   * @param newContent diagram raw content
+   * @param nodeName name of the node to be selected
+   */
+  onDiagramChanged: (newContent: string, nodeName: string) => void;
 }
--- a/packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramService.ts
+++ b/packages/serverless-workflow-diagram-editor-envelope/src/api/DiagramService.ts
@@ -33,4 +33,11 @@ export class DiagramService implements DiagramExposedInteropApi {
   public setContentSuccess() {
     this.envelopeContext.channelApi.notifications.kogitoSwfDiagramEditor__setContentSuccess.send();
   }
+
+  public onDiagramChanged(newContent: string, nodeName: string) {
+    this.envelopeContext.channelApi.notifications.kogitoSwfDiagramEditor__onDiagramChanged.send({
+      newContent,
+      nodeName,
+    });
+  }
 }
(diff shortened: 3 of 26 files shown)


LuboTerifaj:
Hello @handreyrc , could you please resolve conflicts on your branch? Thanks!

romartin (approved):
Hey Handrey!

It looks good to me! Also the "ghost" editor code is clean and readable, such good work! Thanks!

handreyrc:
Hey @LuboTerifaj and @romartin,

I fixed all the rebase issues. We are good again for review.

Thanks!

LuboTerifaj (changes requested):
Hello @handreyrc ,

I have found several issues listed below.

### Common issues

- The order of elements in Text editor is not preserved when editing workflows from diagram.
  - For example: [continuing-as-a-new-execution.sw.json](https://github.com/LuboTerifaj/specification/blob/c2ee1658d1c0aa77db4ee1e855e2b9128daa17a0/examples/resources/continuing-as-a-new-execution.sw.json), [continuing-as-a-new-execution.sw.yaml](https://github.com/LuboTerifaj/specification/blob/c2ee1658d1c0aa77db4ee1e855e2b9128daa17a0/examples/resources/continuing-as-a-new-execution.sw.json)
  - I think we can report this as a separate issue in case it is not easily achievable.

- The issue mentioned above is causing the `name` and `type` of a state is placed below the other state properties when renaming a state.
  - If not the complete order of elements will be fixed, at least this issue could be addressed in this pull-request.
		
### The Toolbox for creating assets issues
#### JSON workflows:
- The states may be placed next to Start node instead of a next level (as a transition) from the original state. Identified cases:
  - If a state contains `"end": true`
  - If you try to create an error transition from any state except switch state.

- New states do not have option to set `"end": true` in diagram.
  - I am okay with addressing it in a separate jira.

- The auto layout always breaks the user layout.
  - I reported it here: https://issues.redhat.com/browse/KOGITO-9719
		
- When you create a new state using Toolbox and select other state. The original one is not deselected.
		
#### Both JSON and YAML workflows:
- It is not possible to delete nodes that have `"end": true` set.
	
#### YAML workflows:	
- Diagram keep freezing after editing workflows from diagram.
  - For example, make an edit in [perform-customer-credit-check.sw.yaml](https://github.com/LuboTerifaj/specification/blob/c2ee1658d1c0aa77db4ee1e855e2b9128daa17a0/examples/resources/perform-customer-credit-check.sw.yaml).
	  
- The states are not created properly using Toolbox.

Because of the issues above, I couldn't have tested YAML workflows properly.

### Other issues
- The diagram is missing scroll bars in Firefox.
  - When diagram grows too much with new states created, it is not possible to change the view, reach the nodes and edit the diagram.
  - I reported the issue here: https://issues.redhat.com/browse/KOGITO-9718

Please let me know which issues can be addressed in this pull-request. For other non-blocker issues I can create separate jiras.

Thank you!

handreyrc:
Hey @LuboTerifaj,

I checked the issues you reported. 

Basically the integration implemented in this PR calls getContent() after a change is performed from the canvas and then forward the new content to refresh the text editor, finally the text editor triggers the reload of the diagram with the new content. We are relying on the marshallers to recover content and we still have issues from the original implementantion and some new after a clean-up. They are hidden because in read-only mode we have only text editor -> diagram comunication. However, in edit mode we have both ways, text editor -> diagram  and diagram -> text editor. 

**Common Issues:**

- The order of elements in Text editor is not preserved when editing workflows from diagram.
> - Part of the problem was introduced by the new marshallers implementation check the jira [KOGITO-8299](https://issues.redhat.com/browse/KOGITO-8299) . Maybe it is time for us to revisit and prioritize this Jira. 
 
**The Toolbox for creating assets issues**
*JSON Workflows:*
- The states may be placed next to Start node instead of a next level (as a transition) from the original state.
> - **Fixed**. The marshallers clean up broke some operations in edit mode. I fixed the ones I found but I missed this one.
- New states do not have option to set "end": true in diagram.
> - There are plans to add "End node" as an option in the toolbox instead. It was not done yet because the end node is a special case, it's just a flag.   
- The auto layout always breaks the user layout.
> - The idea is that the user won't be able to move states freely in edit mode. The auto-layout will always handle node's position automatically. Check the item "2 - UI" in the epic [KOGITO-8457](https://issues.redhat.com/browse/KOGITO-8457). Maybe we should close that Jira or make it a task to disable those behaviors.
- When you create a new state using Toolbox and select other state. The original one is not deselected.
> - IMO it is better to fix this in a new Jira. 

*YAML workflows:*
- Diagram keep freezing after editing workflows from diagram.
> - It seems another issue with the marshallers. This one is more complex but is not related to this PR. In super-dev-mode if you create a node in that diagram, it works, however, if you call getContent() from console and then set the new content back in order to reload the diagram, it breaks. If you agree, I think it is better to address this issue in a new Jira.
- The states are not created properly using Toolbox.
> - Actually they are created propery, the problem is the content reload with the broken YAML recovered from the marshallers. The root cause is the same as previous item. It is part of the same issue.

**Other issues**
- The diagram is missing scroll bars in Firefox.
> - I agree lets fix it in a separate Jira it is not related to this work.

 WYT?

LuboTerifaj (approved):
Hello @handreyrc ,

The most significant issues are the following:

- [KOGITO-8299](https://issues.redhat.com/browse/KOGITO-8299) SWF Editor - JSON property declaration order shouldn't change in the serialization
- [KOGITO-9736](https://issues.redhat.com/browse/KOGITO-9736) Multiplying Architecture integration is broken for YAML files

Other reported issues not reported in my previous [comment](https://github.com/kiegroup/kie-tools/pull/1884#pullrequestreview-1578997216):

- [KOGITO-9734](https://issues.redhat.com/browse/KOGITO-9734) State is not deselected when selecting node after newly created state
- [KOGITO-9735](https://issues.redhat.com/browse/KOGITO-9735) Not possible to delete states with end set to true

Since edit mode is not enabled yet, we can deliver the feature in pieces. Approving the PR.

Thanks!

LuboTerifaj:
Hello @handreyrc , do you plan to continue working on this feature, or can we close the PR? Thanks!

tiagobento:
Closing due to inactivity and conflicts. Let me know if further development will come to this PR. Thanks!

PR #1905, opened 2023-08-07
kie-issues#398: Implement flag to allow self-signed certificates when using CORS Proxy for Git requests

### LINKED ISSUE #398: Implement flag to allow self-signed certificates when using CORS Proxy for Git requests
Users reported not being able to connect to some Github instances. Upon further investigation, we realized they were Github Enterprise instances with self-signed certificates.

The old Git CORS Proxy and the new CORS Proxy packages don't support self-signed certificates for Git requests right now, the capability is already present for other types of requests.

The goal of this task is to make it possible to configure Git Auth Sessions flagged as insecure (the same way we currently do for OpenShift connections) and use this flag to pass the `insecurely-disable-tls-certificate-validation` header to the CORS Proxy instance, allowing for insecure connections to Git providers with self-signed certificates.

thiagoelg:
Closes https://github.com/kiegroup/kie-issues/issues/398


packages/cors-proxy/src/proxy/server.ts
```diff
@@ -31,6 +31,8 @@ export const startServer = (args: ServerArgs): void => {
 
   const app: express.Express = express();
 
+  app.disable("x-powered-by");
```
tiagobento, line 34: 👍

packages/online-editor/build/defaultEnvJson.ts
```diff
@@ -94,6 +94,19 @@ export const defaultEnvJson: EnvJson = {
...
+      domain: "proxy.testing:9091",
+      supportedGitRemoteDomains: [
+        "proxy.testing:9091",
+        "gist.proxy.testing:9091"
+      ],
+      type: AuthProviderType.github,
+      name: "GitHub at Local",
+      group: AuthProviderGroup.GIT,
+      insecurelyDisableTlsCertificateValidation: true
+    }
```
tiagobento: This needs to be removed, right? In any case, I think it would be good to have a guide somewhere (on a `.md` file) explaning how to configure and use a local Git instance with self-signed certificate, for testing and validation purposes.

packages/online-editor/src/accounts/git/ConnectToGitSection.tsx
```diff
@@ -412,6 +429,10 @@ export const fetchAuthenticatedBitbucketUser = async (
       username: bitbucketUsername,
       password: bitbucketToken,
     },
+    proxyUrl: insecurelyDisableTlsCertificateValidation ? proxyUrl : undefined,
+    headers: {
+      ...(insecurelyDisableTlsCertificateValidation ? INSECURELY_DISABLE_TLS_CERTIFICATE_VALIDATION_HEADERS : {}),
```
tiagobento: Why is `INSECURELY_DISABLE_TLS_CERTIFICATE_VALIDATION_HEADERS` only used here? Shouldn't it be used for GitHub requests too?

packages/online-editor/src/authSessions/AuthSessionsContext.tsx
```diff
@@ -137,18 +143,29 @@ export function AuthSessionsContextProvider(props: PropsWithChildren<{}>) {
         const newAuthSessionStatus: [string, AuthSessionStatus][] = await Promise.all(
           [...(authSessions?.values() ?? [])].map(async (authSession) => {
             if (authSession.type === "git") {
-              const authProvider = authProviders.find(({ id }) => id === authSession.authProviderId);
+              const authProvider = authProviders.find(({ id }) => id === authSession.authProviderId) as GitAuthProvider;
```
tiagobento: Although we know this to be true, there's no hard guarantee on the type system that this will be the case. In this scenario, I prefer to throw an exception if `authProvider` is not a GitAuthProvider. WDYT?

packages/online-editor/src/bitbucket/Hooks.tsx
```diff
@@ -135,8 +138,8 @@ export class BitbucketClient implements BitbucketClientApi {
       body: props.body,
     });
   };
-  getApiUrl = () => {
-    return `https://api.${this.domain}/2.0`;
+  getApiUrl = (proxyUrl?: string) => {
+    return `${proxyUrl ? `${proxyUrl}/` : "https://"}api.${this.domain}/2.0`;
```
tiagobento: This template string with ternary is kind of complicated to read. Can you please give it some extra love? :D

packages/online-editor/src/editor/Toolbar/GitIntegration/GitIntegrationContextProvider.tsx
```diff
@@ -168,21 +175,19 @@ export function GitIntegrationContextProvider(props: GitIntegrationContextProvid
...
       );
     }
 
     if (props.workspace.descriptor.origin.kind === WorkspaceKind.BITBUCKET_SNIPPET) {
       return authSessionsSelectFilterCompatibleWithGistOrSnippetUrlDomain(
-        new URL(props.workspace.descriptor.origin.url).hostname,
+        new URL(props.workspace.descriptor.origin.url).host,
         bitbucketSnippet?.owner?.login
       );
     }
```
tiagobento, line 193: This is to include the port, right?

packages/online-editor/src/editor/Toolbar/GitIntegration/GitIntegrationContextProvider.tsx
```diff
@@ -106,6 +106,13 @@ export function GitIntegrationContextProvider(props: GitIntegrationContextProvid
     useState<OctokitRestEndpointMethodTypes["gists"]["get"]["response"]["data"] | undefined>(undefined);
   const [bitbucketSnippet, setBitbucketSnippet] = useState<any>(undefined);
 
+  const insecurelyDisableTlsCertificateValidation = useMemo(() => {
+    if (authProvider?.group === AuthProviderGroup.GIT) {
+      return authProvider.insecurelyDisableTlsCertificateValidation;
+    }
+    return props.workspace.descriptor.gitInsecurelyDisableTlsCertificateValidation;
+  }, [authProvider, props.workspace]);
```
tiagobento, line 114: So... in summary, we only use the flag saved on the Workspace descriptor if there's no AuthProvider associated with the Workspace? I.e., "Unauthenticated"?

packages/online-editor/src/github/Hooks.tsx
```diff
@@ -18,22 +18,58 @@ import { Octokit } from "@octokit/rest";
...
+
+export function getOctokitClient(args: {
+  githubToken?: string;
+  domain?: string;
+  proxyUrl?: string;
+  insecurelyDisableTlsCertificateValidation?: boolean;
+}) {
+  return new Octokit({
+    auth: args.githubToken,
+    baseUrl: args.domain === "proxy.testing:9091" ? "proxy.testing:9092" : getGithubInstanceApiUrl(args.domain),
```
tiagobento: This feels like something to be removed too, right?

packages/online-editor/src/importFromUrl/ImportFromUrlHomePageCard.tsx
```diff
@@ -42,35 +43,56 @@ export function ImportFromUrlCard() {
...
+  const updateInsecurelyDisableTlsCertificateValidation = useCallback(
+    (newAuthSession: AuthSession) => {
+      if (newAuthSession && newAuthSession.type === "git") {
+        const localAuthProvider = authProviders.find((provider) => provider.id === newAuthSession.authProviderId);
+        if (localAuthProvider && localAuthProvider.group === AuthProviderGroup.GIT) {
+          setInsecurelyDisableTlsCertificateValidation(
+            localAuthProvider.insecurelyDisableTlsCertificateValidation ?? false
+          );
+        }
+      }
```
tiagobento:
```suggestion
      if (newAuthSession?.type === "git") {
        const localAuthProvider = authProviders.find((provider) => provider.id === newAuthSession.authProviderId);
        if (localAuthProvider?.group === AuthProviderGroup.GIT) {
          setInsecurelyDisableTlsCertificateValidation(
            localAuthProvider.insecurelyDisableTlsCertificateValidation ?? false
          );
        }
      }
```

^ Isn't it the same? It looks cleaner to me...

packages/workspaces-git-fs/src/services/GitService.tsx
```diff
@@ -19,6 +19,8 @@ import http from "isomorphic-git/http/web";
 import { GIT_DEFAULT_BRANCH } from "../constants/GitConstants";
 import { KieSandboxWorkspacesFs } from "./KieSandboxWorkspaceFs";
 
+const INSECURELY_DISABLE_TLS_CERTIFICATE_VALIDATION = "insecurely-disable-tls-certificate-validation";
```
tiagobento: Ok, so this is duplicated, and this constant doesn't have a very good name, let's say :D 

Or even more than duplicated, as this name is listed at `cors-proxy` too.

If we think about it, this name is owned by the `cors-proxy` package. Since we don't want `online-editor` and `worksapces-git-fs` to depend on `cors-proxy`, my suggestion is that we create a new package called `cors-proxy-api`, and export a this value from there. A lot for a single package, I know, but also sets a standard for dealing with this kind of "soft" dependency between packages. WDYT?

tiagobento (changes requested):
Wow! This PR is awesome. This was no easy task, and you nailed it. I left some minor comments inline, please consider them. Congratulations for the thoroughness and attention to detail. I bet you're tired of typing  the `insecurelyDisableTlsCertificateValidation` string 😝

packages/online-editor/src/accounts/git/ConnectToGitSection.tsx
thiagoelg: It is but in a different place (when creating the Octokit client). I've updated this to be more concise.

packages/online-editor/src/authSessions/AuthSessionsContext.tsx
thiagoelg: Fixed by adding a new type guard and improving the current type checks.

packages/online-editor/src/editor/Toolbar/GitIntegration/GitIntegrationContextProvider.tsx
thiagoelg: Yes

packages/online-editor/src/editor/Toolbar/GitIntegration/GitIntegrationContextProvider.tsx
thiagoelg: Yes, but now that I'm thinking about it, should this be an "OR" comparison? AuthProvider flag OR Workspace flag?

packages/workspaces-git-fs/src/services/GitService.tsx
thiagoelg: I would go further and create a `cors-proxy-client` package, that would work as a drop-in replacement for `fetch`, what do you think?

packages/online-editor/src/editor/Toolbar/GitIntegration/GitIntegrationContextProvider.tsx
tiagobento: Good question. I think the `or` makes more sense.. But I didn't spend a lot of time thinking about it...

packages/workspaces-git-fs/src/services/GitService.tsx
tiagobento: I'm usually not a big fan of "client" packages that "dictate" how the communication with something should be done. I understand that the idea is to make it easier for consumers, but I tend to feel like sometimes callers need a little more control over how the communication is done... Idk, I'm very careful when creating abstractions, but if you feel like we would benefit from this "client" package, go for it.

tiagobento (approved):

zdrapela:
~~Hi @thiagoelg, 
Thank you for this. Unfortunately, it looks QEMU is broken again as mentioned in https://github.com/lima-vm/lima/issues/1742. I noticed that PR checks started to fail (e.g. https://github.com/kiegroup/kie-tools/pull/1926). Could you please take a look again at how it can be fixed? Thanks in advance!~~

zdrapela:
> Hi @thiagoelg, Thank you for this. Unfortunately, it looks QEMU is broken again as mentioned in [lima-vm/lima#1742](https://github.com/lima-vm/lima/issues/1742). I noticed that PR checks started to fail (e.g. #1926). Could you please take a look again at how it can be fixed? Thanks in advance!

Well, you can ignore my previous comment. I figured out that the third workaround mentioned in https://github.com/lima-vm/lima/issues/1742 works and will be opening PR for the fix soon.

thiagoelg:
Thanks @zdrapela! Feel free to let me know if I can help with anything.

This issue kind of sucks and there's not much we can do about it beyond applying these workarounds and waiting for an official fix. :confused:

PR #1912, opened 2023-08-10
kie-issues#381: GitHub Actions workflow on KIE Tools to verify `update-version-to` script

RishiRajAnand:
fixes [kie-issues/issues/381](https://github.com/kiegroup/kie-issues/issues/381)
- Adds a new script on root package.json that verifies all workspace packages for a given version. An error is thrown if there is any version mismatch.
- It covers all package.jsons, maven configs and manifest files
- Git workflow  `verify_update_version` checks whether `update-version-to` script updates all files

### DIFF
--- a/.github/workflows/verify_update_version.yml
+++ b/.github/workflows/verify_update_version.yml
@@ -1,0 +1,44 @@
+name: "CI :: Verify update version"
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    env:
+      TARGET_VERSION: "1.0.0"
+      DEFAULT_VERSION: "0.0.0"
+    steps:
+      - name: "Checkout kie-tools"
+        uses: actions/checkout@v3
+
+      - name: "Setup environment"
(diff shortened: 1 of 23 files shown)


tiagobento:
The goal here was to test wether or not the `update-version-to` script is running properly. By introducing the "verify_version", we're not only duplicating a lot of logic, but we're introducing another convention to packages that I don't think is valuable. In my point of view, the CI job should be the one doing all the verifications, not the packages themselves. 
My idea was to make simpler checks, that didn't introduce new logic, like parsing a version out of a manifest file etc.

A simple check that we can do it looking for the number of changes introduced in Git when going from 0.0.0 to 1.0.0 and back to 0.0.0. Also, there are other checks that are missing, and that would be exploring with different possibilities. For example, using other features of SemVer, like "pre-releases" and "build metadata". What happens if I enter an invalid SemVer string when doing `update-version-to`? This PR doesn't check for those cases.

The logic to verify the `update-version-to` script should be present on the GitHub Action workflow only, concentrated in one place, not spread across packages, so the `verify_version` script should be removed, as it is not necessary for the build itself.

Please consider my comments...

PR #1916, opened 2023-08-11
KOGITO-9704: [yard Editor] Add diagram to YaRD editor

paulovmr:
Screenshot from 2023-08-11 18-52-08


Rikkola:
I see this PR gets the same eclipse errors as mine, is there a way to know if anyone is looking at them?

Rikkola:
The file uses the word "kind" for decision type/kind and the UI calls it "Type", I am not sure if this is really an issue, but bringing it up for consideration. Also I don't think the YaRD schema is set in stone, so we can change it too.

Rikkola (approved):
Looks good, only question is the type of the file, but that does not block this now.

caponetto (approved):

PR #1921, opened 2023-08-17
KOGITO-9712: Use Patternfly V5 in Dashbuilder

jesuino:
KOGITO-9712: Use Patternfly V5 in Dashbuilder
--

In this big change we remove the old widgets and the old libraries Bootstrap and Patternfly V3 in favor of Patternfly V5. A few components will be impacted by this change:

1) The Table component order and column picker needs to be updated. Users requiring such feature can rely on the table external component.
2) Carousel and Slider components needs a new re-design, so it is not re-implemented, but it is not used in any of the samples Dashboards
3) Navigation buttons in Tabs are missing. This needs to be fixed when GWT UI is completely removed (see [1])

The following supported and maintained Dashboards needs to be adjusted:

1) Samples needs to be revisited, specially cards templates and HTML using PF3 classes. Cards used icons and classes from PF3;
2) Serverless Workflow dashboards will work with this change, but when updating the following points need attention:

2.1) The table can use the internal Table to drop the extension component;
2.2) Cards HTML template need to be updated

Finally the Kitchensink example, website and documentation should be updated. This will be done next.

NOTE: There's a memory leak during content refresh. It does not affect runtime, only during edit. The document[2] contains more details;

[1] https://issues.redhat.com/browse/KOGITO-9715
[2] https://docs.google.com/document/d/1UANE7WEQh3YMDygf3X2JAY4XiSlJ_vc8WY6JWh7bsQk/edit


ederign:
William, thank you so much for this work. 

Guys, before merging this, let's do a sync up on the impact on the full repo + the broad dependencies (i.e. swf dev UI)

tiagobento:
Closing due to inactivity. cc @fantonangeli

PR #1927, opened 2023-08-23
kie-issues#512: Removing copied jQuery UI assets in `stunner-editors` module on KIE Tools

### LINKED ISSUE #512: Removing copied jQuery UI  assets in `stunner-editors` module on KIE Tools
In `uberfire-workbench-client-views-patternfly` a copy of JQueryUI version 1.12.1 resources (`jquery-ui.min.css` and `jquery-ui.min.js` ) are copied and part of the codebase. 

Having a copied external dependency in our repo is not a good design choice, and should be avoided.

Check if these resources are still used. If not used, remove them.

yesamer:
Closes: https://github.com/kiegroup/kie-issues/issues/512

Assets removed:
- `jquery-ui.min.css `
- `jquery-ui.min.js`

The js file is never injected, therefore it can be safely removed.
The css file is injected but it seems not actually used, and that makes sense as its related js is not injected. 
However, I checked the `JQuery*` widgets family, and it seems to me they are using bootstrap (eg. `bootstrap-select`, `bootstrap-datepicker`).

So, I think we should be safe to remove them, we can proceed using an "optimistic" approach, which means removing these assets and, in case of unexpected behavior, we can load them as an external dependency.

JQueryDropdownMenu.java is never used.


packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/DMNClientEntryPoint.java
```diff
@@ -28,6 +28,5 @@ public class DMNClientEntryPoint {
     @PostConstruct
     public void init() {
         PatternFlyBootstrapper.ensureMomentIsAvailable();
-        PatternFlyBootstrapper.ensureMonacoEditorLoaderIsAvailable();
```
yesamer, line 31: Not strictly related to the original ticket, I noticed that DMN Editor try to inject Monaco Editor resources even if it doesn't hold anymore the monaco editor internally (now it is loaded in the new BEE), so we can remove it.

tiagobento (approved):

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/DMNClientEntryPoint.java
ederign: What is BEE?

ederign (approved):

packages/stunner-editors/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/DMNClientEntryPoint.java
yesamer: @ederign Boxed Expression Editor

PR #1932, opened 2023-08-29
kie-issues#385: Test Scenario Editor: Integration with scesim marshaller

### LINKED ISSUE #385: Test Scenario Editor: Integration with scesim marshaller
The scope of this ticket is to use the new [JS-XML Marshaller](https://github.com/kiegroup/kie-issues/issues/220) to marshall and unmarshal XML `scesim` data to JSON, in a type-safe way.

yesamer:
Closes: https://github.com/kiegroup/kie-issues/issues/385

The scope of this PR is to lay down the foundation for a reliable XML marshaller / unmarshaller integration for scesim files.

In addition:
- The file content is now managed using the `setContent(..)` and  `getContent()` methods in order to facilitate future integration with the Foundation framework;
- A first version of the Creation Panel has been created. That panel appears when an empty scesim file is opened. The user can determine the asset type and other settings. These are the ONLY fields currently persisted in the XML file. In a future ticket, that panel will be revisited and extended.

Managed cases:

**EMPTY SCESIM FILE**

https://github.com/kiegroup/kie-tools/assets/16005046/5cafb655-7a79-41f2-ab53-55468255efba

**INVALID SCESIM FILE**

https://github.com/kiegroup/kie-tools/assets/16005046/ba6ad6f9-adcf-4f07-9516-50d567db60ae

**UNSUPPORTED SCESIM FILE**

https://github.com/kiegroup/kie-tools/assets/16005046/72b36012-3700-466f-a308-694f5d0c609c

**VALID SCESIM FILE**

https://github.com/kiegroup/kie-tools/assets/16005046/9b750f5f-2386-4deb-9821-057268590e32


tiagobento:
Nice work @yesamer! I'd love to see more details on the "Generic parse error". Users that run into this will most likely report it, and for us having some logs or something will be very important to address the errors.

Regarding the older versions of SceSim, I think we may need to guide users into how they can "upgrade" to continue using the new Tools, if they're coming from BC.

yesamer:
@tiagobento Yep, I'll provide more detailed info for old scesim version files. 
About the parsing error, it seems the `parsererror` field doesn't offer a meaningful message about the error's root cause :/

Screenshot 2023-08-28 at 12 09 35

tiagobento:
@yesamer Oh... But why are we serializing the "parsererror"? I mean, I though this was for when a `.scesim` file was corrupted or something like that.

yesamer:
Yeah, The goal is to detect if there was any error during the parsing phase. 
I noticed that `ScenarioSimulationModel` populates (with un-meaningful data) that "parsererror" object field, when a error with the parser occurs. 
So, I thought to rely on that field to determine if the file is valid or not: if that field exists, an error occurred during the parsing phase. 
Said so, Is not possible to access that field, because the generated Type doesn't have a "parsererror" field.
Screenshot 2023-08-28 at 12 08 48
For that reason, I modified the XSD adding that field. 
I'm aware that's is not a clean solution, because we're introducing a way to persist that field. 
Any other ideas?

tiagobento:
Well.. I think this `parsererror` field should never be persisted. `parsererror` is not relevant to SceSim, but rather to the UI that is trying to display it. If the Marshaller tries to parse a `.scesim` file and fails, the UI should show a screen with relevant information about the error. Users should be able to use that information to either search for a solution online, or report a bug, if they think their SceSim file is valid.

yesamer:
@tiagobento Yep, that's correct and I agree with that. But how can I access at runtime that `parseerror` field if the (auto-generated) Type doesn't hold it?

tiagobento:
You don't use the SceSim model to store errors... If the parser fails, the parser will throw an Error. You catch it and display it nicely on the user's screen... Not sure if I'm missing something...

packages/scesim-editor/src/TestScenarioEditor.tsx
```diff
@@ -0,0 +1,396 @@
...
+  const scesimLoaded: { ScenarioSimulationModel: SceSim__ScenarioSimulationModelType } = useMemo(
+    () => marshaller.parser.parse(),
+    [marshaller.parser]
+  );
+
+  const [scesimModel, setScesimModel] = useState(scesimLoaded);
+
+  const scesimFileStatus = useMemo(() => {
+    if (scesimModel.ScenarioSimulationModel) {
+      if (scesimModel.ScenarioSimulationModel["parsererror"]) {
```
yesamer: @tiagobento

yesamer:
@tiagobento Ready to be reviewed / merged

tiagobento (approved):

PR #1937, opened 2023-08-29
KOGITO-9728: (Apache) SWF Editor - Remove Patternfly files from stunner source code

handreyrc:
Removing Patterfly resources from Stunner.

**JIRA**: [KOGITO-9728](https://issues.redhat.com/browse/KOGITO-9728)

Thanks


ederign (approved):

packages/serverless-workflow-diagram-editor/kie-wb-common-stunner/README.md
```diff
@@ -36,3 +36,14 @@ Change the _inotify_ limits configuration as noted below to avoid further issues
 3. IDEA executable
 
 - The default setup creates a shortcut to idea.exe. Be sure to change to idea64.exe, otherwise you'll be running 32-bits version
+
+## Building Stunner in IntelliJ
+
+In order to build `kie-wb-common-stunner-lienzo` internal module in IntelliJ it is needed to add a `Maven Goal` to the `Run` configurations of the IDE.
```
LuboTerifaj:
```suggestion
In order to build `kie-wb-common-stunner-lienzo`, internal module in IntelliJ, it is needed to add a `Maven Goal` to the `Run` configurations of the IDE.
```

LuboTerifaj (changes requested):
Hello @handreyrc ,

The warning triangle is now shown when editor content has not been set. We already had such issue in past: https://issues.redhat.com/browse/KOGITO-9597.

Could you please have a look at the issue?

Other than that, it looks good to me.

Thanks!

handreyrc:
> Hello @handreyrc ,
> 
> The warning triangle is now shown when editor content has not been set. We already had such issue in past: https://issues.redhat.com/browse/KOGITO-9597.
> 
> Could you please have a look at the issue?
> 
> Other than that, it looks good to me.
> 
> Thanks!

Hey @LuboTerifaj,

About the warning triangle. I used our current main as reference for this implementation and the warning triangle is shown when an empty diagram is loaded.  Check this file [ErrorPage.html](https://github.com/kiegroup/kie-tools/blob/1d94fbf7e124657ec69b6923252af7de7c36a1cf/packages/serverless-workflow-diagram-editor/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/errorpage/ErrorPage.html#L4), the warning triangle is coming from this page.

This PR is important to not block Apache migration. Since it is behaving like our main and if you agree we are not introducing new issues here we could move forward with it.

Maybe the best course of action is to reopen the issue you mentioned if it makes sense.

WDYT?

handreyrc:
Hey @LuboTerifaj,

I fixed the issue with the warning triangle not being shown. 
The problem was caused because the original patternfly files were edited to include fonts in base64 instead of making reference to them using an URL. In my opinion it is not a good idea to temper with those files.

The best solution to avoid editing patternfly files is to use CSS precedence rules. I added the missing fonts in base64 after all css files are included so the fonts references are overridden.

We are good again for review!

Thanks

LuboTerifaj (approved):
It looks good to me now. Thanks!

PR #1946, opened 2023-09-06
Remove self codesign for qemu

zdrapela:


zdrapela:
Opened PR accidentally, need to test first if the walkaround from #1928 is still needed...

EDIT: looks like it is needed.

PR #1954, opened 2023-09-12
kie-issues#500: Replace copyright headers/comments. More leftovers

klbowers:


yesamer (approved):
Thank you @kbowers-ibm, LGTM.
It seems to me that `checkstyle.header.template` is not actually used... but it's correct to update it as well.

tiagobento (approved):

ederign (approved):

PR #1963, opened 2023-09-14
kie-issue#420: Boxed expression component context menu renders both selected and the chosen cell menu

### LINKED ISSUE #391: Context menu sometimes shows the options for the previously selected cell on the DMN Editor's Boxed Expression Editor
Check the videos below for reproducers. The ideal scenario would be to stop showing the Context Menu `onMouseDown` instead of `onClick` with the left-mouse-button. That's only a visual glitch, but should be fixed.

https://github.com/kiegroup/kie-issues/assets/1584568/4b7bfa21-eecf-4c90-8a48-3929ff5f4324

This second problem is more serious, where the options for a cell are wrong.

https://github.com/kiegroup/kie-issues/assets/1584568/6867dd72-142d-4995-9961-c5f7ff3d23bb

### LINKED ISSUE #420: Boxed expression component context menu renders both selected and the chosen cell menu
Currently, the context menu renders both selected and the chosen cell menu. This makes it possible to use the "Add column" capability in the index row context menu (video 2), which shouldn't be able to add columns.

## video 1
https://github.com/kiegroup/kie-issues/assets/24302289/c07b2be6-31ff-40fd-8f5b-019ea98a72e5

## video 2
https://github.com/kiegroup/kie-issues/assets/24302289/87b9ed8f-c57b-4452-84b1-fc23872737fd

## steps
1. open a Relation expression;
2. click on an input cell;
3. right click on the index;
4. It will render both cell and index menus;

```[tasklist]
### Tasks
- [ ] Shouldn't show column options when opening the index context menu;
```

danielzhe:
Closes: https://github.com/kiegroup/kie-issues/issues/391
Closes: https://github.com/kiegroup/kie-issues/issues/420

https://github.com/kiegroup/kie-tools/assets/7305741/29230c5e-1654-4c4d-840b-cb383fcac7e0

### DIFF
--- a/packages/boxed-expression-component/src/contextMenu/PopoverMenu/PopoverMenu.tsx
+++ b/packages/boxed-expression-component/src/contextMenu/PopoverMenu/PopoverMenu.tsx
@@ -137,16 +137,16 @@ export const PopoverMenu = React.forwardRef(
       }
     }, [appendTo]);
 
+    const yPos = appendElement?.getBoundingClientRect().top ?? 0;
     const popupPosition = useMemo(() => {
       if (appendElement) {
-        const yPos = appendElement?.getBoundingClientRect().top ?? 0;
         const availableHeight = document.documentElement.clientHeight;
         if (POPUP_DEFAULT_HEIGHT + yPos + POPUP_DROP_DOWN_RESERVED_HEIGHT > availableHeight) {
           return PopoverPosition.right;
         }
       }
       return PopoverPosition.bottom;
-    }, [appendElement]);
+    }, [appendElement, yPos]);
 
     return (
       <Popover
--- a/packages/boxed-expression-component/src/expressions/BoxedExpressionEditor/BoxedExpressionEditorContext.tsx
+++ b/packages/boxed-expression-component/src/expressions/BoxedExpressionEditor/BoxedExpressionEditorContext.tsx
@@ -22,7 +22,6 @@ import { useContext, useMemo } from "react";
 import { BeeGwtService, DmnDataType, ExpressionDefinition, PmmlParam } from "../../api";
 import { useRef, useState } from "react";
 import "./BoxedExpressionEditorContext.css";
-import * as _ from "lodash";
 import { BoxedExpressionEditorProps } from "./BoxedExpressionEditor";
 
 export interface BoxedExpressionEditorContextType {
--- a/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx
+++ b/packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx
@@ -146,8 +146,6 @@ export function BeeTableCoordinatesContextProvider({
 }: React.PropsWithChildren<{ coordinates: BeeTableCellCoordinates }>) {
   const { activeCell, depth } = useBeeTableSelection();
 
-  //
-
   const { setMaxDepth: setParentMaxDepth } = useBeeTableCoordinatesDispatch();
   const [_maxDepth, _setMaxDepth] = useState<number>(depth);
 
@@ -165,8 +163,6 @@ export function BeeTableCoordinatesContextProvider({
     setMaxDepth((prev) => Math.max(prev, depth));
   }, [coordinates.columnIndex, coordinates.rowIndex, depth, setMaxDepth]);
 
-  //
-
   useEffect(() => {
     if (coincides(activeCell, coordinates)) {
       setCurrentDepth((prev) => ({
@@ -175,8 +171,6 @@ export function BeeTableCoordinatesContextProvider({
       }));
     }
   }, [_maxDepth, activeCell, coordinates, depth, setCurrentDepth]);
-
-  //
 
   const value = useMemo<BeeTableCoordinatesContextType>(() => {
     return {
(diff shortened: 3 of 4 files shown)


packages/boxed-expression-component/src/expressions/BoxedExpressionEditor/BoxedExpressionEditorContext.tsx
```diff
@@ -22,7 +22,6 @@ import { useContext, useMemo } from "react";
 import { BeeGwtService, DmnDataType, ExpressionDefinition, PmmlParam } from "../../api";
 import { useRef, useState } from "react";
 import "./BoxedExpressionEditorContext.css";
-import * as _ from "lodash";
```
danielzhe, line 25: Cleaning the house

packages/boxed-expression-component/src/selection/BeeTableSelectionContext.tsx
```diff
@@ -146,8 +146,6 @@ export function BeeTableCoordinatesContextProvider({
 }: React.PropsWithChildren<{ coordinates: BeeTableCellCoordinates }>) {
   const { activeCell, depth } = useBeeTableSelection();
 
-  //
```
danielzhe, line 149: Cleaning the house

packages/boxed-expression-component/src/table/BeeTable/BeeTableContextMenuHandler.tsx
```diff
@@ -31,9 +31,9 @@ import TrashIcon from "@patternfly/react-icons/dist/js/icons/trash-icon";
 import BlueprintIcon from "@patternfly/react-icons/dist/js/icons/blueprint-icon";
 import CompressIcon from "@patternfly/react-icons/dist/js/icons/compress-icon";
 import * as React from "react";
-import { useCallback, useEffect, useLayoutEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
 import { BeeTableContextMenuAllowedOperationsConditions, BeeTableOperation, BeeTableOperationConfig } from "../../api";
-import { useCustomContextMenuHandler } from "../../contextMenu/Hooks";
+import { useCustomContextMenuHandler } from "../../contextMenu";
```
danielzhe, line 36: This is to stop a warning. We can import "contextMenu" directly instead of "contextMenu/Hooks"

packages/boxed-expression-component/src/table/BeeTable/BeeTableContextMenuHandler.tsx
```diff
@@ -96,34 +94,51 @@ export function BeeTableContextMenuHandler({
...
     DEFAULT_MULTIPLE_ROWS_COLUMNS_INSERTION
   );
 
+  const { activeCell, selectionStart, selectionEnd } = useBeeTableSelection();
+
+  const activeMenu = useRef("");
+
+  const menuId = "menu-" + activeCell?.columnIndex + "-" + activeCell?.rowIndex;
+  const rootMenuId = useMemo(() => {
+    activeMenu.current = menuId;
```
tiagobento: Updating a `ref` inside a `useMemo` is very unexpected... Refs are "local variables" that need to be manually cared for. Updating `ref`s is usually done on `useEffect`s... Can you please try and revert this back to the state? I understand refs are more "forgiving" than states, but I really think that this is a case for a state, not a ref...

tiagobento (changes requested):

packages/boxed-expression-component/src/table/BeeTable/BeeTableContextMenuHandler.tsx
tiagobento: Yeah.. Maybe we should be removing the `index.ts` files from directories... I really think coupling with the most specific path is better...

packages/boxed-expression-component/src/table/BeeTable/BeeTableContextMenuHandler.tsx
danielzhe: It doesn't work with state, because I need it to change "now".  With state it changed only in the next render. What is the issues that we can have keeping it as it is?

tiagobento: `useMemo` is supposed to be side-effect free.

tiagobento: Mutating a `ref` inside a `useMemo` is a "hack".

danielzhe: @tiagobento Change applied!

packages/boxed-expression-component/src/table/BeeTable/BeeTableContextMenuHandler.tsx
```diff
@@ -367,21 +393,21 @@ export function BeeTableContextMenuHandler({
     ]
   );
 
-  const onMinus = () => {
+  const onMinus = useCallback(() => {
     const newValue = (insertMultipleRowColumnsValue || 0) - 1;
     setInsertMultipleRowColumnsValue(newValue);
-  };
+  }, [insertMultipleRowColumnsValue]);
```
tiagobento, line 399:
```suggestion
  const onMinus = useCallback(() => {
    setInsertMultipleRowColumnsValue(prev => (prev || 0) - 1);
  }, []);
```

Avoid a dependency on the useCallback :) Also, why not make the Math.min and Math.max here at those, instead of at `onChange`?

tiagobento: Same applies for `onChange` and `onPlus`. btw

packages/boxed-expression-component/src/contextMenu/PopoverMenu/PopoverMenu.tsx
```diff
@@ -137,16 +137,16 @@ export const PopoverMenu = React.forwardRef(
...
       if (appendElement) {
-        const yPos = appendElement?.getBoundingClientRect().top ?? 0;
         const availableHeight = document.documentElement.clientHeight;
         if (POPUP_DEFAULT_HEIGHT + yPos + POPUP_DROP_DOWN_RESERVED_HEIGHT > availableHeight) {
           return PopoverPosition.right;
         }
       }
       return PopoverPosition.bottom;
-    }, [appendElement]);
+    }, [appendElement, yPos]);
```
tiagobento, line 149: This change is actually pretty interesting... Raises the question: should we be memo-ing things that are obtained imperatively? I mean, what if `document.documentElement.clientHeight` changes too? We won't have the `popupPosition` be recalculated...

tiagobento (approved):

tiagobento:
Left some comments, as there's still a little bit of bending and twisting, but if it works, I guess it's fine for now :) Feel free to send PRs improving this if you want... Not a trivial component, but lots of interesting problems to solve :)

PR #1978, opened 2023-09-26
kie-issues#528: Fix @kie-tools/form-generation-tool: Problems generating forms for schemas that have # in the filename.

### LINKED ISSUE #528: Fix @kie-tools/form-generation-tool: Problems generating forms for schemas that have `#` in the filename.
The `form-generation-tool` fails when generating forms for schemas that contain the `#` character in the name. 

```
Generating form "demo#orderItems"
Cannot generate form "demo#orderItems":  ',' expected. (3:17)
  1 | import React, { useCallback, useEffect, useState }  from "react";
  2 |     import { TextInput, FormGroup, Card, CardBody, Checkbox } from "@patternfly/react-core";
> 3 | const Form__demo#orderItems: React.FC<any> = ( props:any ) => {
    |                 ^
  4 |   const [formApi, setFormApi] = useState<any>();
  5 |   const [ item, set__item ] = useState<string>("");
  6 | const [ order__orderNumber, set__order__orderNumber ] = useState<string>("");
``` 

Kogito codegen creates the file using the BPMN **processId** as the schema name. If the **processId** contains the `.`  character It is replaced by the `#` (eg: if the **processId** is `demo.orderItems` the generated schema is `demo#orderItems.json`).

We must make the `form-generation-tool` shouldn't fail when dealing with `#` in the schema name and should generate clean id's to be used inside of the generated Forms. 

For simplicity, I think the generated form should use the `.` in the file name, since the form-matching mechanism is done by comparing the form name with the id provided by the runtime, and the runtime will always use the `.`.

Steps to reproduce:
1. Generate **Patternfly** forms using *form-generation-tool* for any BPM Quarkus project in kogito-examples (for example [process-usertasks-quarkus-with-console](https://github.com/kiegroup/kogito-examples/tree/stable/kogito-quarkus-examples/process-quarkus-example))

klbowers:
Closes https://github.com/kiegroup/kie-issues/issues/528

### DIFF
--- a/packages/form-generation-tool/src/__tests__/generation/formGenerationCommand.test.ts
+++ b/packages/form-generation-tool/src/__tests__/generation/formGenerationCommand.test.ts
@@ -63,7 +63,7 @@ describe("formGenerationCommand tests", () => {
   it("Generate forms project with schemas", () => {
     const schemas: FormSchema[] = [
       {
-        name: "ApplyFor#Visa",
+        name: "Apply#For#Visa",
         schema: ApplyForVisaSchema,
       },
       {
@@ -84,11 +84,11 @@ describe("formGenerationCommand tests", () => {
     expect(storeFormAssetsMock).toBeCalledTimes(2);
 
     const applyForVisaAsset: FormAsset = storeFormAssetsMock.mock.calls[0][0];
-    expect(applyForVisaAsset.id).toEqual("ApplyFor#Visa");
-    expect(applyForVisaAsset.sanitizedId).toEqual("ApplyFor_Visa");
-    expect(applyForVisaAsset.assetName).toEqual("ApplyFor#Visa.tsx");
-    expect(applyForVisaAsset.sanitizedAssetName).toEqual("ApplyFor_Visa.tsx");
-    expect(applyForVisaAsset.content).toContain("const Form__ApplyFor_Visa");
+    expect(applyForVisaAsset.id).toEqual("Apply#For#Visa");
+    expect(applyForVisaAsset.sanitizedId).toEqual("Apply_For_Visa");
+    expect(applyForVisaAsset.assetName).toEqual("Apply#For#Visa.tsx");
+    expect(applyForVisaAsset.sanitizedAssetName).toEqual("Apply_For_Visa.tsx");
+    expect(applyForVisaAsset.content).toContain("const Form__Apply_For_Visa");
     expect(storeFormAssetsMock.mock.calls[0][1]).toEqual(sourcePath);
     expect(storeFormAssetsMock.mock.calls[0][2]).toBeTruthy();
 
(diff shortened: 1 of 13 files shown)


packages/form-generation-tool/src/generation/tools/uniforms/utils/InputSanitizationUtil.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+export const inputSanitizationUtil = (unsanitizedString: string) => {
+  return unsanitizedString.replace("#", "_");
```
tiagobento: I think we should be using "replaceAll"? 🙈

tiagobento: Probably also adding a test case for multiple `#`?

klbowers: Good catch, thanks Tiago! Had a Java moment, where I expected replace to replace all (without needing to use a regexp and a global flag):D  Modified a test case to check for multiple #'s 👍

packages/form-generation-tool/src/__tests__/generation/formGenerationCommand.test.ts
```diff
@@ -83,15 +84,19 @@ describe("formGenerationCommand tests", () => {
     expect(storeFormAssetsMock).toBeCalledTimes(2);
 
     const applyForVisaAsset: FormAsset = storeFormAssetsMock.mock.calls[0][0];
-    expect(applyForVisaAsset.id).toEqual("ApplyForVisa");
-    expect(applyForVisaAsset.assetName).toEqual("ApplyForVisa.tsx");
-    expect(applyForVisaAsset.content).toContain("const Form__ApplyForVisa");
+    expect(applyForVisaAsset.id).toEqual("Apply#For#Visa");
+    expect(applyForVisaAsset.sanitizedId).toEqual("Apply_For_Visa");
```
yesamer, line 88: @kbowers-ibm Just a minor comment. Usually, is a good practice to don't modify the testing case of a different test scenario. I mean, the original test checked the case of a string with no `#` characters in the string, which is a valid case to test. I think the best approach is to keep the previous test case (to test the no `#` characters in the string) and create a new test case to check the new case you introduced. 
Considering the logic you're testing is not complex, you can consider this comment as optional, so up to you if apply this tip or not.
But for more complex logic, I recommend using that strategy.

yesamer (approved):
LGTM! Thank you @kbowers-ibm, just a minor comment (Optional)

packages/form-generation-tool/src/__tests__/generation/formGenerationCommand.test.ts
klbowers: Thanks @yesamer! I definitely agree that adding a new test case is generally better. In this case there is already another test below using `confirmTravelAsset ` which does essentially the same thing as the above modified test, and contains no # characters, so I felt it would be safe to just modify the above. With more complex logic I will definitely be sure to add additional tests though! 👍

yesamer: Excellent, thank you @kbowers-ibm !

pefernan:
@tiagobento was going to give it a test but you've been faster than me

yesamer:
@pefernan No problem, if you can please review it we would appreciate it a lot!

PR #1966, opened 2023-09-15
NO-ISSUE: Update `go` to `1.21.1` follow-up

zdrapela:
Update also `go` version in `go.mod` and `go.work` files that were missed in #1956.

PR #1998, opened 2023-10-04
NO-ISSUE: Copy samples to e2e directory

tomasdavidorg:
This PR copies Chrome extension test samples from it-tests directory to e2e directory. The it-tests directory will be deleted in https://github.com/kiegroup/kie-tools/pull/1992. 
This is necessary because the Chrome tests use samples from the main branch, so the samples should be present before the tests are executed.

PR #1992, opened 2023-10-03
NO-ISSUE: Rename findings of it tests to e2e tests

tomasdavidorg:
There are some more it tests strings which should be replaced to e2e.

PR #2000, opened 2023-10-04
Revert "NO-ISSUE: Update `go` to `1.21.1` follow-up (#1966)"

dsimansk:
This reverts commit 219907d4a65de2a8bcdffd165521c3a3228f7ca2.

Per my comment on: https://issues.redhat.com/browse/KOGITO-9799.

We should treat Go toolchain and `go.mod` min version as two separate entities. Because the latest Golang might not be available in the internal build systems etc.

### DIFF
--- a/examples/commit-message-validation-service/go.mod
+++ b/examples/commit-message-validation-service/go.mod
@@ -1,6 +1,6 @@
 module github.com/kiegroup/kie-tools/examples/commit-message-validation-service
 
-go 1.21
+go 1.20
 
 require (
 	github.com/gin-contrib/cors v1.4.0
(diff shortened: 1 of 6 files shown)


zdrapela (commented):
Thanks for the fix. I wasn't aware of that.

It looks good. Just so you know, check on this PR misses the IT/E2E tests in `kn-plugin-workflow` and doesn't run them. This will be fixed in PR #1992 (which depends on PR #1998 being merged first). It might be good to merge those first just to make sure.

dsimansk:
> Thanks for the fix. I wasn't aware of that.
> 
> It looks good. Just so you know, check on this PR misses the IT/E2E tests in `kn-plugin-workflow` and doesn't run them. This will be fixed in PR #1992 (which depends on PR #1998 being merged first). It might be good to merge those first just to make sure.

No worries, it's always a bit tricky with Go versions. I can share a few stories that have bitten us so far. 

Anyway for the merge order, your proposal sounds good. But please note that there's a bit of time pressure to pull `kn-plugin-workflow` into `kn` for the next product build cycle. Depending really how much time the other 2 PRs will take.

tiagobento:
Was discussing this with @ljmotta and our understanding is that this PR makes it **possible** to continue building `kie-tools` with any Go version higher than or equal to 1.20, is that correct?

dsimansk:
> Was discussing this with @ljmotta and our understanding is that this PR makes it **possible** to continue building `kie-tools` with any Go version higher than or equal to 1.20, is that correct?

Yes, that's correct. That's the exact motivation for the PR. For that reason it isn't changing any Go version configured in GitHub action. 

https://go.dev/doc/modules/gomod-ref#go

dsimansk:
@caponetto @zdrapela I've rebased the PR. Kindly asking for GH actions run approval, pls. :)

zdrapela (approved):

ljmotta (approved):

PR #2004, opened 2023-10-05
NO-ISSUE: Do not run e2e tests on build image steps in the staging job

caponetto:


paulovmr (approved):

ljmotta (approved):

caponetto:
On second thought, I think we don't need e2e tests enabled in those places. wdyt @paulovmr @ljmotta?

paulovmr:
I think they may be useful in some cases, like when the latest PR(s) only ran a partial build, then they might identify an issue.

tiagobento:
@paulovmr If a PR only ran a partial build, it's 100% guaranteed that the full build is green. We don't run partial builds when there's a chance that we'll break something else that we're not building on the partial build.

paulovmr:
@caponetto My mistake, for the image builds I agree that those tests are not necessary.

My argument would be more for the "full" build without some images, but I guess I was thinking of a case where the problem was two merges in a row, not the partial build failing to catch a problem. So forget that. :)

paulovmr (approved):

PR #2005, opened 2023-10-05
kie-issues#603: Colorize Decision Services call in FEEL expressions in DMN Editor's Boxed Expression Editor

### LINKED ISSUE #603: Colorize Decision Services call in FEEL expressions in DMN Editor's Boxed Expression Editor
This PR introduced a new feature that colorizes variables and Business Knowledge Model calls in DMN Editor's Boxed Expression Editor, but Decision Services currently are not being parsed, making the new colorized mechanism thinks that the Decision Service call is invalid and painting it with the red color, leading the user to think that the call is invalid.

The Decision Services calls should be handled like the Business Knowledge Model calls.

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/603

### DIFF
--- a/packages/dmn-feel-antlr4-parser/src/index.ts
+++ b/packages/dmn-feel-antlr4-parser/src/index.ts
@@ -20,6 +20,6 @@
 export * from "./FeelVariables";
 export * from "./parser/FeelVariable";
 export * from "./parser/ReservedWords";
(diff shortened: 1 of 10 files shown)


yesamer (approved):
LGTM

tiagobento (approved):

PR #2009, opened 2023-10-06
NO-ISSUE: Remove deleted packages on the lock file

caponetto:


thiagoelg (approved):

paulovmr (approved):

tiagobento:
@caponetto Do we have any idea was causes this? I observed that sometimes it occurs on my local... But I'm really not sure what I might be doing wrong.

caponetto:
@tiagobento I think something is up [here](https://github.com/kiegroup/kie-tools/blob/main/scripts/bootstrap/bootstrap.js#L37). 
Especially with `-F !kie-tools-root... ${pnpmFilterStringForInstalling}`.

If you delete a package from the workspace and run `pnpm bootstrap`, the lock file is kept untouched.
On the other hand, if you remove `-F !kie-tools-root... ${pnpmFilterStringForInstalling}` to test and run `pnpm bootstrap`, the deleted package is also removed from the lock file.

I will take a better look at it next week but any ideas are welcome.
Maybe something changed after the upgrade to v8?

ederign (approved):

PR #2017, opened 2023-10-17
kie-issues#366: Translate the Extended Services Go App to a new Java Application.

### LINKED ISSUE #366: Translate the Extended Services Go App to a new Java Application.
Currently the Extended Services is a go application. We should create a new Java based Extended Services App.

Josephblt:
Fixes [kie-issue#366](https://github.com/apache/incubator-kie-issues/issues/366).

This is a shallow copy of the Extended Services Go Application but using the Java Application.
Except for the System Tray support, all functionality was copied. So everything works just the same.


packages/extended-services-java/install.js
```diff
@@ -0,0 +1,30 @@
...
+const buildEnv = require("./env");
+const { setup } = require("@kie-tools/maven-config-setup-helper");
+
+setup(`
+    -Drevision=${buildEnv.env.extendedServices.version}
+    -Dip=${buildEnv.env.extendedServices.ip}
+    -Dport=${buildEnv.env.extendedServices.port}
+    -DkieSandboxUrl=${buildEnv.env.extendedServices.kieSandboxUrl}
+    -Dquarkus.platform.version=${buildEnv.env.quarkusPlatform.version}
+    -Dversion.org.kie.kogito=${buildEnv.env.kogitoRuntime.version}
```
jomarko, line 29: could you please help me to understand, where to find a value that is passed as `buildEnv.env.kogitoRuntime.version` ? I do not see it in https://github.com/kiegroup/kie-tools/pull/2017/files#diff-41cf05cc30d3a4dd3016b6ff2cf1b98a6fc39accbd993ce87bac4381380f0acb

Josephblt: Hello @jomarko. 

They come from this file:
https://github.com/Josephblt/kie-tools/blob/main/packages/root-env/env/index.js
The line that loads the root file is this one:
https://github.com/kiegroup/kie-tools/blob/c299427a9261a0e7a6918f6935b1ef74a6f0c337/packages/extended-services-java/env/index.js#L22C1-L22C68

jomarko: thank you @Josephblt

jomarko (commented):
@Josephblt thank you for the PR. I was able to run DMN runner using this java extended services. Also build of this module was very quick what I think is cool. The only finding I have is related to this scenario:

- Be sure your java extended servies module is stopped. not running
- start `online-editor` module
- try to open `localhost:9001`, it is the `online-editor` homepage
- you will get error screen as attached [1]
- if I start java extended servies and online editor before accessing the `localhost:9001` everything works fine

Maybe somehow related to defaults we set in packages/extended-services-java/env/index.js ? 
Just guess, do not know.

Not sure if this will affect also scenario of running `online-editor` from an image

[1]
Screenshot 2023-10-19 060633

tiagobento:
@jomarko I think that's the same scenario you found on other errors. Webpack was upgraded and now they have this overlay screen for untreated errors. This error is expected, as we do polling for checking if Extended Services is available.

tiagobento (approved):

Josephblt:
Hello @jomarko. This error happens because every couple of seconds the editor checks for the services. This way no user interaction is needed. This happened before with the Go app as well. The difference now is the new overlay.

PR #2031, opened 2023-10-30
kie-issues#633: Change the Extended Services Go App for the new Extended Services Java in the Extended Services Image

### LINKED ISSUE #633: Change the Extended Services Go App for the new Extended Services Java in the Extended Services Image
The [extended-services-image](https://github.com/kiegroup/kie-tools/tree/main/packages/kie-sandbox-extended-services-image) uses the [extended-services (Go Application)](https://github.com/kiegroup/kie-tools/tree/main/packages/extended-services). 
We must change it for the newly created [extended-services-java application](https://github.com/kiegroup/kie-tools/tree/main/packages/extended-services-java).

Josephblt:
Fixes https://github.com/apache/incubator-kie-issues/issues/633.

Unnecessary information in the Extended-Services-Java app Ping Endpoint was removed.
Extended-Services-Image now uses the Extended-Services-Java app.


packages/extended-services-java/env/index.js
```diff
@@ -21,30 +21,23 @@ const { getOrDefault, varsWithName, composeEnv } = require("@kie-tools-scripts/b
...
-    },
-    EXTENDED_SERVICES__ip: {
+    EXTENDED_SERVICES_JAVA__host: {
       default: "0.0.0.0",
-      description: "Extended Services IP",
+      description:
+        "Quarkus HTTP Host. Configures the IP address or host to which a Quarkus application binds for incoming HTTP requests,",
     },
-    EXTENDED_SERVICES__port: {
+    EXTENDED_SERVICES_JAVA__port: {
```
gitgabrio, line 29: is it needed to change the name of that variable ? It could lead to some break and, after all, the meaning/semantic is the same. If it is possible, I would prefer to not change it

packages/extended-services-java/env/index.js
```diff
@@ -21,30 +21,23 @@ const { getOrDefault, varsWithName, composeEnv } = require("@kie-tools-scripts/b
 
 module.exports = composeEnv([require("@kie-tools/root-env/env")], {
   vars: varsWithName({
-    EXTENDED_SERVICES__version: {
-      default: require("../package.json").version,
-      description: "Extended Services Version",
-    },
-    EXTENDED_SERVICES__ip: {
+    EXTENDED_SERVICES_JAVA__host: {
```
gitgabrio, line 24: is it needed to change the name of that variable ? It could lead to some break and, after all, the meaning/semantic is the same. If it is possible, I would prefer to not change it

packages/kie-sandbox-extended-services-image/Containerfile
```diff
@@ -15,17 +15,22 @@
 # specific language governing permissions and limitations
 # under the License. 
 
-FROM --platform=linux/amd64 registry.access.redhat.com/ubi9/ubi-minimal:9.2
+FROM --platform=linux/amd64 registry.access.redhat.com/ubi9/openjdk-11-runtime:1.16-3
```
gitgabrio, line 18: Never realized before, but since this is/should be a "company-agnostic" (open-source) code, is it not strange to rely upon the "redhat" registry ?

packages/kie-sandbox-extended-services-image/package.json
```diff
@@ -18,13 +18,13 @@
     "cleanup": "rimraf dist-dev && mkdir dist-dev",
     "copy:assets": "pnpm copy:extended-services",
     "copy:extended-services": "run-script-os",
-    "copy:extended-services:linux:darwin": "cp -R ../extended-services/dist/linux/kie_sandbox_extended_services_headless ./dist-dev/kie_sandbox_extended_services_headless",
-    "copy:extended-services:win32": "pnpm powershell \"Copy-Item ../extended-services/dist/linux/kie_sandbox_extended_services_headless ./dist-dev/kie_sandbox_extended_services_headless\"",
-    "docker:build": "kie-tools--image-builder -r \"$(build-env extendedServicesImage.registry)\" -a \"$(build-env extendedServicesImage.account)\" -n \"$(build-env extendedServicesImage.name)\" -t \"$(build-env extendedServicesImage.buildTags)\"",
-    "podman:build": "kie-tools--image-builder -r \"$(build-env extendedServicesImage.registry)\" -a \"$(build-env extendedServicesImage.account)\" -n \"$(build-env extendedServicesImage.name)\" -t \"$(build-env extendedServicesImage.buildTags)\" -e podman"
+    "copy:extended-services:linux:darwin": "cp -R ../extended-services-java/dist/extended-services-java ./dist-dev/kie_sandbox_extended_services",
```
gitgabrio: See previous comments: is it needed to change the name of the `/extended-services` directory ? If not, I think it would be better to leave it unchanged - changing the names of files/directories/variables without a need only makes the code instable

packages/kie-sandbox-extended-services-image/package.json
```diff
@@ -18,13 +18,13 @@
...
+    "copy:extended-services:win32": "pnpm powershell \"Copy-Item ../extended-services/dist/extended-services-java ./dist-dev/kie_sandbox_extended_services\"",
+    "docker:build": "kie-tools--image-builder -r \"$(build-env extendedServicesImage.registry)\" -a \"$(build-env extendedServicesImage.account)\" -n \"$(build-env extendedServicesImage.name)\" -t \"$(build-env extendedServicesImage.buildTags)\" --build-arg \"EXTENDED_SERVICES_DEFAULT_PORT=$(build-env extendedServicesJava.port)\" --build-arg \"EXTENDED_SERVICES_DEFAULT_HOST=$(build-env extendedServicesJava.host)\"",
```
gitgabrio, line 23: same comment here about `extendedServicesJava`: the fact that it is a `java` implementation is an implementation detail that does not need to be exposed in the name of that variable

pnpm-lock.yaml
```diff
@@ -5163,9 +5163,9 @@ importers:
 
   packages/kie-sandbox-extended-services-image:
     dependencies:
-      "@kie-tools/extended-services":
+      "@kie-tools/extended-services-java":
```
gitgabrio, line 5166: see previous comments

repo/graph.dot
```diff
@@ -324,7 +324,7 @@ digraph G {
   "@kie-tools/kie-sandbox-distribution" -> "@kie-tools/cors-proxy-image" [ style = "solid", color = "black" ];
   "@kie-tools/kie-sandbox-distribution" -> "@kie-tools/kie-sandbox-extended-services-image" [ style = "solid", color = "black" ];
   "@kie-tools/kie-sandbox-distribution" -> "@kie-tools/kie-sandbox-image" [ style = "solid", color = "black" ];
-  "@kie-tools/kie-sandbox-extended-services-image" -> "@kie-tools/extended-services" [ style = "solid", color = "black" ];
+  "@kie-tools/kie-sandbox-extended-services-image" -> "@kie-tools/extended-services-java" [ style = "solid", color = "black" ];
```
gitgabrio, line 327: see previous comments

repo/graph.json
```diff
@@ -692,7 +692,7 @@
       { "source": "@kie-tools/kie-sandbox-distribution", "target": "@kie-tools/kie-sandbox-image", "weight": 1 },
       {
         "source": "@kie-tools/kie-sandbox-extended-services-image",
-        "target": "@kie-tools/extended-services",
+        "target": "@kie-tools/extended-services-java",
```
gitgabrio, line 695: see previous comments

gitgabrio (changes requested):
Thanks a lot @Josephblt for this PR!
What I would ask you is to avoid  all that variable/file/directory renaming that only adds a lot of instability, creates issues on consumer environments (i.e. ci), and reflect an implementation detail that should not have any impact on the above naming.

packages/extended-services-java/env/index.js
Josephblt: Two things about these variables. They conflict with existing variables from the extended-services go app. Remember they still both exist. So I had to change their names.

That being said, we can consider these new variables. So, I just went ahead and matched the names to their exact function. This way we avoid unnecessary guessing. 

No need to worry about this change as it will not break anything.

packages/extended-services-java/env/index.js
Josephblt: Same as above.

packages/kie-sandbox-extended-services-image/Containerfile
Josephblt: There was a discussion about this, last week, but yeah, I'm open to suggestions. What we need here is a Linux image that has Java 11 Runtime Environment on it. 

The reason I went with that one is because it has minimal overhead, provides the right JRE, and we already used an image from that same registry.

packages/kie-sandbox-extended-services-image/package.json
Josephblt: This is related to the above comment.

pnpm-lock.yaml
Josephblt: This here, is necessary because of the existence of both apps.

repo/graph.dot
Josephblt: Same as above.

repo/graph.json
Josephblt: Same as above.

packages/kie-sandbox-extended-services-image/package.json
Josephblt: Here again, we still have to take into account the fact that the extended-services go app still exists. To keep clarity and prevent any misidentification between the two apps I thought it was important to differentiate. I don't see how this can be a problem.

packages/kie-sandbox-extended-services-image/Containerfile
tiagobento: If the image we chose has a compatible license and meets out requirements, I don't see any problem this image coming from Red Hat.

tiagobento:
@gitgabrio We're gradually moving away from the Go-based Extended Services. The first step is to have a Quarkus app that does the same as it does. This was addressed on #2017. Now this PR goes a step further and changes the Extended Services used inside the Extended Services image to the new Java-based one. The Go-based one is still around, as we need it for the community production deployment at sandbox.kie.org. Once we have an environment that we can trust, we will stop offering users the possibility to download Extended Services, and we will maintain an Extended Services online for community users to use. Until we don't have this environment, we will keep both, that's why now there's the `extended-services` package, and the `extended-services-java` package. In the future, when we only have one, we can rename `extended-services-java` to `extended-services` again. For now, both will co-exist.

tiagobento (approved):

PR #2033, opened 2023-10-30
KOGITO-9701: Implement Workflow Instances section

fantonangeli:
**Jira:** https://issues.redhat.com/browse/KOGITO-9701

**Description:**
Implement the Workflow Instances section as we have on Serverless Logic Web Tools.

This will require one additional (optional) property for the Data Index URL in the data JSON file (`dataIndexExternalUrl`).

If the data index URL property is not present or the given URL cannot be accessible, the Workflow Instances section should not be shown (or disable the section and warn the user about it).

Since the data index can be embedded in the deployment, we can also have a boolean property `embeddedDataIndex`. If true, `dataIndexExternalUrl` is ignored and we must try to fetch the embedded data index.

**Notes:**
- Instead having`embeddedDataIndex` and `dataIndexExternalUrl` in the data JSON file, the app uses `dataIndexExternalUrl`, if not set uses the embedded data index service, if the service is not available a message is shown on the page with the link to https://kiegroup.github.io/kogito-docs/serverlessworkflow/latest/data-index/data-index-core-concepts.html. I'm available to change this behaviour.
- in `/src/runtimeTools` there are files duplicated from `Serverless Logic Web Tools` . Maybe we can evaluate how to reuse the code in a different PR, as we did in https://github.com/kiegroup/kie-tools/pull/1995

**Preview:**
[KOGITO-9701.webm](https://github.com/kiegroup/kie-tools/assets/17780574/0a39c084-dadb-4319-be29-8d8881c473f0)

Data Index Service not available:
image


tomasdavidorg (approved):

packages/sonataflow-deployment-webapp/src/context/AppContextProvider.tsx
```diff
@@ -19,30 +19,40 @@
...
 import { DEFAULT_APPDATA_VALUES } from "../AppConstants";
-import { AppData } from "../data";
+import { AppData, verifyDataIndex } from "../data";
 import { useAppDataPromise } from "../hooks/useAppDataPromise";
 import { AppContext } from "./AppContext";
 
 export function AppContextProvider(props: PropsWithChildren<{}>) {
   const appDataPromise = useAppDataPromise();
   const [data, setData] = useState<AppData>(DEFAULT_APPDATA_VALUES);
+  const [dataIndexAvailable, setDataIndexAvailable] = useState(true);
```
caponetto:
```suggestion
  const [dataIndexAvailable, setDataIndexAvailable] = useState(false);
```

Isn't it better to assume that the data index is unavailable until we are sure via the effect below?

packages/sonataflow-deployment-webapp/src/data/index.ts
```diff
@@ -22,9 +22,26 @@ import { routes } from "../routes";
 export interface AppData {
   appName: string;
   showDisclaimer: boolean;
+  dataIndexExternalUrl?: string;
+  dataIndexUrl: string;
```
caponetto: This looks confusing. Can we have only one URL?

packages/sonataflow-deployment-webapp/src/runtimeTools/workflows/RuntimeToolsWorkflowInstances.tsx
```diff
@@ -0,0 +1,84 @@
...
+          {`Data Index service not available`}
+        </Title>
+        <EmptyStateBody>
+          <TextContent>
+            <Text>
+              Start by setting the Data Index in the config file.
+              <br />
+              Read more on &nbsp;
+              <a
+                href="https://kiegroup.github.io/kogito-docs/serverlessworkflow/latest/data-index/data-index-core-concepts.html"
```
caponetto: Could you please extract this URL to a const in `AppConstants`?

packages/sonataflow-deployment-webapp/src/runtimeTools/workflows/RuntimeToolsWorkflowInstances.tsx
```diff
@@ -0,0 +1,84 @@
...
+    <Bullseye>
+      <EmptyState>
+        <EmptyStateIcon icon={CubesIcon} />
+        <Title headingLevel="h4" size="lg">
+          {`Data Index service not available`}
+        </Title>
+        <EmptyStateBody>
+          <TextContent>
+            <Text>
+              Start by setting the Data Index in the config file.
```
caponetto: I think we can show here what is the current data index URL set (if any).

caponetto (commented):
Great work, @fantonangeli! Thanks!
Regarding the duplicated code from web tools, can you please create a JIRA for us to track it? Also, it would be nice to list in the JIRA all the files that are duplicated.

packages/sonataflow-deployment-webapp/src/context/AppContextProvider.tsx
fantonangeli: If I set `dataIndexAvailable` to false, the page will show the error until the service answers with a 200 code, like in the recording below.
I would set it to undefined as initial state.
[Peek 2023-11-06 16-48.webm](https://github.com/kiegroup/kie-tools/assets/17780574/c323fb99-2e1e-4406-97cc-74d2f8a32097)

packages/sonataflow-deployment-webapp/src/data/index.ts
fantonangeli: The dataIndexUrl can be external or embedded:
- `dataIndexExternalUrl` comes from the JSON data file
- `dataIndexUrl` is "/graphql" if `dataIndexExternalUrl` is not set.

I would suggest having only `dataIndexUrl` in both code and JSON data file. WDYT?

packages/sonataflow-deployment-webapp/src/context/AppContextProvider.tsx
caponetto: `undefined` sounds good

packages/sonataflow-deployment-webapp/src/data/index.ts
caponetto: +1 for having only `dataIndexUrl`

fantonangeli:
> Great work, @fantonangeli! Thanks! Regarding the duplicated code from web tools, can you please create a JIRA for us to track it? Also, it would be nice to list in the JIRA all the files that are duplicated.

Thank you @caponetto
Jira for reuse of components: https://issues.redhat.com/browse/KOGITO-9935

caponetto (approved):

PR #2044, opened 2023-11-09
KOGITO-9929: Filter and Timeout Tooltips show null instead of undefined

treblereel:
JIRA: https://issues.redhat.com/browse/KOGITO-9929

### DIFF
--- a/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/definition/StateDataFilter.java
+++ b/packages/serverless-workflow-diagram-editor/sw-editor/sw-editor-api/src/main/java/org/kie/workbench/common/stunner/sw/definition/StateDataFilter.java
@@ -32,16 +32,11 @@ public class StateDataFilter {
     public String output;
 
     public final String getInput() {
-        if(input == null) {
-            return "undefined";
-        }
         return input;
     }
 
     public final void setInput(String input) {
-        if(!input.equals("undefined")) {
-            this.input = input;
-        }
+        this.input = input;
     }
 
     public final String getOutput() {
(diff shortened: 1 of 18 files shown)


handreyrc (changes requested):
Hey @treblereel,

I noticed there are more cases involving the null. It is the same issue in other places.
Check the Operation state and Parallel state  if you have the following properties defined as below:
 - "actions": [{}]
 - "functionRef": {}
 
Maybe it's an oportunity to check all the information we are presenting from Icon's tooltips,  we may find more similar cases.

Please, could you take a look at those as well?

Thanks!

treblereel:
> Hey @treblereel,
> 
> I noticed there are more cases involving the null. It is the same issue in other places. Check the Operation state and Parallel state if you have the following properties defined as below:
> 
> * "actions": [{}]
> * "functionRef": {}
> 
> Maybe it's an oportunity to check all the information we are presenting from Icon's tooltips, we may find more similar cases.
> 
> Please, could you take a look at those as well?
> 
> Thanks!

In case we have multiple 'null' messages instead of 'undefined', we should rework our rendering. Otherwise, our models will be flooded with Js.undefined() checks, which isn't great.

@LuboTerifaj What do you think?

LuboTerifaj:
Hello @handreyrc , [KOGITO-9929](https://issues.redhat.com/browse/KOGITO-9929) is caused by [KOGITO-9831](https://issues.redhat.com/browse/KOGITO-9831). The issues you are mentioning were also present on main before. It may be beneficial to go through all icons and identify potential issues. However, I would report it in a separate jiras. There are already some issues reported, for example: [KOGITO-8762](https://issues.redhat.com/browse/KOGITO-8762), so we might want to revisit it, and decide what should be, or what shouldn't be shown in the diagram.

LuboTerifaj (changes requested):
Hello @treblereel ,

There is still null shown in some places within filters and timeouts:

- In stateDataFilter: `output: null`	
- In timeouts: 
  - `EventTimeout: null`
  - `ActionExecTimeout: null`

Could you please have a look at it?

Thanks!

LuboTerifaj:
Hello @treblereel , regarding rework of rendering I agree. See my [comment](https://github.com/kiegroup/kie-tools/pull/2044#issuecomment-1825496952) above. The current fix looks more like a temporary solution for [KOGITO-9929](https://issues.redhat.com/browse/KOGITO-9929) regression.

LuboTerifaj (approved):
Hello @treblereel , the fix looks good to me. Thanks!

handreyrc (approved):
Hey @treblereel and @LuboTerifaj ,

It looks good to me as well lets move forward with it!

Thanks!

PR #2049, opened 2023-11-13
[0.30.4] NO-ISSUE: Fix CVEs on dmn-dev-deployment-base-image and update Quarkus and Kogito versions

thiagoelg:


ederign (approved):

porcelli (approved):
Lgtm

yesamer:
@thiagoelg @tiagobento It seems failing because of this https://stackoverflow.com/a/76731553
Updating selenium should do the trick. Here's the patch https://github.com/kiegroup/kie-tools/pull/2051, hope it will solve the issue.
I expect we have the same problem in the `main` branch.

yesamer:
@thiagoelg There's another issue to solve.
Our Java CodeCompletion module is referring to a "stale" version, 1.25.0, that is no longer valid.
I updated it in this PR https://github.com/kiegroup/kie-tools/pull/2053.
`main` is already up-to-date

yesamer (commented):
I guess we need to update https://github.com/kiegroup/kie-tools/blob/0.30.4-prerelease/packages/stunner-editors/pom.xml#L263 as well. 
I see 8.40.1 is available, that's probably the engine version wrapped in Kogito 1.40.2, right?

tiagobento:
The tests that broke are on the Chrome Extension. As much as I hate to say it, I think we should go ahead and merge it like that, given that 0.30.4 is going to be a special-purpose release only for `DMN Dev deployment image`. WDYT @ederign @thiagoelg ?

ederign:
@tiagobento go ahead. Please let me know if anything is necessary on main in our side.

PR #2067, opened 2023-11-23
KOGITO-9962: Add Temporarily unexpire M118 flags to Chrome extension tests

tomasdavidorg:
https://issues.redhat.com/browse/KOGITO-9962

Enables "Temporarily unexpire M118 flags" for Chrome extension tests.
Currently it does not have any effect because the `--allow-insecure-localhost` works even without the `--enable-features=UnexpireFlagsM118` flag but in the future, new versions of Chrome may require it.

So it is good to have here for preventive reasons and when it comes to the local testing, these steps need to be done:

1. Enable "Temporarily unexpire M118 flags. chrome://flags/#temporary-unexpire-flags-m118
2. Restart browser
3. Enable "Allow invalid certificates for resources loaded from localhost."
4. Restart browser


paulovmr (approved):

tomasdavidorg:
Probably random fail on MacOs. Rerunning.

LuboTerifaj (approved):
It looks good to me. Thanks!

PR #2078, opened 2023-12-05
kie-issues#605: Apache CI migration

rodrigonull:
ISSUE: https://github.com/apache/incubator-kie-issues/issues/605

1. Added new Jenkins jobs to replace the existing GitHub workflows
2. Disabled Github workflows triggers for the jobs that we were able to migrate to Jenkins
3. Pinned some Github actions versions (ASF has strict requirements for using Github actions)
4. Increased E2E test timeouts to solve issues we have experienced when running the build / tests on ASF Jenkins

NOTE: ASF Jenkins does not have MacOS / Windows runners so we had to keep the Github workflows for these operating systems.


.ci/jenkins/shared-scripts/pipelineVars.groovy
```diff
@@ -0,0 +1,37 @@
...
+    String githubRepositoryName = 'kie-tools';
+    String githubRepositorySlug = 'apache/kie-tools';
+
+    String quayPushCredentialsId = 'quay-io-kie-tools-token';
+    String openshiftCredentialsId = 'openshift-kie-tools-token';
+    String kieToolsBotGithubCredentialsId = 'kie-tools-bot-gh';
+    String kieToolsBotGithubTokenCredentialsId = 'kie-tools-bot-gh-token';
+    String kieToolsGithubCodeQLTokenCredentialsId = 'kie-tools-gh-codeql-token';
+    String chromeStoreCredentialsId = 'kie-tools-chome-store';
+    String chromeStoreRefreshTokenCredentialsId = 'kie-tools-chome-store-refresh-token';
```
thiagoelg: Small typo

```suggestion
    String chromeStoreCredentialsId = 'kie-tools-chrome-store';
    String chromeStoreRefreshTokenCredentialsId = 'kie-tools-chrome-store-refresh-token';
```

.ci/kie-tools-ci-build.Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+vim \
+make \
+zip \
+unzip \
+bzip2 \
+xvfb \
+fluxbox
+
+# Cleanup
+RUN apt-get clean autoclean && apt-get autoremove --yes && rm -rf /var/lib/{apt,dpkg,cache,log}/
```
ljmotta: The "Cleanup" step should be within the "Install" step. After the RUN ends, it creates a new layer in the image that can't be changed afterward, if we have a separate RUN for the "Cleanup" it will not erase the content of the previous layer.

rodrigonull: Good catch, I've applied the suggestion

tiagobento (approved):

PR #2087, opened 2023-12-08
kie-issues#129: Import private repository into sandbox via chrome-extension [backend-part]

jomarko:
Related to https://github.com/apache/incubator-kie-issues/issues/129

This PR doesn't not fix completely https://github.com/apache/incubator-kie-issues/issues/129 as it is split into two problems:
- importing private repository via chrome extension
- displaying `Open in Sandbox` button in the chrome extension

This PR solves only the first 'backend' issue, importing a private repository. If there is a compatible session configured in sandbox, this PR tries to use that one for importing automatically.

### DIFF
--- a/packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
+++ b/packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
@@ -30,9 +30,9 @@ import { useRoutes } from "../navigation/Hooks";
 import { QueryParams } from "../navigation/Routes";
 import { OnlineEditorPage } from "../pageTemplate/OnlineEditorPage";
 import { useQueryParam, useQueryParams } from "../queryParams/QueryParamsContext";
-import { useSettingsDispatch } from "../settings/SettingsContext";
 import {
   ImportableUrl,
+  isCertainlyGit,
   isPotentiallyGit,
   isSingleFile,
   UrlType,
@@ -44,7 +44,7 @@ import { AdvancedImportModal, AdvancedImportModalRef } from "./AdvancedImportMod
 import { fetchSingleFileContent } from "./fetchSingleFileContent";
 import { useGitHubClient } from "../github/Hooks";
 import { AccountsDispatchActionKind, useAccountsDispatch } from "../accounts/AccountsContext";
-import { useAuthSession, useAuthSessions } from "../authSessions/AuthSessionsContext";
+import { AuthInfo, useAuthSession, useAuthSessions } from "../authSessions/AuthSessionsContext";
 import { useAuthProvider, useAuthProviders } from "../authProviders/AuthProvidersContext";
 import { getCompatibleAuthSessionWithUrlDomain } from "../authSessions/CompatibleAuthSessions";
 import { useWorkspaces } from "@kie-tools-core/workspaces-git-fs/dist/context/WorkspacesContext";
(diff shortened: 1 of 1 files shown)


packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
```diff
@@ -96,6 +98,25 @@ export function NewWorkspaceFromUrlPage() {
...
+      authSessions,
+      authSessionStatus,
+      urlDomain: new URL(importableUrl.url).hostname,
+    });
+    setAuthInfo({
+      username: (compatible as GitAuthSession[])[0].login,
+      uuid: (compatible as GitAuthSession[])[0].uuid,
+      password: (compatible as GitAuthSession[])[0].token,
+    });
+  }
```
tiagobento: Using an `if` statement directly on the component's function implementation is not something we really do, since it will potentially lead to very strange behaviors, as this `if` statement will be executed every time `NewWorkspaceFromUrlPage` re-renders.

That's one of the reasons why we have the general rule of always encapsulating code inside hooks, like `useEffect`.

More than that, this code waits for an error to happen, checks for the error string (which is very weak in terms of reliability) and tries to isolate a scenario so that `authInfo` can be set.

Instead of trying to find a way to really "ignore" everything else going around the fix, we always need to try and understand how the current code is working, and how the current abstractions can be adapted so that the bug can be fixed.

We have a very similar code to this one on the effect marked with `// Startup the page. Only import if those are set.`.

packages/online-editor/src/importFromUrl/NewWorkspaceFromUrlPage.tsx
```diff
@@ -76,7 +76,9 @@ export function NewWorkspaceFromUrlPage() {
 
   const authProviders = useAuthProviders();
   const { authSessions, authSessionStatus } = useAuthSessions();
-  const { authSession, gitConfig, authInfo } = useAuthSession(queryParamAuthSessionId);
+  const authSessionInfo = useAuthSession(queryParamAuthSessionId);
+  const { authSession, gitConfig } = authSessionInfo;
+  const [authInfo, setAuthInfo] = useState<AuthInfo | undefined>(authSessionInfo.authInfo);
```
tiagobento: Creating this state introduces an ambiguity, as `authInfo` is already determined by the `queryParamAuthSessionId`. We should be setting `queryParamAuthSessionId` correctly and re-rendering the page instead of introducing a new state.

tiagobento (changes requested):
Left some comments inline. Please take a careful look at them and let me know if there's something you don't understand.

This page is already able to import projects using authSessionIds. If it's not working for private repositories, we should be trying to make the page set its `authSessionId` correctly, using the mechanisms it already has.

Introducing a carefully tailored `if` statement that isolates a specific error so that the bug can be fixed might result in things working as expected, but it is a harmful practice that will make this code harder and harder to maintain, making more bugs appear until it snowballs into an incomprehensible component that will need a full rewrite eventually.

jomarko:
I have incorporated the review feedback and tried the following scenarios:

### Public repository
Importing a public repository using link `https://localhost:9001/#/import?url=https://github.com/apache/incubator-kie-kogito-examples/tree/stable` worked. See the attached screenshot.
image

### Private repository, token available but not compatible
Using link `https://localhost:9001/#/import?url=https://github.com/jomarko/dmn-sample-private/tree/PR` (where `dmn-sample-private` is my private repository) opened and advanced import modal
image


### Private repository, no token available
Using same link as above resulted into the same result as above == an advanced import modal was opened.

### Private repository, compatible token available
Using link `https://localhost:9001/#/import?url=https://github.com/jomarko/dmn-sample-private/tree/PR` Worked successfully without opening an advanced import modal.

image

tiagobento (approved):
Nice job @jomarko!

Issue #2095, opened 2023-12-18
Investigate why SLWT UI breaks with `@patternfly/react-core` version `4.276.12`


caponetto:
This behavior is not seen on version `4.276.6`

<img width="500" alt="image" src="https://github.com/apache/incubator-kie-tools/assets/638737/95dc4eda-a948-4f7d-abd0-3a23fe253086">

<img width="500" alt="image" src="https://github.com/apache/incubator-kie-tools/assets/638737/69decf03-3f49-4f36-9c18-54823b922c6c">

fantonangeli:
Closed in favour of: 
- https://github.com/apache/incubator-kie-tools/issues/2641

PR #1982, opened 2023-09-26
KOGITO-5611: Move SWF runtime tools packages from kogito-apps to kie-tools

paulovmr:
This PR migrates packages related to serverless workflow runtime tools from `kogito-apps` to `kie-tools`. It is the first of many that will deal with that migration, since there are still features and tests to be migrated. 

The new packages are now integrated in the Serverless Logic Web Tools, as demonstrated in the video below:

https://github.com/kiegroup/kie-tools/assets/832830/1888e6cc-6eb4-4671-a785-45e92deabef4

PR #2096, opened 2023-12-18
NO-ISSUE: Remove duplicated version of `@patternfly/react-core` from the lock file

caponetto:
A PR has introduced a second version of `@patternfly/react-core` in the lock (`4.276.12`) and set this version for some packages. This version (`4.276.12`) breaks the UI of SLWT.

Considering that we already had `4.276.6` being used, I'm not sure why the lock has included this new version.
To fix that, I simply ran `pnpm bootstrap` from the version of the lock file before the PR.
Now, all packages are back to `4.276.6`.

Anyway, I opened an issue to investigate why SLWT UI breaks on `4.276.12`. See https://github.com/apache/incubator-kie-tools/issues/2095

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -800,13 +800,13 @@ importers:
         version: 7.16.12
       "@babel/preset-env":
         specifier: ^7.16.0
-        version: 7.18.10(@babel/core@7.16.12)
+        version: 7.16.11(@babel/core@7.16.12)
       "@babel/preset-react":
         specifier: ^7.16.0
         version: 7.16.0(@babel/core@7.16.12)
       "@babel/preset-typescript":
         specifier: ^7.22.5
-        version: 7.22.5(@babel/core@7.16.12)
+        version: 7.23.0(@babel/core@7.16.12)
       "@jest/globals":
         specifier: ^26.6.2
         version: 26.6.2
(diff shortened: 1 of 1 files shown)


thiagoelg (approved):
My bad!
Nice catch though :)

paulovmr (approved):

tiagobento:
Any idea how that happened even with `CI :: Dependencies consistency`?

caponetto:
The version specified in the `package.json` files was correct. `syncpack` does not look at the lock file, does it?
Now, why did bootstrap choose to install a new version instead of using the one we already had?

tiagobento:
@caponetto Right... that's the mystery D:

caponetto:
@thiagoelg do you remember doing anything differently that could have resulted in that behavior?

thiagoelg:
Not really, but my PR took a while, and multiple merges with main... maybe in one of those merges a conflict was not solved properly :confused:

paulovmr:
@tiagobento @caponetto @thiagoelg I had changed the version of this dependency from `4.276.12` to `4.276.6` during my [PR](https://github.com/apache/incubator-kie-tools/pull/1982) migrating components from kogito-apps. So if this task was started before its merge and some conflicts happened, that might have been the issue indeed.

PR #2099, opened 2023-12-20
kie-issues#399: chrome-extension: Some DMN Files are not correctly loaded

### LINKED ISSUE #399: chrome-extension: Some DMN Files are not correctly loaded
Step to reproduce:

1. Install the KIE chrome-extension version 0.31.0
2. Try to open any DMN file with spaces in the file name (eg. "Traffic Violation.dmn") part of any repository
3. The editor works, but data is not loaded (DRG and Data Types are not loaded)

Files with no space (eg. "TrafficViolation.dmn") work correctly.

https://github.com/kiegroup/kie-issues/assets/16005046/a363bf3a-1598-4419-b76b-7c458f4dd780

klbowers:
Closes https://github.com/apache/incubator-kie-issues/issues/399
I setup loggers and tried multiple scenarios in GitHub, but no matter what I tried commitSHA was always undefined (hence why the URL contained undefined in Luiz' image). I think it makes more sense to use the ref here, as that seems to more accurately reflect the GitHubURL.

https://github.com/apache/incubator-kie-tools/assets/92726146/0290c265-98ca-4f79-8e39-105707a64c42


packages/chrome-extension/src/app/github/api.ts
```diff
@@ -39,7 +45,7 @@ export function fetchFile(
     .then((res) => (contentType === ContentType.BINARY ? (res.data as any).content : atob((res.data as any).content)))
     .catch((e) => {
       console.debug(`Error fetching ${path} with Octokit. Fallback is 'raw.githubusercontent.com'.`);
-      return fetch(`https://raw.githubusercontent.com/${org}/${repo}/${commitSHA}/${path}`).then((res) =>
+      return fetch(modifiedRawGithubUserContentUrl.toString()).then((res) =>
         res.ok ? res.text() : Promise.resolve(undefined)
       );
     });
```
tiagobento, line 44: I don't see why we would need `commitSHA` in this case.... I mean, there's only one place that looks for the `commitSHA` and it is here:

> @ prEditors.tsx#parsePrInfo
`  const commitSHA = dependencies.all.getViewFileButton()?.href.split("/")[6];`

Now, let's analyze everything...

1. Looking at the GitHub PR page, it looks like `dependencies.all.getViewFilebutton()` is working as expected. 
image

2. Since there can be multiple "View file button"s in the PR page, I already think we should change this name to `getFirstViewFileButton`. Getting the `commitSHA` out of it is ok too, as it doesn't matter how many buttons will there be, since they'll always have the same `commitSHA` to be extracted.

3. Taking a step further, and seeing if the `commitSHA` extraction works as expected... it does:
image

4. Ok, so why would `commitSHA` not exist? Maybe because we're looking at an Editor that's not inside a PR? :D From your video it looks like we're dealing with the <SingleEditorViewApp>, which uses `fetchFile` without `commitSHA`.

---

Anyway...... Since it does work with `ref` and `commitSHA` is only used in that specific place.. I think we should be removing this whole `commitSHA` thing. WDYT? :)

packages/chrome-extension/src/app/github/api.ts
```diff
@@ -29,6 +29,12 @@ export function fetchFile(
   contentType?: ContentType,
   commitSHA?: string
 ) {
+  const rawGithubUserContentUrl = "https://raw.githubusercontent.com";
+  const modifiedRawGithubUserContentUrl = new URL(rawGithubUserContentUrl);
+  modifiedRawGithubUserContentUrl.pathname = `${org}/${repo}/${ref}/${path}`;
+  if (commitSHA) {
+    modifiedRawGithubUserContentUrl.pathname = `${modifiedRawGithubUserContentUrl.pathname}/${commitSHA}`;
+  }
```
tiagobento: If you agree with removing the `commitSHA` altogether, this code will change a lot, but as I'm here, here's some suggestions on how to make this better.

1. If `commitSHA` exists, in this case your final URL would be ``${org}/${repo}/${ref}/${path}/${commitSHA}`. Which is NOT preserving the old behavior.
2. The `rawGithubUserContentUrl` variable is totally unnecessary. You could inline it to avoid having an extra line and having to name one extra thing as well (although this is perfectly named, IMHO).
3. 
```typescript
modifiedRawGithubUserContentUrl.pathname = `${org}/${repo}/${commitSHA ?? ref}/${path}`
``` 
4. All this logic is being executed before you're even sure you'll need it. Always try to put stuff inside the smallest scope possible (unless you have bigger plans). In this case, this entire logic only applies to the block inside the `catch` there.

would be much better, IMHO.

Anyway, if we agree to remove the commitSHA stuff, this can be reduced a lot.

tiagobento (changes requested):
Please consider my comments inline.

tiagobento:
@kbowers-ibm Another comment... there are a lot of warnings on the Chrome Extension package related to missing dependencies on Hooks. Addressing them would be a good addition, as you're already dealing with this code anyway. Feel free to do it later, though. Just saying ;)

klbowers:
Appreciate the detailed review Tiago, thank you! :) Will push an update next week with the changes you suggested👍 
> there are a lot of warnings on the Chrome Extension package related to missing dependencies on Hooks. Addressing them would be a good addition

Will do, good idea!

packages/chrome-extension/src/app/github/api.ts
klbowers: Thanks Tiago, I defined the const inline in the catch block, and removed the commitSHA stuff entirely

packages/chrome-extension/src/app/components/common/customEffects.ts
```diff
@@ -60,5 +60,5 @@ export function useInitialAsyncCallEffect<T>(promise: () => Promise<T>, callback
     return () => {
       canceled = true;
     };
-  }, []);
+  }, [callback, promise]);
```
tiagobento, line 65: This looks correct now, but we need to make sure the arguments are being memoized.. See `IsolatedPrEditor`.. We need to wrap the first argument there in a `useCallback` otherwise this effect will run multiple times.

packages/chrome-extension/src/app/components/common/customEffects.ts
```diff
@@ -27,7 +27,7 @@ export function useEffectAfterFirstRender(func: () => ReturnType<EffectCallback>
     } else {
       firstRender.current = false;
     }
-  }, deps);
+  }, [func]);
```
tiagobento: This one is not right as well... as the `deps` parameter actually is what should be used. I think the correct would be:
```suggestion
  }, [func, ...deps]);
```

packages/chrome-extension/src/app/github/api.ts
```diff
@@ -39,7 +38,10 @@ export function fetchFile(
     .then((res) => (contentType === ContentType.BINARY ? (res.data as any).content : atob((res.data as any).content)))
     .catch((e) => {
       console.debug(`Error fetching ${path} with Octokit. Fallback is 'raw.githubusercontent.com'.`);
-      return fetch(`https://raw.githubusercontent.com/${org}/${repo}/${commitSHA}/${path}`).then((res) =>
+      const modifiedRawGithubUserContentUrl = new URL(
+        `https://raw.githubusercontent.com/${org}/${repo}/${ref}/${path}`
+      );
```
tiagobento: I would inline this variable as well to avoid having to give it a name.

tiagobento (changes requested):
Great job @kbowers-ibm! I left some comments inline.

Please solve the conflicts and it'll be good to merge when green!

packages/chrome-extension/src/app/components/common/customEffects.ts
klbowers: Thanks @tiagobento, I've made the suggested changes. However ESLint is complaining about this specific change with the following, "React Hook useEffect has a spread element in its dependency array. This means we can't statically verify whether you've passed the correct dependencies." Should I just do [func, deps] instead of spreading the deps in the dependency array?

tiagobento: Ahm, I don't think so. To be honest, I haven't done research on how we're supposed to deal with custom hooks dependencies checks... but I guess it's okay for specific cases like this if you ignore the warning with a ESLint directive and provide a nice comment explaining that `deps` is an extension mechanism to allow callers to define their own custom dependencies.

klbowers: Fair enough! Have added the comment and disabled the ESLint directive for that line, thanks @tiagobento!

tiagobento (approved):

PR #2103, opened 2023-12-27
kie-issues#240: Create Helm Chart to install KIE Sandbox and its related services

thiagoelg:
References: https://github.com/apache/incubator-kie-issues/issues/240

This is the first step to getting Helm charts for KIE Sandbox and its related services.

This PR adds a new `kie-sandbox-helm-chart` package with templates and default values fully customizable to install KIE Sandbox, CORS Proxy, and Extended Services on any Kubernetes provider.

Not included in this PR:
- Versioning and publishing the Helm chart (should be tackled in a new PR);
- Automated testing;

[README with a lot more information and possible configurations](https://github.com/thiagoelg/kie-tools/blob/kie-sandbox-helm-charts/packages/kie-sandbox-helm-chart/README.md).


packages/kie-sandbox-helm-chart/install.js
```diff
@@ -0,0 +1,17 @@
+const buildEnv = require("./env");
+const yaml = require("js-yaml");
+const fs = require("fs");
+
+const file = "src/Chart.yaml";
+const doc = yaml.load(fs.readFileSync(file, "utf8"));
+doc.version = buildEnv.env.kieSandboxHelmChart.tag;
```
ljmotta: The dependencies version shouldn't be updated as well? Or updating the `kie_sandbox_helm_chart` version will reflect on the dependencies?

.github/actions/setup-env/action.yml
```diff
@@ -97,6 +97,16 @@ runs:
...
 
+    - name: "Setup Helm (Ubuntu)"
+      if: runner.os == 'Linux'
+      shell: bash
+      env:
+        HELM_VERSION: "v3.13.3"
+      run: |
+        echo "STEP: Install Helm (Ubuntu only)"
+        sudo wget -q https://get.helm.sh/helm-$HELM_VERSION-linux-amd64.tar.gz -O - | tar -xzO linux-amd64/helm > /usr/local/bin/helm
+        sudo chmod +x /usr/local/bin/helm
```
ljmotta, line 108: It's better to remove the tar.gz? or it doesn't make a difference here?

packages/kie-sandbox-helm-chart/src/templates/NOTES.txt
```diff
@@ -0,0 +1,147 @@
+{{- if not .Values.global.ingressSource }}
+
+In order to get KIE sandbox running you need to run these commands:
```
ljmotta:
```suggestion
In order to get KIE Sandbox running you need to run these commands:
```

ljmotta (commented):
Nice @thiagoelg! It looks like you have put a lot of work into this PR! I have just a couple of general observations and others that I'll do inline.
- It have any particular reason to choose the `yaml.helm` instead of `helm.yaml`? I ask this because GitHub and VS Code would have code highlight and interaction out of the box with the second option.
- The Apache header is missing in some files.

thiagoelg:

The files are named `.yaml.helm` because they are not true YAMLs, and prettier tends to accuse formatting errors (blocking commits with the pre-commit hook). The solution was to name them this way and set the VSCode settings.json to use the Kubernetes extension when loading `.helm` files.

.github/actions/setup-env/action.yml
thiagoelg: There's no need, the contents of the download are being piped to the `tar` command, so the .tar.gz is not stored.

packages/kie-sandbox-helm-chart/install.js
thiagoelg: I think they should, nice catch.

We will know more when it's time to publish this chart, but for now, I've updated the install script to change all versions (from the Chart and Subcharts).

tiagobento (approved):

ljmotta (approved):

PR #2105, opened 2023-12-27
NO-ISSUE: Use the DMN marshaller on the DMN Language Service and temporarily disable the Validation panel for DMN 1.3+ on KIE Sandbox

ljmotta:
This PR disables the Validation panel for DMNs spec version 1.3 and above. It also introduce the DMN marshaller on the DMN language service, changing how it's checked for imported models.

This PR fixes a bug where recursive imported models would cause a loop on the KIE Sandbox.


packages/online-editor/src/editor/EditorPage.tsx
```diff
@@ -331,34 +331,48 @@ export function EditorPage(props: Props) {
...
+              },
+            })
+          );
+        });
       },
-    });
-  }, [workspaces, workspaceFilePromise.data?.workspaceFile]);
+      [workspaces, workspaceFilePromise.data?.workspaceFile]
+    )
+  );
```
tiagobento: I don't understand why this conversion to state + effect was necessary. It doesn't look like you're using `modelContent`....

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -109,27 +213,14 @@ export class DmnLanguageService {
...
-      if (importedModelsResources.length > 0) {
-        return [...importedModelsResources, ...(await this.getAllImportedModelsResources(importedModelsResources))];
-      }
-      return [...importedModelsResources];
+  public getDmnSpecVersion(modelContent: string) {
+    if (modelContent === "") {
+      return;
+    }
+    try {
+      return getMarshaller(modelContent)?.originalVersion;
```
tiagobento: `getMarshaller` never returns null or undefined...

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+  modelResources: DmnLanguageServiceImportedModelResources[];
+  normalizedPosixPathRelativeToWorkspaceRoot?: undefined;
+}
+
+interface ImportedModelsByNormalizedPosixPathRelativeToWorkspaceRoot {
+  modelResources?: undefined;
+  normalizedPosixPathRelativeToWorkspaceRoot: string;
+}
+
+type ImportedModelsParams = ImportedModelsByModelResources | ImportedModelsByNormalizedPosixPathRelativeToWorkspaceRoot;
```
tiagobento: These types are really confusing.. and it looks like they're not being reused. Giving names to types can be really hard, especially when they are single-purposed. Please consider inlining them or changing some functions' signatures... it's really hard to understand why you'd need an XOR between a list of models and a path...

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
-        path.posix.join(
-          path.dirname(modelResources.pathRelativeToWorkspaceRoot),
-          path.normalize(importedModel.getAttribute(LOCATION_URI_ATTRIBUTE) ?? "")
-        )
-      )
+    if (!modelResources.content) {
+      return [];
+    }
+
+    const marshaller: DmnMarshaller = getMarshaller(modelResources.content);
```
tiagobento: Unnecessary explicit type.

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+        normalizedPosixPathRelativeToWorkspaceRoot: string;
+      }) => Promise<DmnLanguageServiceImportedModelResources | undefined>;
     }
   ) {}
 
-  private getImportedModelPathRelativeToWorkspaceRoot(
-    modelResources: DmnLanguageServiceImportedModelResources
+  private getImportedModelByNormalizedPosixPathRelativeToWorkspaceRoot(
+    modelResources: DmnLanguageServiceImportedModelResources,
+    importedModelsNormalizedPosixPathRelativeToWorkspaceRoot: Set<string>
```
tiagobento: I know we kind of came up with this name together, but it's really not readable. Let's change it to something more explicit like "visited" or "seen". It stands out enough, and since its scope is really narrow, it's okay to leave out the format of the path, IMHO. There's very little indirection here, and more than that, you can add a nice TSDoc entry here. Let's leave the big names for variables that are part of the domain, not part of some programming mechanism  we need (in this cache, a cycle prevention set)

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+        return importedModelNormalizedPosixPathRelativeToWorkspaceRoot;
+      })
       .filter((e) => e !== null) as string[];
   }
 
-  public getImportedModelPathsRelativeToWorkspaceRoot(
-    modelResources: DmnLanguageServiceImportedModelResources[] | DmnLanguageServiceImportedModelResources
+  private getImportedModelsByNormalizedPosixPathRelativeToWorkspaceRoot(
+    modelResources: DmnLanguageServiceImportedModelResources[] | DmnLanguageServiceImportedModelResources,
+    importedModelsNormalizedPosixPathRelativeToWorkspaceRoot: Set<string>
```
tiagobento: Same here.

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+  private async getImportedModelsByModelResources(
+    modelsResources: DmnLanguageServiceImportedModelResources[]
+  ): Promise<DmnLanguageServiceImportedModelResources[]> {
+    return this.internalGetImportedModelsByModelResources(modelsResources, new Set<string>());
+  }
+
+  public async getImportedModels(args: ImportedModelsParams) {
+    if (args.modelResources) {
+      return this.getImportedModelsByModelResources(args.modelResources);
+    }
```
tiagobento: From the type, this is not optional nor nullable.

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+
+    const importedModelsNormalizedPosixPathRelativeToWorkspaceRoot = new Set<string>();
```
tiagobento: Same here for the big name. I think naming this `seen` or `visited` or even `cyclePrevention` would be much better.

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+    importedModelsNormalizedPosixPathRelativeToWorkspaceRoot.add(args.normalizedPosixPathRelativeToWorkspaceRoot);
+
+    const modelResource = [
+      (await this.args.getModelContent({
+        normalizedPosixPathRelativeToWorkspaceRoot: args.normalizedPosixPathRelativeToWorkspaceRoot,
+      })) ?? ([] as DmnLanguageServiceImportedModelResources[]),
+    ].flatMap((e) => e);
```
tiagobento: Singular name for an array variable. You can inline this one. It will allow you to remove the casting too, as type inference will kick in.

packages/online-editor/src/dmnRunner/DmnRunnerContextProvider.tsx
```diff
@@ -216,14 +216,14 @@ export function DmnRunnerContextProvider(props: PropsWithChildren<Props>) {
 
       const decodedFileContent = decoder.decode(fileContent);
       const importedModelsResources =
-        (await props.dmnLanguageService?.getAllImportedModelsResources([
-          { content: decodedFileContent, pathRelativeToWorkspaceRoot: props.workspaceFile.relativePath },
-        ])) ?? [];
+        (await props.dmnLanguageService?.getImportedModels({
+          normalizedPosixPathRelativeToWorkspaceRoot: props.workspaceFile.relativePath,
+        })) ?? [];
```
tiagobento: I feel like this has an explicit `any` maybe?

packages/online-editor/src/editor/EditorPage.tsx
tiagobento: I think I remember you went through an iteration where the DMN LS instance needed the file content... I think maybe you can revert this change?

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+  normalizedPosixPathRelativeToWorkspaceRoot: string;
+}
+
+type ImportedModelsParams = ImportedModelsByModelResources | ImportedModelsByNormalizedPosixPathRelativeToWorkspaceRoot;
+
 export interface DmnLanguageServiceImportedModelResources {
   content: string;
-  pathRelativeToWorkspaceRoot: string;
+  normalizedPosixPathRelativeToWorkspaceRoot: string;
 }
```
tiagobento: Plural name for an interface that represents a single unit of resource.

tiagobento: I know this was already like that, but since you touched it... :)

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+    }
+
+    return definitions.import
+      .flatMap((importedModel) => {
+        const importedModelNormalizedPosixPathRelativeToWorkspaceRoot = path.posix.normalize(
+          path.posix.join(
+            path.dirname(modelResources.normalizedPosixPathRelativeToWorkspaceRoot),
+            path.normalize(importedModel["@_locationURI"] ?? "")
+          )
+        );
```
tiagobento: Mixing `path.posix` and `path` usages. I'm really not sure why we would use `path.posix`, to be honest. Can you please double-check at the `Windows vs. POSIX` section here? --> https://nodejs.org/api/path.html#pathposix:~:text=COPY-,Windows%20vs.%20POSIX,-%23

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+
+    const marshaller: DmnMarshaller = getMarshaller(modelResources.content);
+    const definitions = marshaller.parser.parse()?.definitions;
+
+    if (!definitions) {
+      return [];
+    }
+    if (!definitions.import) {
+      return [];
+    }
```
tiagobento: Isn't it the same as 

```suggestion
    if (!definitions?.import) {
      return [];
    }
```

?

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
+          )
+        ) {
+          return [];
+        }
+        importedModelsNormalizedPosixPathRelativeToWorkspaceRoot.add(
+          importedModelNormalizedPosixPathRelativeToWorkspaceRoot
+        );
+        return importedModelNormalizedPosixPathRelativeToWorkspaceRoot;
+      })
       .filter((e) => e !== null) as string[];
```
tiagobento: This filter became useless after the change, no? I mean, we're not returning any `null` elements on `flatMap`, right?

packages/dmn-language-service/src/DmnLanguageService.ts
```diff
@@ -20,61 +20,165 @@
...
   ): string[] {
     if (Array.isArray(modelResources)) {
-      return modelResources.flatMap((modelResource) => this.getImportedModelPathRelativeToWorkspaceRoot(modelResource));
+      return modelResources.flatMap((modelResource) =>
+        this.getImportedModelByNormalizedPosixPathRelativeToWorkspaceRoot(
+          modelResource,
+          importedModelsNormalizedPosixPathRelativeToWorkspaceRoot
+        )
+      );
     }
```
tiagobento: Actually, I think we can always force an array here, right? I mean, there's only one usage of this method, and it's private, and this usage always passed an array as argument.

tiagobento (changes requested):
Great contribution @ljmotta! Since we're still kind of breaking ground on this, I left a lot of comments. As we talked, there are big plans for the DMN LS, and I really want to avoid creating more trouble from the start. Thanks!

packages/dmn-language-service/src/DmnLanguageService.ts
ljmotta: A join without the `posix` creates a path with `\` separators, I tested it locally with a node.js script on my Windows machine. But the `path.posix.normalize` is unnecessary. 😅

packages/online-editor/src/editor/EditorPage.tsx
ljmotta: You're right. I made a change to set the model content on the class constructor and I've changed back to avoid one instance per file. I'll change it back to a `useMemo`.

packages/dmn-language-service/src/DmnLanguageService.ts
ljmotta: Indeed, the naming is not helping. This XOR type is used on the `getImportedModel` method. I guess it wasn't a good idea to support two different behaviors in the same method? 😓

packages/dmn-language-service/src/DmnLanguageService.ts
ljmotta: This was necessary to avoid the `DmnMarshaller<DmnMarshallerVersions>` default type that is returned from the `getMarshaller`. I'll add the `upgradeTo: latest` flag.

packages/dmn-language-service/src/DmnLanguageService.ts
tiagobento: Right, right, that makes total sense. Yep, we should be careful then to only use 'path.posix' when dealing with POSIX paths then. The DMN LS especially runs both on Node and browsers, so it makes sense to force POSIX everywhere. As we've been discussing as part of the Paths Standardizatin initiative, dealing with FS paths should be restricted to the last moment possible, when directly interaction with the platform's FS. 

Thx!

tiagobento (approved):

PR #2112, opened 2024-01-05
kie-issues#783: Standardize the use of path strings on Apache KIE Editors

### LINKED ISSUE #783: Standardize the use of path strings on Apache KIE Editors
The Apache KIE Editors have been relying on the "Resource Content API" from `@kie-tools/workspace` from the start. We have several use cases for this that range from DMN Included Models to Serverless Workflow's `spec` directories.

Until now, each Editor has come up with its own rules that define how paths should be:
1.  sent around the "wire" (between Channel and Envelopes)
2. serialized in files
3. interpreted by requests such as `kogitoWorkspace_openFile`.

This has become somewhat unmanageable, because as the relationship between Editors tightens, different components may have different expectations when it comes to paths. This leads to maintenance/debugging hell and ultimately, frustation and a lot of time spent.

This issues aims to address this problem by creating a convention that ALL Apache KIE Editors follow. Here's my proposal:

1. Envelopes that communicate with Channels using the "Resource Content API" must always talk in terms of "paths relative to the workspace root". Whatever "workspace root" is. This is up to the Channel to decide, and Envelope's should not care about that. This includes requests such as `kogitoWorkspace_openFile`, but also `kogitoEditor_contentChanged` (sometimes referred to as `setContent` as well) and of course `kogitoWorkspace_resourceContentRequest` and `kogitoWorkspace_resourceListRequest`.
2. Editor components and files should always serialize paths as relative to its own directory, using the `./` and `../` notations at the start. This removes ambiguity as to where those paths are relative to and remove the need to define a "workspace root" in every context that interacts with this file.
3. How channels treat their paths in storage/git is not in scope here.
4. How paths are displayed on the screen is not in scope here, as different Editors may have different needs.
5. Editor envelopes will know the `workspaceRootAbsolutePath` through the `EditorInitArgs` object that is sent as argument to `kogitoEditor_initRequest`
6. Path strings are ALWAYS POSIX-compliant, meaning that directory separators are forward slashes (`/`). Even on Windows.

---

This spreadsheet tries to summarize all the use-cases involving paths in all Apache KIE Editors. It is a live spreadsheet and will be updated as work on this front progresses.

https://docs.google.com/spreadsheets/d/1XZLP7Zt-1QpQqOrUOTW0SDU-SHUFoDMIp-g4k1ZeCZw/edit#gid=0

tiagobento:
Closes https://github.com/apache/incubator-kie-issues/issues/783

---

In this PR, @ljmotta and I implement the new path strings standard defined by https://github.com/apache/incubator-kie-issues/issues/783, and also some other things:

> NOTE: Chrome Extension and Standalone Editors were NOT tested for DMN, since they'll be upgraded to the new DMN Editor on the next PR.

See the testing spreadsheet: https://docs.google.com/spreadsheets/d/1XZLP7Zt-1QpQqOrUOTW0SDU-SHUFoDMIp-g4k1ZeCZw/edit#gid=486638114

- Configured all VS Code Extensions to be debuggable. See the `watch` scripts on their package.json files.
- Fixed some warnings on `packages/examples`
- Changed the webpack `target` for all VS Code Extensions so that they properly run on all platforms, including Web.
- Fixed minor alignment issues on KIE Sandbox's Editor Page
- Slightly updated the position of some nodes on `Sample.dmn` to make them render better.
- Fixed Boxed Context expressions with undefined `<result>` entries.
- Fixed PMML and JAVA Boxed Functions with empty values.
- Refactored VS Code backends to not depend on Envelope/Channel APIs anymore, but instead have its own default concrete implementation of them. (`WorkspaceChannelApi` --> `VsCodeWorkspaceChannelApiImpl` and `NotificationsChannelApi` --> `VsCodeNotificationsChannelApiImpl`)
- Updated `editor.wordBasedSuggestions` to default to `"off"` instead of `false`, as per VS Code's recommendation.
- Updated the legacy DMN Editor to properly work with PMML files when upgrading to DMN 1.5 and the new DMN Editor.
- Made it possible for DMN files on the legacy DMN Editor to include any file in the workspace.
- Fixed JsInterop binding of GWT/J2CL Editors with the Notifications API.
- Started filtering out files based on `.gitignore` files on VS Code. Especially useful for Java projects which tend to duplicate assets on their `target` folders. We use isomorphic-git, so it works on VS Code web too! Nice one @ljmotta!

----

#### TODO
- [x] Address all FIXME / TODO references introduced on this PR.
- [x] Investigate SceSim error on files at `test` folder
- [ ] Finish testing all cases on the spreadsheet.


tiagobento:
@caponetto or @paulovmr May I ask for your help with executing the tests for SWF/YARD/DASHBUILDER?

yesamer (approved):
@tiagobento @ljmotta I tested scesim-editor (the old one) in VSCode with your changes, my considerations are below:
- The scesim engine is not affected by the changes, the filenames are retrieved stripping away the rest of the path. The model are stored in the dmn-engine memory and the file name is enough to find them.
- The existing scesim files will become stale after this change, and the user will manually need to fix it. That's because the editor will check for the existence of the stored DMN path when opening the editor, and the new logic will fail using the old root-based path.
We agreed to create a PR to simply as match as possible this patching action, and eventually inform the user about the change (changelog, release note, blog post, etc..)

packages/dashbuilder-editor/dev-webapp/App.tsx
```diff
@@ -33,12 +33,13 @@ import "./App.scss";
...
 export type ServerlessWorkflowType = "yml" | "yaml";
 
-const FILE = {
+const FILE: EmbeddedEditorFile = {
   fileName: "test.dash.yaml",
   fileExtension: "dash.yaml",
   getFileContents: function (): Promise<string | undefined> {
     return Promise.resolve("");
   },
+  normalizedPosixPathRelativeToTheWorkspaceRoot: "",
```
caponetto:
```suggestion
  normalizedPosixPathRelativeToTheWorkspaceRoot: "test.dash.yaml",
```
Otherwise, we see `No Editor available for 'dash.yaml' extension` when running the test webapp.

packages/dashbuilder-viewer/dev-webapp/App.tsx
```diff
@@ -39,12 +39,13 @@ pages:
...
     - html: <h1>Welcome to Dashbuilder!</h1>`;
 
-const FILE = {
+const FILE: EmbeddedEditorFile = {
   fileName: "test.dash.yaml",
   fileExtension: "dash.yaml",
   getFileContents: function (): Promise<string | undefined> {
     return Promise.resolve(TEST_CONTENT);
   },
+  normalizedPosixPathRelativeToTheWorkspaceRoot: "",
```
caponetto:
```suggestion
  normalizedPosixPathRelativeToTheWorkspaceRoot: "test.dash.yaml",
```
Otherwise, we see `No Editor available for 'dash.yaml' extension` when running the test webapp.

packages/operating-system/src/paths.ts
```diff
@@ -0,0 +1,9 @@
+import * as __path from "path";
```
caponetto: Can we move this file to `vscode-extension` package? I see this is being used only in there.
Or use `path-browserify` instead of `path`.
My concern is other core packages starting to use this file, which will make polyfill mandatory for webapps. Sometimes polyfill are not under our control.

caponetto (commented):
Thanks for this effort @tiagobento @ljmotta!

packages/operating-system/src/paths.ts
tiagobento: We already has the polyfill configured on our common Webpack config.. Do you mean sometimes we are not using the common Webpack config?

caponetto: We are using some SWF packages outside kie-tools and the webapp in question disables most of polyfills, including the one for path.

tiagobento: Ok, got it. I guess it's fair to have this only in VS Code, given all other channels are not platform-specific, so we don't really have to deal with Windows paths :)

I'll apply the change.

caponetto: Great, thank you ❤️

PR #2123, opened 2024-01-17
kie-issues#424: Add form-generation-tool 'bin' entry point

### LINKED ISSUE #424: Adjust forms-generation-tool to properly release the cli for Linux, macOS and Windows

jomarko:
Closes: https://github.com/apache/incubator-kie-issues/issues/424

For testing this PR, only these methods should be available, as we are about to test ci command of the unpublished package

# 01 npx
See this [documentation](https://exploringjs.com/nodejs-shell-scripting/ch_installing-packages.html#npx). It is enough to run:
- `cd kie-tools/packages/form-generation-tool`
- `npx form-generation-tool` 

# 02 npm link
See this [documentation](https://exploringjs.com/nodejs-shell-scripting/ch_installing-packages.html#npm-link-installing-a-globally-linked-package-locally).
- `cd kie-tools/packages/form-generation-tool`
- `npm link`
- `form-generation-tool`

Once done with testing, do `npm uninstall -g` from `kie-tools/packages/form-generation-tool`

# 03 npm install [local path]
[updated] See this [docuementation](https://exploringjs.com/nodejs-shell-scripting/ch_installing-packages.html#installing-unpublished-packages-via-local-paths) and hint form the [comment](https://github.com/apache/incubator-kie-tools/pull/2123#pullrequestreview-1828045799).

We need to update `version` in packages that are dependencies of the `form-generation-tools` and use `pnpm pack`

-  update `"version": "0.0.0",` to `"version": "0.32.0",` in `kie-tools/packages/uniforms-bootstrap4-codegen/package.json` 
-  update `"version": "0.0.0",` to `"version": "0.32.0",` in `kie-tools/packages/uniforms-patternfly-codegen/package.json` 
- `cd kie-tools/packages/form-generation-tool`
- `pnpm pack`
- `npm i -g kie-tools-form-generation-tool-0.0.0.tgz`
- `form-generation-tool`

Once done with testing, do `npm uninstall -g` from `kie-tools/packages/form-generation-tool`

### DIFF
--- a/packages/form-generation-tool/README.md
+++ b/packages/form-generation-tool/README.md
@@ -30,13 +30,26 @@ After the command has finished, go to `packages/form-generation-tool/dist` folde
 
 ## Running the CLI
 
-In the command line just execute the CLI binary:
+If you built the `form-generation-tool` package as described above locally, then in the command line just execute the CLI binary:
 
 ```shell script
 ./form-generation-cli-linux
 ```
 
-This command will start a wizard to help you generate the forms:
+For those, who do not want to build `form-generation-tool` package locally, they can install last published version and then run it:
+
+```shell script
+npm i -g @kie-tools/form-generation-tool
+form-generation-tool
+```
+
+For those, who want to try `form-generation-tool` without instalation they can start it as:
+
+```shell script
+npx @kie-tools/form-generation-tool
+```
+
+All commands will start a wizard to help you generate the forms:
 
 1. First set the path to your Kogito Project.
 
--- a/packages/form-generation-tool/bin.js
+++ b/packages/form-generation-tool/bin.js
@@ -21,4 +21,4 @@
 /**
  * This file is used as entry point of the form-generation-tool cli command
  */
-require("../../dist/index");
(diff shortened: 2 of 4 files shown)


tiagobento (changes requested):
Hi @jomarko. Thanks for this PR. I especially like the investigation you did. My only request is that we move `src/cli/cli.js` to `bin.js` at the root of the package and change the reference to `./dist/index`.

Regarding case 3, `npm install`, what is happening is that you are installing the package without "npmzing" it first. Let me explain:

We use `pnpm`'s feature of linking packages inside the monorepo through the special version `workspace:*`. This is not something NPM recognizes, so it fails to install the package directly. To test it, you can use `pnpm pack`. If I'm not mistaken, this will resolve the `workspace:*` version to whatever version is declared in the root package.json, which will be `0.0.0`.

jomarko:
@tiagobento feedback incorporated, also was able now to manually check 3th way of installation. Thank you many times!

tiagobento (approved):

PR #2124, opened 2024-01-17
kie-issues#831: Make the new DMN Editor actually generate an SVG representing the currently open DRD

### LINKED ISSUE #831: Make the new DMN Editor actually generate an SVG representing the currently open DRD.
During #577 we didn't implement an SVG-generation mechanism for the new DMN Editor. This task is to fix that, making it possible to generate SVGs from the currently open DRD.

tiagobento:
Closes https://github.com/apache/incubator-kie-issues/issues/831

### DIFF
--- a/packages/dmn-editor-envelope/src/DmnEditorFactory.tsx
+++ b/packages/dmn-editor-envelope/src/DmnEditorFactory.tsx
@@ -55,28 +55,7 @@ export class DmnEditorInterface implements Editor {
   // Not in-editor
 
   public getPreview(): Promise<string | undefined> {
-    return Promise.resolve(`
-<svg
-  xmlns="http://www.w3.org/2000/svg"
-  xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1"
-  width="540" height="540" viewBox="0 0 540 540" xml:space="preserve">
-  <g transform="matrix(1 0 0 1 270 270)" id="ee1530d3-d469-49de-b8ad-62ffb6e5db7a"></g>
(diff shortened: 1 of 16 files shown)


thiagoelg (approved):

jomarko (commented):
Thank you for the PR @tiagobento . I focused more on manual check in VS Code. Should be this generate svg feature tested also on some other channel? If yes, please let me know how,

My findings are:
# 01 node text color 
Is ignored. See the attached screenshot. In the SVG all node text are black, while most nodes in original diagram uses different node text color.
image

# 02 node styling applied unexpectedly
is probably unrelated to your PR, but not sure, so sharing it here. If you confirm it is unrelated, we can report, if the team will agree, separately.

- Open some diagram that contains some non default nodes, it means their colors and fonts are not only black and white. 
- Then start clicking these non default nodes. 
- They will be automatically re-styled do default black and white.

The same is doable also vice versa
- Open some node styling properties
- Change the node color or font
- Select another node
- This last selected node will be automatically re-styled as the previously selected node

ljmotta:
@jomarko I could reproduce this issue on the daily-dev environment. Changing the node fill color, stroke color and font color resets to the default. It appears the DMN Editor is not saving the colors anymore.

tiagobento:
@jomarko Thanks a lot for testing it. It should be fixed now.

ljmotta:
@tiagobento @jomarko  I've create an [issue](https://github.com/apache/incubator-kie-issues/issues/842) to track the node style bug

jomarko (approved):
Thank you, I think I can not help with code review in this case, as it is quite big for my current skills. Approving as findings from the manual check were addressed.

yesamer (approved):

PR #2126, opened 2024-01-19
kie-issues#151: Make browser extensions work on other GitHub instances, like github.ibm.com

### LINKED ISSUE #151: Make browser extensions work on other GitHub instances, like github.ibm.com
Our browser extensions currently have a fixed domain on their manifests, which makes it only work for github.com and its subdomains.

When accessing other GitHub instances, like github.ibm.com, we should provide a way to run our browser extensions and inject our custom Editors there.

jomarko:
Closes: https://github.com/apache/incubator-kie-issues/issues/151

Changes introduced here:
- chrome-extension user can now setup token for any github instance, not only the https://github.com
- chrome-extension user can now browse any github instance, not only the the https://github.com
- chrome-extension user can open extension settings and list the allowed websites for the extension
- chrome extension user will see logs, if the extension doesn't find required github ui web elements

------
Please notice, user is blocked from using private repositories on his/her custom github instance, as the sandbox currently support only the https://github.com instance
Screenshot 2024-01-23 131605
Screenshot 2024-01-23 131554


packages/chrome-extension/src/app/Dependencies.ts
```diff
@@ -62,7 +62,7 @@ export class Dependencies {
 
   public readonly openRepoInExternalEditor = {
     buttonContainerOnRepoHome: () => {
-      return document.querySelector("#repository-details-container .pagehead-actions") as HTMLElement | null;
+      return document.querySelector(".pagehead-actions") as HTMLElement | null;
```
jomarko, line 65: this change to a less specific locator is an effort to increase a chance the extension will work on multiple instances, even if some of them is not the latest github instance

packages/chrome-extension/src/app/components/common/GitHubContext.tsx
```diff
@@ -70,10 +84,10 @@ export const GitHubContextProvider: React.FC<{}> = (props) => {
 
   useEffect(() => {
     if (token) {
-      octokitInstance = new Octokit({ auth: token });
+      octokitInstance = new Octokit({ auth: token, baseUrl: getGitHubApiBaseUrl(window.location.origin) });
       console.debug("Token found");
     } else {
-      octokitInstance = new Octokit();
+      octokitInstance = new Octokit({ baseUrl: getGitHubApiBaseUrl(window.location.origin) });
```
jomarko, line 87: from now on, we must to setup the `Octokit` library to use proper url, by default it uses `https://api.github.com`

packages/chrome-extension/src/index.ts
```diff
@@ -91,37 +91,66 @@ function init(globals: Globals) {
   const fileInfo = extractFileInfoFromUrl();
   const pageType = discoverCurrentGitHubPageType();
 
+  if (!globals.dependencies.all.octiconMarkGitHub()) {
+    globals.logger.warn("This is not the GitHub web page.");
+    return;
+  }
```
jomarko: if there is no github icon on the page, we are sure user is not browsing github web page

packages/chrome-extension/src/index.ts
```diff
@@ -91,37 +91,66 @@ function init(globals: Globals) {
...
-      ...globals,
-      pageType,
-      className: "btn ml-2 d-none d-md-block",
-      container: () => globals.dependencies.openRepoInExternalEditor.buttonContainerOnRepoFilesList()!,
-      fileInfo,
-    });
+    if (!globals.dependencies.openRepoInExternalEditor.buttonContainerOnRepoFilesList()) {
+      globals.logger.warn(
+        "The extension can not display all buttons properly. Please be sure you use the extension with the latest GitHub instance."
+      );
```
jomarko: here and bellow I added 4 similar `if` statements. that I feel is not ideal. however I was not able to do it differently.

My original idea was to iterate all `globals.dependencies.openRepoInExternalEditor` locators in single place and check if some of them does not exist. in such case I wanted to inform user, some features may be not available.

but for example `buttonContainerOnRepoFilesList` doesn't exist until proper page is opened.

So in final I added 4 checks, where I check only locator required by given page view.

jomarko:
### Checklist
With extension allowed on all sites:
- [x] open public github instance, repository view
- [x] open public github instance, asset view
- [x] open public github instance, PR view
- [x] open public github instance, home screen
- [x] open public github instance, settings view
- [x] open custom github instance, repository view
- [x] open custom github instance, asset view
- [x] open custom github instance, PR view
- [x] open custom github instance, home screen
- [x] open custom github instance, settings view
- [x] open non github instance page

With extension allowed on specific site:
- [x] open this specific site
- [x] open non allowed site

packages/chrome-extension-pack-kogito-kie-editors/manifest.dev.json
```diff
@@ -7,8 +7,8 @@
     {
       "run_at": "document_idle",
       "js": ["content_scripts/github.js"],
-      "matches": ["https://*.github.com/*"],
-      "all_frames": true
+      "matches": ["https://*/*", "http://*/*"],
+      "all_frames": false
```
jomarko, line 11: `all_frames` as `false` helps us to not produce similar logs, in other words do not try to inject 'Open in KIE Sandbox' button into nested frames, where we do not need it.

image

packages/chrome-extension/src/app/components/common/GitHubContext.tsx
```diff
@@ -53,6 +53,20 @@ export function getCookie(name: string) {
...
+    apiBaseUrl = "https://api." + baseUrl.slice(8);
+  } else if (baseUrl.indexOf("http://") === 0) {
+    apiBaseUrl = "http://api." + baseUrl.slice(7);
+  } else {
+    console.error("Not supported protocol");
+    return;
+  }
+
+  return apiBaseUrl;
+}
```
tiagobento:
```suggestion
export function getGitHubApiBaseUrl(baseUrl: string) {
  if (baseUrl.startsWith("https://")) {
    return "https://api." + baseUrl.slice(8);
  } else if (baseUrl.startsWith("http://")) {
    return "http://api." + baseUrl.slice(7);
  } else {
    console.error(`Unsupported protocol for GitHub API base URL: ${baseUrl}`);
    return;
  }
}
```

tiagobento (approved):
I have one single suggestion that doesn't have a lot of impact, I just think looks nicer and is easier to read...

Overall, thank you for this PR, Jozef. I am especially impressed with the clarity of the code and the comments you left here to aid reviewers. Nice job! And keep digging deep into stuff, the whole process you went through to get here was very nice to see as well. Great work!

jomarko:
Thank you @tiagobento , I pushed your proposed changes. I like them and I agree, your version is shorter and better readable.

yesamer (approved):

PR #2131, opened 2024-01-26
kie-issues#785: Integrate the Boxed Expression Editor into the new DMN Editor properties panel making it possible to edit its cells

### LINKED ISSUE #403: Using DMN Runner with Included Model fails to run
When using an included model with the DMN runner within Sandbox, you cannot use a DMN that incorporates an included model in  the DMN model you're currently on.

Test Scenario within Sandbox
1. Create DMN that is to be used just as a model to be shared across multiple DMN files (dmn1.dmn)
2. Create the DMN that a decision (dmn2) will be created from and include dmn1.dmn as an included model
3. Go to Sandbox/Canvas and try to run a decision that uses the included model and it will fail saying it cannot find the model.

Error returned in JavaScript console:
```log
Uncaught (in promise) Error: File '/fs_v1__ee23443a-fba5-493d-a252-60878b077e79/dmn1.dmn' doesn't exist
    at e.callback (index.js:8:2210641)
    at e.receive (index.js:8:2212425)
    at Object.receive (index.js:8:2207714)
    at ready.e.workspacesWorker.port.onmessage (index.js:8:4615565)
```

In best practices of models, this is a common pattern to use shared models amongst multiple decisions.

### LINKED ISSUE #785: Integrate the Boxed Expression Editor into the new DMN Editor properties panel making it possible to edit its cells
On the DMN Editor's Boxed Expression Editor, the properties panel should allow editing properties of Boxed Expression cells. When nothing is selected, the properties of the node containing the opened Boxed Expression should be displayed.

Currently, the selected cell of the Boxed Expression Editor isn't correctly exposed, this should be fixed before developing the properties panel.
Examples of incorrect behaviors: 

-  selecting a Boxed Invocation parameter will return an empty `selectedObjectId`.
- selecting a cell of a Boxed Context (e.g. its context entry name) and then selecting the boxed expression associated with it, will first return an `undefined` on the `selectedObjectId`, and then the correct id.

Tasks
- [ ] Correctly expose the selected Boxed Expression Editor cell.
- [ ] Implement the Boxed Expression Properties Panel enabling the user to edit the cells from the panel.

### LINKED ISSUE #1018: React-Based DMN Editor: Data type not imported correctly
if the `itemDefinition` does not contain the id attribute, it is not correctly shown/managed in the editor

Unknown

Unknown-2

From above, it seems optional, so editor should be able to work without it (being a namedElement, editor should rely on required “name” attribute)

[XSD](https://github.com/apache/incubator-kie-drools/blob/4004fe9bf745c5eae18223cd1d3a42d8efd4566e/kie-dmn/kie-dmn-validation/src/main/resources/org/kie/dmn/validation/org/omg/spec/DMN/20230324/DMN15.xsd) excerpt (`tItemDefinition → tNamedElement → tDMNElement`)

```xml
<xsd:complexType name="tItemDefinition">
	<xsd:complexContent>
		<xsd:extension base="tNamedElement">
...
<xsd:complexType name="tNamedElement">
		<xsd:complexContent>
			<xsd:extension base="tDMNElement">
              <xsd:attribute name="name" type="xsd:string" use="required"/>
...
<xsd:complexType name="tDMNElement">
	...
	<xsd:attribute name="id" type="xsd:ID" use="optional"/>
```

### LINKED ISSUE #1122: DMN Editor: It is possible to reorder **Input data** and **Input decisions** for the included Decision Service node
To reproduce the issue please follow these steps:
- Open DMN Editor
- Add a Decision Service node - **DS Node**
- Add single Decision node into its **output** section - **Decision**
- Add two Input Data nodes to the diagram - **Input One** and **Input Two**
- Connect both **Input One** and **Input Two** with the **Decision**
- Create new DMN model in the same folder
- Include the previous DMN model into it, use **Included models** tab for it
- Add the **DS Node** to the current diagram
- Open the properties panel for the **DS Node**

Current: User can drag and drop **Input data** items in the second diagram
Expected: User can not drag and drop **Input data** items in the second diagram

### Fixed by
- apache/incubator-kie-tools#2131

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/785; and
Closes: https://github.com/apache/incubator-kie-issues/issues/1018; and
Closes: https://github.com/apache/incubator-kie-issues/issues/403; and
Closes: https://github.com/apache/incubator-kie-issues/issues/1122

Deployed version on https://ljmotta.github.io/kie-tools-demos/kie-issues-785/#/

This PR integrates the boxed expression editor into the new dmn editor properties panel.

Some general changes:
- Improve the read-only constraint enum component
- Fix small bugs on beeToDmn.ts
- Regenerates all IDs and adds IDs to all missing xsd:ID objects on a DMN file. This action prevents cells with an `"@_id" === undefined`.
- Fix Storybook source maps.
- Disable shortcuts on `boxed-expression-component` Storybook.
- Fix small bugs on `boxed-expression-component`.


jomarko:
Hello @ljmotta is available some deployed version of editor with your changes please? Would like to do some manual tests. Or do I need to build it locally on my own?

ljmotta:
@jomarko The PR is not ready for review yet. As soon it's ready for review I'll publish it.

jomarko:
### 01 Decision Table Header Constraints
I have an observation of this difference. In the legacy editor, user has an option to define Decision table header constraints. See **high, medium, low** in the attached picture below. It is a way to 'constraint' column values without having defined a custom data type with constraint. Also I think DMN Runner and extended services uses this information for displaying validation tips and messages to the user.
Screenshot 2024-04-15 104744

While in the new DMN Editor we do not have **Constraint** in the properties panel. We have **Input values** instead. Is this a replacement of the **Constraint** from the legacy editor? If yes, what is your opinion about bringing some tooltip, placeholder, .... to make easier for users to fill in this input fields?
Screenshot 2024-04-15 104806

jomarko:
### 02 Output Constraint
For Output header cells, we have a **Constraint** section, however only if the column is bound to a **Structure**? Is that expected? Because currently for columns bound to built -in data types, we do display **Constraint** section at all.

Maybe of options below would make the situation more 'clear':
- Do not display **Constraint** for any column, if that is currently never supported
- Display **Constraint** always, and say explicitly if that is supported or not for a given column
- maybe some other option

Screenshot 2024-04-15 110321

jomarko:
### 03 Data Type Constraint and Monaco Suggestion
probably unrelated to the changes in the PR, however wanted to mention it here so we track it somehow.
Please notice adding data type constraint in combination of monaco suggestion causes an 'invalid Z index' issue? we see a button over suggestions.
Screenshot 2024-04-15 111623

jomarko:
### 04 Commit the change
Imagine a situation like below, there is a cell selected and user put some description, **a** in our screenshot. Now imagine user want to proceed to the next cell, so he clicks to the next cell, the problem, maybe, maybe it is intended, is if user immediately click to another cell, then the **a** is committed into this second selected cell.

**Initial state**
Screenshot 2024-04-15 114114

**Click to next cell**
Screenshot 2024-04-15 114140

jomarko:
### 05 Function parameters
Are in the properties panel written the same font as other text, maybe small change of font would make the properties panel more readable?

Screenshot 2024-04-15 115808

vs. using bold
Screenshot 2024-04-15 121244

jomarko:
### 06 included Nodes
I have two questions about included nodes.

#### 06.01
Is it expected the data type of the included node can be changed, please?
Screenshot 2024-04-15 121746

#### 06.02
Is it expected the boxed expression of the included node can not be displayed?

packages/boxed-expression-component/.storybook/manager.ts
```diff
@@ -22,4 +22,6 @@ import { KieToolsTheme } from "@kie-tools/storybook-base/dist/config/theme";
 
 addons.setConfig({
   theme: KieToolsTheme,
+  // Shortcuts conflicts with the cell behavior
+  enableShortcuts: false,
```
jomarko, line 27: What shortcuts do we disable please? I am able to use arrow keys, or undo.

packages/dmn-editor/src/propertiesPanel/BoxedExpressionPropertiesPanelComponents/DecisionTableRootCell.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { BoxedExpressionIndex } from "../../boxedExpressions/boxedExpressionIndex";
+import { DMN15__tDecisionTable } from "@kie-tools/dmn-marshaller/dist/schemas/dmn-1_5/ts-gen/types";
+import { useBoxedExpressionUpdater } from "./useUpdateBee";
+import { ClipboardCopy } from "@patternfly/react-core/dist/js/components/ClipboardCopy";
+import { FormGroup } from "@patternfly/react-core/dist/js/components/Form";
+import { useDmnEditorStore } from "../../store/StoreContext";
+
+type DecisionTableRoot = Pick<
+  DMN15__tDecisionTable,
+  "@_label" | "description" | "@_typeRef" | "@_outputLabel" | "@_aggregation" | "@_hitPolicy"
```
jomarko, line 32: I would like to ask why `description` does not need to be `@_description`

jomarko (commented):
### 07 Tests
Do we track somehow e2e playwright tests, when will we add them?

jomarko:
### 08 Back Compatibility
Opening an old model [1] has different result on legacy editor vs the new editor. please notice the decision table output header cell.

**Legacy Editor**
Screenshot 2024-04-15 124051

vs. **New Editor**
Screenshot 2024-04-15 131001



[1] https://raw.githubusercontent.com/redhat-cop/businessautomation-cop/master/testing-examples/cucumber-quarkus-example/src/main/resources/pricing.dmn

ljmotta:
@jomarko Thanks a lot for your review! I'll answer the raised observations inline:

1 - We have opted to use only constraints from custom data types, instead of letting create constraints on the Decision Table header. 
2 - As mentioned on 1, to create a constraint it’s necessary to create a custom data type, and not a `Structure`.
3 - It’s not related to this PR, I've confirmed on the `dailydev`. I guess we need to create a separated issue, WDYT?
4 - This is a serious problem. I will take a look into it.
5 - I guess it would be good to incorporate this.
6 -
  - It shouldn’t be possible to change the node data type! I could confirm this bug on the `dailydev`. I guess we'll need to track this on a different issue as well as it's not related to the boxed expression editor. I'll open the issue for it.
  - Regarding the boxed expression, we have disabled the option to open it.

7 -
  - We’re not tracking then, but I’ll add it to the 945 epic.
  - I’ll improve the comment. The disabled shortcuts are from Storybook and not from the Boxed Expression. It was causing a conflict with the Monaco editor.
  - The `description` is a XSD `element`, and the `@_` is a XSD `attribute`.

8 - I could confirm this bug on the `dailydev` environment too. It looks like a problem on the expression object. I'll open the issue for it.

ljmotta:
@jomarko I've incorporated all your comments (4,5,7). Also, I've fixed all issues you've pointed out (3,6,8). The deployed page was updated. Please refresh the cookies to ensure you have the latest version (ctrl+shift+R)

jomarko:
Thank you for all updates and quick fix @ljmotta . I can confirm all points are fixed, however **08** seems still a little bit not ideal. When I open the model, in the output header of the decision table, the name is correct - **Best Proce** however it doesn't show **number** as data type, it still points to **Undefined**.
Screenshot 2024-04-16 145526

Is that expected somehow? I mean do older DMN versions specify DT output data type somehow incompatibly with DMN 1.5 spec so we can not derive it?

ljmotta:
@jomarko I overlooked the issue, sorry. I've already addressed and updated the deployed version.

tiagobento (approved):

jomarko (approved):
Thank you, @ljmotta! I think the original issue/comment **08** is solved very well now.

However, I found an interesting behavior, probably an issue with decision table top level output header cell. To me it seems to worth investigate in a separate ticket to not extend scope of this PR.

### DT-simple-01
Screenshot 2024-04-18 195903

in this screenshot is an issue, that once node with DT expression is opened, the properties panel show the node data type, **Any** in the screenshot, but it is not synced with the boxed expression, see **<Undefined>** there.

### DT-simple-02
Screenshot 2024-04-18 200015
In this screenshot is demoed, that the sync mechanism works well, if the change is done in boxed expression editor, we see the properties panel was synced. See the  **boolean**

However, what is strange in properties panel, we have all of these: **Decision Name**, **Decision Type**, **Column Name** and **Column Type**

### DT-complex-01
Screenshot 2024-04-18 200100
Is similar sync problem as DT-simple-01, again see **Any** type. 


The three listed issues/observations should be tracked/addressed separately form this PR to not extend the PR scope and make potential next PRs verification easier. I am approving this PR. Thank you, @ljmotta !

ljmotta:
@jomarko The `DT-simple-01` and `DT-complex-01` were a bug I've introduced on the previous commits, and it affected all expressions. Thanks for spotting it. Regarding `DT-simple-02`, when a Decision Table has only one output column, the Decision Table Output Header cell represents two cells. The Decision and the Column itself, thus we have the `Decision Name`, `Decision Type`, `Column Name` and `Column Type`. This becomes evident when we create a new output column, and now we will not have the `Decision Name`, `Decision Type` anymore.

tiagobento:
@ljmotta It just occurred to me that we might be better off waiting for 

- https://github.com/apache/incubator-kie-issues/issues/236

before merging this PR. We need to make sure the Properties Panel works for all expression types. Meanwhile, please do revisit Conditional and Iterator expressions introduced recently.

jomarko:
@ljmotta thank you for explanation of the `DT-simple-02`. I have rechecked `DT-simple-01` and `DT-complex-02` using [1] and the sync mechanism works without an issue.

@tiagobento thank you for reminder about new expression types. I can confirm using [1] the **For**, **Every** and **Some** cause an issue [2]. On the other hand, **If/Conditional** is working ok.


[1]
https://ljmotta.github.io/kie-tools-demos/kie-issues-785

[2]
image

ljmotta:
@jomarko This issue was on main, I need to rebase and update my version. But I'll need to fix a few more things on the `idRandomizer` file.

jomarko:
I can still see the issue for: **For**, **Every** and **Some** expression. I guess there are still some pending updates on your site? Please let me know once I should recheck the behavior.

ljmotta:
@jomarko I've updated the deployed version. The last time you've tried it was missing a few adjustments. Now, it should work with the Iterator expressions.

ljmotta:
@jomarko I've finish with the changes, and I've re-deployed the PR.

PR #2136, opened 2024-01-30
kie-issues#808: kn-plugin cli version 10.0.0

ederign:
Guys, I think we are ready to merge it. There are some pending tasks yet (moving to new artifacts also Tiago comment related to images), but those can be addressed on later PRs.


ederign:
Binaries: https://drive.google.com/drive/folders/1nY-GfnpyRETSj2AxN8qB25YsfLvAQJ7Q?usp=drive_link

packages/kn-plugin-workflow/pkg/command/gen_manifest.go
```diff
@@ -42,6 +43,9 @@ func NewGenManifest() *cobra.Command {
 	# Persist the generated Operator manifests on a default path (default ./manifests)
 	{{.Name}} gen-manifest
 
+	# Specify a custom target namespace. (default: kubeclt current namespace; --namespace "" to don't set namespace on your manifest)
```
caponetto, line 46: Wondering if it'd be better to provide a `--skip-namespace` arg or something like that.
It might avoid confusion since it's self-explanatory.
I understand this adds extra work, so it's just an idea.

ederign: I'll do that. Actually, I was also considering this extra option. So it would be --namespace {name} or --skip-namespace. That is what you mean?

caponetto: Exactly!

masayag:
Can we add two more flags to `gen-manifests`:
```
--image - Specify an image  to be used as the workflow's container image ( sets a value for .spec.podTemplate.container.image)
```
and an option to specify the value for the `sonataflow.org/profile` annotation (any of prod/dev/nil or future gitops when/if be added).

masayag:
Can the generated configmaps use the `|` to allow a multi-line string to improve readability both for the -props and -resources configmaps?
cc @dmartinol

ederign:
@masayag, we can accommodate both on the next releases. https://issues.redhat.com/browse/KOGITO-10052

packages/kn-plugin-workflow/env/index.js
```diff
@@ -34,17 +34,17 @@ module.exports = composeEnv([require("@kie-tools/root-env/env")], {
...
     KN_PLUGIN_WORKFLOW__quarkusVersion: {
       name: "KN_PLUGIN_WORKFLOW__quarkusVersion",
-      default: "2.16.10.Final",
+      default: "3.2.9.Final",
       description: "Quarkus version to be used when creating the SonataFlow project",
     },
     KN_PLUGIN_WORKFLOW__devModeImage: {
       name: "KN_PLUGIN_WORKFLOW__devModeImage",
-      default: "quay.io/kiegroup/kogito-swf-devmode:1.44",
+      default: "quay.io/kiegroup/kogito-swf-devmode:latest",
```
tiagobento: We should use the `999-20240218-SNAPSHOT` tag here too, no?

tiagobento: Alright, so there is no timestamped snapshot for this particular image, and we would need to use the `kogito-swf-devmode-nightly` version of it. However, with https://github.com/apache/incubator-kie-issues/issues/1001, we won't need to deal with it anymore, just pointing to the local image built on `kie-tools` directly.

packages/kn-plugin-workflow/package.json
```diff
@@ -22,7 +22,7 @@
     "build:darwin:arm": "pnpm setup:env make build-darwin-arm64",
     "build:dev": "rimraf dist && pnpm build",
     "build:linux": "pnpm setup:env make build-linux-amd64",
-    "build:prod": "rimraf dist && run-script-os && pnpm test",
+    "build:prod": "rimraf dist && run-script-os && pnpm test && pnpm test:e2e",
```
ederign: @tiagobento ^

paulovmr (approved):

caponetto (approved):

tiagobento (approved):

PR #2139, opened 2024-02-01
NO-ISSUE: Fix Windows `boxed-expression-component` Storybook start script

ljmotta:
## On this PR

This PR creates a specific start script for Windows using `powershell`.

## Problem
To evaluate the `$(build-env ... )` on windows is required the usage of the `powershell` script, which requires the command to be wrapped on double quotes. Wrapping the `storybookArgs` on single quotes prevents the `$(build-env ... )` evaluation. Therefore, the usage of an environment variable.


tiagobento (approved):

jomarko (approved):
Thank you @ljmotta 

Windows run seems playwright tests were started, some skipped, but no failures. So probably we are fine.

https://github.com/apache/incubator-kie-tools/actions/runs/7743209943?pr=2139

PR #2146, opened 2024-02-07
kie-issues#669: The DMN Editor should allow using the new alternative representation of InputData nodes (DMN 1.5) 

### LINKED ISSUE #669: The DMN Editor should allow using the new alternative representation of InputData nodes (DMN 1.5)
As described at
- #231 

DMN 1.5 allows changing the shape of InputData nodes to an alternative representation that is more aligned with the one of BPMN.

The DMN Editor should provide a new entry on the Properties panel for users to check whether an InputData node should use this alternative representation or not. If true, the shape of the node should change.

ljmotta:
Closes https://github.com/apache/incubator-kie-issues/issues/669

## Description
This PR adds the alternative input data node shape into the new DMN Editor. Now, it's possible to change the input data node shape on each DRD separately.
To accommodate those changes, the DRD selector was changed, showing the selected DRD properties. Also, the dropdown doesn't close on the DRD select, giving the user the option to check the properties before manually closing it.

## Video
https://github.com/apache/incubator-kie-tools/assets/24302289/bc3e3bc4-6f84-4a4d-8f1d-4bf4c6c38fcc


packages/dmn-editor/src/boxedExpressions/BoxedExpression.tsx
```diff
@@ -273,7 +273,11 @@ export function BoxedExpression({ container }: { container: React.RefObject<HTML
 
   ////
 
-  const Icon = expression ? NodeIcon(getNodeTypeFromDmnObject(expression.drgElement)) : () => <></>;
+  const Icon = useMemo(
+    () =>
+      expression?.drgElement ? NodeIcon({ nodeType: getNodeTypeFromDmnObject(expression.drgElement) }) : () => <></>,
+    [expression?.drgElement]
+  );
```
tiagobento: IMHO we should be throwing in this case. This is really, really exceptional behavior.

packages/dmn-editor/src/diagram/Diagram.tsx
```diff
@@ -511,7 +530,7 @@ export const Diagram = React.forwardRef<DiagramRef, { container: React.RefObject
           state.diagram.ongoingConnection = undefined;
         });
       },
-      [dmnEditorStoreApi, container, reactFlowInstance, externalModelsByNamespace]
+      [dmnEditorStoreApi, container, reactFlowInstance, externalModelsByNamespace, isAlternativeInputDataShape]
```
tiagobento: Looks like you don't really need to reactivelly change this hook when `isAlternativeInputDataShape`, as you have the state in your hands. You can access `state.computed(state).isAlternativeInputDataShape()`, right?

packages/dmn-editor/src/diagram/DrdSelectorPanel.tsx
```diff
@@ -21,66 +21,147 @@ import * as React from "react";
...
+              </button>
+              <br />
+            </React.Fragment>
+          ))}
+        </div>
+
+        <div style={{ gridArea: "header-properties" }}>
+          <Title headingLevel="h3" style={{ height: "36px" }}>
+            {getDefaultDrdName({ drdIndex })}
+          </Title>
```
tiagobento: This will always render either "Default DRD" or "Untitled DRD". We want the actual name of the DRD showing, right? See line 100.

packages/dmn-editor/src/diagram/DrdSelectorPanel.tsx
```diff
@@ -21,66 +21,147 @@ import * as React from "react";
...
+        </div>
+        <div style={{ gridArea: "content-list" }} className={"kie-dmn-editor--drd-list"}>
+          {thisDmn.model.definitions["dmndi:DMNDI"]?.["dmndi:DMNDiagram"]?.map((drd, i) => (
+            <React.Fragment key={drd["@_id"]!}>
+              <button
+                className={i === diagram.drdIndex ? "active" : undefined}
+                onClick={() => {
+                  dmnEditorStoreApi.setState((state) => {
+                    state.diagram.drdIndex = i;
+                    // state.diagram.drdSelector.isOpen = false;
```
tiagobento: Either uncomment or delete.

packages/dmn-editor/src/diagram/DrdSelectorPanel.tsx
```diff
@@ -21,66 +21,147 @@ import * as React from "react";
...
+              <FormGroup label={"Input Data node shape"}>
+                <ToggleGroup
+                  aria-label="Tweak the shape of the input data node"
+                  className={"kie-dmn-editor--drd-properties--input-data-node-shape"}
+                >
+                  <ToggleGroupItem
+                    text="Classic"
+                    icon={<InputDataIcon padding={"2px 0 0 0"} height={22} />}
+                    buttonId="classic-input-node-shape"
+                    isSelected={isAlternativeInputDataShape === false}
```
tiagobento, line 130: I was writing the comment to ask to change it to `!isAlternativeInputDataShape`, but really something about this specific line makes it easier to write it this way. Well done :)

packages/dmn-editor/src/diagram/maths/DmnMaths.ts
```diff
@@ -194,13 +194,13 @@ export function getContainmentRelationship({
   container: DC__Bounds;
   divingLineLocalY?: number;
   snapGrid: SnapGrid;
-  containerMinSizes: (snapGrid: SnapGrid) => DC__Dimension;
-  boundsMinSizes: (snapGrid: SnapGrid) => DC__Dimension;
+  containerMinSizes: (args: { snapGrid: SnapGrid; isAlternativeInputDataShape?: boolean }) => DC__Dimension;
+  boundsMinSizes: (args: { snapGrid: SnapGrid; isAlternativeInputDataShape?: boolean }) => DC__Dimension;
```
tiagobento: Why are those optional here?

packages/dmn-editor/src/diagram/nodes/DefaultSizes.ts
```diff
@@ -24,36 +24,52 @@ import { NodeType } from "../connections/graphStructure";
...
-export const MIN_NODE_SIZES: Record<NodeType, (snapGrid: SnapGrid) => DC__Dimension> = {
-  [NODE_TYPES.inputData]: (snapGrid) => {
+export const MIN_NODE_SIZES: Record<
+  NodeType,
+  ({
+    snapGrid,
+    isAlternativeInputDataShape,
+  }: {
+    snapGrid: SnapGrid;
+    isAlternativeInputDataShape?: boolean;
```
tiagobento: Why optional?

packages/dmn-editor/src/diagram/nodes/DefaultSizes.ts
```diff
@@ -91,57 +107,73 @@ export const MIN_NODE_SIZES: Record<NodeType, (snapGrid: SnapGrid) => DC__Dimens
...
-export const DEFAULT_NODE_SIZES: Record<NodeType, (snapGrid: SnapGrid) => DC__Dimension> = {
-  [NODE_TYPES.inputData]: (snapGrid) => {
+export const DEFAULT_NODE_SIZES: Record<
+  NodeType,
+  ({
+    snapGrid,
+    isAlternativeInputDataShape,
+  }: {
+    snapGrid: SnapGrid;
+    isAlternativeInputDataShape?: boolean;
```
tiagobento: Why optional?

packages/dmn-editor/src/diagram/nodes/EditableNodeLabel.tsx
```diff
@@ -68,6 +69,7 @@ export function EditableNodeLabel({
   skipValidation?: boolean;
   onGetAllUniqueNames: (s: State) => UniqueNameIndex;
   fontCssProperties?: React.CSSProperties;
+  setAlternativeEditableNodeHeight?: React.Dispatch<React.SetStateAction<number>>;
```
tiagobento: What does `alternative` have to do here in the context of this component? I mean, this is really decoupled from anything related to the alternative representation of the InputData node. We need a better name, or a better mechanism to satisfying the new node's needs.

packages/dmn-editor/src/diagram/nodes/NodeStyle.ts
```diff
@@ -192,9 +192,12 @@ export function getFontCssProperties(dmnFontStyle?: DmnFontStyle): React.CSSProp
   };
 }
 
-export function getNodeLabelPosition(nodeType: NodeType): NodeLabelPosition {
+export function getNodeLabelPosition(nodeType: NodeType, isAlternativeInputDataShape?: boolean): NodeLabelPosition {
```
tiagobento: Why optional?

packages/dmn-editor/src/diagram/nodes/Nodes.tsx
```diff
@@ -1193,18 +1266,22 @@ function useNodeDimensions(
   type: NodeType,
   snapGrid: SnapGrid,
   shape: DMNDI15__DMNShape,
-  isExternal: boolean
+  isExternal: boolean,
+  isAlternativeInputDataShape?: boolean
```
tiagobento: Why optional?

packages/dmn-editor/src/mutations/addOrGetDrd.ts
```diff
@@ -32,6 +33,7 @@ export function addOrGetDrd({ definitions, drdIndex }: { definitions: DMN15__tDe
   definitions["dmndi:DMNDI"]["dmndi:DMNDiagram"][drdIndex] ??= {};
 
   const defaultDiagram = definitions["dmndi:DMNDI"]["dmndi:DMNDiagram"][drdIndex];
+  defaultDiagram["@_id"] ??= generateUuid();
```
tiagobento, line 36: 👍

packages/dmn-editor/src/store/computed/computeIsAlternativeInputDataShape.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import { State } from "../Store";
+
+export function computeIsAlternativeInputDataShape(
+  drdIndex: State["diagram"]["drdIndex"],
+  dmnDi: State["dmn"]["model"]["definitions"]["dmndi:DMNDI"]
+) {
+  return dmnDi?.["dmndi:DMNDiagram"]?.[drdIndex]["@_useAlternativeInputDataShape"] ?? false;
+}
```
tiagobento: Why are we using the `dmnDi` as parameter here and not `State["dmn"]["model"]["definitions"]["dmndi:DMNDI"]["dmndi:DMNDiagram"]`?

tiagobento: Also, feel free to inline this file directly at `Store.ts`... it's so small and I don't see this changing to be more complex in the future.

packages/dmn-editor/src/svg/DmnDiagramSvg.tsx
```diff
@@ -250,6 +250,17 @@ const SVG_NODE_LABEL_TEXT_ADDITIONAL_PADDING_TOP_LEFT = 8;
...
+    case "center-bottom":
+      const cbTx = n.position.x! + n.width! / 2;
+      const cbTy = n.position.y! + n.height!;
+      const cbWidth = n.width! - 2 * SVG_NODE_LABEL_TEXT_PADDING_ALL;
+      return {
+        verticalAnchor: "middle",
+        textAnchor: "middle",
+        transform: `translate(${cbTx},${cbTy})`,
+        width: cbWidth,
+      } as const;
```
tiagobento: Nice. I would've completely forgotten this 🙈

tiagobento (changes requested):
@ljmotta Awesome PR! Thanks a lot for this work. Indeed a great addition to the DMN Editor.

My two most important comments here are related to two things:

1. Unnecessary use of reactivity when you could rely on the state object directly. (This is something I think we should reassess.. I mean, for `useCallback` hooks, reacvitity need is really rare... as we could simply do `dmnEditorStoreApi.getState()` and use the most up-to-date state, right?)
2. Optional positional boolean parameter. This really is a great source of bugs, and I think we should really avoid them. It's really easy to convert positional parameters to object parameters in VS Code too! [1]). Also, for non-positional parameters, we should also try and make it not optional. It's too easy to forget it somewhere.



[1] 

https://github.com/apache/incubator-kie-tools/assets/1584568/ccf86fa1-9b90-4828-85f6-f20fe3c88d98

packages/dmn-editor/src/diagram/nodes/DefaultSizes.ts
ljmotta: The MIN_NODE_SIZES is used by all nodes, which doesn't have an alternative shape. I made it optional to not polute the codebase with a specific node property on all function calls.

tiagobento: @ljmotta You could work some TypeScript "magic" to make calls with a constant argument resolve to a type that doesn't need this parameter, however, from the entire domain's perspective, this parameter is mandatory, since `InputData` nodes require it. Making it optional for all cases, also makes it optional for `InputData`, which is incorrect. We always need to know whether or the `InputData` node is in its alternative or classic form.

tiagobento (approved):

PR #2154, opened 2024-02-12
kie-issues#234: Identify DMN DRD elements with collapsed dependencies with ellipsis (...) notation

### LINKED ISSUE #234: Identify DMN DRD elements with collapsed dependencies with ellipsis (...) notation
DRG Elements with requirements not displayed on the current DRD SHOULD be notated with an ellipsis (...) to
show that this is the case.

See the "Routing" decision at the DRD below.

image

From the [**DMN 1.4 spec**](https://www.omg.org/spec/DMN/1.4/PDF)  , page 167 (181 on the PDF)

ljmotta:
Closes https://github.com/apache/incubator-kie-issues/issues/234

## Description
- Adds the ellipsis into the Decision, BKM and Knowledge Source nodes;
- Fix the delete Text Annotation node from DRD;
- This PR removes the optionality of the "isCollection" on Input Data and Decision Node;

## Video

https://github.com/apache/incubator-kie-tools/assets/24302289/bd4d02c2-da5d-4d3c-9e69-f2da99192568

### DIFF
--- a/packages/dmn-editor/src/diagram/nodes/NodeSvgs.tsx
+++ b/packages/dmn-editor/src/diagram/nodes/NodeSvgs.tsx
@@ -78,7 +78,6 @@ export function InputDataNodeSvg(__props: NodeSvgProps & { isCollection?: boolea
     height,
     fillColor,
     strokeColor,
-    hasHiddenNodes,
     props: { isCollection, ...props },
   } = normalize(__props);
 
@@ -112,7 +111,6 @@ export function InputDataNodeSvg(__props: NodeSvgProps & { isCollection?: boolea
         ry={ry}
       />
       {isCollection && <NodeCollectionMarker {...__props} anchor={"bottom"} />}
-      {hasHiddenNodes && <NodeHiddenInformationMarker {...__props} />}
     </>
   );
 }
--- a/packages/dmn-editor/src/diagram/nodes/NodeSvgs.tsx
+++ b/packages/dmn-editor/src/diagram/nodes/NodeSvgs.tsx
@@ -29,7 +29,6 @@ export type NodeSvgProps = RF.Dimensions &
     fillColor?: string;
     strokeColor?: string;
     strokeWidth?: number;
-    hasHiddenSource?: boolean;
   };
 
 export const ___NASTY_HACK_FOR_SAFARI_to_force_redrawing_svgs_and_avoid_repaint_glitches = { flag: false };
@@ -44,7 +43,6 @@ export function normalize<T extends NodeSvgProps>(_props: T) {
     height: _height,
     fillColor: _fillColor,
     strokeColor: _strokeColor,
-    hasHiddenSource: _hasHiddenSource,
     ...props
   } = _props;
 
@@ -64,12 +62,11 @@ export function normalize<T extends NodeSvgProps>(_props: T) {
     height: height + (___NASTY_HACK_FOR_SAFARI_to_force_redrawing_svgs_and_avoid_repaint_glitches.flag ? 0 : 0.1),
     fillColor: _fillColor,
     strokeColor: _strokeColor,
-    hasHiddenSource: _hasHiddenSource,
     props,
   };
 }
 
-export function InputDataNodeSvg(__props: NodeSvgProps & { isCollection?: boolean }) {
+export function InputDataNodeSvg(__props: NodeSvgProps & { isCollection: boolean }) {
   const {
     strokeWidth,
     x,
@@ -116,7 +113,7 @@ export function InputDataNodeSvg(__props: NodeSvgProps & { isCollection?: boolea
 }
 
 export function AlternativeInputDataNodeSvg(
-  __props: NodeSvgProps & { isCollection?: boolean; isIcon: boolean; transform?: string }
+  __props: NodeSvgProps & { isCollection: boolean; isIcon: boolean; transform?: string }
 ) {
   const {
     strokeWidth,
@@ -171,7 +168,7 @@ export function AlternativeInputDataNodeSvg(
   );
 }
 
-export function DecisionNodeSvg(__props: NodeSvgProps & { isCollection?: boolean }) {
+export function DecisionNodeSvg(__props: NodeSvgProps & { isCollection: boolean; hasHiddenSource: boolean }) {
   const {
     strokeWidth,
     x,
@@ -180,8 +177,7 @@ export function DecisionNodeSvg(__props: NodeSvgProps & { isCollection?: boolean
     height,
     fillColor,
     strokeColor,
-    hasHiddenSource,
-    props: { isCollection, ...props },
+    props: { isCollection, hasHiddenSource, ...props },
   } = normalize(__props);
 
   return (
@@ -203,8 +199,17 @@ export function DecisionNodeSvg(__props: NodeSvgProps & { isCollection?: boolean
   );
 }
 
-export function BkmNodeSvg(__props: NodeSvgProps) {
-  const { strokeWidth, x, y, width, height, fillColor, strokeColor, hasHiddenSource, props } = normalize(__props);
+export function BkmNodeSvg(__props: NodeSvgProps & { hasHiddenSource: boolean }) {
+  const {
+    strokeWidth,
+    x,
+    y,
+    width,
+    height,
+    fillColor,
+    strokeColor,
+    props: { hasHiddenSource, ...props },
+  } = normalize(__props);
   const bevel = 25;
   return (
     <>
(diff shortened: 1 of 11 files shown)


packages/dmn-editor/src/mutations/deleteNode.ts
```diff
@@ -107,10 +107,11 @@ export function deleteNode({
...
     if (nodeNature === NodeNature.ARTIFACT) {
-      if (mode === NodeDeletionMode.FROM_DRG_AND_ALL_DRDS) {
-        const nodeIndex = (definitions.artifact ?? []).findIndex((a) => a["@_id"] === dmnObjectId);
-        dmnObject = definitions.artifact?.splice(nodeIndex, 1)?.[0];
-      }
+      const nodeIndex = (definitions.artifact ?? []).findIndex((a) => a["@_id"] === dmnObjectId);
+      dmnObject =
+        mode === NodeDeletionMode.FROM_DRG_AND_ALL_DRDS
+          ? definitions.artifact?.splice(nodeIndex, 1)?.[0]
+          : definitions.artifact?.[nodeIndex];
```
tiagobento: Not sure we should allow this. Hiding `artifacts` from a DRD won't ever allow them to come back, as we don't have the option to add `artifacts` to a DRD. This will make the DMN file contain phantom elements, which won't ever be removed. I think we should throw in this case and disallow hiding a node from a DRD when it is not a `drgElement`.

packages/dmn-editor/src/svg/DmnDiagramSvg.tsx
```diff
@@ -59,6 +59,7 @@ import { Text } from "@visx/text";
 import { TypeOrReturnType } from "../store/ComputedStateCache";
```
tiagobento: Lint warning here. Missing hook dependencies.

packages/dmn-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -327,6 +343,7 @@ export function computeDiagramData(
 
   return {
     drgEdges,
+    drgNodeIdsBySourceNodeId,
```
tiagobento: Although correct, I think we could have a better name for this map. I would call it `directDepenentsByNodeId`. Or even `drgAdjacencyList`. Note that the DRG of a DMN can include nodes frm other DMNs too, so a DRG is not simply `definitions.drgElement`, as `definitions.drgElement` represents the local elements of the DRG, but since there can be references to external DRG elements, a DRG includes both local and external nodes.

packages/dmn-editor/src/svg/DmnDiagramSvg.tsx
```diff
@@ -78,7 +81,18 @@ export function DmnDiagramSvg({
...
+  drgElementsWithoutVisualRepresentationOnCurrentDrd: string[];
 }) {
+  const nodesWithHiddenSource = useMemo(() => {
+    return drgElementsWithoutVisualRepresentationOnCurrentDrd.reduce((acc, hiddenElementId) => {
+      drgNodeIdsBySourceNodeId.get(hiddenElementId)?.forEach((targetIds) => {
+        acc.add(targetIds);
+      });
+      return acc;
+    }, new Set<string>());
+  }, [drgElementsWithoutVisualRepresentationOnCurrentDrd, drgNodeIdsBySourceNodeId]);
```
tiagobento: Why do we need this when we already have `data.hasHiddenSource` assigned to nodes?

packages/dmn-editor/src/diagram/nodes/NodeSvgs.tsx
```diff
@@ -499,3 +520,44 @@ function NodeCollectionMarker(__props: NodeSvgProps & { anchor: "top" | "bottom"
...
   );
 }
+
+function NodeHiddenInformationMarker(__props: NodeSvgProps & { anchor: "middle" | "left" }) {
+  const { strokeWidth, x, y, width, height, fillColor, strokeColor, props } = normalize(__props);
+
+  const dotRadius = 1;
+  const xPosition = props.anchor === "middle" ? x + width / 2 : x + width / 4;
+  const xSpacing = 7;
+  const yPosition = props.anchor === "middle" ? y + height - 18 : y + height - 11;
```
tiagobento: I wouldn't mind some comments here regarding the magic numbers...

packages/dmn-editor/src/diagram/nodes/NodeSvgs.tsx
```diff
@@ -499,3 +520,44 @@ function NodeCollectionMarker(__props: NodeSvgProps & { anchor: "top" | "bottom"
     </>
   );
 }
+
+function NodeHiddenInformationMarker(__props: NodeSvgProps & { anchor: "middle" | "left" }) {
```
tiagobento: We're reusing `NodeSvgProps` here for a component that doesn't represent a Node. I think we shouldn't do this. Or we should have a specific prop that is called `nodeProps`, making it explicit that this component needs to know its container node props.

packages/dmn-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -160,6 +170,7 @@ export function computeDiagramData(
     const _shape = indexes.dmnShapesByHref.get(id);
     if (!_shape) {
       drgElementsWithoutVisualRepresentationOnCurrentDrd.push(id);
+      drgNodeIdsBySourceNodeId.get(id)?.forEach((id) => nodeHasHiddenSource.add(id));
```
tiagobento: I don't think this is correct. As you can see below, `ackEdge` is called for external DMNs, so `drgNodeIdsBySourceNodeId` is not complete yet.

Actually, you don't really need `nodeHasHiddenSource`. Let's think about it:

You have an indexed adjacency list representation of the graph with  `drgNodeIdsBySourceNodeId`. 

After all edges have been acked, we're sure that `drgNodeIdsBySourceNodeId` is complete. You can iterate through the nodes, and query `nodesById` to see if all direct dependents exist in the graph. If not, you assign `data.hasHiddenSource.`

packages/dmn-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -171,6 +182,7 @@ export function computeDiagramData(
       dmnObject,
       shape,
       index,
+      hasHiddenSource: false,
       parentRfNode: undefined,
```
tiagobento: Can you leave a comment saying that these two get overridden further along?

packages/dmn-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -244,6 +256,10 @@ export function computeDiagramData(
     } else if (localNodes[i].type === NODE_TYPES.decisionService) {
       localNodes[i].zIndex = NODE_LAYERS.DECISION_SERVICE_NODE;
     }
+
+    if (nodeHasHiddenSource.has(localNodes[i].id)) {
+      localNodes[i].data.hasHiddenSource = true;
+    }
```
tiagobento: External nodes can have hidden source nodes too. Please see my comment above to understand what I think is the best strategy here.

packages/dmn-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -88,6 +95,8 @@ export function computeDiagramData(
   const edges: RF.Edge<DmnDiagramEdgeData>[] = [];
 
   const drgEdges: DrgEdge[] = [];
+  const drgNodeIdsBySourceNodeId: DrgNodeIdsBySourceNodeId = new Map();
```
tiagobento: See my comment below regarding these names.

packages/dmn-editor/src/store/computed/computeDiagramData.ts
tiagobento: Yeah I think `drgAdjacencyList` is the best one :P Since we're dealing with a directed graph too, I would have the type be: 

```ts
export type DrgAdjacencyList = Map<string, { dependents: Set<string>, dependencies: Set<string> }>;
```

If you don't want to populate the `dependencies` part of it, just leave it out of the type for now.

tiagobento: At some point we're probably going to need the transitive closure, but for now we only need direct edges mapped.

tiagobento (changes requested):
Left some comments inline regarding names and external nodes.

Let me make a point here about the names you chose.

"Hidden source", although correct, uses ReactFlow and generic diagrams terminology. In DMN, there's no such thing as a "source node". What we have are requirements. Information/Knowledge/Authority requirements.

A Decision which doesn't have all of its direct requirements listed in the DRD should display an ellipsis marker.

I would name everything to reflect that. Instead of `hiddenSources`, I would say `hiddenRequirements`.  You even had `HiddenInformation` in some places, which makes things a little harder, even.

packages/dmn-editor/src/store/computed/computeDiagramData.ts
ljmotta: I've totally missed the external nodes. I'll make the adjustments.

tiagobento (approved):

PR #2163, opened 2024-02-16
kie-issues#944: On the DMN Editor, all buttons that have no displayed text should have a title.

### LINKED ISSUE #944: On the DMN Editor, all buttons that have no displayed text should have a title.
Examples:

image
image
image

tiagobento:
Closes https://github.com/apache/incubator-kie-issues/issues/944

### DIFF
--- a/packages/dmn-editor/src/dataTypes/DataTypes.tsx
+++ b/packages/dmn-editor/src/dataTypes/DataTypes.tsx
@@ -164,6 +164,8 @@ export function DataTypes() {
 
   const [isAddDataTypeDropdownOpen, setAddDataTypeDropdownOpen] = useState(false);
 
+  const extraPropsForDropdownToggleAction = { title: "New Data Type" };
+
   return (
     <>
       {(dataTypesTree.length <= 0 && (
@@ -197,6 +199,7 @@ export function DataTypes() {
                           id="add-data-type-toggle"
                           splitButtonItems={[
                             <DropdownToggleAction
+                              {...extraPropsForDropdownToggleAction}
                               key="add-data-type-action"
                               aria-label="Add Data Type"
                               onClick={() =>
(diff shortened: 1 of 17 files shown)


packages/dmn-editor/src/dataTypes/DataTypes.tsx
```diff
@@ -197,6 +199,7 @@ export function DataTypes() {
                           id="add-data-type-toggle"
                           splitButtonItems={[
                             <DropdownToggleAction
+                              {...extraPropsForDropdownToggleAction}
```
ljmotta, line 203: Why not inline it?

jomarko (commented):
@tiagobento thank you for adding this improvement. It will be beneficial as for human users so for playwright testing.

I have one question about general titles we add:
- "Add"
- "Remove"
- "Close"

Would it be possible to add more detail into similar titles?
- "Add [something]"
- "Remove [something]"
- "Close [something]"

I understand it is probably not needed for human users, who understand from context, what will be "Added", "Removed" or "Closed", however if we do "page.getByTitle('Add')", it can match multiple elements and it may be difficult to write short tests then.

What is your opinion please?

tiagobento:
@jomarko Yeah, I preferred to leave it as inferable from context, but I see your point. I can change it to be more specific, of course. However, I must say that we shouldn't start compromising on UI/UX in order for tests to be more easily written. I think we should of course keep tests in mind with everything that we do. IMHO we should always think what's the best UX possible for users, then figure out how to test it.

packages/dmn-editor/src/dataTypes/DataTypes.tsx
tiagobento: Added a comment...

jomarko (approved):
@tiagobento thank you.

ljmotta (approved):

PR #2164, opened 2024-02-16
KOGITO-10018: SWF Editor - Integrate SWF diagram editor theme API with VsCode and Combined Editor

handreyrc:
SWF Diagram Editor theme support. 

**JIRA**: [KOGITO-10018](https://issues.redhat.com/browse/KOGITO-10018)
**JIRA**: [KOGITO-10016](https://issues.redhat.com/browse/KOGITO-10016)
**JIRA**: [KOGITO-10015](https://issues.redhat.com/browse/KOGITO-10015)

Screenshots:
combined_theme
vscode_theme

### DIFF
--- a/packages/editor/src/envelope/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
+++ b/packages/editor/src/envelope/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
@@ -31,13 +31,15 @@ import {
 import { Modal } from "@patternfly/react-core/dist/js/components/Modal";
 import { KeyboardIcon } from "@patternfly/react-icons/dist/js/icons/keyboard-icon";
 import { OperatingSystem } from "@kie-tools-core/operating-system";
-import { useKogitoEditorEnvelopeContext } from "../../api";
+import { EditorTheme, useKogitoEditorEnvelopeContext } from "../../api";
 import { useEditorEnvelopeI18nContext } from "../i18n";
+import { useSharedValue } from "@kie-tools-core/envelope-bus/dist/hooks";
 
 export function KeyBindingsHelpOverlay() {
   const [showing, setShowing] = useState(false);
   const envelopeContext = useKogitoEditorEnvelopeContext();
   const { i18n } = useEditorEnvelopeI18nContext();
+  const [theme] = useSharedValue(envelopeContext.channelApi?.shared.kogitoEditor_theme);
 
   const toggle = useCallback(() => {
     setShowing(!showing);
@@ -82,11 +84,13 @@ export function KeyBindingsHelpOverlay() {
     }
   }, [showing]);
 
+  const themeCss = !theme! && theme === EditorTheme.DARK ? " dark" : "";
+
   return (
     <>
       <div
         onClick={() => setShowing(!showing)}
-        className={"kie-tools--keyboard-shortcuts kie-tools--keyboard-shortcuts-icon"}
+        className={"kie-tools--keyboard-shortcuts kie-tools--keyboard-shortcuts-icon" + themeCss}
         data-ouia-component-id="keyboard-shortcuts-icon"
         data-testid={"keyboard-shortcuts-help-overlay-icon"}
       >
@@ -100,7 +104,7 @@ export function KeyBindingsHelpOverlay() {
         width={"60%"}
         onClose={toggle}
         data-testid={"keyboard-shortcuts-help-overlay"}
-        className={"kie-tools--keyboard-shortcuts"}
+        className={"kie-tools--keyboard-shortcuts" + themeCss}
       >
         <TextContent>
           <TextList component={TextListVariants.dl}>
--- a/packages/editor/src/envelope/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
+++ b/packages/editor/src/envelope/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
@@ -84,7 +84,7 @@ export function KeyBindingsHelpOverlay() {
     }
   }, [showing]);
 
-  const themeCss = !theme! && theme === EditorTheme.DARK ? " dark" : "";
+  const themeCss = getCssThemeModifier(envelopeContext.supportedThemes, theme!);
 
   return (
     <>
@@ -126,6 +126,21 @@ export function KeyBindingsHelpOverlay() {
       </Modal>
     </>
   );
+}
+
+function getCssThemeModifier(supportedThemes: EditorTheme[], theme: EditorTheme) {
+  if (supportedThemes.includes(theme)) {
+    switch (theme) {
+      case EditorTheme.DARK: {
+        return " dark";
+      }
+      default: {
+        return "";
+      }
+    }
+  }
+
+  return "";
 }
 
 function handleMacOsCombination(combination: string, os?: OperatingSystem) {
--- a/packages/editor/src/api/KogitoEditorEnvelopeContext.ts
+++ b/packages/editor/src/api/KogitoEditorEnvelopeContext.ts
@@ -24,6 +24,7 @@ import { KogitoEditorChannelApi } from "./KogitoEditorChannelApi";
 import { I18nService } from "@kie-tools-core/i18n/dist/envelope";
 import { OperatingSystem } from "@kie-tools-core/operating-system";
 import { KeyboardShortcutsService } from "@kie-tools-core/keyboard-shortcuts/dist/envelope/KeyboardShortcutsService";
+import { EditorTheme } from "./EditorTheme";
 
 export interface KogitoEditorEnvelopeContextType<
   ChannelApi extends KogitoEditorChannelApi & ApiDefinition<ChannelApi>
@@ -34,6 +35,7 @@ export interface KogitoEditorEnvelopeContextType<
     keyboardShortcuts: KeyboardShortcutsService;
     i18n: I18nService;
   };
+  supportedThemes: EditorTheme[];
 }
 
 export const KogitoEditorEnvelopeContext = React.createContext<KogitoEditorEnvelopeContextType<any>>({} as any);
--- a/packages/editor/src/envelope/KogitoEditorEnvelope.tsx
+++ b/packages/editor/src/envelope/KogitoEditorEnvelope.tsx
@@ -62,6 +62,7 @@ export class KogitoEditorEnvelope<
         keyboardShortcuts: keyboardShortcutsService,
         i18n: i18nService,
       },
+      supportedThemes: [],
     }
   ) {}
 
--- a/packages/editor/src/envelope/KogitoEditorEnvelopeApiImpl.ts
+++ b/packages/editor/src/envelope/KogitoEditorEnvelopeApiImpl.ts
@@ -24,6 +24,7 @@ import {
   EditorContent,
   EditorFactory,
   EditorInitArgs,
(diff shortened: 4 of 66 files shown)


tiagobento:
@handreyrc Instead of creating `kogitoEditor_created`, couldn't we add a new property to `EditorInitArgs` instead?

packages/editor/src/embedded/embedded/EmbeddedEditor.tsx
```diff
@@ -124,6 +128,7 @@ const RefForwardingEmbeddedEditor: React.ForwardRefRenderFunction<EmbeddedEditor
             channel: props.channelType,
             workspaceRootAbsolutePosixPath:
               props.workspaceRootAbsolutePosixPath ?? DEFAULT_WORKSPACE_ROOT_ABSOLUTE_POSIX_PATH,
+            theme: theme,
```
tiagobento: Missing hook dependency here..

caponetto:
@tomasdavidorg could you please help us validate the channels?
The only expected channel to react to a theme change is VS Code.

packages/editor/src/embedded/embedded/EmbeddedEditor.tsx
handreyrc: Hey @tiagobento,
 
Its is fixed! 
BTW thanks for the tip with the EditorInitArgs.

tiagobento:
@handreyrc I deleted my last comment because I think we need to make for changes to this PR.

As you saw, there's already a Shared Value on `KogitoEditorChannelApi`. This Shared Value is accessible on both ends -- Channel and Envelope. For this reason, we don't need a new `kogitoEditor_setTheme` on `KogitoEditorEnvelopeApi`. In fact, we don't even need to add the `theme` property to the EditorInitArgs at all.

If you do a short-lived subscription of the shared value, you can read its value asyncly. 

```ts
const subscription = envelopeServerApi.shared.kogitoEditor_theme.subscribe(newValue => {
    doSomething(newValue);
    envelopeServerApi.shared.kogitoEditor_theme.unsubscribe(subscription);
});
```

Of course, now looking at your use-case, it might've made sense to add an `envelopeApi.shared.kogitoEditor_theme.get()`, returning a Promise with the current value on the other side of the bus, like a `peek` on a stream.

---

Hope it made some sense, and yeah, really hard to know that given we have zero documentation :/

tomasdavidorg (commented):
Hi @handreyrc,
I tested the changes I have these 4 comments:
1. After creating of a new SW file, the content is not set and the diagram is always in the light mode 
Screenshot from 2024-02-27 08-49-25
2. The zoom percentages are always in the light mode
Screenshot from 2024-02-27 08-49-40
3. Keyboards shortcuts are always in the light mode
Screenshot from 2024-02-27 08-53-35
4. When the SW file is changed to some incorrect form and the mode is switched, the diagram remains in the previous mode. (This is probably caused by the fact that the diagram is not reloaded when there is an error in the editor)
Screenshot from 2024-02-27 08-59-21

handreyrc:
Hi @tiagobento,
 
I made the changes you suggested and it works well. Basically I added a subscription for the Theme value right after the editor is created and I removed the kogitoEditor_setTheme.
However,  like in your code sample above, we are unsubscribing for the theme right after consuning it in the initialization, so it won't listen to theme changes if the diagram is already loaded (KogitoEditor_setTheme was my hook for changing the theme in this scenario). 
Considering the subscription approach, I removed the "unsubscribe" and it solves the issue. What we need is some sort of  permanent subscription to keep watching theme changes.

The question is: 

- Is there some sort of issue if we do not unsubscribe for that shared value?

tiagobento:
@handreyrc Glad to hear it worked! No, you're definitely right, my point with unsubscribing right away was for the case where you would just look at the value once, but for this particular use case, indeed we need a "permanent" subscription that will alter the theme on the Editor every time is changes on the Shared Value. Well done! 

One thing I would do, though, is trying to find where would it be a good place to put the unsubscription. Maybe on the "componentDidUnmount" (or the returned lambda of an `useEffect` without dependencies) of the `EditorEnvelopeView` component?

tiagobento:
Not that this is a good place to put it, but we don't have a "destructor" or a "dispose" mechanism for Editor envelopes...

handreyrc:
> Hi @handreyrc, I tested the changes I have these 4 comments:
> 
> 1. After creating of a new SW file, the content is not set and the diagram is always in the light mode
>    Screenshot from 2024-02-27 08-49-25
> 2. The zoom percentages are always in the light mode
>    Screenshot from 2024-02-27 08-49-40
> 3. Keyboards shortcuts are always in the light mode
>    Screenshot from 2024-02-27 08-53-35
> 4. When the SW file is changed to some incorrect form and the mode is switched, the diagram remains in the previous mode. (This is probably caused by the fact that the diagram is not reloaded when there is an error in the editor)
>    Screenshot from 2024-02-27 08-59-21

Hey @tomasdavidorg,

I worked on the items you pointed above.

1- Dark theme was added to the Error Page.  The canvas was implemented in a way that if there are errors, the Error page is presented and no further rendering happens until we have a valid diagram. So if the Error Page is presented with dark theme set and then the theme is changed to light, the error page will remain dark and vice-versa. Changing this behavior may have implications because it is liked to wokflow loading as well, so if we really need to change the theme in this scenario we could consider doing it in a separate PR.

2- Dark Theme was added to the zoombar combobox.

3- Grey background was added to the Keyboard shortcuts modal on dark mode. 

4- Theme change in workflows with errors. I didn't consider the behavior of not be possible to reload a broken workflow in my previous implementation, good catch.  It was looking alful because the diagram is not reloaded but the background and  state selection color pattern changes. In fact the non reloading behavior is on purpose, so if the worflow is broken it remains visible and functional on the canvas. Following this behavior, no theme at all is applied to the broken workflows, however, when it gets valid again the theme applied.

Please, could take a look again at these items and share your thoughts?

Cc @caponetto  

Thanks!!!

handreyrc:
> @handreyrc Glad to hear it worked! No, you're definitely right, my point with unsubscribing right away was for the case where you would just look at the value once, but for this particular use case, indeed we need a "permanent" subscription that will alter the theme on the Editor every time is changes on the Shared Value. Well done!
> 
> One thing I would do, though, is trying to find where would it be a good place to put the unsubscription. Maybe on the "componentDidUnmount" (or the returned lambda of an `useEffect` without dependencies) of the `EditorEnvelopeView` component?

Hey @tiagobento,

We tested your idea of implemening the unsubscribe in a returned lambda of an useEffect without dependencies of the EditorEnvelopeView or some other place. Turns out that after further investigation those hooks are never called. In our tests the whole editor iFame is dropped before those lifecycle hooks have a chance to be executed. Looking at this context its not very likely that we are introducing a memory leak by not unsubscribing that shared value. On top of that there are other implications in making that subscription in other places such as random nullpointers when the setTheme is called and the GWT Editor is not completely initialized.

Please, take a look at my latest.

WDYT?

BTW, Thanks a lot @paulovmr for the helping hand.

The Keyboard Shortcuts modal is a common element between the editors. By adding theme there we are changing it for all editors. Is that a problem? Any suggestion?

This is how it looks on VsCode Dark Theme:
image

tomasdavidorg (approved):
Hi, @handreyrc, thanks for the updates. Looks good to me. Np with the error pages.

Could you please check one more scenario?

1. Switch to the dark mode.
2. Create a new file (File -> New File -> aaa.sw.json)
3. Click Create a Serverless Workflow -> Serverless workflow example
4. The Diagram view is rendered by the Light mode.


image

handreyrc:
> tomasdavidorg

Hey @tomasdavidorg,

It is fixed! 
There was an issue with a flag hasErrors that was not being set to false if the session is reused.
Both, the error page and the new diagram must follow the current theme.

Please, could you check it again?

tomasdavidorg:
> > tomasdavidorg
> 
> Hey @tomasdavidorg,
> 
> It is fixed! There was an issue with a flag hasErrors that was not being set to false if the session is reused. Both, the error page and the new diagram must follow the current theme.
> 
> Please, could you check check it again?

@handreyrc Works fine. Thanks 🙂

handreyrc:
Hey @tiagobento and @caponetto,

I addressed the issue with the dark theme being applied to the keyboad shortcuts modal in all editors.  Now  only the SWF editor will listen to theme changes.

Basically I added the property " supportedThemes: EditorTheme[]"  to the KogitoEditorEnvelopeContextType so it's available everywhere. If the editor supports multiple themes just list the themes in the respective factory for that editor otherwise do nothing. It also allows to make the theme subscription only if multiple themes are supported. 

Please, could you take a look?

Thanks.

packages/editor/src/envelope/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
```diff
@@ -82,11 +84,13 @@ export function KeyBindingsHelpOverlay() {
     }
   }, [showing]);
 
+  const themeCss = envelopeContext.supportedThemes.length > 1 && !theme! && theme === EditorTheme.DARK ? " dark" : "";
```
caponetto: Since you've introduced the `supportedThemes` array, I think this line needs some attention. Maybe extract it into a more generic function? I mean, if we add more themes, we'd need to change this line too (currently, the value is either empty string or `dark`). And what happens if I provide only one supported theme, and it's the dark one? Also, I think the part `&& !theme!` is unnecessary.

caponetto (commented):
Hey @handreyrc, thanks for the updates. I checked the DMN Editor and I'm not seeing any reaction to theme changes anymore. I've left one comment on the PR though, please check if it makes sense.

Also, please see my notes:

_Bugs found during my tests_:
- [VS Code > Diagram editor] If you change the theme while the workflow has errors, the diagram theme never changes accordingly, even if you fix the workflow and continue to edit it.

_Regressions found during my tests in comparison with the previous tests_:
- [Combined editor > Text Editor] The keyboard shortcut icon is kept in light mode (see the image, it practically disappears)
Screenshot 2024-03-22 at 09 18 12


_Follow up tasks (I suggest to group them in a new epic on `kie-issues`)_:
- [VS Code > Diagram editor] There is a white blink on each update. It also happens with the light theme but it is more evident on the dark theme.
- [Diagram editor] The theme doesn't change when the workflow contains errors or is empty
- [Combined editor > Diagram editor] The scrollbars are kept in light mode
- [Combined editor] The divider between the Text and Diagram editor is kept in light mode
- [Combined editor] The loading screen is kept in light mode

packages/editor/src/envelope/KeyBindingsHelpOverlay/KeyBindingsHelpOverlay.tsx
handreyrc: @caponetto;

Sure, it makes sense, I moved this logic into a function and I'm also testing if the theme is among the supported themes this way we dont need to test if we support more than one theme anymore.
The "!theme!" thing is fixed. It was sort of a java (not !) lapse. it is fixed as well.
The two issues you reported above are fixed as well.

Thank you!

caponetto (approved):

handreyrc:
Hey @caponetto,

I created a new epic so we can close this one and still keep track of the issues and improvements you reported above.

**Epic**: [KOGITO-10071](https://issues.redhat.com/browse/KOGITO-10071) 

Thanks

paulovmr (approved):

tiagobento (approved):

PR #2165, opened 2024-02-19
kie-issues#469: Static Validation for DMN and BPMN Editors on VS Code extensions - Part 1

Josephblt:
Part 1 of: [incubator-kie-issues/issues/469](https://github.com/apache/incubator-kie-issues/issues/469).

All functionality implemented. A new VS Code extension project was created. It is possible to validate BPMN and DMN files. You can connect to any Extended-Service or choose to locally run one.

A separate PR will be sent with tests for all features that this one introduced.


jomarko (commented):
@Josephblt hello, thank you for this new extension `extended-services-vscode-extension`.  From the ticket description it seems to be very important and useful, however I would like to ask few questions about its usage.

- Once it is installed into VS Code, are validation messages automatically shown to the user? 
- Or does use need to do any additional setting? 
- Does user needs to open some specific panel in VS Code?
- Does user need to start validation by some command?
- Is user prompted to install this new extension once old `DMN` or `BPMN` extension is installed and the new is not detected?

Josephblt:
Hello @jomarko.

- Once it is installed into VS Code, are validation messages automatically shown to the user?
Yes. Once installed, Bpmn and Dmn files behave like any other supported language. So any time the user opens a file or changes something, the files are validated. If any error is detected it is reported to the Problems tab.
- Or does use need to do any additional setting?
It works right out of the box. But it supports a few configurations. You can run the services locally or point to any address you want. You can also set how long between each connection check.
- Does user needs to open some specific panel in VS Code?
Nothing, it is intended to work just as any other script/language validator. If you change a file it reports the errors.
- Does user need to start validation by some command?
You can, but it's not necessary. It detects file activity, both from VS Code itself and outside sources (Git changes for example).
- Is user prompted to install this new extension once old DMN or BPMN extension is installed and the new is not detected?
No, but this is a cool feature that we already discussed. However, this is a change that needs to be implemented in the DMN and BPMN extensions themselves. So not part of this PR.

packages/vscode-extension/src/VsCodeRecommendation.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    vscode.window.showInformationMessage(message, action).then((selection) => {
+      if (selection === action) {
+        vscode.commands.executeCommand("bpmn-vscode-extension.installExtendedServices");
+      }
+    });
+
+    context.subscriptions.push(disposable);
+  }
+}
```
tiagobento: BPMN-specific commands here...

packages/extended-services-vscode-extension/src/watchers/kie-files/kie-file.ts
```diff
@@ -0,0 +1,56 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import * as vscode from "vscode";
+
+export class KIEFile {
+  public readonly fsPath: string;
```
tiagobento: Please take this comment into consideration when naming variables that represent paths https://github.com/apache/incubator-kie-issues/issues/783#issuecomment-1871525975

packages/extended-services-vscode-extension/src/local-service-manager.ts
```diff
@@ -0,0 +1,163 @@
...
+
+import * as childProcess from "child_process";
+import * as path from "path";
+import * as vscode from "vscode";
+
+export class LocalServiceManager {
+  readonly startLocalServiceCommandUID: string = "extended-services-vscode-extension.startLocalService";
+  readonly stopLocalServiceCommandUID: string = "extended-services-vscode-extension.stopLocalService";
+
+  private readonly relativePath: string = "/dist-dev/extended-services-java";
```
tiagobento: We shouldn't be using the `dist-dev` directory in production code.

packages/extended-services-vscode-extension/src/validation-helper.ts
```diff
@@ -0,0 +1,131 @@
...
+      } catch (error) {
+        vscode.window.showErrorMessage("Validate DMN error: " + error.message);
+      }
+    }
+  }
+
+  private async validate(serviceURL: URL): Promise<void> {
+    const bpmnFiles: vscode.TextDocument[] = await KIEFileWatcher.findOpenBPMNFiles();
+    const dmnFiles: vscode.TextDocument[] = await KIEFileWatcher.findOpenDMNFiles();
+    await new Promise((f) => setTimeout(f, 50));
```
tiagobento: Can you please comment on the purpose of this blocking await?

packages/extended-services-vscode-extension/src/jit-executor/jit-commands.ts
```diff
@@ -0,0 +1,113 @@
...
+import {
+  BPMNValidationResponse,
+  BPMNValidationResponseParser,
+  DMNValidationResponse,
+  DMNValidationResponseParser,
+  PingResponse,
+} from "./responses";
+import { KIEValidateBodyPayload, KIEValidateResourcePayload } from "./payloads";
+
+export class JITCommands {
```
tiagobento:
```suggestion
export class ExtendedServicesClient {
```

packages/extended-services-vscode-extension/src/watchers/configurations/configuration.ts
```diff
@@ -0,0 +1,103 @@
...
+import * as vscode from "vscode";
+
+export class Configuration {
+  static readonly autoRunConfigurationID: string = "extendedServices.enableAutorun";
+  static readonly serviceURLConfigurationID: string = "extendedServices.serviceURL";
+  static readonly connectionHeartbeatIntervalConfigurationID: string = "extendedServices.connectionHeartbeatInterval";
+
+  readonly autoRun: boolean;
+  readonly serviceURL: URL;
+  readonly connectionHeartbeatInterval: number;
```
tiagobento: Always a good idea to make the unit clear here, so:

```suggestion
  readonly connectionHeartbeatIntervalInMs: number;
```

packages/extended-services-vscode-extension/src/watchers/configurations/configuration.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import * as vscode from "vscode";
+
+export class Configuration {
+  static readonly autoRunConfigurationID: string = "extendedServices.enableAutorun";
+  static readonly serviceURLConfigurationID: string = "extendedServices.serviceURL";
+  static readonly connectionHeartbeatIntervalConfigurationID: string = "extendedServices.connectionHeartbeatInterval";
+
+  readonly autoRun: boolean;
+  readonly serviceURL: URL;
```
tiagobento:
```suggestion
  readonly extendedServicesUrl: URL;
```

?

packages/extended-services-vscode-extension/src/watchers/configurations/configuration.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+import * as vscode from "vscode";
+
+export class Configuration {
+  static readonly autoRunConfigurationID: string = "extendedServices.enableAutorun";
+  static readonly serviceURLConfigurationID: string = "extendedServices.serviceURL";
+  static readonly connectionHeartbeatIntervalConfigurationID: string = "extendedServices.connectionHeartbeatInterval";
+
+  readonly autoRun: boolean;
```
tiagobento:
```suggestion
  readonly enableAutoRun: boolean;
```

packages/extended-services-vscode-extension/src/local-service-manager.ts
```diff
@@ -0,0 +1,163 @@
...
+    this.localServiceRunningHandler = null;
+  }
+
+  public unsubscribeLocalServicesTerminated() {
+    this.localServiceTerminatedHandler = null;
+  }
+
+  private runLocalService(serviceURL: URL): void {
+    const absolutePath = this.context.extensionPath;
+    const jarFilePath = path.join(absolutePath, this.relativePath, this.jarFileName);
```
tiagobento: Path variables... need better names.

packages/extended-services-vscode-extension/src/local-service-manager.ts
```diff
@@ -0,0 +1,163 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import * as childProcess from "child_process";
+import * as path from "path";
+import * as vscode from "vscode";
+
+export class LocalServiceManager {
```
tiagobento: Why not `LocalExtendedServicesManager`?

packages/extended-services-vscode-extension/src/local-service-manager.ts
tiagobento: Also, this variable is named `relativePath` and starts with a `/`..

packages/extended-services-vscode-extension/src/jit-executor/responses.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export class BPMNValidationResponseParser {
+  public static parse(jsonResponse: string[]): BPMNValidationResponse[] {
+    if (jsonResponse.length > 0) {
+      return jsonResponse.map((response) => {
+        const uriMarker = "Uri: ";
+        const processIdMarker = " - Process id: ";
+        const nameMarker = " - name : ";
+        const errorMarker = " - error : ";
```
tiagobento: Those can all be constants.

packages/extended-services-vscode-extension/src/extension/extension-main.ts
```diff
@@ -0,0 +1,217 @@
...
+    if (configuration.autoRun) {
+      vscode.commands.executeCommand(localServiceManager.startLocalServiceCommandUID, configuration.serviceURL);
+    } else {
+      vscode.commands.executeCommand(
+        connectionManager.startConnectionHeartbeatCommandUID,
+        configuration.serviceURL,
+        configuration.connectionHeartbeatInterval
+      );
+    }
+  }
```
tiagobento: What happens if there's no configuration?

packages/extended-services-vscode-extension/src/extension/extension-main.ts
```diff
@@ -0,0 +1,217 @@
...
+          vscode.commands.executeCommand(localServiceManager.startLocalServiceCommandUID, configuration.serviceURL);
+        } else {
+          vscode.commands.executeCommand(
+            connectionManager.startConnectionHeartbeatCommandUID,
+            configuration.serviceURL,
+            configuration.connectionHeartbeatInterval
+          );
+        }
+      }
+    }
```
tiagobento, line 165: IMHO there's a lot of nested branches here. I always prefer to use guard clauses with `throw`s for exceptional cases. Helps catch errors and prevent the system from failing silently.

tiagobento: Please do apply that reasoning to other parts of the code as well.

packages/extended-services-vscode-extension/src/jit-executor/jit-commands.ts
```diff
@@ -0,0 +1,113 @@
...
+        URI: document.fileName,
+      };
+      bodyPayload.resources.push(resourcePayload);
+    }
+
+    try {
+      return JSON.stringify(bodyPayload);
+    } catch (error) {
+      vscode.window.showErrorMessage("Error building body payload: ", error.message);
+      return "";
```
tiagobento: Is this a good idea?

packages/extended-services-vscode-extension/src/jit-executor/jit-commands.ts
```diff
@@ -0,0 +1,113 @@
...
+      resources: [],
+    };
+
+    for (const document of documents) {
+      const resourcePayload: KIEValidateResourcePayload = {
+        content: document.getText(),
+        URI: document.fileName,
+      };
+      bodyPayload.resources.push(resourcePayload);
+    }
```
tiagobento:
```suggestion
    const bodyPayload: KIEValidateBodyPayload = {
      mainURI: "VS Code KIE files",
      resources: documents.map(document => ({
        content: document.getText(),
        URI: document.fileName,
      })),
    };
```

?

packages/extended-services-vscode-extension/src/jit-executor/payloads.ts
tiagobento, line 1: I think `requests` would be a better name for it, no? :D

packages/extended-services-vscode-extension/src/validation-helper.ts
tiagobento, line 1: I'm not a big fan of "helper", "util", "common" terminology... My experience tells me those easily become catch-all places with a lot of unrelated stuff, creating massive coupling and unmaintainable code. 

In this single-class file, you can already see three mixed responsibilities:
1. Managing the diagnostic collection of VS Code.
2. Actually calling Extended Services to validate stuff.
3. Registering the VS Code commands that will handle the validation.

Please do restructure to reduce coupling.

tiagobento (changes requested):
Did a first round of review and left some comments inline.

General comments:

- Please do use PascalCase acronyms. (e.g., `KIEFile` becomes `KieFile`)
- JITExecutor is not exposed to us in any form. This extension depends on Extended Services. The fact that Extended Services embeds JIT Executor is an implementation detail we don't need to expose.
- Be mindful of error treatment / exceptional case handling. There's a lot of `if` statements without a matching `else`, potentially hiding errors and making the system fail silently.
- Try and keep names more explicit and consistent. `service` is too generic to be used without a discriminator. `command` may refer to VS Code commands or JITCommands, so you have a name clash there... `Extended-Services` with a dash is not something we use anywhere else... This kind of thing. 
- For "weird" parts of the code where context is required, don't be afraid to leave comments. E.g., `Range(0,0,0,0)`
- The directory structure of `packages/extended-services-vscode-extension/src` looks odd IMHO. The two main files are nested inside the `extension` directory, while some "util" files like "validation-helper.ts" are at the top-level. I'm not a big fan of too much nesting, so feel free to put all files at the root. Start with a flat structure, and let the grouping show itself. `configuration` being inside `watchers` is a symptom of this premature nesting :)


Other than that looks solid. Fair separation of concerns, except from the ValidationHelper class of course. Great use/de-use of resources. Nice attention to detail.

The programming style is very Java-like, but I feel like you'll get TypeScriptized soon enough 😜

Thanks @Josephblt!!! Nice job! Looking forward to round 2 of the review.

packages/vscode-extension/src/VsCodeRecommendation.ts
Josephblt: Great catch. Changed "bpmn-vscode-extension.installExtendedServices" to "vscode-xtension.installExtendedServices".

packages/extended-services-vscode-extension/src/watchers/kie-files/kie-file.ts
Josephblt: After the refactory, the variable type was changed. It is now typed as a vscode.Uri.

packages/extended-services-vscode-extension/src/local-service-manager.ts
Josephblt: This is variable now:
 esNormalizedFsPathRelativeToWorkspace = dist-dev/extended-services-java
 private readonly esNormalizedFsPathRelativeToWorkspace: string = "dist/extended-services-java";

packages/extended-services-vscode-extension/src/validation-helper.ts
Josephblt: This was left over from testing, sorry about this.

packages/extended-services-vscode-extension/src/extension/extension-main.ts
Josephblt: Removed all logic nesting and changed to fail-first approach. I agree this way is more elegant when dealing with errors.

For all the other array-loop nesting cases, I changed those to functional-style.

packages/extended-services-vscode-extension/src/jit-executor/jit-commands.ts
Josephblt: No, it is not. This block now shows the error message window and then throws the error.

packages/extended-services-vscode-extension/src/jit-executor/jit-commands.ts
Josephblt: After the refactory, due to how the Diagnostics work, it was better to create individual payloads. So no collection mapping is necessary.

packages/extended-services-vscode-extension/src/jit-executor/payloads.ts
Josephblt: This is now called /request/ValidationRequests.ts

packages/extended-services-vscode-extension/src/watchers/configurations/configuration.ts
Josephblt: Added unit to the variable name.

packages/extended-services-vscode-extension/src/watchers/configurations/configuration.ts
Josephblt: Agreed.

packages/extended-services-vscode-extension/src/watchers/configurations/configuration.ts
Josephblt: Done.

packages/extended-services-vscode-extension/src/local-service-manager.ts
Josephblt: I went with LocalExtendedServices.

packages/extended-services-vscode-extension/src/local-service-manager.ts
Josephblt: Yes. Changed to this:
private start(extendedServicesURL: URL, extensionAbsoluteFsPath:string): void {
        const jarAbsoluteFilePath:string = path.join(extensionAbsoluteFsPath, this.esNormalizedFsPathRelativeToWorkspace, this.esJarFileName);

packages/extended-services-vscode-extension/src/extension/extension-main.ts
Josephblt: Caso não haja configuração neste caso não é necessário fazer nada. O tratamento para uma configuração inválida é feito acima e simplesmente encerra a conexão.

packages/extended-services-vscode-extension/src/validation-helper.ts
Josephblt: Agreed. I changed it to Validator and the class has only 2 distinct methods for validation: one for BPMN and DMN. It receives a KieFile and returns the diagnostics for that file.

jomarko:
### Manual verification checkilst
- [x] multiple files opened, switching between them
- [x] turn on/off using indicator bottom right corner 🔴 
- [x] execute validation using command
- [x] change extension settings 🔴 
- [x] non supported file opened
- [x] corrupted dmn/bpmn file opened 🔴 
- [x] change opened file manually
- [x] change opened file externally , using Notepad
- [x] change opened file using KIE extension 🔴 
- [x] undo / redo changes in opened file

jomarko:
### 01 Validation stops to work
- Imagine user is working with a project with dmn and bpmn files. 
- The problems tab properly display messages
- Then user modify the project to contain a file [1] and also open that file [1]. FYI, this file was created manually by me erasing some lines manually, we do not have issue in KIE Editors.
- The VSCode displays an error [2]. From the liveness indicator, orange color, we can see, the validation/extended services are stopped.

The problem is, even erasing the problematic file and opening other, the extended-services extension is not restarted. Or better to say, The liveness indicator is still like not online. And messages are not refreshed.

[1]
[Untitled copy.bpmn.txt](https://github.com/apache/incubator-kie-tools/files/15469242/Untitled.copy.bpmn.txt)

[2]
Screenshot 2024-05-28 140245

jomarko:
### 02 Restart extended-services
Once I click the button to disconnect Apache kie extended services, I am not able to re-connect them using the same button in the bottom right corner

jomarko:
### 03 Automatic run of the local instance

It seems the extended services validation are automatically started even if my configuration for automatic start of the local instance is turned off.

The screenshot is taken just after closing and reopening the VSCode

Screenshot 2024-05-28 144502

jomarko:
### 04 KIE Editor changes effect delay
When user edit the dmn/bpmn file with the KIE editor, not plain xml editor in VSCode, the problems tab is updated with a delay. See the steps below.

- Please open the file [1]. You will see two validation messages [2].
- Now change **New Input Data** data type to **Any** and save the file **A.dmn**
- The problems tab will still display two validation messages
- Now change **New Decision** data type to **Any** and save the file **A.dmn**
- The problems tab will display now one message [3], but it should be already empty [4]

[1]
[A.dmn.txt](https://github.com/apache/incubator-kie-tools/files/15470863/A.dmn.txt)

[2]
image

[3]
Screenshot 2024-05-28 155835


[4]
image

jomarko (commented):
My review is completed. Please take a look at the points I have raised after trying the extension manually.

Josephblt:
Hello @jomarko. Thanks for testing things out. As I progress, I'll review the reported failures individually and address them in the comments. Thanks.

Josephblt:
> 01 Validation stops to work

Yes, I managed to find the problems with this. First, there was a problem happening due to a different response format. Second I changed the behaviour to not close the connection when an error happens during validation.

Josephblt:
> ### 03 Automatic run of the local instance
Nice catch. This is fixed now.

Josephblt:
> ### 02 Restart extended-services
After some tinkering, I managed to reproduce this error using WSL. I changed the way the process is started and finished. This is fixed now.

Josephblt:
> ### 04 KIE Editor changes effect delay

Hello @jomarko. Nice catch. According to an [old bug reported](https://github.com/microsoft/vscode/issues/72831) in the VS Code repository,  it is possible to get stale data from tracked textDocuments. The last commit fixes that.

danielzhe:
I found an issue regarding the "02 Restart extended-services", reported before: if you click to disable the extended-service and don't wait for it to disable and click again in the icon at the right bottom of the VSCode, the extended-service doesn't come back live again, being necessary to kill the process manually in task manager.

In my opinion, I don't see why the user would like to completely disable it. Maybe just an option to "ignore the errors" instead of disable and enable? I don't know.

Josephblt:
Hello @danielzhe. I got an idea of what could be causing this. I'll let you know when I fix this. Thanks.

Josephblt:
Hello @danielzhe. All done. I could only reproduce the issue you reported by fast clicking the status bar icon. This last commit fix that issue.

jomarko (commented):
@Josephblt thank you for updates. I think the points I raised during my previous review are addressed now. I have one more question. I tried to run extended services externally, I mean I set **Automatically run a local instance of the service** as false. It worked fine, I was doing updates for the dmn file and the problems tab was updated correctly, however then when I stopped that external extended service, the VS Code displayed this, an error message and the livenes indicator as live. Not sure, just asking, if this is intended behavior, or the extended services vscode extension should automatically detect the external extended service is offline and change the indicator to offline?
Screenshot 2024-06-14 135212

Josephblt:
Hello @jomarko. To update the liveness indicator we use a heartbeat check. When the connection is lost the check fails, all diagnostics are cleared, the icon changes, and the following error appears:
Connection Lost
In the image you sent, there is a validation error. So I assume you tried to validate before the heartbeat check, so the validation failed and that error appeared instead.

jomarko (approved):
Ok, thank you @Josephblt , this PR is fine for me, I will get in touch in private channel with you to chat a little bit about the documentation for this extension.

tiagobento (approved):

PR #2168, opened 2024-02-20
kie-issues#954: Extended Services Runner erases data on dates with Decision Update

### LINKED ISSUE #954: Extended Services Runner erases data on dates with Decision Update
When creating a decision and looking at dates that are in the generated form, if you make any change to the DMN Decision (add or remove ()s for instance, not changing the data types), the form wipes out all data. This makes the process of updating rules and validating much harder if you have to keep putting in the dates. The ideal state with decision changes are that the data should remain in place and the update not affect the data types. This is how you would typically test a decision to validate it is working correctly with the data you know should render the appropriate answer. See attached gif for example.

ljmotta:
Closes https://github.com/apache/incubator-kie-issues/issues/954

## Description
The `DateTimeField` component saves its value as a `Date` object and AJV handles the `date-time` type as a `string`. This discrepancy caused the `Date` value to be removed after a new edit. This PR skips the error keyword `type` for the `date-time` type when the data is a `Date`.

https://github.com/apache/incubator-kie-tools/assets/24302289/8c1f17a8-e7ff-44c7-8743-6875a9465389

### DIFF
--- a/packages/dmn-runner/src/jsonSchema.ts
+++ b/packages/dmn-runner/src/jsonSchema.ts
@@ -77,8 +77,8 @@ export function removeChangedPropertiesAndAdditionalProperties<T extends Validat
         return;
       }
 
-      // uniforms-patternfly saves date-time component value as a Date object and
-      // AJV handles data-time as an string, causing an error with keyword type.
+      // uniforms-patternfly saves the DateTimeField component value as a Date object and
+      // AJV handles data-time as a string, causing an error with keyword type.
       // Also, the ajv.ErrorObject doesn't correctly type the parentSchema property
       if (
         error.keyword === "type" &&
--- a/packages/dmn-runner/src/jsonSchema.ts
+++ b/packages/dmn-runner/src/jsonSchema.ts
@@ -77,6 +77,17 @@ export function removeChangedPropertiesAndAdditionalProperties<T extends Validat
         return;
       }
 
+      // uniforms-patternfly saves date-time component value as a Date object and
+      // AJV handles data-time as an string, causing an error with keyword type.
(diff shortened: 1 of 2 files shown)


tiagobento (approved):

packages/dmn-runner/src/jsonSchema.ts
```diff
@@ -77,6 +77,17 @@ export function removeChangedPropertiesAndAdditionalProperties<T extends Validat
         return;
       }
 
+      // uniforms-patternfly saves the DateTimeField component value as a Date object and
+      // AJV handles data-time as a string, causing an error with keyword type.
+      // Also, the ajv.ErrorObject doesn't correctly type the parentSchema property
+      if (
+        error.keyword === "type" &&
+        (error.parentSchema as any)?.format === "date-time" &&
```
jomarko: Shouldn't we use `DATE_TIME` constraint instead of `"date-time"` hardcoded string in this code snippet?

```
enum DmnAjvSchemaFormat {
  DATE = "date",
  TIME = "time",
  DATE_TIME = "date-time",
  DAYS_TIME_DURATION = "days and time duration",
  YEARS_MONTHS_DURATION = "years and months duration",
}
```

jomarko (commented):
Thank you for a quick fix, one question from me @ljmotta

packages/dmn-runner/src/jsonSchema.ts
ljmotta: Great catch! I'll make the change.

jomarko (approved):
Thank you for updates @ljmotta .

PR #2195, opened 2024-03-12
kie-issues#959: Create example apps that use the Quarkus Dev UIs directly on the examples directory of `kie-tools`

### LINKED ISSUE #959: Create example apps that use the Quarkus Dev UIs directly on the examples directory of `kie-tools`

fantonangeli:
Closes: https://github.com/apache/incubator-kie-issues/issues/959

**Notes:**
- This PR starts from https://github.com/apache/incubator-kie-tools/pull/2167 to use sonataflow-quarkus-devui-extension
- I created a spreadsheet with the new kie-tools examples, with old and new GAVs: 
https://docs.google.com/spreadsheets/d/1sDF18E4CEmuP5FZa71y7B_o5ReG_4AO0Ryg-fUPpaI4/edit?usp=sharing

**Migrated kogito-examples:**
- drools-process-usertasks-quarkus-example
- sonataflow-greeting-quarkus-example

Please let me know if we need more examples.

**Preview:**
[kie-issues#959.webm](https://github.com/apache/incubator-kie-tools/assets/17780574/36802b05-d111-4d8d-aa95-23d5ee1857e3)


tiagobento:
@fantonangeli Not a big fan of the multiple parents..... Can't we do a single "examples" parent? What are the configurations that these "parents" are passing down to its children modules?

tiagobento:
@fantonangeli Having the parent `pom.xml` at the `examples` directory is misleading, as not all examples are parented by that `pom.xml` file. More than that, this `pom.xml` wouldn't be owned by any package in particular, making it impossible to tell, using our current setup, what to build when there's a change made to it.

To successfully create this parenting mechanism inside `kie-tools`, the best way is to create a new package. Please take a look at what Pere did here #2223 with the `maven-base` package. The parenting strategy for Maven packages is not well defined yet, but I feel like the best way forward for us is having everything concentrated at the `maven-base` package, to act as a centralized place for us to manage our Maven versions. I also want to avoid a complex hierarchy between Maven packages. Flat structures FTW :)

For now, though, don't worry about this Maven parent unification, so please create a new `quarkus-examples-maven-parent` package.

Also, WDYT about:
1. Renaming `decisiontable-quarkus-example` to `drools-xls-decision-table-quarkus-example`;
2. Renaming `sonataflow-greeting-quarkus` to `sonataflow-greeting-quarkus-example`

so we keep a consistency `[domain]-[feature/name]-quarkus-example`...

cc @paulovmr @caponetto

fantonangeli:
Hi @tiagobento, thanks a lot for your advice and help here.
+1 about naming consistency.
+1 avoid having a pom directly in the `examples` dir.

 Regarding creating a new package `quarkus-examples-maven-parent `, I don't think we need this package to have an `env/index.js` and a `install.js` as the children modules have it defined already. 
 This package doesn't also need a build configuration on a `package.json` as the children modules have their build configuration.
 So IMHO we don't need a new pnpm package but this can be just a simple dir `examples/quarkus-examples-maven-parent ` with a pom inside. 
 But
WDYT?

tiagobento:
@fantonangeli The only problem I see with this suggestion is in the future... let me exaplain.

A directory inside `packages`/`examples`/`scripts` without a `package.json` and a clear reference to it on their dependent packages opens the precedence of "directories" becoming a concept, while we're actually aiming for the 'atom' of the KIE Tools build system to be a "package". 

Also, even if `examples/quarkus-examples-maven-parent` is really simple, and it really doesn't have much inside it, if we make a change to its `pom.xml` and break it, we wouldn't find out, since we wouldn't be building it. We'd only find out when building the packages that depend on it, which IMHO is not a great precedence to have too.

Don't worry about repeating the `env/index.js` everywhere, it's expected to work like that, and it is really easy to make changes to it, as pretty much everything related to our` build-env` mechanism is grep-able.

tiagobento:
Btw, I think we should wait for #2223 so we can have the examples point to it, WDYT? cc @pefernan 

Or do you want to do the opposite? Merge the examples and Pere fixes it on #2223 too?

pefernan:
let me do one minute change on my PR

fantonangeli:
@tiagobento thank you for your clear explanation. I understand now and agree on the possible issues we can have in the future by having a simple directory and not a package.
As https://github.com/apache/incubator-kie-tools/pull/2223 has already one approval, maybe it's better to wait for @pefernan's PR, then I will update this PR accordingly

fantonangeli:
Hi @tiagobento, I updated my PR following https://github.com/apache/incubator-kie-tools/pull/2223. 
The build error seems to be related to missing snapshots for the version `999-20240331-SNAPSHOT` here: 
https://repository.apache.org/content/groups/public/org/kie/kogito/kogito-addons-quarkus-process-management

The build works locally if I change the `KOGITO_RUNTIME_version` to `999-20240218-SNAPSHOT` in `packages/root-env/env/index.js`

Am I missing something in my poms?

pefernan:
@fantonangeli this addon has been renamed to `org.kie:kie-addons-quarkus-process-management` I think that's why you have this issue. 
You can check https://docs.google.com/spreadsheets/d/1IN9bH2rCxbV7dcehItqwELfTNaGrXKFT8Ppj6a5Jeb4/edit#gid=0 if you have similar problems with other addons.

tiagobento:
@fantonangeli I know it doesn't look right, but Drools and Kogito Runtimes versions now are the same thing.... you can use the same property to configure Drools too. There's 0 chance we'll need different versions for these two things.

Another thing is.... maybe now that we have packages/maven-base we can use it instead of creating a new "parent" just for Quarkus examples? I mean, the idea in the future is that all Maven packages inherit this `packages/maven-base` package anyway. WDYT? cc @pefernan @paulovmr

fantonangeli:
Hi @tiagobento, thank you for your last suggestion. I also agree using only one parent as `quarkus-examples-maven-parent` adds only a few properties. I updated this PR with these last suggestions.

tiagobento:
@fantonangeli Thanks for the changes! This is getting in a very good shape.

Please do remove the `DROOLS_version` variable from `root-env` and use the same version of Kogito Runtimes for Drools dependencies. We'll give it a better name in the future, I promise :)

As for the maven-base changes on the pom.xml, I think we can list the dependencies at the <dependencyManagement> section so we can refer to them in the <dependencies> of the examples without specifying a version. It will make the pom.xml files of Maven-based packages look a lot nicer, and it will be much easier to maintain too :)

fantonangeli:
> As for the maven-base changes on the pom.xml, I think we can list the dependencies at the section so we can refer to them in the of the examples without specifying a version. It will make the pom.xml files of Maven-based packages look a lot nicer, and it will be much easier to maintain too :)

@tiagobento I can only find a dependency with a version. Please let me know if I'm wrong.
examples/drools-process-quarkus-example/pom.xml
Commit:  https://github.com/apache/incubator-kie-tools/pull/2195/commits/a0b8eb46c78f839301210810a38cc7ea6b8a0409

tiagobento:
@fantonangeli The commit you sent is the correct approach. Dependencies with versions should be inside <dependencyManagement> on `maven-base`. All other Maven-based packages should be able to declare dependencies without specifying a version.

tiagobento:
@fantonangeli Since we are introducing a lot of Maven-based packages on this PR, I'm taking the opportunity to make our Maven configurations better here. One point that I need to address is the fact that we're declaring the <parent> using a relative path to the `maven-base` directory.

This has the potential of making a package work without declaring a dependency to the `maven-base` package in the package.json. Since `package.json` is the file that controls de build order and everything, we need to use the mechanisms it provides to only access stuff that we declare. Please apply this change not only to the new Maven packages you're creating here, but to other Maven packages declaring `maven-base` as parent:

`<relativePath>../../packages/maven-base/pom.xml</relativePath>`

becomes

`<relativePath>./node_modules/@kie-tools/maven-base/pom.xml</relativePath>`

This way, packages that fail to declare `maven-base` as a dependency on their package.json files will throw an error, keeping our build safe.

fantonangeli:
@tiagobento relativePaths updated. Thank you

tiagobento:
@fantonangeli There's a lot of properties on the `maven-base` pom.xml file, but I can't see them being reused... are they really necessary?

fantonangeli:
@tiagobento thank you, I reviewed the maven-base/pom.xml properties and together with this I also removed some plugin versions (https://github.com/apache/incubator-kie-tools/pull/2195/commits/8dc54c3b6eaba750e4e07a239baf0f0f8a7bf80e).
Pls let me know if this was ok

tiagobento:
@fantonangeli Some versions were being used, as they are properties that are picked up by some plugins. Other, that were user-defined, though, weren't...

```
    <maven.compiler.source>17</maven.compiler.source>
    <maven.compiler.target>17</maven.compiler.target>
```
 
for example, I think we need them there :)

tiagobento:
@fantonangeli Noticed that the build scripts for the new example modules are not parameterized by `KIE_TOOLS_BUILD__buildExamples`... can you please take a look at the existing examples and apply the same strategy to the new ones? I'm talking about using `run-script-if` and `build-env examples.build`.

This is important for local development where people _usually_ don't want to build/test all the examples when doing a full build.

fantonangeli:
@tiagobento @paulovmr as agreed, I replaced the following examples:
- pmml-quarkus-example 
- drools-process-quarkus-example 
- drools-dmn-quarkus-example
- drools-xls-decision-table-quarkus-example

in favour of:
- drools-process-usertasks-quarkus-example

paulovmr (approved):
Nice, thanks @fantonangeli !

tiagobento:
@fantonangeli Can you please rebase/merge this PR with `main`? There was a problem with `macos-latest` that will only reflect on this PRs CI if you rebase/merge...

fantonangeli:
@tiagobento I updated this PR branch and now the CI checks are green.
Thanks again for your review

tiagobento (approved):

tiagobento:
@fantonangeli Sorry I know this PR is taking a lot of hits... There's just too many things going on. Do you mind rebasing it please?

fantonangeli:
@tiagobento I updated this PR and all checks are green

PR #2199, opened 2024-03-14
kie-issues#1012: Removing `kie-soup-commons` dependency in DashBuilder

### LINKED ISSUE #1012: Removing `kie-soup-commons` dependency in DashBuilder
Dashbuild still depends on a 7.59.0 version of kie-soup-commons. 
That dependency is required to perform null (and empty) checks on Objects only, so we can safely remove it and replace it with native Java API.

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/1012

Dashbuild still depends on a 7.59.0 version of kie-soup-commons. 
That dependency is required to perform null (and empty) checks on Objects only, so we can safely remove it and replace it with native Java API.

### DIFF
--- a/packages/dashbuilder/appformer/uberfire-api/pom.xml
+++ b/packages/dashbuilder/appformer/uberfire-api/pom.xml
@@ -44,11 +44,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.kie.soup</groupId>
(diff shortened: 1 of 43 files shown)


porcelli (approved):
good job @yesamer - towards a sustainable codebase!

paulovmr (approved):

PR #2193, opened 2024-03-11
kie-issues#965: Make sure `kie-tools` is using the latest timestamped SNAPSHOT from Kogito (999-20240403-SNAPSHOT)

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/965

PR #2207, opened 2024-03-25
kie-issues#1043: Use shared DMN models for testing purpose

### LINKED ISSUE #1043: React-Based DMN Editor: use shared DMN models for testing purpose
DMN models used for testing purposes are moving to the [kie-dmn-test-resources](https://github.com/apache/incubator-kie-drools/tree/main/kie-dmn/kie-dmn-test-resources) module.
[Readme](https://github.com/apache/incubator-kie-drools/blob/32c2ac248937c0e6b56b5d80d6253e2469a3d618/kie-dmn/kie-dmn-test-resources/Readme.md) file gives a quick explanation of its structure.
It would be better to start using such files:
1. to be sure that both editor and engines works and behaves the same way
2. such files are tested/validated engine side, so we are sure about their validity

@tiagobento @yesamer

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/1043

Requires 
- https://github.com/apache/incubator-kie-tools/pull/2193 
- https://github.com/apache/incubator-kie-tools/pull/2222
- https://github.com/apache/incubator-kie-tools/pull/2229

This PR:

1. import models from `org.kie:kie-dmn-test-resources` in the newly-created package `dmn-testing-models`
2. import JBang scripts to perform checks against backend code (Validation, Compilation, and SemanticComparison) in the newly-created package `dmn-marshaller-backend-compatibility-tester`
3. refactor idempotency.test.ts to use the models imported in point 1
4. Introduced dmnValidation.test that relies on the DMNValidation JBang script (using the models imported in point 1)
5. Introduced dmnSemanticComparison.test that relies on the DMNSemanticComparison JBang script (using the models imported in point 1)
6. Introduced xsdValidation .test to validate DMN files produced by the dmn-parser against its XSD schema (using the models imported in point 1)


packages/dmn-testing-models/README.md
```diff
@@ -0,0 +1,10 @@
+## DMN testing models
+
+This package is meant to contain all the DMN models published inside `org.kie:kie-dmn-test-resources` to make them available for testing purposes.
```
jomarko: Could we please add also a link for github repository? https://github.com/apache/incubator-kie-drools/tree/main/kie-dmn/kie-dmn-test-resources

packages/dmn-testing-models/env/index.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+//const { getOrDefault, varsWithName, composeEnv } = require("@kie-tools-scripts/build-env");
```
jomarko: we can remove this line probably

packages/dmn-testing-models/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+Simply issue `mvn clean generated-resources` to download the jar and extract the models under `target/generated-resources/valid_models`.
```
jomarko:
```suggestion
Simply issue `mvn clean generate-resources` to download the jar and extract the models under `target/generated-resources/valid_models`.
```

packages/dmn-marshaller/tests-data--manual/other/decisionAndInputWithAddition.dmn
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jomarko, line 1: Maybe we should move also this model under https://github.com/apache/incubator-kie-drools/tree/main/kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5 ?

packages/dmn-marshaller/tests/idempotency.test.ts
```diff
@@ -20,30 +20,61 @@
...
 
 const files = [
-  "../tests-data--manual/other/attachment.dmn",
-  "../tests-data--manual/other/empty13.dmn",
-  "../tests-data--manual/other/sample12.dmn",
-  "../tests-data--manual/other/list.dmn",
-  "../tests-data--manual/other/list2.dmn",
-  "../tests-data--manual/other/external.dmn",
-  "../tests-data--manual/other/weird.dmn",
-  "../tests-data--manual/dmn-1_4--examples/Chapter 11 Example 1 Originations/Chapter 11 Example.dmn",
```
jomarko: shouldn't this PR physically remove these models? are they still used elsewhere?

packages/dmn-marshaller/tests/idempotency.test.ts
```diff
@@ -20,30 +20,61 @@
...
+  }
+});
 
-      expect(xml_firstPass).toStrictEqual(xml_secondPass);
-    });
+describe("idempotency_all", () => {
+  for (const models_paths of testing_models_paths) {
+    const parent_path = path.join(__dirname, models_paths);
+    testDirectory(parent_path);
   }
```
jomarko: sorry, it may be silly question, I am not expert of reading directories in typescript, I also believe this does the expected work, I am just concerned if in the future for some reason we move the directory or models of the directory to a new location, this will silently still pass reading zero files?

maybe some assertion of how many files were tested would make slightly more explicit, what models do we test? I do not know, or in different test suites I saw a hard-coded list of files intended to be tested.

jomarko (commented):
thank you for the PR

packages/dmn-marshaller/tests/idempotency.test.ts
gitgabrio: @jomarko They are used elsewhere (e.g. typeSafety.test.ts, versions.test.test)

packages/dmn-marshaller/tests-data--manual/other/decisionAndInputWithAddition.dmn
gitgabrio: @jomarko Scope of the different version folders is to have models showing features specific for that version.
IINW that model has nothing specific for DMN 1.5 (beside the tag)
Unfortunately, until now we did not enforced a strong classification/organization of models, result being that it is near to impossible to sort out the models based on the version-specific features they demonstrate.
From now on, we plan to keep such organization, so that whenever a new DMN version comes out, version-specific models will be stored under correct folder, and will be easier to manage/use them for testing purposes.

packages/dmn-testing-models/env/index.js
gitgabrio: Thanks for spotting this ! 👍

packages/dmn-marshaller/tests/idempotency.test.ts
gitgabrio: @jomarko I see the point, but IMO:

1. these class has two testing methods, one of which uses explicitily defined files
2. one of the goal is to allow "transparent" inclusion of new models, i.e. whenever we add a model to the original jar, it will immediately goes under test, enhancing bug-discovery time; having to synchronize the expected number (or also name) of files would impair noticeably that, making PR much more cumbersome to create
3. if we move the test elsewhere, the method will throw a directory not found (or similar) error
4. the only way to move files to different directory would be changing the pom.xml itself, in which case I'm expecting people to be aware of consequences
5. the other possibility would be to change the structure of the jar itself, but there seems no reason to do that and, eventually, the modification should be reflected to the pom.xml, to keep the same destination folder

Does this make sense ? WDYT ?

packages/dmn-marshaller/tests-data--manual/other/decisionAndInputWithAddition.dmn
jomarko: Thank you, yes, then keep this file as it is currently

packages/dmn-marshaller/tests/idempotency.test.ts
jomarko: Yes, it sounds you are confident this is robust enough, so I think we can keep as you proposed. I agree that people contributing and doing reviews will probably spot if some new model is not tested.

jomarko:
@gitgabrio please fix the pom.xml formatting, see the logs:
```
> kie-tools-root@0.0.0 format:check /home/runner/work/incubator-kie-tools/incubator-kie-tools
> prettier --check . '**/*.xml'

Checking formatting...
[warn] packages/dmn-testing-models/pom.xml
[warn] Code style issues found in the above file. Forgot to run Prettier?
 ELIFECYCLE  Command failed with exit code 1.
 ```

File: packages/dmn-marshaller/tests/idempotency.test.ts [hunk 7]
```diff
@@ -20,30 +20,61 @@
...
+  "../node_modules/@kie-tools/dmn-testing-models/dist/valid_models/DMNv1_5/ListReplaceEvaluate.dmn",
+  "../node_modules/@kie-tools/dmn-testing-models/dist/valid_models/DMNv1_5/TypeConstraintsChecks.dmn",
+  "../node_modules/@kie-tools/dmn-testing-models/dist/valid_models/DMNv1_x/OneOfEachType.dmn",
+  "../node_modules/@kie-tools/dmn-testing-models/dist/valid_models/DMNv1_x/allTypes.dmn",
+];
+
+const testing_models_paths = [
+  "../node_modules/@kie-tools/dmn-testing-models/dist/valid_models/DMNv1_5",
+  "../node_modules/@kie-tools/dmn-testing-models/dist/valid_models/DMNv1_x",
 ];
```
tiagobento: In order not make packages dependent on the directory structure of the monorepo, for these cases we use `require.resolve("@kie-tools/dmn-testing-models")`, using the package name, instead of the package directory. The command above is like an import, but for a directory inside a package. You can use `path.join` then to concatenate the result of `require.resolve` with the path string you're pointing to. Keep in mind that we need to support these tests on Windows as well, if I'm not mistaken.

packages/dmn-marshaller/tests/idempotency.test.ts
```diff
@@ -20,30 +20,61 @@
...
+      testFile(child_path);
+    } else {
+      testDirectory(child_path);
+    }
+  });
+}
+
+function testFile(fullPathOfFile: string) {
+  test(fullPathOfFile.substring(fullPathOfFile.lastIndexOf("/") + 1), () => {
+    console.log("Testing " + fullPathOfFile);
```
tiagobento: We don't need this log here, as the name of the test already makes it clear what's being tested.

packages/dmn-marshaller/tests/idempotency.test.ts
```diff
@@ -20,30 +20,61 @@
...
+    const stats = fs.statSync(child_path);
+    if (stats.isFile()) {
+      testFile(child_path);
+    } else {
+      testDirectory(child_path);
+    }
+  });
+}
+
+function testFile(fullPathOfFile: string) {
```
tiagobento: We usually name `path` string variables with some more details.

Please take a look at this comment: https://github.com/apache/incubator-kie-issues/issues/783

packages/dmn-testing-models/env/index.js
```diff
@@ -0,0 +1,28 @@
...
+const { varsWithName, composeEnv } = require("@kie-tools-scripts/build-env");
+
+module.exports = composeEnv([require("@kie-tools/root-env/env")], {
+  vars: varsWithName({}),
+  get env() {
+    return {};
+  },
+});
```
tiagobento: You need the package.json version here to be exposed as a variable (*)....

packages/dmn-testing-models/install.js
```diff
@@ -0,0 +1,25 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+const buildEnv = require("./env");
+const { setup } = require("@kie-tools/maven-config-setup-helper");
+
+setup(`
+    -Dversion.org.kie.kogito=${buildEnv.env.kogitoRuntime.version}
```
tiagobento, line 26: (*)... so you can add a `-Drevision` here. In `kie-tools`, the version of a package is determined by the package.json, and all technology-specific files are generated/parameterized from there. That's why you see `-Drevision` configured for other Maven packages.

packages/dmn-testing-models/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.kie.tools</groupId>
+    <artifactId>dmn-testing-models</artifactId>
+    <version>${revision}</version>
+    <packaging>jar</packaging>
+
+    <name>DMN Testing Models</name>
+    <description>Testing models for DMN editor</description>
```
tiagobento: Not sure if that's the better description here, as these models are currently only being used by the marshaller... Maybe we can leave this as a generic description...

packages/dmn-testing-models/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+        <developerConnection>scm:git:git@github.com:apache/incubator-kie-tools.git</developerConnection>
+        <url>scm:git:git@github.com:apache/incubator-kie-tools.git</url>
+    </scm>
+
+    <developers>
+        <developer>
+            <name>All developers are listed in the KIE GitHub organization</name>
+            <url>https://github.com/orgs/kiegroup/people</url>
+        </developer>
+    </developers>
```
tiagobento: Maybe remove this?

packages/dmn-testing-models/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+        </dependency>
+    </dependencies>
+
+    <build>
+<!--        <resources>-->
+<!--            &lt;!&ndash; DMN generated resources unpacked from org.kie:kie-dmn-test-resources &ndash;&gt;-->
+<!--            <resource>-->
+<!--                <directory>${project.build.directory}/generated-resources/valid_models</directory>-->
+<!--            </resource>-->
+<!--        </resources>-->
```
tiagobento: Remove?

packages/dmn-testing-models/package.json
```diff
@@ -0,0 +1,43 @@
...
+  ],
+  "scripts": {
+    "build:dev": "run-script-os",
+    "build:dev:linux:darwin": "mvn clean generate-resources",
+    "build:dev:win32": "pnpm powershell \"mvn clean generate-resources\"",
+    "build:prod": "pnpm build:dev",
+    "copy:app": "run-script-os",
+    "copy:app:linux:darwin": "copyfiles -u 1 \"target/generated-resources/valid_models/**/*.dmn\" dist/",
+    "copy:app:win32": "copyfiles -u 1 \"target/generated-resources/valid_models/**/*.dmn\" dist/",
+    "dist": "rimraf dist && mkdir dist && pnpm copy:app",
```
tiagobento: These 4 scripts aren't being used, from what I understood. Can you please check and remove them if that's true?

packages/dmn-testing-models/README.md
tiagobento: I guess this paragraph is not true, as we're unpacking directly at `dist`, no?

packages/dmn-marshaller/tests/idempotency.test.ts
```diff
@@ -20,30 +20,61 @@
...
-      const xml_original = fs.readFileSync(path.join(__dirname, file), "utf-8");
-
-      const { parser, builder } = getMarshaller(xml_original, { upgradeTo: "latest" });
-      const json = parser.parse();
-
-      const xml_firstPass = builder.build(json);
-      const xml_secondPass = builder.build(getMarshaller(xml_firstPass, { upgradeTo: "latest" }).parser.parse());
+    testFile(path.join(__dirname, file));
+  }
+});
```
tiagobento: Why stop testing the models that are currently being tested?

packages/dmn-marshaller/tests/idempotency.test.ts
```diff
@@ -20,30 +20,61 @@
 import * as fs from "fs";
 import * as path from "path";
 import { getMarshaller } from "@kie-tools/dmn-marshaller";
+import { DmnMarshaller, DmnMarshallerOpts, DmnMarshallerVersions } from "../src";
```
tiagobento: Please do not import from `src` directly, and instead use `@kie-tools/dmn-marshaller`.

packages/dmn-marshaller/tests/xsdSequence.test.ts
```diff
@@ -21,74 +21,57 @@ import * as fs from "fs";
...
-    </dmndi:DMNDiagram>
-  </dmndi:DMNDI>
-</definitions>
-`);
-    });
-  }
+    const xmlRetrieved = marshaller.builder.build(json).replace(/\s/g, "");
+    const xmlExpected = fs.readFileSync(path.join(__dirname, fileExpected.path), "utf-8").replace(/\s/g, "");
+    expect(xmlRetrieved).toStrictEqual(xmlExpected);
+  });
```
tiagobento: Several comments here:

1. 
Using an array for `fileSource` and `fileExpected` is not the way to go here. I would inline the paths, to be honest.

2.
Having one `describe` for each `test` looks weird too. I would do one describe like this:

`describe("build always produces elements in the same order"`

and two tests like this:

`test("after adding elements")`
`test("after reading denormalized file")`


3. 
Replacing blank spaces with an empty string is tempering with the tests. My suggestion is that we match the files char by char, without exceptions.

tiagobento (changes requested):
Hey @gitgabrio ! Amazing to see this PR! I left a few comments inline, mostly for things we are used to doing that are not written anywhere.... 

Anyway, thanks A LOT for doing this, I guess this will help us achieve amazing parity between Runtimes and Tools.

packages/dmn-marshaller/tests/idempotency.test.ts
gitgabrio: (I was sure to have done something extemely naive, thanks for suggestion)

packages/dmn-marshaller/tests/idempotency.test.ts
gitgabrio: Not sure to understand the question

packages/dmn-marshaller/tests/xsdSequence.test.ts
gitgabrio: > Several comments here:
> 
> 
> Using an array for `fileSource` and `fileExpected` is not the way to go here. I would inline the paths, to be honest.
> 

Could you provide an example ? It is not clear to me (is that already used somewhere else ?)

> 
> Having one `describe` for each `test` looks weird too. I would do one describe like this:
> 
> `describe("build always produces elements in the same order"`
> 
> and two tests like this:
> 
> `test("after adding elements")` `test("after reading denormalized file")`
> 

Fine

> 
> Replacing blank spaces with an empty string is tempering with the tests. My suggestion is that we match the files char by char, without exceptions.

The `toStrictEqual` method fails here, because the generated xml has different formatting/white space then the original one.
One of the caveat of this test is to be sure that the generated  xml is absolutely identical to the original one as it is: if the "litmus test" xml is somehow modified by the very code that is under test, then the result of the test could be "invalid".
e.g.

```javascript
1 const xml_original = fs.readFileSync(fullPathOfFile, "utf-8");
2 const { parser, builder } = getMarshaller(xml_original, { upgradeTo: "latest" });
3 const json = parser.parse();
4 const xml_firstPass = builder.build(json);
5 const xml_secondPass = builder.build(getMarshaller(xml_firstPass, { upgradeTo: "latest" }).parser.parse());
6 expect(xml_firstPass).toStrictEqual(xml_secondPass);
```

1. we are testing the "marshaller" (the parser and the build const)
2. the string "xml_firstPass"  (line 4)  is a "product" of the marshaller, created from the original string
3. the string "xml_secondPass"  (line 5)  is another "product" of the marshaller, created from the  "xml_firstPass" string
4. at line 6, the test verify if two products of the marshaller are exactly the same
5. in that case the `toStrictEqual` succeed because both strings have been generated by the marshaller, but at the same time it does not guarantee that the marshaller itself does not introduce/modify/whatever something wrong in both of them
6. that's why I wanted to test the raw, original xml with the one created by the marshaller, and only way to do that comparison is to strip out whitespaces.
If there is a nicer command, fine for me, but what it is important IMO is the logic and princiuple of the test itself.
Does this make sense ? Am I clear ?

packages/dmn-marshaller/tests/idempotency.test.ts
tiagobento: This comment is at the wrong place, let me go to the correct place.

packages/dmn-marshaller/tests/idempotency.test.ts
tiagobento: Ok, but why should we stop using those files in the idempotency tests?

packages/dmn-marshaller/tests/xsdSequence.test.ts
tiagobento: Yep, I got it, you have a good point, and we indeed need a good strategy to make sure we're covered. My observation was about `xsdSequence.test.ts`, where you introduced the `/\s/g` replace. If we ignore spaces and new lines, we are ignoring the new line treatment for FEEL expressions, for example, which is very important.

IMHO we should normalize the files by formatting them with Prettier, programatically. We've done it in other places, no reason why we can't do it here as well. We trust Prettier won't break anything when formatting, so doing:

1. Read unformatted file
2. Formatting it with Prettier, using the XML rules.
3. Parsing it with our Marshaller.
4. Building it with our Marshaller, and formatting it with Prettier.
5. Comparing formatted original XML with formatted built XML.

should be the way to go, IMHO. WDYT?

tiagobento: Btw our Marshaller will sometimes produce an XML that it NOT a 1:1 match with the original.

Examples include:

1. Self-closing tags.
2. CDATA blocks get flattened and all values get escaped using XML entities.
3. Comments
4. xsd:sequence order (We can read in any order, but will always build in the correct order declared by xsd:sequences on the XSD)

yesamer (approved):
Blocked by https://github.com/apache/incubator-kie-tools/pull/2193

yesamer (changes requested):
Blocked by https://github.com/apache/incubator-kie-tools/pull/2193

packages/dmn-marshaller/tests/idempotency.test.ts
gitgabrio: @tiagobento 
The important thing is the quality/kind of tests, and not the number of them.
As it currently is, the idempotency test has a very focused meaning, which does not change if executed with 10 or 100 files that basically has the same features.
On the other side, the scope of "centralize DMN testing files" is exactly to have and use the centralized ones, slowly removing all the others (beside very specific ones and use cases.
Anyway, if you want them back, no prblm to me 👍

packages/dmn-marshaller/tests/xsdSequence.test.ts
gitgabrio: @tiagobento 
About last comment, I get your point, but currently I see no alternatives, and TBH this string comparison is by itself a bad way to verify the behavior of this kind of code.
We spent a lot of time setting up a testing strategy for code-generation, backend side, and if you think about it this is more or less the same thing. We actually need something that is able to verify semantically the two objects (expected and retrieved), but currently there is not. Until then,  IMO the current proposal is an improvement, anyway, at least for the moment being.

gitgabrio:
@yesamer Just pushed last modifications.

packages/dmn-testing-models/env/index.js
yesamer: Done

packages/dmn-testing-models/install.js
yesamer: Done

packages/dmn-testing-models/pom.xml
yesamer: Done

packages/dmn-testing-models/pom.xml
yesamer: Done

packages/dmn-testing-models/pom.xml
yesamer: Done

packages/dmn-marshaller/tests/idempotency.test.ts
yesamer: Done

packages/dmn-marshaller/tests/idempotency.test.ts
yesamer: @tiagobento I applied your suggestion, but I need to introduce an empty index.js in the src and assign it as `main` parameter in package.json. Without that, `require.resolve("@kie-tools/dmn-testing-models")` is not resolved as module. Is that fine or there's a nicer way to do that ?

packages/dmn-marshaller/tests/idempotency.test.ts
yesamer: Done (I hope correctly)

packages/dmn-marshaller/tests/xsdSequence.test.ts
tiagobento: Well, if we normalize the generated XML, and test for known discrepancies like the ones I shared above, I think we'll be fine. As for semantically verifying, we can always use the DOM. It's the most detailed semantic representation of XML documents, and it looks like there's a comparison function available. See https://developer.mozilla.org/en-US/docs/Web/API/Node/isEqualNode.

packages/dmn-marshaller/tests/idempotency.test.ts
tiagobento: I agree, but some files I put there have some specific behavior that was being tested before, and the new files may not reproduce those behaviors. I would prefer to keep them for the time being.

packages/dmn-marshaller/tests/idempotency.test.ts
tiagobento: Right, so the entire `src` folder is not necessary. If you remove the `main` property from the `package.json` it might work. If that doesn't work, I would leave an empty index.js somewhere so `main` can point to this file. Doesn't have to be inside `src`, can be at the root of the package.

packages/dmn-marshaller/tests/idempotency.test.ts
yesamer: Restored

jomarko:
the build fails seems to be related to the changes in this PR, please let us know once  the PR is ready for a re-review

```
Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.6.1:unpack (unpack) on project dmn-testing-models: Unable to find/resolve artifact. The following artifacts could not be resolved: org.kie:kie-dmn-test-resources:jar:tests:999-20240218-SNAPSHOT (absent): Could not find artifact org.kie:kie-dmn-test-resources:jar:tests:999-20240218-SNAPSHOT in apache-public-repository-group 
```

gitgabrio:
@jomarko 
Yes, @yesamer mentioned [it](https://github.com/apache/incubator-kie-tools/pull/2207#pullrequestreview-1958703432)

packages/dmn-marshaller/tests/xsdSequence.test.ts
yesamer: I'm trying to set in place the steps described by @tiagobento  [here](https://github.com/apache/incubator-kie-tools/pull/2207#discussion_r1538040788) and I finally found a good configuration for Prettier to solve the formatting issue found by @gitgabrio  (eg. trailing whitespaces).
After that, the configuration revealed what Tiago already anticipated [here](https://github.com/apache/incubator-kie-tools/pull/2207#discussion_r1538044196). Because of the following cases, we don't have a 1:1 copy:

**kie namespace**: If already present in the original xml file, the one generated by our parser replaces it with a new version. If is missing, the parser adds it.
Screenshot 2024-04-03 at 16 35 01
Screenshot 2024-04-03 at 16 35 15

**xml prolog**: The parser automatically adds it, if not present in the original file
Screenshot 2024-04-03 at 16 44 28

**Special characters** Some special characters are converted (eg. `"` is converted to `&quote`)
Screenshot 2024-04-03 at 16 48 30

Self-closing tags case seems correctly managed with Prettier.

packages/dmn-marshaller/tests/jbang/dmnValidation.java
```diff
@@ -27,16 +26,16 @@ public static void main(String... args) throws Exception {
 
     public static int validate(String... args) throws Exception {
         if (args.length > 2 || args.length <= 0) {
-            throw new IllegalArgumentException("Validation requires 1 or 2 xmls");
+            throw new IllegalArgumentException("Validation requires 1 or 2 DMN file paths");
```
gitgabrio: @yesamer 
Please keep in mind that there could be more then two models in the "import" scenario; e.g.

1. model A import both model B and model C
2. model A import model B that, in turn, import model C
3. ecc

Please remove the `args.length > 2 ` constraint, and adapt the exception message accordingly.

packages/dmn-marshaller/tests/dmnValidation.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      const importedMarshalledXMLFilePath = parseXMLAndWriteInFile(normalizedFsPathRelativeToTheFiles.imported);
+      const importerMarshalledXMLFilePath = parseXMLAndWriteInFile(normalizedFsPathRelativeToTheFiles.importer);
+
+      try {
+        executeJBangScript(jbangDmnValidationScriptPath, importedMarshalledXMLFilePath, importerMarshalledXMLFilePath);
+      } catch (error) {
+        const fileName = normalizedFsPathRelativeToTheFiles.importer.substring(
+          normalizedFsPathRelativeToTheFiles.importer.lastIndexOf(path.sep) + 1
+        );
+        fail("Validation of " + error + " failed! Please scroll up to DMN VALIDATION ERROR message to see the reason");
```
jomarko:
```suggestion
        fail("Validation of " + fileName + " failed! Please scroll up to DMN VALIDATION ERROR message to see the reason");
```

packages/dmn-marshaller/tests/jbang/dmnValidation.java
jomarko, line 1: what is the difference between this file, `dmnValidation.java` and `dmnSemanticComparision.java` please?

packages/dmn-marshaller/tests/jbang/jbangManager.ts
```diff
@@ -0,0 +1,33 @@
...
+const jbang = require("@jbangdev/jbang");
+const buildEnv = require("../../env");
+
+export function executeJBangScript(scriptPath: string, ...args: string[]) {
+  /* Windows requires double quotes to wrap the argument, while in POSIX it must be wrapped by single quotes */
+  const isWindowsPath = path.sep !== "/";
+  const quoteChar = isWindowsPath ? '"' : "'";
+  jbang.exec("--java 17", "properties@jbangdev", "java.version");
+  jbang.exec(
+    "-Dkogito-runtime.version=" + buildEnv.env.kogitoRuntime.version,
```
jomarko: am I correct `kogito-runtime.version` as property, its name, is defined in `drools` repository? maybe some link/comment for it would help in case it is changed in future?

packages/dmn-testing-models/install.js
```diff
@@ -0,0 +1,26 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+const buildEnv = require("./env");
+const { setup } = require("@kie-tools/maven-config-setup-helper");
+
+setup(`
+    -Drevision=${buildEnv.env.dmnTestingModel.version}
+    -Dversion.org.kie.kogito=${buildEnv.env.kogitoRuntime.version}
```
jomarko, line 26: is it correct here we do not use `kogito-runtime.version`?

jomarko (commented):
@yesamer congrats to green build, some questions inline from me

packages/dmn-marshaller/tests/jbang/dmnValidation.java
yesamer: @jomarko dmnSemanticComparision.java is still a WIP, that will compare the semantic of two different XML files

packages/dmn-marshaller/tests/jbang/jbangManager.ts
yesamer: @jomarko No, that property name is used in the jBang script (`dmnValidation.java`)

packages/dmn-testing-models/install.js
yesamer: @jomarko  Yep, it's correct. `version.org.kie.kogito` is used in the module pom.xml file, while `kogito-runtime.version` is used in the JBang script.

gitgabrio (commented):
HI @yesamer - thanks for your work on that!
There are different details that could be improved, IMO.
1. use picocli to have better control over parameters
2. it would be possible to use one single class, switching method executed depending on one of the above parameter (we used `--command` somewhere else)
3. with picocli, you may pass arbitrary parameters, also lists
4. if you prefer to keep two different classes, I would suggest to create a parent one, to avoid duplication of `//DEPS` and `//REPO`
5. about dmnSemanticComparison.compare - currently seems a little bit frail; I would suggest to make two different methods, one that accept two files (models without import), and another that accept two lists of files (for models with import; again, with picocli this is easily doable)

Last note (trivial): being those classes actual java classes, it would be nice to have them named starting with uppercase letter

yesamer:
@gitgabrio Thank you!
I intentionally ignored the `picocli` part so far, my first goal is to make the test work and then I'll improve that part.

About the class name, I followed the same convention I saw in the [JBangdemo](https://github.com/jbangdev/jbang-npm/tree/main/demo). Honestly, I don't know if that is a JBang convention, a mistake, or just a way to say that is "script". I don't have any strong preference for that, if In our group we managed all the JBang scripts in that way, no problem to use Uppercase there.

yesamer:
@gitgabrio @jomarko @tiagobento Ready for another review round.

yesamer (approved):
Approved @gitgabrio

gitgabrio (commented):
I think this is fine!

yesamer:
@tiagobento This is ready for another review round.

- I added the new module that holds the JBang stuff, as agreed
- I implemented the fetching mechanism as agreed. We can't directly call `jbang` in the `package.json` file, because after installing JBang with the preinstall hook, it requires a new Shell for changes to take effect. So I created a new TS file to manage this case. The file is then called in the package.json "fetch" script.

For further info, please ping me or just read the READMEs I wrote.

packages/dmn-testing-models/README.md
```diff
@@ -0,0 +1,21 @@
...
+### Usage
+
+The command `mvn clean generate-resources` downloads the jar and extract the models under `dist/`.
+
+To make them available for testing purpose:
+
+1. set the `KOGITO_RUNTIME_version` in the terminal (if different from the default one)
+2. `pnpm bootstrap` on the root directory
+3. cd dmn-testing-models
+4. `pnpm install` (this creates the required mvn configuration)
```
tiagobento: Nto necessary. `install` scripts are executed as part of the `bootstrap` script, when `bootstrap` does `pnpm install`.

tiagobento: Thanks for the README, though! Looks great.

packages/dmn-testing-models/env/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+const { varsWithName, composeEnv } = require("@kie-tools-scripts/build-env");
+
+module.exports = composeEnv([require("@kie-tools/root-env/env")], {
+  vars: varsWithName({}),
+  get env() {
+    return {
+      dmnTestingModel: {
+        version: require("../package.json").version,
+      },
```
tiagobento:
```suggestion
      dmnTestingModels: {
        version: require("../package.json").version,
      },
```

packages/dmn-testing-models/index.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/** Empty Entrypoint of this module - No TS/JS code here, it just holds DMN files for testing purposes */
```
tiagobento, line 20: We don't need this file at all, do we?

packages/dmn-testing-models/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+>
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.kie.tools</groupId>
+    <artifactId>dmn-testing-models</artifactId>
+    <version>${revision}</version>
+    <packaging>jar</packaging>
+
+    <name>DMN Testing Models</name>
+    <description>Testing models for DMN marshaller</description>
```
tiagobento, line 40: Can you please make this pom.xml file declare `maven-base` as parent? Please see https://github.com/apache/incubator-kie-tools/pull/2195#issuecomment-2061743588

packages/dmn-marshaller/tests/xsdValidation.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import * as fs from "fs";
+import * as path from "path";
+import { getMarshaller } from "@kie-tools/dmn-marshaller";
+import * as validator from "xsd-schema-validator";
+
+/**
+ * This test suite validates the xml produced (parsed and built) by the marshaller against DMN 1.5 XML Schema.
+ * The validation is performed by a 3rd party library (xsd-schema-validator)
+ */
```
tiagobento: Outdated comment?

packages/dmn-marshaller/package.json
```diff
@@ -51,6 +53,7 @@
     "jest-when": "^3.5.0",
     "rimraf": "^3.0.2",
     "ts-jest": "^26.5.6",
-    "typescript": "^4.6.2"
+    "typescript": "^4.6.2",
+    "xsd-schema-validator": "^0.9.0"
```
tiagobento: Not needed anymore, I suppose?

packages/dmn-marshaller-backend-compatibility-tester/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  const scriptPath = dmnBackendCompatibilityScriptPaths.get(script);
+
+  const jbangArgs = [] as string[];
+  jbangArgs.push("-Dkogito-runtime.version=" + buildEnv.env.kogitoRuntime.version);
+  jbangArgs.push(scriptPath!);
+  args?.forEach((arg) => jbangArgs.push(quoteChar + arg + quoteChar));
+
+  jbang.exec("properties@jbangdev", "java.version");
+  jbang.exec(jbangArgs.join(" "));
+}
```
tiagobento: We discussed making the JBang completely abstracted from the caller, no? This would then become a method that accepts parameters using an object parameter, instead of exposing the command itself. Let's make JBang an implementation detail of this package, instead of leaking it to the consumers.

packages/dmn-marshaller-backend-compatibility-tester/package.json
```diff
@@ -0,0 +1,33 @@
+{
```
tiagobento, line 1: I guess this package needs to be private... we don't want to publish this to NPM, do we?

tiagobento: The fact that it depends on `build-env` to execute is something that prevents this from being publishable to NPM. I would say make it private or change the strategy to have `build-env` abstracted from its runtime. Creating a configuration file from it can be an option, for example.

packages/dmn-marshaller-backend-compatibility-tester/README.md
```diff
@@ -0,0 +1,36 @@
+### Module Description
```
tiagobento:
```suggestion
### DMN Marshaller backend compatibility tester
```
?

packages/dmn-marshaller-backend-compatibility-tester/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+The scope of this module is to provide a reliable scripting mechanism that tests DMN Files generated by the `dmn-marshaller` module against the backend counterpart code that lives in the drools repository.
```
tiagobento:
```suggestion
The scope of this module is to provide a reliable scripting mechanism that tests DMN Files generated by the `dmn-marshaller` package against the backend counterpart code that lives in the drools repository.
```

packages/dmn-marshaller-backend-compatibility-tester/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+The entry point of this module is the `index.js` file that exports:
+
+- `executeJBangScript` that requires as parameters: the Script to execute (see below) and its required args.
+- The enumeration of the currently available Script
+  That represents all the required elements for consuming this module externally.
+
+To execute a specific JBang script present in this module you must:
+
+- Import the `"@kie-tools/dmn-marshaller-backend-compatibility-tester": "workspace:*"` dependency;
```
tiagobento:
```suggestion
- Import the `"@kie-tools/dmn-marshaller-backend-compatibility-tester" dependency;
```

No need to expose the inner mechanisms of `pnpm` (`workspace:*`) here.

packages/dmn-marshaller-backend-compatibility-tester/README.md
```diff
@@ -0,0 +1,36 @@
...
+- Import the `DmnBackendCompatibilityScript` Enum and `executeJBangScript` in your JS/TS file;
+- Execute the JBang script: eg. `executeJBangScript(DmnBackendCompatibilityScript.DMN_VALIDATION, args...)`
+- For the precise args list, please check inside the Java Class Script of your interest and take note of all `@Option` annotated fields.
```
tiagobento, line 36: As we discussed, I think we're much better off encapsulating JBang as an implementation detail of this package.

packages/dmn-marshaller/tests/dmnValidation.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  executeJBangScript,
+} from "@kie-tools/dmn-marshaller-backend-compatibility-tester";
+
+/**
+ * This test suite validates the xml produced (parsed and built) by the marshaller relying on KIE DMN Validator
+ * (https://github.com/apache/incubator-kie-drools/tree/main/kie-dmn/kie-dmn-validation).
+ * A JBang script is used to actually call the KIE DMN Validator Java code.
+ */
+
+const dmnTestingModelsPath = require.resolve("@kie-tools/dmn-testing-models");
```
tiagobento: Ok, so I guess that's what you need the `index.js` for. You can change this to 

```ts
path.dirname(require.resolve("@kie-tools/dmn-testing-models/package.json"))
```

so that you don't need the index.js there :)

tiagobento: This will also remove the necessity for `../`, I guess.

packages/dmn-marshaller/tests/dmnValidation.test.ts
```diff
@@ -0,0 +1,156 @@
...
+          "--importedDmnFilesPaths=" + importerGeneratedXMLFilePath
+        );
+      } catch (error) {
+        fail(error.cause);
+      }
+    }
+  );
+}
+
+function parseXMLAndWriteInFile(normalizedFsPathRelativeToTheFile: string): string {
```
tiagobento: We usually do camelCase acronyms.. So `XML` would become `Xml` here.

packages/dmn-marshaller/tests/idempotency.test.ts
```diff
@@ -21,6 +21,8 @@ import * as fs from "fs";
 import * as path from "path";
 import { getMarshaller } from "@kie-tools/dmn-marshaller";
 
+const dmnTestingModels = require.resolve("@kie-tools/dmn-testing-models");
```
tiagobento: Same as above.

tiagobento (changes requested):
This is getting pretty good! Thanks for all the effort @yesamer!

As a general comment, we usually don't do snake_case... Some of the tests I wrote, though, include variables named like `xml_original`. Probably a bad choice from my part, as I was trying to convey the feeling that they should be the same thing, so I didn't want to name them `xmlOriginal` and `xmlFirstPass`, you know.. Anyway, feel free to remove the snake_case altogether or use them only where you find necessary.

Thanks!

packages/dmn-testing-models/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "main": "dist/index.js",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build:dev": "rimraf dist && copyfiles index.js dist/ && run-script-os",
+    "build:dev:linux:darwin": "mvn clean generate-resources",
+    "build:dev:win32": "pnpm powershell \"mvn clean generate-resources\"",
+    "build:prod": "pnpm build:dev",
```
tiagobento: This package, albeit not a proper Maven module producing a JAR and everything, is still a Maven package. If any other Maven package decides to depend on it as a Maven dependency, not installing it will make the build fail. You can simply change this to `mvn clean install` (or `mvn clean verify`!) to solve this problem.

packages/dmn-marshaller/tests/idempotency.test.ts
yesamer: @tiagobento Isn't this already in camelCase?

packages/dmn-marshaller/tests/dmnValidation.test.ts
yesamer: Done!

packages/dmn-testing-models/README.md
yesamer: Done!

packages/dmn-testing-models/env/index.js
yesamer: Done!

packages/dmn-marshaller/tests/xsdValidation.test.ts
yesamer: I removed the test suite at all, as it's a duplication of dmnValidation.test

packages/dmn-marshaller/package.json
yesamer: Based on the previous comment, yes. removed.

packages/dmn-testing-models/index.js
yesamer: After I applied the suggestion of comment https://github.com/apache/incubator-kie-tools/pull/2207#discussion_r1569493358, yes. Removed.

packages/dmn-testing-models/pom.xml
yesamer: Done!

packages/dmn-marshaller-backend-compatibility-tester/package.json
yesamer: @tiagobento Thank you, set it as private.

packages/dmn-marshaller/tests/dmnValidation.test.ts
yesamer: With `path.dirname(require.resolve("@kie-tools/dmn-testing-models/package.json"))` we need to replace `../` with `dist/`
Done!

packages/dmn-testing-models/package.json
yesamer: Done!

packages/dmn-marshaller-backend-compatibility-tester/src/index.ts
yesamer: @tiagobento I've encapsulated all the details related to JBang, and implemented a new API to be externally consumed.

yesamer: (Still WIP)

yesamer: Done!

packages/dmn-marshaller/tests/idempotency.test.ts
tiagobento: I was talking about the require.resolve() with the package.json here :P

packages/dmn-marshaller-backend-compatibility-tester/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "dist",
+    "src"
+  ],
+  "scripts": {
+    "build:dev": "rimraf dist && tsc -p tsconfig.json && pnpm prefetch",
+    "build:prod": "pnpm build:dev",
+    "prefetch": "node dist/dependenciesFetch.js"
+  },
+  "devDependencies": {
+    "@jbangdev/jbang": "0.2.0",
```
tiagobento: Based on a previous conversation we had, I think this needs to be in the `dependencies` section, since it is required to run.

packages/dmn-marshaller-backend-compatibility-tester/src/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  executeScript(parentScriptPath);
+}
+
+export function checkDMNValidation(dmnFilePath: string) {
+  const command = "--command=no_imports";
+  const dmnFilePathCommand = "--dmnFilePath=" + dmnFilePath;
+  const importedDmnFilesPaths = "--importedDmnFilesPaths=";
+
+  executeScript(dmnValidationScriptPath, [command, dmnFilePathCommand]);
+}
```
tiagobento:
```suggestion
export function checkDmnValidation(data: { dmnFilePath: string }) {
  const command = "--command=no_imports";
  const dmnFilePathCommand = "--dmnFilePath=" + data.dmnFilePath;
  const importedDmnFilesPaths = "--importedDmnFilesPaths=";

  executeScript(dmnValidationScriptPath, [command, dmnFilePathCommand]);
}
```

?

packages/dmn-testing-models/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Models are separated between < 1.5 version and 1.5 version; such classification is based on actual version-specific features, and not on the referenced tag itself.
+
+For future DMN versions there will be version specific folders.
+
+The original `org.kie:kie-dmn-test-resources` also contains _invalid_ models, but for the moment being we use only the valid ones to verify round-trip validation.
+
+### Usage
+
+The command `mvn clean generate-resources` downloads the jar and extract the models under `dist/`.
```
tiagobento: Outdated.

packages/dmn-testing-models/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+To make them available for testing purpose:
+
+1. set the `KOGITO_RUNTIME_version` in the terminal (if different from the default one)
+2. `pnpm bootstrap` on the root directory
+3. cd dmn-testing-models
+4. `pnpm build:dev` or `pnpm build:prod` (they behave the same way - this executes `mvn clean generate-resources`)
```
tiagobento: Outdated.

packages/dmn-testing-models/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+  xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd"
+>
+
+  <parent>
+    <groupId>org.kie</groupId>
+    <artifactId>kie-tools-maven-base</artifactId>
+    <version>${revision}</version>
+    <relativePath>./node_modules/@kie-tools/maven-base/pom.xml</relativePath>
```
tiagobento, line 30: 👍

tiagobento (changes requested):
Thx Yeser! Please double-check the READMEs contents, and there are still some acronyms in all caps. Other than that, just a few comments inline. Looking great.

packages/dmn-marshaller-backend-compatibility-tester/package.json
yesamer: Done

packages/dmn-marshaller-backend-compatibility-tester/src/index.ts
yesamer: Done

packages/dmn-testing-models/README.md
yesamer: Done

packages/dmn-testing-models/README.md
yesamer: Done

tiagobento (approved):
Thank you @yesamer and @gitgabrio !

packages/dmn-marshaller/tests/dmnSemanticComparison.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  "dist/valid_models/DMNv1_x/allTypes.dmn",
+  "dist/valid_models/DMNv1_x/dtevent.dmn",
+  "dist/valid_models/DMNv1_x/habitability.dmn",
+  "dist/valid_models/DMNv1_x/loan.dmn",
+  "dist/valid_models/DMNv1_x/LoanEligibility.dmn",
+  "dist/valid_models/DMNv1_x/OneOfEachType.dmn",
+  "dist/valid_models/DMNv1_x/Prequalification.dmn",
+  "dist/valid_models/DMNv1_x/testWithExtensionElements.dmn",
+  "dist/valid_models/DMNv1_x/Traffic Violation Simple.dmn",
+  "dist/valid_models/DMNv1_x/Traffic Violation.dmn",
```
jomarko, line 57: is it intentional we do not have all content from https://github.com/apache/incubator-kie-drools/tree/main/kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models ?

packages/dmn-marshaller/tests/dmnValidation.test.ts
```diff
@@ -0,0 +1,150 @@
...
+const dmnTestingModels = [
+  "dist/valid_models/DMNv1_5/AllowedValuesChecksInsideCollection.dmn",
+  "dist/valid_models/DMNv1_5/DateToDateTimeFunction.dmn",
+  "dist/valid_models/DMNv1_5/ForLoopDatesEvaluate.dmn",
+  "dist/valid_models/DMNv1_5/Imported_Model_Unamed.dmn",
+  "dist/valid_models/DMNv1_5/ListReplaceEvaluate.dmn",
+  "dist/valid_models/DMNv1_5/NegationOfDurationEvaluate.dmn",
+  "dist/valid_models/DMNv1_5/TypeConstraintsChecks.dmn",
+  "dist/valid_models/DMNv1_x/multiple/Financial.dmn",
+  "dist/valid_models/DMNv1_x/multiple/Imported_Traffic_Violation.dmn",
```
jomarko, line 46: is it intentional we introduce the same constants as in the file above?

packages/dmn-marshaller-backend-compatibility-tester/src/DmnSemanticComparison.java
```diff
@@ -0,0 +1,259 @@
...
+import org.kie.internal.io.ResourceFactory;
+
+/**
+ * JBang script that performs DMN files' XML (in string format) validation
+ * relying on KIE DMN Validator
+ * (https://github.com/apache/incubator-kie-drools/tree/main/kie-dmn/kie-dmn-validation).
+ * The script can manage one or two (in case of imported model) DMN file paths.
+ * The XSD SCHEMA, DMN COMPLIANCE and DMN COMPILATION are validated.
+ */
+class DmnSemanticComparison extends DmnMarshallerBackendCompatibilityTesterScript {
```
jomarko, line 60: My comment to this jbang class is we mix a terminology. In class `Option` we speak about `generated` files while in the method implementation and logs we speak about `parsed` files, I think we could unify the naming. What is your opinion?

jomarko (commented):
Thank you for the PR!

packages/dmn-marshaller/tests/dmnSemanticComparison.test.ts
yesamer: @jomarko This is a good question. 
The reason for this behavior is intentional and it's because we need to discriminate between DMN models that import a DMN model and DMN models that don't.
In the first case, if want to validate / or check for semantic comparison, you need to pass both models (the importer and the imported) together to the JBang script.
So, the first array holds all the DMN files that don't have any imported DMN files.
The second array holds the DMN files that import another DMN file (both are specified).

packages/dmn-marshaller/tests/dmnValidation.test.ts
yesamer: Yes, even if dmnValidation and dmnSemanticComparison are currently declaring the same arrays, I would prefer to keep them well separated.

packages/dmn-marshaller-backend-compatibility-tester/src/DmnSemanticComparison.java
yesamer: I agree with you. Done.

jomarko (approved):
Thank you @yesamer

PR #2209, opened 2024-03-26
kie-issues#1053: restore of e2e tests in kn workflow cli

### LINKED ISSUE #1053: restore of e2e tests in kn workflow cli
E2E tests were disabled because some quarkus dependencies are located in the apache snapshot repository.
The generated test application (pom.xml) must specify the correct repository to re-enable the tests.

treblereel:
Closes: apache/incubator-kie-issues#1053

### DIFF
--- a/packages/kn-plugin-workflow/e2e-tests/helper_test.go
+++ b/packages/kn-plugin-workflow/e2e-tests/helper_test.go
@@ -39,7 +39,7 @@ import (
 
 // ExecuteCommand executes a command with the given arguments and returns an error if the command fails.
 func ExecuteCommand(command string, args ...string) error {
-	cmd := exec.Command(command, args...)
+	cmd := exec.Command("MAVEN_ARGS=\"-B\""+command, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	return cmd.Run()
--- a/packages/kn-plugin-workflow/e2e-tests/helper_test.go
+++ b/packages/kn-plugin-workflow/e2e-tests/helper_test.go
@@ -31,34 +31,11 @@ import (
 	"strings"
 	"syscall"
 	"testing"
-	"time"
 
 	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/command/quarkus"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/require"
 )
-
-// let's pre-fetch kogito-quarkus-serverless-workflow-devui-deployment before running the tests
-func init() {
-    started := time.Now()
-    fmt.Println("🚀 Starting dependency pre-fetch...")
-    defer func(started time.Time) {
-        takes := time.Since(started)
-        fmt.Printf("✅ Dependency fetch completed in %.2f seconds.\n", takes.Seconds())
-    }(started)
-
-    args := []string{
-        "dependency:get",
-        "-Dartifact=org.kie.kogito:kogito-quarkus-serverless-workflow-devui-deployment:999-20240218-SNAPSHOT",
-        "-DremoteRepositories=https://repository.apache.org/content/repositories/snapshots",
-    }
-
-	cmd := exec.Command("mvn", args...)
-    output, err := cmd.CombinedOutput()
-    if err != nil {
-        fmt.Printf("❌ Failed to execute command: %v\nOutput:\n%s", err, string(output))
-    }
-}
 
 // ExecuteCommand executes a command with the given arguments and returns an error if the command fails.
 func ExecuteCommand(command string, args ...string) error {
--- a/packages/kn-plugin-workflow/e2e-tests/helper_test.go
+++ b/packages/kn-plugin-workflow/e2e-tests/helper_test.go
@@ -39,7 +39,7 @@ import (
 
 // ExecuteCommand executes a command with the given arguments and returns an error if the command fails.
 func ExecuteCommand(command string, args ...string) error {
-	cmd := exec.Command("MAVEN_ARGS=\"-B\""+command, args...)
+	cmd := exec.Command(command, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	return cmd.Run()
@@ -179,7 +179,6 @@ func AddSnapshotRepositoryDeclarationToPom(t *testing.T, projectDir string) {
 
 	file, err := os.Open(pomFilePath)
 	require.NoErrorf(t, err, "Expected nil error, got: %v", err)
-	defer file.Close()
 
 	content, err := io.ReadAll(file)
 	require.NoErrorf(t, err, "Expected nil error, got: %v", err)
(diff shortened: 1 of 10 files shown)


tiagobento:
@treblereel I think listing the Apache repo is making the tests considerably slower. Please do list Maven Central as the first repo, and Apache snapshots in second. I guess this will speed things up a little bit. Also, might be worth running Maven with the `-B` flag on these tests not to pollute the build logs with the artifact download progress.

treblereel:
Looks like the apache repo is very slow. Here is my download speed:

```
Saving to: ‘kogito-quarkus-serverless-workflow-devui-deployment-999-20240218-20240218.073833-1.jar’

kogito-quarkus-serverless-workflow-devui-depl 100%[=======================================================================>]  70.37M   401KB/s    in 2m 43s  
```
I don't like it, but it seems we should increase the timeout

tiagobento:
@treblereel What about pre-fetching the dependencies during the `install` script? We can create a pom.xml that contains the dependencies needed for testing and run a `mvn dependency:tree` to fetch everything.

treblereel:
@tiagobento yeap, think the same way, maybe we can fetch jar in parallel (or make maven do that). I'm investigation the possible options

treblereel:
atm `999-20240218-SNAPSHOT` is hardcoded, but it must be received from  index.js as env var. As i can see, this part in plugin was started but not finished, I'm going to fix it as separate commit.

treblereel:
retest this please

treblereel:
Since the kogito dependencies are not available for version `999-20240331-SNAPSHOT`, I switched to Apache.

For example:
org.kie.kogito:kogito-quarkus-serverless-workflow-devui -> org.apache.kie.sonataflow:sonataflow-quarkus-devui

I was planning to do this as a separate task, but this way it's even more convenient.

yesamer (approved):

packages/kn-plugin-workflow/pkg/command/quarkus/build.go
```diff
@@ -232,7 +232,7 @@ func runBuildImage(cfg BuildCmdConfig) (out string, err error) {
 	builderConfig := getBuilderConfig(cfg)
 	executableName := getExecutableNameConfig(cfg)
 
-	build := common.ExecCommand("mvn", "package",
+	build := common.ExecCommand("mvn", "-B","package",
```
tomasdavidorg:
```suggestion
	build := common.ExecCommand("mvn", "-B", "package",
```

packages/kn-plugin-workflow/e2e-tests/helper_test.go
```diff
@@ -39,7 +39,7 @@ import (
 
 // ExecuteCommand executes a command with the given arguments and returns an error if the command fails.
 func ExecuteCommand(command string, args ...string) error {
-	cmd := exec.Command(command, args...)
+	cmd := exec.Command("MAVEN_ARGS=\"-B\""+command, args...)
```
tomasdavidorg: For `mvn` as `command` and `clean`, `install` as `args...`, this will create command:
 ```
MAVEN_ARGS="-B"mvn clean install
```
so I think this should not work.

Also the `ExecuteCommand` function is not used for maven execution, see https://github.com/apache/incubator-kie-tools/blob/main/packages/kn-plugin-workflow/e2e-tests/main_test.go#L180

File: packages/kn-plugin-workflow/e2e-tests/helper_test.go [hunk 3]
```diff
@@ -172,3 +172,46 @@ func CleanUpAndChdirTemp(t *testing.T) {
...
 	}
 }
+
+func AddSnapshotRepositoryDeclarationToPom(t *testing.T, projectDir string) {
+	VerifyFilesExist(t, projectDir, []string{"pom.xml"})
+	pomFilePath := filepath.Join(projectDir, "pom.xml")
+
+	file, err := os.Open(pomFilePath)
+	require.NoErrorf(t, err, "Expected nil error, got: %v", err)
+	defer file.Close()
```
tomasdavidorg: This defers the calling of the `file.Close()` at the end of this function but we rewrite the `pom.xml` file before the end. I think this might cause some issues.

tomasdavidorg (changes requested):

packages/kn-plugin-workflow/e2e-tests/helper_test.go
```diff
@@ -172,3 +172,45 @@ func CleanUpAndChdirTemp(t *testing.T) {
...
+func AddSnapshotRepositoryDeclarationToPom(t *testing.T, projectDir string) {
+	VerifyFilesExist(t, projectDir, []string{"pom.xml"})
+	pomFilePath := filepath.Join(projectDir, "pom.xml")
+
+	file, err := os.Open(pomFilePath)
+	require.NoErrorf(t, err, "Expected nil error, got: %v", err)
+
+	content, err := io.ReadAll(file)
+	require.NoErrorf(t, err, "Expected nil error, got: %v", err)
+
```
tomasdavidorg, line 188:
```suggestion

	err = file.Close()
	require.NoErrorf(t, err, "Expected nil error, got: %v", err)
	
```

tomasdavidorg (commented):
But I think we still need to close the file after the reading is done and before `err = os.WriteFile(pomFilePath, []byte(modifiedXml), 0644)` starts.

Sorry I know only some basics Go, so I am thinking out loud. 😅

packages/kn-plugin-workflow/e2e-tests/helper_test.go
treblereel: @tomasdavidorg updated

tomasdavidorg (approved):
@treblereel Thanks 🙂

tiagobento:
So we have #2237, but this PR passed, even with the typo. Are we not using the QUARKUS_VERSION env var?

tiagobento:
Ok, so I think what happens is:
1. The typo was only on Windows scripts.
2. The build succeedded, but running the binaries on Windows would be broken, as QuarkusVersion would be `nil`.
3. Since we don't run E2E tests on Windows, we this PR didn't catch it!

Other than enabling the E2E tests on Windows, I'm not sure how we could've caught this. I found it by chance in logs.. Anyway, all good for now, despite the gap in correctness we have...

tiagobento (approved):

treblereel:
> Ok, so I think what happens is:
> 
> 1. The typo was only on Windows scripts.
> 2. The build succeedded, but running the binaries on Windows would be broken, as QuarkusVersion would be `nil`.
> 3. Since we don't run E2E tests on Windows, we this PR didn't catch it!
> 
> Other than enabling the E2E tests on Windows, I'm not sure how we could've caught this. I found it by chance in logs.. Anyway, all good for now, despite the gap in correctness we have...

Do we have an issue for this bug? If yes, I can take care of it

tiagobento:
@treblereel I'm afraid this is much more complicated, as we skip E2E tests for all packages on Windows. We could have `kn-plugin-workflow` ignore the flag that makes this skipping, but then it would hurt the consistency of not running E2E tests on Windows altogether... What are your thoughts?

PR #2225, opened 2024-04-04
kie-issues#389: Include environment variables on all KIE Sandbox-related images to configure the port used

### LINKED ISSUE #389: Include environment variables on all KIE Sandbox-related images to configure the port used
Right now all our images (`kie-sandbox-image`, `git-cors-proxy-image`, and `kie-sandbox-extended-services`) run on specific ports and can't be configured at build or runtime.

Adding support to environment variables will enable the services running on the container to use different ports as well as expose different ports on the container, this is important for scenarios where there are port conflicts.

The running applications AND the Containerfile should support the same environment variable, for easy configuration.

Josephblt:
Closes: https://github.com/apache/incubator-kie-issues/issues/389

Added variables for the configuration of images:

- Kie-Sandbox-Image:
  - KIE_SANDBOX_PORT
- Cors-Proxy-Image:
  - CORS_PROXY_HTTP_PORT
  - CORS_PROXY_ORIGIN
  - CORS_PROXY_VERBOSE


packages/cors-proxy-image/Containerfile
```diff
@@ -17,12 +17,17 @@
 
 FROM --platform=linux/amd64 registry.access.redhat.com/ubi9/ubi-minimal:9.3
 
+ARG CORS_PROXY_DEFAULT_HTTP_PORT
+ARG CORS_PROXY_DEFAULT_ORIGIN
+ARG CORS_PROXY_DEFAULT_VERBOSE
```
thiagoelg: @Josephblt What do you think about setting a default value for these ARGs? I know we pass them via the `build:*` scripts in our package.json, but setting a default value means we can just run `docker build` and it will work just fine.

thiagoelg:
Could these changes affect the `kie-sandbox-distribution` and `kie-sandbox-helm-chart` packages?

packages/kie-sandbox-image/package.json
```diff
@@ -24,8 +24,8 @@
...
-    "image:docker:build": "kie-tools--image-builder build -r \"$(build-env kieSandbox.image.registry)\" -a \"$(build-env kieSandbox.image.account)\" -n \"$(build-env kieSandbox.image.name)\" -t \"$(build-env kieSandbox.image.buildTags)\"",
-    "image:podman:build": "kie-tools--image-builder build -r \"$(build-env kieSandbox.image.registry)\" -a \"$(build-env kieSandbox.image.account)\" -n \"$(build-env kieSandbox.image.name)\" -t \"$(build-env kieSandbox.image.buildTags)\" -e podman"
+    "image:docker:build": "kie-tools--image-builder build -r \"$(build-env kieSandbox.image.registry)\" -a \"$(build-env kieSandbox.image.account)\" -n \"$(build-env kieSandbox.image.name)\" -t \"$(build-env kieSandbox.image.buildTags)\" --build-arg \"KIE_SANDBOX_DEFAULT_PORT=$(build-env onlineEditor.dev.port)\"",
+    "image:podman:build": "kie-tools--image-builder build -r \"$(build-env kieSandbox.image.registry)\" -a \"$(build-env kieSandbox.image.account)\" -n \"$(build-env kieSandbox.image.name)\" -t \"$(build-env kieSandbox.image.buildTags)\" --build-arg \"KIE_SANDBOX_DEFAULT_PORT=$(build-env onlineEditor.dev.port)\" -e podman"
```
tiagobento: This is not the same, as we're going from 8080 to 9001...

packages/cors-proxy-image/Containerfile
```diff
@@ -17,12 +17,17 @@
 
 FROM --platform=linux/amd64 registry.access.redhat.com/ubi9/ubi-minimal:9.3
 
+ARG CORS_PROXY_DEFAULT_HTTP_PORT
```
tiagobento: Can we use `CORS_PROXY_DEFAULT_PORT` instead of `HTTP_PORT`? I mean, I prefer `HTTP_PORT`, but all other places have these set as `PORT` only, so let's keep it consistent...

packages/kie-sandbox-image/package.json
tiagobento: I'd keep the default internal port as 8080...

tiagobento (changes requested):

packages/kie-sandbox-image/entrypoint.sh
```diff
@@ -2,8 +2,11 @@
 
 # Copying the KIE Sandbox assets here is essential for when the container is running with the readOnlyRootFilesystem flag.
 # But, just like any other directory modified during runtime, the /var/www/html must be a mounted volume in the container in this case.
```
thiagoelg: This comment is related to the `cp -R` command below, we should keep them together

thiagoelg:
Hey @Josephblt, I gave this a better thought (and testing) and I don't think using Environment Variables for the port makes sense (I know... I was the one that wrote that in the original issue, my bad).

It doesn't make sense because the `EXPOSE` instruction is not executed during runtime, so the exposed port will always be the one defined in the ARG.
So let's say we set the ARG port as 8080 during build time, but we set the ENV as 9022 during runtime. The exposed port will still be 8080, but the app will be running at 9022.
Now, according to the Docker documentation, the [EXPOSE](https://docs.docker.com/reference/dockerfile/#expose) instruction is only a way to document the port that is expected to be available, but in practice, you can still map any container port to any host port. The problem is that Kubernetes (and OpenShift) usually expects the exposed port to be the one where the app is available, so by default, all Services are created pointing to these ports.

Having said all that, having the ARGs for defining the ports is enough, and we don't need to change them during runtime.

What do you think?

PS1.: When using ARGs we can revert those `sed` commands back to the Containerfile.

PS2.: I discovered that OpenShift creates a bunch of environment variables by itself on its containers, potentially overwriting the ENVs we defined here, take a look:
```
KUBERNETES_SERVICE_PORT_HTTPS=443
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_SERVICE_PORT_8080_TCP=8080
WORKSPACEB8CB941D988C49CF_SERVICE_SERVICE_HOST=172.30.106.115
KUBERNETES_SERVICE_PORT=443
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_PORT_8080_TCP_PROTO=tcp
HOSTNAME=runtime-tools-task-console-webapp-image-6b5df5fcbc-bn2jw
WORKSPACEB8CB941D988C49CF_SERVICE_PORT_4444_TCP=tcp://172.30.106.115:4444
NSS_SDB_USE_CACHE=no
WORKSPACEB8CB941D988C49CF_SERVICE_PORT_4444_TCP_PORT=4444
PWD=/
WORKSPACEB8CB941D988C49CF_SERVICE_SERVICE_PORT_EXEC=4444
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_SERVICE_PORT=8080
container=oci
HOME=/
KUBERNETES_PORT_443_TCP=tcp://172.30.0.1:443
WORKSPACEB8CB941D988C49CF_SERVICE_PORT=tcp://172.30.106.115:4444
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_PORT_8080_TCP_ADDR=172.30.151.27
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_PORT_BLA=8080
TERM=xterm
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_SERVICE_HOST=172.30.151.27
WORKSPACEB8CB941D988C49CF_SERVICE_SERVICE_PORT=4444
SHLVL=1
KUBERNETES_PORT_443_TCP_PROTO=tcp
KUBERNETES_PORT_443_TCP_ADDR=172.30.0.1
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_PORT_8080_TCP_PORT=8080
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_PORT_8080_TCP=tcp://172.30.151.27:8080
KUBERNETES_SERVICE_HOST=172.30.0.1
KUBERNETES_PORT=tcp://172.30.0.1:443
KUBERNETES_PORT_443_TCP_PORT=443
WORKSPACEB8CB941D988C49CF_SERVICE_PORT_4444_TCP_ADDR=172.30.106.115
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_PORT=tcp://172.30.151.27:8080
WORKSPACEB8CB941D988C49CF_SERVICE_PORT_4444_TCP_PROTO=tcp
_=/usr/bin/printenv
```
Out of all of these, only `RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_PORT_BLA` was defined by me, as `RUNTIME_TOOLS_TASK_CONSOLE_WEBAPP_IMAGE_PORT` was being overwritten by OpenShift.

thiagoelg:
Here's an example of what that would look like: https://github.com/apache/incubator-kie-tools/pull/2226/commits/f5a525d5295ec86f67924c424e05fae01a9c0e23

packages/cors-proxy-image/Containerfile
Josephblt: I agree. I'll set the default values according to the index.js files.

packages/cors-proxy-image/Containerfile
Josephblt: Agreed. Changing it to CORS_PROXY_DEFAULT_PORT.

packages/kie-sandbox-image/entrypoint.sh
Josephblt: Most definitely, that one slipped by, sorry.

packages/kie-sandbox-image/Containerfile
```diff
@@ -17,17 +17,19 @@
...
   && microdnf --disableplugin=subscription-manager clean all \
   && echo "Mutex posixsem" >> /etc/httpd/conf/httpd.conf \
-  && sed -i -e 's/Listen 80/Listen 8080/' /etc/httpd/conf/httpd.conf \
-  && sed -i -e 's/#ServerName www.example.com:80/ServerName 127.0.0.1:8080/' /etc/httpd/conf/httpd.conf \
-  && sed -i -e "/ServerName 127.0.0.1:8080/aHeader set Content-Security-Policy \"frame-ancestors 'self';\"" /etc/httpd/conf/httpd.conf \
-  && sed -i -e 's/Options Indexes FollowSymLinks/Options -Indexes +FollowSymLinks/' /etc/httpd/conf/httpd.conf \
-  && sed -i '$ a ServerTokens Prod' /etc/httpd/conf/httpd.conf \
-  && sed -i '$ a ServerSignature Off' /etc/httpd/conf/httpd.conf \
+  && sed -i "s/Listen 80/Listen $KIE_SANDBOX_DEFAULT_PORT/g" /etc/httpd/conf/httpd.conf \
+  && sed -i "s/#ServerName www.example.com:80/ServerName www.example.com:$KIE_SANDBOX_DEFAULT_PORT/g" /etc/httpd/conf/httpd.conf \
```
thiagoelg:
```suggestion
  && sed -i "s/#ServerName www.example.com:80/ServerName 127.0.0.1:$KIE_SANDBOX_DEFAULT_PORT/g" /etc/httpd/conf/httpd.conf \
```

Without this the next `sed` command won't find a match.

packages/kie-sandbox-extended-services-image/Containerfile
```diff
@@ -20,18 +20,15 @@ ARG BUILDER_IMAGE_ARG
 
 FROM --platform=linux/amd64 ${BUILDER_IMAGE_ARG}
 
-ARG EXTENDED_SERVICES_DEFAULT_HOST
-ARG EXTENDED_SERVICES_DEFAULT_PORT
-
-ENV EXTENDED_SERVICES_HOST=$EXTENDED_SERVICES_DEFAULT_HOST
-ENV EXTENDED_SERVICES_PORT=$EXTENDED_SERVICES_DEFAULT_PORT
+ARG EXTENDED_SERVICES_DEFAULT_HOST = 0.0.0.0
+ARG EXTENDED_SERVICES_DEFAULT_PORT = 21345
```
thiagoelg:
```suggestion
ARG EXTENDED_SERVICES_DEFAULT_HOST=0.0.0.0
ARG EXTENDED_SERVICES_DEFAULT_PORT=21345
```

I don't think there should be spaces here

Josephblt:
Hello @tiagobento, @thiagoelg, I think we are set here. Thanks.

packages/kie-sandbox-distribution/env/index.js
```diff
@@ -67,11 +67,11 @@ module.exports = composeEnv([buildEnv, extendedServicesImageEnv, corsProxyImageE
       description: "",
     },
     KIE_SANDBOX_DISTRIBUTION__extendedServicesContainerPort: {
-      default: "21345",
+      default: extendedServicesImageEnv.env.extendedServicesImage.port,
       description: "",
     },
     KIE_SANDBOX_DISTRIBUTION__extendedServicesExposedPort: {
-      default: "21345",
+      default: extendedServicesImageEnv.env.extendedServicesImage.port,
```
thiagoelg: I think this should be fixed to `21345` or whatever port it was before since it is the host port.

(Maybe calling these variables `...ExposedPort` was a mistake on my part, they should be called `...HostPort` to avoid confusion, but we can leave that for later)

thiagoelg (approved):

Josephblt:
Hello @tiagobento. I think this is good to merge. Thanks.

tiagobento (approved):

PR #2229, opened 2024-04-07
NO-ISSUE: Make sure `kie-tools` is using the latest timestamped SNAPSHOT from Kogito (999-20240411-SNAPSHOT)

yesamer:
Update all Kogito references to `999-20240411-SNAPSHOT`


packages/stunner-editors/pom.xml
```diff
@@ -326,7 +326,7 @@
     <version.zanata.plugin>2.3.0</version.zanata.plugin>
 
     <!-- KIE Libraries -->
-    <version.org.kie>999-20240331-SNAPSHOT</version.org.kie>
+    <version.org.kie>999-20240407-SNAPSHOT</version.org.kie>
```
tiagobento: Can we start using the version from `build-env` here? Adding an `install.js` script etc.

packages/kn-plugin-workflow/env/index.js
```diff
@@ -39,12 +39,12 @@ module.exports = composeEnv([require("@kie-tools/root-env/env")], {
...
-      default: "quay.io/kiegroup/kogito-swf-devmode-nightly:main-2024-04-03",
+      default: "quay.io/kiegroup/kogito-swf-devmode-nightly:main-2024-04-05",
       description: "SonataFlow dev mode image (used on cli run)",
     },
     KN_PLUGIN_WORKFLOW__kogitoVersion: {
       name: "KN_PLUGIN_WORKFLOW__kogitoVersion",
-      default: "999-20240331-SNAPSHOT",
+      default: "999-20240407-SNAPSHOT",
       description: "Kogito version to be used when creating and converting to Quarkus Projects",
     },
```
tiagobento: I think we can remove this env and start using `kogitoRuntime.version` from root-env.. WDYT? cc @paulovmr @caponetto

tiagobento (changes requested):
I noticed we have different dates for the images... Could we unify?

yesamer:
@tiagobento At this time, the latest version of the image https://quay.io/repository/kiegroup/kogito-base-builder-nightly?tab=tags&tag=latest is 2024-04-05

packages/stunner-editors/pom.xml
yesamer: I guess yes, good idea.

tiagobento:
@ricardozanini Any idea what could've happened with the nightly build of this image?

packages/stunner-editors/pom.xml
yesamer: Done :)

packages/kn-plugin-workflow/env/index.js
paulovmr: Sure, makes sense.

ricardozanini:
@tiagobento I'll take a look on Jenkin

ricardozanini:
@yesamer @tiagobento this image has been removed as part of the old operator archiving: https://github.com/apache/incubator-kie-kogito-images/pull/1748

ricardozanini:
If you want, I think I can help review these PRs related to images. Let me know!

yesamer:
@tiagobento @thiagoelg That means we need a replacement for that... Should I open a ticket?

ricardozanini:
@yesamer why do you need a replacement? That image won't be used anywhere as far as I know.

tiagobento:
@ricardozanini It is used by Serverless Logic Web Tools, it seems like... cc @caponetto

paulovmr:
@ricardozanini @tiagobento yes this image was the base for the web tools builder and dev-mode images. @ricardozanini could it be restored?

ricardozanini:
@paulovmr yes, we can. I can't do it today, if some of you have some room, here's the PR where I removed it: https://github.com/apache/incubator-kie-kogito-images/pull/1748/files

ricardozanini:
@paulovmr @caponetto honestly, I saw the image now and I'm not seeing anything that it's not on openjdk-17 images. Why not use it directly?

ricardozanini:
Here: https://github.com/apache/incubator-kie-kogito-images/pull/1754

rodrigonull:
@yesamer I'm running the weekly jobs to generate new timestamped snapshots. We should have the 999-20240411-SNAPSHOTs available by tomorrow morning.

yesamer:
@rodrigonull Great!!!! Are you launching the job for all our repos? jit-executor as well? Thank you!

rodrigonull:
@yesamer yes, my plan is to trigger all jobs including the JIT executor.

tiagobento:
@yesamer For images, please change the tags to 999-YYYYMMDD instead of main-*, as the first comes from the weekly jobs (dependencies are also timestamped SNAPSHOTs) and the second is from daily jobs (dependencies are 999-SNAPSHOT)

yesamer:
@tiagobento Done!

tiagobento (approved):

gitgabrio (commented):
Hi @yesamer 
Out of curiosity: would not be possible to store that variable in one single place, and maybe refer to it wherever is needed ?

tiagobento:
@gitgabrio That's the idea behind https://github.com/apache/incubator-kie-tools/blob/main/packages/root-env/env/index.js#L53-L60, but it's not always simple to do it, especially when the format changes slightly :)

yesamer:
@gitgabrio I guess @tiagobento can give you a better answer than me, anyway my feeling is that is it possible to remove some of those places, and the tooling team is already committed to that. In this PR I removed one place :)

yesamer:
@tiagobento Can I merge it? :)

tiagobento:
@yesamer There's only one approve :( We need 2.

In the mean time...... could we remove `KN_PLUGIN_WORKFLOW__quarkusVersion` and `KN_PLUGIN_WORKFLOW__kogitoVersion`? We can use the one from `root-env` instead of having two variables dedicated to the `kn-plugin-workflow`.. and we @paulovmr's blessing :)

yesamer:
@tiagobento Right.

So should I just remove KN_PLUGIN_WORKFLOW__kogitoVersion and replace it with `kogitoRuntime.version`, correct?

gitgabrio (approved):

tiagobento:
@yesamer Yes!

klbowers (approved):

tiagobento:
@yesamer Looks like this one can be removed too! `SERVERLESS_LOGIC_WEB_TOOLS__kogitoVersion` cc @paulovmr

tiagobento:
@yesamer And this one too! `SERVERLESS_LOGIC_WEB_TOOLS__quarkusPlatformVersion`.

This is the last comment I promise 😅

tiagobento:
I bet @gitgabrio is thrilled with us removing env vars, though :)

yesamer:
BINGO!

yesamer:
Done! @tiagobento @gitgabrio @kbowers-ibm

PR #2230, opened 2024-04-08
kie-issues#883: On the DMN Editor, allow deleting DRDs

### LINKED ISSUE #883: On the DMN Editor, allow deleting DRDs

tiagobento:
Closes https://github.com/apache/incubator-kie-issues/issues/883

image

---

Also on this PR:
- Little improvements to the Documentation links component. (https instead of http, and padding)
- More consistent strokeWidth on the alternative InputData node icon
- Allow changing InputData node shape even on the default DRD.
- Added EmptyState to the DRG nodes panel
- Added EmptyState to the DRD selector panel

---

image

image

image


thiagoelg:
About the CI failing:

My guess is that `stream-browserify` is not a direct dependency of `dmn-vscode-extension`, BUT it's an indirect dependency of `dev-deployment-dmn-form-webapp`. Since this is a partial build, the partial bootstrap is not installing `dev-deployment-dmn-form-webapp` dependencies, therefore not installing `stream-browserify`.

The fix would be to include `"stream-browserify": "ˆ3.0.0"` as a `dmn-vscode-extension` devDependency.

tiagobento:
@thiagoelg Yeah... it makes sense that it is not working for the partial build.... but it doesn't make sense that it does work for the full build....

packages/dmn-editor/src/diagram/DrdSelectorPanel.tsx
```diff
@@ -30,19 +30,32 @@ import { Title } from "@patternfly/react-core/dist/js/components/Title";
...
 
   const dmnEditorStoreApi = useDmnEditorStoreApi();
 
+  const drds = thisDmn.model.definitions["dmndi:DMNDI"]?.["dmndi:DMNDiagram"] ?? [];
+
+  const removeDrd = useCallback(() => {
+    dmnEditorStoreApi.setState((s) => {
+      const nextDrds = s.dmn.model.definitions["dmndi:DMNDI"]?.["dmndi:DMNDiagram"];
+      nextDrds?.splice(s.diagram.drdIndex, 1);
+      s.diagram.drdIndex = Math.max(0, Math.min(s.diagram.drdIndex, (nextDrds?.length ?? 0) - 1));
```
ljmotta, line 55: In the case we have 3 DRDs (DRD1, DRD2, DRD3), deleting the DRD2, will select the DRD3. I'm inclined to think that we should select the previous (DRD1) instead of the next one, I'm not sure why... but I'm fine with this logic too. Just saying :)

ljmotta (approved):
It's looking great! nice additions @tiagobento !

thiagoelg (approved):

PR #2239, opened 2024-04-13
kie-issues#947: Expansion of the dmn-editor playwright test suite - part 01

jomarko:
This PR belongs to the expansion phase of the dmn-editor playwright test suite, see the [ticket apache/incubator-kie-tools#3905](https://github.com/apache/incubator-kie-tools/issues/3905). 

This PR especially focus on:
- properties panel tests
- resize node tests

Next PRs for the [ticket apache/incubator-kie-tools#3905](https://github.com/apache/incubator-kie-tools/issues/3905) will focus on these areas:
- DRDs
- Decision Services


packages/dmn-editor/src/diagram/nodes/Nodes.tsx
```diff
@@ -1312,6 +1321,7 @@ export function NodeResizerHandle(props: NodeResizeHandleProps) {
   return (
     <RF.NodeResizeControl style={resizerControlStyle} minWidth={minSize["@_width"]} minHeight={minSize["@_height"]}>
       <div
+        data-testid={`${props.nodeName}-resize-handle`}
```
ljmotta: Looking at it now, I guess we could create more specific `data-testids` as our codebase can increase considerably.

We could use `kie-tools--dmn-editor--*`, WDYT?

packages/dmn-editor/tests/e2e/__fixtures__/propertiesPanel/decisionPropertiesPanel.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  public async getQuestion() {
+    return await this.panel().getByPlaceholder("Enter a question...").inputValue();
+  }
+
+  public async setAllowedAnswers(args: { newAllowedAnswers: string }) {
+    await this.panel().getByPlaceholder("Enter allowed answers...").fill(args.newAllowedAnswers);
+    // commit changes by click to the diagram
+    await this.diagram.resetFocus();
```
ljmotta: For the `Add Edge - *` tests we explicity put this on the test with a `annotation`, so we don't "hide" this bug inside the fixture. We should do the same here.

packages/dmn-editor/tests/e2e/drdArtifacts/changeGroupProperties.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      striketrough: true,
+      color: "#f12200",
+      fontFamily: "Verdana",
+    });
+
+    await expect(diagram.get()).toHaveScreenshot("change-group-font.png");
+  });
+
+  test("should reset the Group node font", async ({ nodes, groupPropertiesPanel }) => {
+    test.skip(true, "https://github.com/apache/incubator-kie-issues/issues/1076");
```
ljmotta, line 74: Please add an `annotation` so after the issue is fixed we can keep track of what this particular test ensure

packages/dmn-editor/tests/e2e/drdArtifacts/changeGroupProperties.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    await expect(nodes.get({ name: DefaultNodeName.GROUP })).toHaveScreenshot("reset-group-font.png");
+  });
+
+  test("should change the Group node shape - fill color", async ({ nodes, groupPropertiesPanel }) => {
+    await nodes.select({ name: DefaultNodeName.GROUP, position: NodePosition.TOP });
+    await groupPropertiesPanel.setFillColor({ color: "#f12200" });
+
+    // It's necessary to pick the parent element ".." to have access to the SVG.
+    await expect(nodes.get({ name: DefaultNodeName.GROUP }).locator("..").locator("rect").nth(0)).toHaveAttribute(
```
ljmotta: Could we put this inside a fixture method?

ljmotta (changes requested):
@jomarko The tests and fixtures are looking great. I've added a few observations inline that should be addressed. My comments are applied to all instances, so please double-check other files as well. For the `data-testid`, you don't need to address the ones that are not on this PR. We'll address them afterward.

packages/dmn-editor/src/diagram/nodes/Nodes.tsx
jomarko: yes, I am fine with that, it is a good idea!

packages/dmn-editor/tests/e2e/__fixtures__/propertiesPanel/decisionPropertiesPanel.ts
jomarko: Here I think we are in a different situation. I do not think in this PR we are doing a workaround due to a bug.

I agree what we spot in `Add Edge -*` tests is an issue we should address. However, here, `allowed answers` property of the decision node is a multiline text area that I can not be committed simply by pressing enter key. That works only for text inputs - single line inputs.

So decided to click into the canvas, to commit the changes. I am not interested much, where to the canvas the click is done, that is why I used the existing `diagram.resetFocus()` fixture.

Should I implement committing the multiline text area in some better/smarter way?

packages/dmn-editor/tests/e2e/drdArtifacts/changeGroupProperties.spec.ts
jomarko: Thank you for the comment. It is a topic I had also in mind, I mean usage `test.skip` and `test.info().annotations.push()`.

To me it seemed not needed to use both, because in the test report I found and information about the skipped and the link for the issue.

I mean for the `changeGroupProperties.spec.ts` the test report looks like below. There is link for the issue, that was a reason for skipping the test.
Screenshot 2024-04-16 172552

Then I compare it with `addAssociation.spec.ts` for example, I see there are two links, but I think there we really do two things, one workaround and one skip.

Screenshot 2024-04-16 172841


Maybe there is something that I miss in the understanding? To me it seems `changeGroupProperties.spec.ts` always need to do a skip and we do not do a workaround, or is your point I should use `test.info().annotations` but with different `TestAnnotations` constant? I see there are `REGRESSION` and `AFFECTED_BY` next to `WORKAROUND_DUE_TO`

packages/dmn-editor/tests/e2e/drdArtifacts/changeGroupProperties.spec.ts
jomarko: Yes, good point, thank you!

packages/dmn-editor/tests/e2e/__fixtures__/propertiesPanel/decisionPropertiesPanel.ts
ljmotta: I overlook this. Thanks for the explanation @jomarko ! As it's related to the properties panel, I guess pressing the `Tab` key would be a better approach.

packages/dmn-editor/tests/e2e/drdArtifacts/changeGroupProperties.spec.ts
ljmotta: The idea behind this is to have the annotation in place for when the issue is solved.

For this case it would be a "regression" annotation as we create the issue and the test ensures the issue was fixed (if it passes). The "affected_by" is when we have a test that doens't work because of an issue. It's very similar but on the regression the test is specific for the issue, and the other is not specific and it's affect by the issue. The workaround is when we don't skip the test and perform a workaround so it can pass.

jomarko: ok, added `REGRESSION` annotations

packages/dmn-editor/tests/e2e/__fixtures__/propertiesPanel/decisionPropertiesPanel.ts
jomarko: thank you for this tip, I like usage of the "Tab", changes pushed now

packages/dmn-editor/tests/e2e/__fixtures__/nodes.ts
```diff
@@ -91,7 +91,7 @@ export class Nodes {
     targetPosition: { x: number; y: number };
     thenRenameTo?: string;
   }) {
-    await this.hover({ name: args.from, position: NodePosition.TOP });
+    await this.select({ name: args.from, position: NodePosition.TOP });
```
ljmotta, line 94: Why do we need this change? Hovering the mouse on top of the node should work for this case.

packages/dmn-editor/tests/e2e/__fixtures__/nodes.ts
```diff
@@ -102,6 +102,21 @@ export class Nodes {
...
+
+  public async getPathAttribute(args: { nodeName: string; attribute: "fill" | "stroke" }) {
+    // It's necessary to pick the parent element ".." to have access to the SVG.
+    return await this.get({ name: args.nodeName }).locator("..").locator("path").nth(0).getAttribute(args.attribute);
+  }
+
+  public async getPolygonAttribute(args: { nodeName: string; attribute: "fill" | "stroke" }) {
+    // It's necessary to pick the parent element ".." to have access to the SVG.
+    return await this.get({ name: args.nodeName }).locator("..").locator("polygon").nth(0).getAttribute(args.attribute);
+  }
```
ljmotta: I guess we need a `private getParentElement(...)` returning the `this.get({ name: args.nodeName }).locator("..")`. WDYT?

packages/dmn-editor/tests/e2e/__fixtures__/propertiesPanel/parts/documentationProperties.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { Locator, Page } from "@playwright/test";
+
+export class DocumentationProperties {
+  constructor(public panel: Locator, public page: Page) {}
+
+  public async addDocumentationLink(args: { linkText: string; linkHref: string }) {
+    await this.panel.getByTitle("Add documentation link").click();
+    await this.panel
+      .locator(".kie-dmn-editor--documentation-link--row")
```
ljmotta: Please use a `data-testid` for the documentation link row instead of the class name.

ljmotta (changes requested):

packages/dmn-editor/tests/e2e/__fixtures__/nodes.ts
jomarko: @ljmotta hi, this is needed because of the **Decision Service** nodes. For those, `hover` is not enough to display their contextual menus. The fixture `dragNewConnectedNode` worked for **Decision Service** nodes tests until now, just because we have a pseudocode
```
addDecisionServiceNode() // this caused Decision service node is selected and its contextual menu is visible
dragNewConnectedNode() // worked , however its internal `hover` had no effect
```

ljmotta (approved):
@jomarko Thanks for your patience! The code is looking great.

tiagobento (approved):

PR #2246, opened 2024-04-17
NO-ISSUE: Add Nix-based development environment using devbox

thiagoelg:
## Nix-based development environment

Our polyglot monorepo structure supports all kinds of programming languages and project structures, but with each new build dependency or CLI tool configuring our local development environment is getting harder and more complex.

[Devbox](https://www.jetify.com/devbox/docs/) presented itself as a good solution that relies on [Nix](https://nixos.org/) packages to install and configure all of these build dependencies, spinning up a new shell with everything required to build and run kie-tools.

Below are some demos showing how easy and powerful it is:

<table>
<tr><td>Devbox Shell
<tr><td>

https://github.com/apache/incubator-kie-tools/assets/10515135/ff6820d5-b2fc-42d7-89a4-1ebf17b73803
</table>

<table>
<tr><td>Automatic shell with <a href="https://direnv.net/">direnv</a>
<tr><td>

https://github.com/apache/incubator-kie-tools/assets/10515135/1fb2ca78-ddbf-4444-835f-e150c022c95a
</table>

<table>
<tr><td>Changing dependency version on the fly
<tr><td>

https://github.com/apache/incubator-kie-tools/assets/10515135/8a63c833-8862-4b3e-ae24-33ec392d2adc
</table>

More information and how-to: https://github.com/thiagoelg/kie-tools/blob/nix-shell/NIX_DEV_ENV.md


devbox.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetpack-io/devbox/0.10.4/.schema/devbox.schema.json",
+  "packages": {
+    "temurin-bin-17": "17.0.9",
```
LightGuard, line 4: Is this the JDK version we want to be using? Should we use Seamer or OpenJDK?

devbox.json
```diff
@@ -0,0 +1,30 @@
...
+    "nodejs": "18.19.1",
+    "maven": "3.9.6",
+    "kubernetes-helm": "3.13.3",
+    "gnumake": "4.4.1",
+    "go": "1.21.8",
+    "python": "3.12.2",
+    "operator-sdk": "1.34.1",
+    "kubebuilder": "3.14.0"
```
jomarko, line 12: Should we have some sort of check to be sure we sync these versions with other places like `repo/build-dependencies-versions.json`

Maybe it could be some github action that read/parse json and compare the versions?

jomarko (commented):
Thank you for this PR, I believe it will improve developer getting started experience. I have only one question inline.

tiagobento (approved):

devbox.json
tiagobento: That would be great, but it's a bit cumbersome to do :/ Unfortunately the names are not a 1:1 to match, so having a check that more or less works doesn't bring much to us, IMHO. It's one more place to remember to update, yes, but not too big of a concern IMHO right now. In any case, thanks for bringing this up @jomarko!

porcelli (approved):
this is fantastic! I've adopted nix in my dev environment and I it has been a game changer, and since then I became vert vocal about it...

really glad to see this **optional** thing - that can boost the developer experience!

devbox.json
thiagoelg: Thanks for looking into it @jomarko! As Tiago said, right now it's just one more place to update, but shouldn't be too hard, we just need to make sure that whatever package we are upgrading is available in nixpkgs.

devbox.json
thiagoelg: Hmm.. not sure, I've always used Temurin, since it's built on top of OpenJDK, and never had problems.

porcelli: I think we've been using temurin for awhile now. I don't know what Seamer distro is.

LightGuard: Sorry, that may have been an autocorrect from the Mac. Semeru, not Seamer.

PR #2254, opened 2024-04-23
kie-issues#1074 & kie-issues#1075: Update 'Shape' part of the properties panel

### LINKED ISSUE #1074: DMN Editor: width and height properties are not validated
DMN node width and height properties can be set using Properties panel component. The issue is the properties panel allows users to put any value.
- Good thing is that that real node is not resized below its minimal size
- Bad is user is not informed that invalid value was put into the form

See the attached screenshot. There is selected a **Decision** node. Its minimal size is 160x80. The node keeps this minimal size even lower value was put into the form (20x10). However we should propbably:
- reset properties panel form inputs to minimal value
- or mark form input values as invalid, red background or sth similar
- or not have a minimal size for nodes
- or ...

Screenshot 2024-04-10 091929

### LINKED ISSUE #1075: DMN Editor: Shape Reset button ignores node size
Node Shape properties have a button for **Reset**ing to the default state. This button reset node colors, however the node size (width and height) is not reset to the default size.

- Create any node
- Resize it
- Open the properties panel for the given node
- Click Shape reset button - **node size is not reset** 

Screenshot 2024-04-10 121431

The current `onReset` [callback](https://github.com/apache/incubator-kie-tools/blob/main/packages/dmn-editor/src/propertiesPanel/ShapeOptions.tsx#L241) should probably include logic from `DefaultSizes.ts` [file](https://github.com/apache/incubator-kie-tools/blob/main/packages/dmn-editor/src/diagram/nodes/DefaultSizes.ts).

jomarko:
In this PR we address apache/incubator-kie-issues#1074 and apache/incubator-kie-issues#1075 as both are tightly related.

Summary of this PR is:
- 1074: We prevent user from setting width/height lower than minimal size allowed, if such value is put in, it is automatically reset to minimal allowed
- 1075: This PR further unify the position of the 'Reset font' and 'Reset shape' buttons.  
- 1075: This PR fixes 'Reset shape' button
- 1074, 1075: Re-enabled and added playwright tests


ljmotta:
@jomarko To use the `onBlur` you will need to create a new `width` and `height` states to be used by the `TextInput` field.

```ts
  const [width, setWidth] = useState<number>(boundWidth);
  const [height, setHeight] = useState<number>(boundHeight);
```

With this in place, you can change the `onChange={onChangeWidth}` and `onChange={onChangeHeight}` to `onBlur={onBlurWidth}` and `onBlur={onBlurHeight}`. The `value` and `onChange` on the `TextInput` will be:

```
value={isDimensioningEnabled ? width : undefined}
onChange={(newWidth) => setWidth(+newWidth)}
```

On the `onBlurWidth` you will use the `width` state instead of the callback parameter to set the bounds, or you will reset the `width` state to the minimal value.

ljmotta:
Also, it's not related to this issue, but could you please add a margin bottom on the Shape section?

image

packages/dmn-editor/src/propertiesPanel/ShapeOptions.tsx
```diff
@@ -105,17 +110,36 @@ export function ShapeOptions({
...
+        });
+
+        if (parseFloat(newWidth) < minNodeSize["@_width"]) {
+          bounds["@_width"] = minNodeSize["@_width"];
+        } else {
+          bounds["@_width"] = +parseFloat(newWidth).toFixed(2);
+        }
       });
     },
     [setBounds]
```
ljmotta: Missing dependencies

packages/dmn-editor/src/propertiesPanel/ShapeOptions.tsx
```diff
@@ -105,17 +110,36 @@ export function ShapeOptions({
...
+          isAlternativeInputDataShape: state.computed(state).isAlternativeInputDataShape(),
+        });
+        if (parseFloat(newHeight) < minNodeSize["@_height"]) {
+          bounds["@_height"] = minNodeSize["@_height"];
+        } else {
+          bounds["@_height"] = +parseFloat(newHeight).toFixed(2);
+        }
       });
     },
     [setBounds]
```
ljmotta: Missing dependencies

packages/dmn-editor/src/propertiesPanel/ShapeOptions.tsx
```diff
@@ -250,6 +274,17 @@ export function ShapeOptions({
...
+              snapGrid: state.diagram.snapGrid,
+              isAlternativeInputDataShape: state.computed(state).isAlternativeInputDataShape(),
+            });
+            shape["dc:Bounds"]!["@_width"] = minNodeSize["@_width"];
+            shape["dc:Bounds"]!["@_height"] = minNodeSize["@_height"];
+          }
+        }
       });
     });
   }, [setShapeStyles]);
```
ljmotta, line 331: Missing dependencies

ljmotta (changes requested):

packages/dmn-editor/src/propertiesPanel/ShapeOptions.tsx
```diff
@@ -63,6 +68,9 @@ export function ShapeOptions({
   const boundPositionX = useMemo(() => +(shapeBound?.["@_x"]?.toFixed(2) ?? ""), [shapeBound]);
   const boundPositionY = useMemo(() => +(shapeBound?.["@_y"]?.toFixed(2) ?? ""), [shapeBound]);
 
+  const [width, setWidth] = useState<number>(boundWidth);
+  const [height, setHeight] = useState<number>(boundHeight);
```
jomarko, line 74: This is now the only issue, imagine a scenario:
- two nodes added to the canvas
- open properties panel
- select first node
- change its width or height
- **keep properties panel open and select second node**
- the properties panel is all correctly updated, but **width** and **height** are not, the values of the first node are shown

If we do click diagram between node selection or we close and reopen properties panel, correct **width** and **height** are initialized.

ljmotta: @jomarko that's true, we will have this problem. We'll also have another problem: resizing the node using the node handler instead of using the properties panel will not update the properties panel as well. Meaning we need to keep the state in sync with the `boundWidth`/`boundHeight`. To do that, we can use an `useEffect`.

jomarko:
All changes incorporated, let's keep fingers crossed for the CI!

tiagobento:
Maybe we should keep the REGRESSION annotation in the tests? I mean, it's always nice to have the history persisted in the code. That annotation being there tells the story of why those tests exist. WDYT?

tiagobento:
@jomarko Please follow the recommendations of ESLint regarding hook dependencies. There's a lot of warnings in the new code...

jomarko:
@tiagobento thank you for the warning, about the warnings, I believe I fixed all I introduced in `ShapeOptions.tsx`. I pushed also a new code comparing yesterday status, the `onBlur` was problematic in this scenario.
- add two nodes, change their sizes randomly
- open properties panel, **Shape** section
- select first node
- change its **width**, using properties panel input box, and confirm by click into **height** properties panel input box
- select second node
- ❗  here the second node **height** was incorrectly updated, because **onBlur** was invoked, but already different node was selected

packages/dmn-editor/src/propertiesPanel/ShapeOptions.tsx
```diff
@@ -238,8 +289,22 @@ export function ShapeOptions({
...
+        setWidth(boundWidth);
+        setHeight(boundHeight);
+        setCurrentlyReShapedNodeId(nodeIds[0]);
+      }
+    }, 0);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [boundWidth, boundHeight, currentlyReShapedNodeId, nodeIds]);
```
ljmotta: Why we need the `setTimeout` here? One important concept of the `useEffect` is it runs everytime a value of the dependency array change. Meaning if the `boundWidth` change, it will re-run, calling the `setHeight` and `setCurrentlyReShapedNodeId`. To avoid this, we need to break this useEffect into multiple ones.

```ts
useEffect(() => {
  setWidth(boundWidth);
}, [boundWidth])
```

packages/dmn-editor/src/propertiesPanel/ShapeOptions.tsx
```diff
@@ -63,6 +68,12 @@ export function ShapeOptions({
   const boundPositionX = useMemo(() => +(shapeBound?.["@_x"]?.toFixed(2) ?? ""), [shapeBound]);
   const boundPositionY = useMemo(() => +(shapeBound?.["@_y"]?.toFixed(2) ?? ""), [shapeBound]);
 
+  // this variable was introduced because of scenarios, when user switches between nodes without closing and reopening the properties panel
+  const [currentlyReShapedNodeId, setCurrentlyReShapedNodeId] = useState<string>(nodeIds[0]);
```
ljmotta: I'm not sure we need this, could you share which valeus are not updated when we select another node? Changing the selected node should update all the `bound*` constants.

ljmotta (changes requested):

packages/dmn-editor/src/propertiesPanel/ShapeOptions.tsx
ljmotta: I guess I understand the reason behind this. The way to solve this, is to have the `nodes` outside of the `onBlur` callback. It would look like the `shape` constant.

```ts
const node= useDmnEditorStore((s) =>
    s.computed(s).getDiagramData(externalModelsByNamespace).nodesById.get(nodeIds[0])
  );
```

This way `node` will be updated everytime the `nodeIds` are updated.

jomarko: I am sorry, I was not able to incorporate this, I tried a lot, but still the corner case was not working. Will describe the corner case more here.

- Imagine we have a diagram with two nodes, lest say **A** and **B**
- Lets resize them, you can use drag handler in the corner of the node, to make the behavior obvious, for the **A** lets increase its width and for the **B** lets increase its height
- Now, please open **A** properties panel, its Shape section, set focus to its **width** field, i.e. click into it, or use **Tab** until the cursor is in the **width**
- Keep the focus in the **width** and keep the Properties panel open please
- Now, click the node **B**, i.e. select the node **B**
- ❗ here you will see that the node **B** **width** was automatically set to the **width** of the node **A**. If we put breakpoints into the **onBlurWidth**, we will see that in the time of its execution the blur **event.target.value** contain **width** of the node **A**, however the **nodeIds[0]** is already equal of the node **B**

ljmotta:
@jomarko I've created a PR[1] for your branch with some fixes.
On it, I've fixed the onBlur with a reference. I save the `nodeId` so when the `onBlur` fires it will have the previous `nodeId` (if the selected node changed).
I could see a problem on the `onReset` too. The bug was happening the second time the reset button was pressed. To fix that I passed the `minNodeSize` from the `callback`, and I had to spread the `DEFAULT_*_COLOR`, which was being overwritten. 

[1] https://github.com/jomarko/kie-tools/pull/10

jomarko:
@ljmotta thank you for your help! The `useRef` hook usage with string was knowledge that I was missing. In examples, it is often used more with some `input` element. I like we use it for string variable.

Also, I like the change we added `nodeId` parameter to `setBounds` and we pass it from `onBlur` handlers.

The only change I needed to do, was a place we 'sync' `previousNodeId`, because during a manual checks I found another corner scenario, not original one, that was not working, I describe it here so you can double-check my changes are appropriate.

- Imagine we have a diagram with two nodes, lest say **A** and **B**
- Lets resize them, you can use drag handler in the corner of the node, to make the behavior obvious, for the **A** lets increase its width and for the **B** lets increase its height
- Now, please open **A** properties panel, its Shape section, set focus to its **width** field, i.e. click into it, or use **Tab** until the cursor is in the **width**
- Keep the focus in the **width** and keep the Properties panel open please
- Now, click the node **B**, i.e. select the node **B**
- Repeat the previous focus exercise for this node **B**. Set focus to its **width** or **height** field. However, just set focus to them, do not change the value.
- Now, again without closing or reopening the Properties panel, select node **A**
- ❗ here you will see that the node **A** would be inappropriately resized using **width**/**height** values of the node **B**

ljmotta (approved):
@jomarko Thanks! I've made some tests and it's working.

PR #2257, opened 2024-04-23
NO-ISSUE: Fix typo on `cors-proxy` package scripts

caponetto:


thiagoelg (approved):
Nice catch!

tiagobento:
@caponetto Thanks a lot for this! I guess we weren't seeing a failure because we ended up using the default value for `origin` in this case, as the `$(build-env ...)` command didn't have any output, due to the typo...

This highlights a problem we have on shell command substitution usage, which unfortunately can't be solved on the shells themselves. I created

- https://github.com/apache/incubator-kie-issues/issues/1113

and would love to hear your input!

paulovmr (approved):

PR #2259, opened 2024-04-23
NO-ISSUE: Fix Maven Wrapper version and install tar/gzip to dev-deployment-base-image

thiagoelg:
With the [3.3.0 release of Maven Wrapper](https://github.com/apache/maven-wrapper/releases/tag/maven-wrapper-3.3.0) the default method of installing/downloading the wrapper is via `-Dtype=only-script`, which means that the .jar file required to run the wrapper os downloaded during runtime, as before it was downloaded during build time with `-Dtype=bin`.

This change breaks the build inside container images that don't have the required binaries to download and extract the Maven Wrapper dependencies (e.g. dev-deployment-kogito-quarkus-blank-app-image didn't have `gzip` installed from its parent image, causing `./mvnw` to fail when attempting to extract the wrapper binary).

This PR fixes the Maven Wrapper version to 3.3.0 and installs the required dependencies on the dev-deployment-base-image.

More info: https://maven.apache.org/wrapper/


packages/dev-deployment-base-image/Containerfile
```diff
@@ -31,7 +31,10 @@ USER root
 ENV DEV_DEPLOYMENT__UPLOAD_SERVICE_EXTRACT_TO_DIR=$HOME_PATH/app
 ENV DEV_DEPLOYMENT__UPLOAD_SERVICE_PORT=8080
 
-RUN mkdir -p -m 777 $HOME_PATH/app && mkdir -p -m 777 /tmp/app && mkdir -p -m 777 /.m2
+RUN microdnf install -y tar-2:1.34-6.el9_1.x86_64 gzip-1.12-1.el9.x86_64 \
```
LightGuard, line 34: I doubt it'll make that much of a difference, but does microdnf also cleanup when it is done?

LightGuard (approved):

packages/dev-deployment-base-image/Containerfile
thiagoelg: Nice catch, adding `microdnf clean all` should make the image a bit smaller.

tiagobento (approved):

PR #2261, opened 2024-04-24
kie-tools-issues#2260: Use python3 instead of python

### LINKED ISSUE #2260: Use python3 instead of python
`python` usually refers to Python 2 and `python3` refers to Python 3. Some Linux distributions still does not link `python` to `python3`.

tomasdavidorg:
Fixes #2260


devbox.json
```diff
@@ -7,7 +7,7 @@
     "kubernetes-helm": "3.13.3",
     "gnumake": "4.4.1",
     "go": "1.21.8",
-    "python": "3.12.2",
+    "python3": "3.12.2",
```
thiagoelg: This can stay as `python`, as there's no `python3` on Nixpkgs.

rodrigonull:
As an alternative you can run:
`sudo update-alternatives --install /usr/local/bin/python python $(which python3) 1`
`sudo update-alternatives --install /usr/local/bin/pip pip $(which pip3) 1`
on linux systems to make `python3` link to `python`

devbox.json
tomasdavidorg: Thanks. Fixed.

tomasdavidorg:
> As an alternative you can run: `sudo update-alternatives --install /usr/local/bin/python python $(which python3) 1` `sudo update-alternatives --install /usr/local/bin/pip pip $(which pip3) 1` on linux systems to make `python3` link to `python`

Yes, there is also a `python-is-python3` package for some distributions but I would say from user experience we should lower the number of commands needed for building this repo. There is already a lot of required preinstalled software but I understand that this is caused by the latest changes and in the future, some improvements will be necessary.

tomasdavidorg:
I am just suggesting these changes because on some distributions there is no default linking `python` to `python3` and I also found some older recommendation like this one https://stackoverflow.com/questions/62214293/is-python3-always-installed-with-python-3. If you think this is not necessary we can close this PR.

thiagoelg (approved):

tiagobento (approved):

PR #2274, opened 2024-04-29
kie-issues#1019: React-Based DMN Editor: allowedValues not allowed for custom type

### LINKED ISSUE #1019: React-Based DMN Editor: allowedValues not allowed for custom type
It is not possible to set constraints for custom types

Unknown-3

Unknown-4

Image

Both `allowedValues` and `typeConstraint` refers to the base type. So, a custom type seems allowed to define such constraints, that will be applied to the base type. 

This would make sense because:

1. a custom type “A” extend a base type “string” (e.g.)
2. both custom type “B” and “C” extends “A”, but they could have different constraints

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/1019

## Description
This PR enables constraints for several cases:
 - a refecence to struct (complex item definition);
image

 - a collection of a reference to struct (complex item definition);
image

 - a reference for a data type;
image

 - a collection of a reference for a data type;
image

 - a collection of a reference for a collection of a data type;
image

 - all cases above for the item components of a complex type
image

Additional fixes:
 - Changing between two or more data types that had an `expression`  constriant would not update the `expression` constraint according.

before:
https://github.com/apache/incubator-kie-tools/assets/24302289/e696532a-07bf-4451-89de-bb0bf4e35c46

after:
https://github.com/apache/incubator-kie-tools/assets/24302289/a40eeda9-e88d-4241-92b3-911fe7983b5c

### DIFF
--- a/packages/dmn-editor/src/dataTypes/Constraints.tsx
+++ b/packages/dmn-editor/src/dataTypes/Constraints.tsx
@@ -18,7 +18,7 @@
  */
 
 import * as React from "react";
-import { useMemo, useCallback } from "react";
+import { useMemo, useCallback, useState, useEffect } from "react";
 import { ConstraintsExpression } from "./ConstraintsExpression";
 import {
   DMN15__tItemDefinition,
@@ -311,6 +311,7 @@ export function useConstraint({
 }) {
   const constraintValue = constraint?.text.__$$text;
   const kieConstraintType = constraint?.["@_kie:constraintType"];
+  const itemDefinitionId = itemDefinition["@_id"]!;
 
   const isConstraintEnum = useMemo(
     () =>
@@ -359,7 +360,7 @@ export function useConstraint({
     };
   }, [enabledConstraints, enumToKieConstraintType, isCollectionConstraintEnabled, itemDefinition]);
 
-  const selectedConstraint = useMemo<ConstraintsType>(() => {
+  const initialSelectedConstraint = useCallback(() => {
     if (isConstraintEnabled.enumeration && kieConstraintType === "enumeration") {
       return ConstraintsType.ENUMERATION;
     }
@@ -389,6 +390,13 @@ export function useConstraint({
     kieConstraintType,
   ]);
 
+  const [selectedConstraint, setSelectedConstraint] = useState<ConstraintsType>(initialSelectedConstraint);
+  // Changing the `itemDefinitionId` should re-calculate the initial state of `selectedConstraint`;
+  useEffect(() => {
+    setSelectedConstraint(initialSelectedConstraint());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [itemDefinitionId]);
+
   return {
     constraintValue,
     isConstraintEnum,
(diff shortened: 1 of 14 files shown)


tiagobento (approved):

jomarko (commented):
Thank you for this PR @ljmotta. I did some manual checks and have some interesting findings, however not sure how to proceed with them.

### 01 Clearing the Collection constraint expression
I found out that if user is typing and removing characters into **Collection constraint** expression input, then the app will automatically switch to  **None** constraint type if user removed all characters.

Imagine user is in a state like below:
Screenshot 2024-04-30 124203

Then press backspace 5times to remove all characters, editor will automatically switch to **None**
Screenshot 2024-04-30 124213

### 02 Constraints and Runner
I built locally your PR as `online-editor... build:dev` and started both `online-editor` and `extended-services`

Then the problems appear if user firstly starts to use a Data type in the diagram and consequently start to add constraints.

The diagrams may be very simple:
[structure constraints 2.dmn.txt](https://github.com/apache/incubator-kie-tools/files/15162303/structure.constraints.2.dmn.txt)

Once you open such ^ diagram, switch to the **Data types** tab and try to add enumeration **Collection item constraint** to the **tStringCustom**. This will cause in few seconds this error screen appears:
Screenshot 2024-04-30 124650

The same issue can be seen, if user starts to edit the **Collection constraint** expression, Simply starts typing there.

ljmotta:
@jomarko Thanks a lot for your review! I've addressed the first problem on my latest commit. Could you please review again?

Regarding the second problem, it was addressed on https://github.com/apache/incubator-kie-drools/pull/5858 . To check it yourself, you can change the `KOGITO_RUNTIME_version` env variable on `root-env/env/index.js`  to `999-20240428-SNAPSHOT`, build the `extended-services-java` with `pnpm build:dev` and start it.

jomarko:
@ljmotta thank you for the updates.

### 01 Clearing the Collection constraint expression
Now works, if the user clears the expression, the toggle group button is not switched to none anymore. 
However, there is one interesting issue, in the `Expression` constraint widget. The monaco suggestions are displayed just every odd hit of the keyboard. Imagine you want to type `count` word:
- c: monaco is shown
- o: monaco is hidden
- u: monaco is shown
- n: monaco is hidden
- t: monaco is shown

### 02 Constraints and runner
Thank you for the guidance of using latest available `KOGITO_RUNTIME_version`. I can confirm I was able to put a `Collection item constraint` with extended services connected to the sandbox. However now I see the same behavior if I try to put some `Collection constraint` expression in a moment, when given data type is already in use (assigned as data type to any node). Is that expected? I mean there is no `KOGITO_RUNTIME_version` available that would contain a fix for it?

ljmotta:
@jomarko I've addressed the issue. It took more than I thought. I've also made a deploy of it: https://ljmotta.github.io/kie-tools-demos/kie-issues-1019/#/

jomarko (approved):
Thank you @ljmotta. I re-checked the 01 point. I will check the 02 in the future.

tiagobento (approved):

PR #2290, opened 2024-05-03
NO-ISSUE: Remove bin directory from kogito-serverless-operator before install

thiagoelg:
Why?

Installed binaries can have dependencies or references to old go versions and will break if go version changes.
This guarantees that binaries are added during bootstrap/install.


pefernan (approved):

porcelli (approved):
lgtm - thank you for spotting this

yesamer (approved):

PR #2302, opened 2024-05-07
kie-issues#1173: Retrieve workflow definitions from data index instead of openapi.json

### LINKED ISSUE #1173: Retrieve workflow definitions from data index instead of openapi.json

paulovmr:
Closes: https://github.com/apache/incubator-kie-issues/issues/1173

### DIFF
--- a/packages/runtime-tools-swf-gateway-api/src/gatewayApi/apis.tsx
+++ b/packages/runtime-tools-swf-gateway-api/src/gatewayApi/apis.tsx
@@ -518,42 +518,28 @@ export const triggerCloudEvent = (event: CloudEventRequest, baseUrl: string, pro
   return doTriggerCloudEvent(event, baseUrl, proxyEndpoint);
 };
 
-export const createWorkflowDefinitionList = (
-  workflowDefinitionObjs: WorkflowDefinition[],
-  url: string
-): WorkflowDefinition[] => {
-  const workflowDefinitionList: WorkflowDefinition[] = [];
-  workflowDefinitionObjs.forEach((workflowDefObj) => {
-    const workflowName = Object.keys(workflowDefObj)[0].split("/")[1];
-    const endpoint = `${url}/${workflowName}`;
-    workflowDefinitionList.push({
-      workflowName,
-      endpoint,
-    });
-  });
-  return workflowDefinitionList;
-};
-
-export const getWorkflowDefinitionList = (baseUrl: string, openApiPath: string): Promise<WorkflowDefinition[]> => {
-  return new Promise((resolve, reject) => {
-    SwaggerParser.parse(`${baseUrl}/${openApiPath}`)
-      .then((response) => {
-        const workflowDefinitionObjs: any[] = [];
-        const paths = response.paths;
-        const regexPattern = /^\/[^\n/]+\/schema/;
-        Object.getOwnPropertyNames(paths)
-          .filter((path) => regexPattern.test(path.toString()))
-          .forEach((url) => {
-            let workflowArray = url.split("/");
-            workflowArray = workflowArray.filter((name) => name.length !== 0);
-            /* istanbul ignore else*/
-            if (Object.prototype.hasOwnProperty.call(paths![`/${workflowArray[0]}`], "post")) {
-              workflowDefinitionObjs.push({ [url]: paths![url] });
-            }
-          });
-        resolve(createWorkflowDefinitionList(workflowDefinitionObjs, baseUrl));
+export const getWorkflowDefinitions = (client: ApolloClient<any>): Promise<WorkflowDefinition[]> => {
+  return new Promise<WorkflowDefinition[]>((resolve, reject) => {
+    client
+      .query({
+        query: GraphQL.GetProcessDefinitionsDocument,
+        fetchPolicy: "network-only",
+        errorPolicy: "all",
       })
-      .catch((err: any) => reject(err));
+      .then((value) => {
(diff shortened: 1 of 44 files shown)


packages/runtime-tools-swf-gateway-api/src/gatewayApi/apis.tsx
```diff
@@ -518,42 +518,28 @@ export const triggerCloudEvent = (event: CloudEventRequest, baseUrl: string, pro
...
+        errorPolicy: "all",
       })
-      .catch((err: any) => reject(err));
+      .then((value) => {
+        const workflowDefinitions = value.data.ProcessDefinitions;
+        resolve(
+          value.data.ProcessDefinitions.map((item: { id: string; endpoint: string }) => {
+            return {
+              workflowName: item.id,
+              endpoint: item.endpoint,
```
fantonangeli, line 539: I think we should continue composing the `endpoint` using the `kogitoServiceUrl` to send the correct `Target-Url` to cors-proxy

packages/serverless-logic-web-tools/src/runtimeTools/contexts/WebToolsWorkflowDefinitionListContextProvider.tsx
```diff
@@ -20,12 +20,17 @@
...
 import { useSettings } from "../../settings/SettingsContext";
+import { useEnv } from "../../env/EnvContext";
 
 export function WebToolsWorkflowDefinitionListContextProvider(props: React.PropsWithChildren<{}>) {
   const settings = useSettings();
+  const { env } = useEnv();
 
   return (
-    <WorkflowDefinitionListContextProvider kogitoServiceUrl={settings.runtimeTools.config.kogitoServiceUrl}>
+    <WorkflowDefinitionListContextProvider
```
fantonangeli, line 30: Following my previous comment about `kogitoServiceUrl` I think we still need to pass it to `WorkflowDefinitionListContextProvider`

fantonangeli (changes requested):
@paulovmr Thank you for your PR.
I left 2 comments about `kogitoServiceUrl` which should be needed to send the data to the workflow, but overall looks good.

caponetto:
@paulovmr do we still need to set up this field? 
All the info you need is present on the Data Index, including the Service URLs.

image

packages/runtime-tools-swf-gateway-api/src/gatewayApi/apis.tsx
paulovmr: Hi @fantonangeli , if you have `kogito.service.url` set in the application.properties, the endpoint recorded in the data-index will come correctly with the base URL. Like this:
```
{
  "data": {
    "ProcessDefinitions": [
      {
        "id": "expression",
        "endpoint": "http://localhost:8480/expression"
      }
    ]
  }
}
```

caponetto:
@fantonangeli I think we'll need a similar improvement to the SonataFlow web app. WDYT?

fantonangeli:
> @fantonangeli I think we'll need a similar improvement to the SonataFlow web app. WDYT?

Yes, I agree

packages/runtime-tools-swf-gateway-api/src/gatewayApi/apis.tsx
fantonangeli: I confirm it works. Thanks for your instruction.

pefernan:
hi @paulovmr ! Would it be possible to bring this to the `jbpm-quarkus-devui` too??

paulovmr:
hi @pefernan ! Sure, I'm doing some other changes in the web tools UI to wrap up this PR, but after I can send one with similar changes to the current ones for jbpm as well.

packages/serverless-logic-web-tools/src/navigation/Routes.ts
```diff
@@ -148,14 +148,20 @@ export const routes = {
...
-  }>(({ workflowId }) => `/runtime-tools/workflow-details/${workflowId}`),
+  }>(({ workflowId }) => `/runtime-tools/workflow-instances/${workflowId}`),
   runtimeToolsWorkflowForm: new Route<{
     pathParams: PathParams.WORKFLOW_NAME;
   }>(({ workflowName }) => `/runtime-tools/workflow-definition/${workflowName}`),
-  runtimeToolsTriggerCloudEvent: new Route<{}>(() => `/runtime-tools/trigger-cloud-event`),
-  runtimeToolsTriggerCloudEventForWorkflow: new Route<{
+  /*  runtimeToolsTriggerCloudEvent: new Route<{
+    queryParams: QueryParams.URL;
+  }>(() => `/runtime-tools/trigger-cloud-event`),*/
```
caponetto:
```suggestion
```

caponetto (commented):
LGTM. Left only one minor comment.

fantonangeli:
LGTM and tested locally. 
Only one question: do we want to update SonataFlow Deployment Webapp in this PR or in a separate PR?
https://github.com/apache/incubator-kie-tools/blob/e35af74d9bda0b444e712294d29363426330e1ff/packages/sonataflow-deployment-webapp/src/pages/Workflows/Workflows.tsx#L43

caponetto (approved):

paulovmr:
@fantonangeli Thanks, I've created https://github.com/apache/incubator-kie-issues/issues/1276 to tackle the `sonataflow-deployment-webapp` separately.

fantonangeli (approved):

PR #2304, opened 2024-05-07
kie-issues#937: On the DMN Editor, adding a DRD then pressing "Undo", makes the Editor will crash

### LINKED ISSUE #937: On the DMN Editor, adding a DRD then pressing "Undo", makes the Editor will crash.
`diagram.drdIndex` must never point to a non-existent DRD. We can use `state.computed` to make sure the index is never invalid based on the array's length.

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/937

## Description
This PR changes the `diagram.drdIndex` to `diagram.__unsafeDrdIndex`. Now, to retrieve the `drdIndex` it's required to use the computed `getDrdIndex()` method. This method will check if the current `__unsafeDrdIndex` is pointing to a valid DRD index. In a positive case, it will return the current value. Otherwise, it will return the last available index.


tiagobento (approved):

yesamer (approved):

porcelli (approved):
I understand that we still don't have the proper setup to add to this PR a test coverage that would prevent this issue in the future.

this is tracked by this https://github.com/apache/incubator-kie-issues/issues/1174

PR #2309, opened 2024-05-09
kie-issues#407: KIE Sandbox Accelerators break the current workspace if they fail to initialize

### LINKED ISSUE #407: KIE Sandbox Accelerators break the current workspace if they fail to initialize
On the KIE Sandbox Editor Page, if the "Apply Accelerator" fails, we have two different behaviors:
- In an workspace without commits it will break losing all files on it;
- In an workspace with commits, it will rollback to the last commit;

## Video
https://github.com/kiegroup/kie-issues/assets/24302289/1cc9f05c-679d-4608-baac-e0a313a11400

## Steps to reproduce
A simple way to reproduce it, is to use the KIE Sandbox without running the cors-proxy.
 - Open the DMN Editor;
 - Click in the Apply Accelerator dropdown, and in Quarkus;
 - The workspace will break.

```[tasklist]
### Acceptance Criteria
- [ ] Before applying the Accelerator it should perform a commit action;
- [ ] If the Accelerator fails it should rollback to the previous state;
```

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/407

Worked out this solution with @kbowers-ibm. Before creating a commit to the backup branch, we `add` the files to the staging area.

https://github.com/apache/incubator-kie-tools/assets/24302289/1e113b55-d131-475a-b3ff-3485b24a619c


packages/workspaces-git-fs/src/worker/WorkspacesWorkerApiImpl.ts
```diff
@@ -389,6 +389,19 @@ export class WorkspacesWorkerApiImpl implements WorkspacesWorkerApi {
 
   //git
 
+  public async kieSandboxWorkspacesGit_add(args: { workspaceId: string; relativePath: string }): Promise<void> {
```
thiagoelg: There's already a method called `kieSandboxWorkspacesGit_stageFile` that does exactly this and it's also callable via `workspaces.stageFile` on KIE Sandbox.

packages/online-editor/src/accelerators/AcceleratorsHooks.tsx
```diff
@@ -162,11 +162,16 @@ export function useAcceleratorsDispatch(workspace: ActiveWorkspace) {
...
 
         const workspaceFiles = await workspaces.getFiles({ workspaceId });
-
-        // Create new temporary branch with current files, but stay on main
+        // Create a backup branch with the current files, but stay on main
         await workspaces.branch({ workspaceId, name: BACKUP_BRANCH_NAME, checkout: false });
 
-        // Commit moved files to moved files branch (this commit will never be pushed, as this branch will be deleted)
+        // Adds all files to the staging area
+        await workspaces.add({
```
thiagoelg:
```suggestion
        await workspaces.stageFile({
```

packages/workspaces-git-fs/src/worker/WorkspacesWorkerApiImpl.ts
thiagoelg: Maybe `stageFile` is not the best name for this method since it's capable of staging all files within a path, so it's your call to rename `stageFile` or delete it and keep the `add`, as they work the same.

ljmotta:
@thiagoelg Thanks for your review. I think the best approach would to add a nested object into the ` workspaces` context isolating the `git` methods. As this would require a bigger refactor I've rolled back to the `stageFile` as `add`  can be ambiguous.

thiagoelg (approved):

tiagobento (approved):

PR #2316, opened 2024-05-10
kie-issue#1149: DMN Editor produces invalid typeRef attributes

### LINKED ISSUE #1149: DMN Editor produces invalid typeRef attributes
This issue will describe the situation using `for` expression, however probably all DMN 1.5. expressions may have this issue.

In the attached screenshots we see two semantically same for expression. The good thing is that models can be executed, as we see in the DMN Runner panel on the right side, however the extended-services report validation issues to us.

Screenshot 2024-05-03 132440
Screenshot 2024-05-03 123506

If we take a look into produced xml code, we will notice `typeRef="&lt;Undefined&gt;"` on few places, especially in the nested parts of the `<in>` and `<return>`. 

- [for-expression.dmn.txt](https://github.com/apache/incubator-kie-issues/files/15199976/for-expression.dmn.txt)
- [for-more-complex.dmn.txt](https://github.com/apache/incubator-kie-issues/files/15199975/for-more-complex.dmn.txt)

To solve the problem `typeRef` pointing to `<Undefined>` shouldn't be present in the xml **or** the `typeRef` should point to a correct data type. In the attached models:
- `<in>` nested expression should point to `tNumHolderList`
- `<return>` nested expression should point to `tNumHolder`

here are version of file, without validation issues:

- [for-expression-fixed.dmn.txt](https://github.com/apache/incubator-kie-issues/files/15199990/for-expression-fixed.dmn.txt)
- [for-more-complex-fixed.dmn.txt](https://github.com/apache/incubator-kie-issues/files/15200006/for-more-complex-fixed.dmn.txt) <- contains validation issues but not related to `typeRef` probably

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1149

Some places related to the user interface are using "<Undefined>".
The places where the data is marshaled are using `undefined`.

The "<Undefined>" value is still present in `DmnBuiltInDataType` because it is still a "valid value".

### DIFF
--- a/packages/dmn-editor/src/boxedExpressions/getDefaultBoxedExpression.tsx
+++ b/packages/dmn-editor/src/boxedExpressions/getDefaultBoxedExpression.tsx
@@ -18,30 +18,30 @@
  */
 
 import {
-  DmnBuiltInDataType,
-  BoxedExpression,
-  generateUuid,
-  BoxedLiteral,
-  BoxedFunction,
-  BoxedContext,
-  BoxedList,
-  BoxedInvocation,
-  BoxedRelation,
-  BoxedDecisionTable,
   BoxedConditional,
-  BoxedFor,
-  BoxedSome,
+  BoxedContext,
+  BoxedDecisionTable,
   BoxedEvery,
+  BoxedExpression,
   BoxedFilter,
+  BoxedFor,
+  BoxedFunction,
+  BoxedInvocation,
+  BoxedList,
+  BoxedLiteral,
+  BoxedRelation,
+  BoxedSome,
+  DmnBuiltInDataType,
+  generateUuid,
 } from "@kie-tools/boxed-expression-component/dist/api";
 import {
-  LITERAL_EXPRESSION_MIN_WIDTH,
+  BEE_TABLE_ROW_INDEX_COLUMN_WIDTH,
   CONTEXT_ENTRY_VARIABLE_MIN_WIDTH,
+  DECISION_TABLE_ANNOTATION_DEFAULT_WIDTH,
   DECISION_TABLE_INPUT_DEFAULT_WIDTH,
   DECISION_TABLE_OUTPUT_DEFAULT_WIDTH,
-  DECISION_TABLE_ANNOTATION_DEFAULT_WIDTH,
+  LITERAL_EXPRESSION_MIN_WIDTH,
   RELATION_EXPRESSION_COLUMN_DEFAULT_WIDTH,
-  BEE_TABLE_ROW_INDEX_COLUMN_WIDTH,
 } from "@kie-tools/boxed-expression-component/dist/resizing/WidthConstants";
 import {
   DECISION_TABLE_INPUT_DEFAULT_VALUE,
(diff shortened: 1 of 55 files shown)


packages/dmn-editor/src/boxedExpressions/getDefaultBoxedExpression.tsx
```diff
@@ -68,13 +65,13 @@ export function getDefaultBoxedExpression({
...
   logicType: BoxedExpression["__$$element"] | undefined;
-  typeRef: string;
+  typeRef: string | undefined;
   allTopLevelDataTypesByFeelName: DataTypeIndex;
   getInputs?: () => { name: string; typeRef: string | undefined }[] | undefined;
   getDefaultColumnWidth?: (args: { name: string; typeRef: string | undefined }) => number | undefined;
   widthsById: Map<string, number[]>;
 }): BoxedExpression {
-  const dataType = allTopLevelDataTypesByFeelName.get(typeRef);
+  const dataType = allTopLevelDataTypesByFeelName.get(typeRef ?? "<Undefined>");
```
ljmotta: `"<Undefined>"` is value used in a few places. Maybe extract to a variable? WDYT?

packages/stunner-editors-dmn-loader/src/index.tsx
```diff
@@ -98,7 +98,7 @@ const BoxedExpressionEditorWrapper: React.FunctionComponent<BoxedExpressionEdito
   const beeGwtService: BeeGwtService = {
     getDefaultExpressionDefinition(logicType, dataType, isRoot) {
       const defaultExpression = gwtExpressionToDmnExpression(
-        window.beeApiWrapper?.getDefaultExpressionDefinition(gwtLogicType(logicType), dataType)
+        window.beeApiWrapper?.getDefaultExpressionDefinition(gwtLogicType(logicType), dataType ?? "<Undefined>")
```
ljmotta: The `dataType` parameter of `getDefaultExpressionDefinition` accepts an `undefined` value, why fallback to `"<Undefined>"`? Maybe should we fallback inside the method?

ljmotta (commented):
@danielzhe Thanks for this change! I've added some comments inline

packages/dmn-editor/src/boxedExpressions/getDefaultBoxedExpression.tsx
danielzhe: You're right. I'm going to use the `DmnBuiltInDataType.Undefined` for this.

packages/stunner-editors-dmn-loader/src/index.tsx
danielzhe: Because this is to the GWT layer and since it will be removed in the future I think we should avoid spending time on it.

tiagobento: This is not for display only, though, right? I mean, this is getting into the XML.

danielzhe: As we talked in pvt, I changed it to the proposed change by @ljmotta :)

danielzhe:
Changes applied! Ready for your review @ljmotta @tiagobento

ljmotta (approved):

tiagobento (approved):

tiagobento (changes requested):
Actually... Just checked an there are a lot of references to `DmnBuiltInDataType.Undefined` in the `stunner-editors-dmn-loader` component. Those are used in the `beeToGwt` function, and certainly are leaking to the XML. Please check that too...

tiagobento (approved):

PR #2341, opened 2024-05-16
kie-issues#451: Implement autolayout on the new React-based DMN Editor for DMN files without Diagram information - Part 1

ljmotta:
This PR is part of https://github.com/apache/incubator-kie-issues/issues/451

## Description
This PR adds a new empty modal for DMN models that have a DRG and doens't have a DRD. 

image

Clicking in "Auto-generate Diagram" will auto-generate a DRD with the auto-layout applied. Clicking in "I'll create the Diagram myself" will keep the DRD empty. If the user wants to interact with the DRG nodes, it will need to use the "DRG Nodes" option from the palette.

This PR also:
- Split the Auto layout logic from the `AutoLayoutButton` to the `autoLayouy.ts` and `AutoLayoutHook.ts`.
- Add a new example without a DRD on the dev webapp


packages/dmn-editor/src/diagram/Diagram.tsx
```diff
@@ -1182,6 +1186,131 @@ export const Diagram = React.forwardRef<DiagramRef, { container: React.RefObject
...
+          }}
+        />
+
+        <EmptyState>
+          <EmptyStateIcon icon={MousePointerIcon} />
+          <Title size={"md"} headingLevel={"h4"}>
+            {`This DMN's Diagram is empty`}
+          </Title>
+          <EmptyStateBody>{`The current DMN doens't have a DRD associated with the DRG.`}</EmptyStateBody>
+          <EmptyStateBody>{`Do you want to auto generated it?`}</EmptyStateBody>
```
jomarko:
```suggestion
          <EmptyStateBody>{`The current DMN doesn't have a DRD associated with the DRG.`}</EmptyStateBody>
          <EmptyStateBody>{`Do you want to auto-generate it?`}</EmptyStateBody>
```

packages/dmn-editor/src/normalization/autoGenerateDrd.ts
```diff
@@ -0,0 +1,205 @@
...
+  };
+}) {
+  // Create DRD
+  args.model.definitions["dmndi:DMNDI"] = {
+    ...args.model.definitions["dmndi:DMNDI"],
+    "dmndi:DMNDiagram": [
+      {
+        "@_id": generateUuid(),
+        "@_name": getDefaultDrdName({ drdIndex: 0 }),
+        "@_useAlternativeInputDataShape": false,
```
jomarko, line 53: we invoke `autoGenerateDrd` from the `Diagram.tsx`. There on many places we dynamically check for this alternative shape setting like `state.computed(state).isAlternativeInputDataShape()`. is it expected, here, in `autoGenerateDrd.ts` we hard-code alternative shape to false?

jomarko (commented):
Thank you for the PR, my only point for discussion is the terminology we use. In the new modal we inform user about some missing DRD in DRG, what is aligned with DMN standard terminology I think.

However, what is the actual problem in human language. IIUC, there is no information, about nodes placement, their shapes, sizes, is that correct? I do not have a clear answer but from my point of view it is a point that somehow should share opinion about it, if we should be closer to the standard terminology or closer to human language.

Recently I know there was requirement from Gabriele for the data types tab, to remove the term `Structure` and start to use the term `Complex` to be aligned with standard. There I think both are quite understandable for the human.

Here I am not sure if the term DRD is enough understandable for the user. I mean, Does user confirming this new modal understand what will happen in the diagram? Or, does the user understand what is consequence of clicking yes and what is a consequence of clicking no?

packages/dmn-editor/src/normalization/autoGenerateDrd.ts
ljmotta: The modal will popup as soon as the user open the DMN model (e.g. first time opening the Editor). As `false` is the default value for any new model, I've harded code this value. The user still can change this option on the DRD dropdown.

ljmotta:
@jomarko Thanks for bringing this up. I agree, I will add more information to the user, like the node placement, sizes and consequences of the yes/no choice. Now, regarding the specific terminology, the editor is specific to DMN models so I think we can assume the user will know a few things at least. It's a gray area for me, and in this case I'd say it's not necessary to explain what is DRD and DRG as both are very important DMN concepts. What you guys think? @tiagobento @baldimir @yesamer @gitgabrio

gitgabrio:
@ljmotta 
Let me give my POV.  DMN is mostly about "execution", while the graphical part is just a way to easily define/visualize them.
An indirect proof is that a model without any kind of graphical information is perfectly valid and executable. 
What I'm driving at is that while DMN user may (or should) know all the details related to execution (e.g. difference between our most beloved `typeConstraint` and `allowedValues`) because they actually change the behavior, the graphical part are just inner details, whose knowledge does not have any impact on the execution, and could be happily ignored.
So, I also would prefer something simpler, like 
"The current DMN does not have any diagram associated"

wdyt ?

yesamer:
About the terminology: I guess it's fine to use the proposed sentence by Gabriele, but I'm not against using the DRG and DRD acronyms. Maybe, we can use the extended form as an alternative (eg. Decision Requirement Diagram / Graph), or use the info icon with a popup to better clarify it.

About the current implementation, I like it, well done @ljmotta. In addition, I would consider just to manage this scenario:
- The user opens a DMN without the DMNDI;
- The user chooses no when dealing with the AutoGeneration popup introduced in that PR
- After a while, the user changes his/her mind and wants to autogenerate the Diagram.

Based on that, can we consider ALWAYS showing the modal, without a yes/no option, until the user explicitly wants to autogenerate it? Eg.

Screenshot 2024-05-22 at 11 47 18

In that way, the users:

- Don't need to take any actions if he/she is not interested in autogenerating the Diagram;
- We don't force the user to make that choice just after opening the file in the editor
- The user can easily autogenerate the Diagram at any moment;

WDYT?

baldimir:
Hi, as discussed in the typeConstraint and allowedValues issue, I think the best would be to add clarifications in a hint (or the "i" icon). Similar approach should be applied everywhere throughout the application I think, so it starts to be self-documented to some extent.

ljmotta:
@yesamer @gitgabrio @jomarko I've updated the PR description with the new modal. I've used a piece of each idea, and I'm very happy with the result. WDYT?

tiagobento:
I love it. Nicely done! Looks great. +1 from me!

gitgabrio:
Very nice, thanks @ljmotta ! 👍

danielzhe:
I'm late for the discussion but I don't like the terms DRG and DRD and I agree with @gitgabrio 

In the old editor, I was the one who created the modal for this case. We had some discussion about at that time and the sentence we chose to use at the time was: "The diagram does not have layout information". We never used the terms "DRG" or "DRD".

I saw you changed the message @ljmotta and I think the new terms are great and very clear to the user what is happening.

yesamer:
It looks great!!! Thank you @ljmotta for listening to our feedback and implementing a nice solution!
Great teamwork!!

danielzhe (approved):
Great work! I didn't find any issue with the DMN models without layout I had.

tiagobento (approved):

PR #2344, opened 2024-05-17
NO-ISSUE: Add range and enum DMN Runner validation for the time, date, date-time and duration types of the new Type Constraint property

ljmotta:
The new DMN editor uses the new `type constraints` property to set the Item Definition/Item Components constraint. This new property is mapped to the  JSON Schema `x-dmn-type-constraints` keyword. This PR adds the property to the AJV validator. This PR also break the file `constants` into multiple specific ones.

Before:
image

After:
image

### DIFF
--- a/packages/dmn-runner/src/ajv.ts
+++ b/packages/dmn-runner/src/ajv.ts
@@ -26,20 +26,24 @@ import {
   DATE_AND_TIME_FEEL_REGEXP,
   DATE_ENUM_REGEXP,
   DATE_FEEL_REGEXP,
-  DAYS_AND_TIME_DURATION_FORMAT,
-  DAYS_AND_TIME_DURATION_REGEXP,
-  RECURSION_KEYWORD,
-  RECURSION_REF_KEYWORD,
-  X_DMN_ALLOWED_VALUES_KEYWORD,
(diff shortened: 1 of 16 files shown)


tiagobento (approved):

jomarko:
Looking

jomarko (commented):
Thank you for this PR. For me it seems working for most of the data types. However I see some issue for `date and time`

This attached model is simple id function, it takes the input (date and time) and return the same value.

[dateandtime.dmn.txt](https://github.com/apache/incubator-kie-tools/files/15378323/dateandtime.dmn.txt)

If the dmn built in `date and time` is used, it works well, see:
Screenshot 2024-05-20 162933

However if the input data node data type is changed to `tDateAndTime`, it crashes.
Screenshot 2024-05-20 162903


Please notice the extended servies needs to be running.

ljmotta:
Apparently, the string is converted to a Date class after a new edit, and I'm not sure why. I've added a new line to handle this case. @jomarko Thanks for spotting this one.

jomarko (approved):
Thank you for the updates, the scenario with the `date and time` seems to work fine now!

tiagobento (approved):

PR #2345, opened 2024-05-17
NO-ISSUE: Apache KIE™ names for VS Code and Chrome extensions

tiagobento:
The Chrome Extensions' descriptions need to be edited directly on the Chrome Web Store.

### DIFF
--- a/packages/bpmn-vscode-extension/README.md
+++ b/packages/bpmn-vscode-extension/README.md
@@ -15,7 +15,7 @@
    under the License.
 -->
 
-## BPMN Editor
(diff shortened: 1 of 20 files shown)


yesamer (approved):
Thank you @tiagobento.
In parallel, we should deprecate the current extensions signed with the KIE group and and add redirection to this new one.

porcelli (approved):

klbowers (approved):

tiagobento:
It's only the displayName that is changing @yesamer

PR #2351, opened 2024-05-21
NO-ISSUE: DMN Editor: Fix random 'object not extensible' errors by typing everything with the new `Normalized` type, making the `@_id` property non-optional

tiagobento:
cc @jomarko @kbowers-ibm @ljmotta. Please see the attached video with a known reproducer you shared, where the error doesn't happen anymore.

https://github.com/apache/incubator-kie-tools/assets/1584568/cc332e46-b168-42d6-961a-ef7ada846daa

### DIFF
--- a/packages/dmn-editor/src/normalization/normalize.ts
+++ b/packages/dmn-editor/src/normalization/normalize.ts
@@ -17,11 +17,22 @@
  * under the License.
  */
 
+import { DmnLatestModel } from "@kie-tools/dmn-marshaller";
 import { getNewDmnIdRandomizer } from "../idRandomizer/dmnIdRandomizer";
 import { addMissingImportNamespaces } from "../mutations/addMissingImportNamespaces";
 import { State } from "../store/Store";
 
-export function normalize(model: State["dmn"]["model"]) {
+export type Normalized<T> = WithRequiredDeep<T, "@_id">;
+
+export type WithRequiredDeep<T, K extends keyof any> = T extends undefined
+  ? T
+  : T extends Array<infer U>
+  ? Array<WithRequiredDeep<U, K>>
+  : { [P in keyof T]: WithRequiredDeep<T[P], K> } /** */ & (K extends keyof T //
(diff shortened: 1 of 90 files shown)


packages/dmn-editor/src/normalization/normalize.ts
```diff
@@ -17,11 +17,22 @@
...
 
+import { DmnLatestModel } from "@kie-tools/dmn-marshaller";
 import { getNewDmnIdRandomizer } from "../idRandomizer/dmnIdRandomizer";
 import { addMissingImportNamespaces } from "../mutations/addMissingImportNamespaces";
 import { State } from "../store/Store";
 
-export function normalize(model: State["dmn"]["model"]) {
+export type Normalized<T> = WithRequiredDeep<T, "@_id">;
+
+export type WithRequiredDeep<T, K extends keyof any> = T extends undefined
```
jomarko: Do we need to `export` this type? It is used only internally in this file I think.

packages/dmn-editor/src/normalization/normalize.ts
```diff
@@ -17,11 +17,22 @@
...
+  ? T
+  : T extends Array<infer U>
+  ? Array<WithRequiredDeep<U, K>>
+  : { [P in keyof T]: WithRequiredDeep<T[P], K> } /** */ & (K extends keyof T //
```
jomarko:
```suggestion
  : { [P in keyof T]: WithRequiredDeep<T[P], K> } & (K extends keyof T
```

?

jomarko (approved):
some questions inline, otherwise I didn't find issues during manual check

thank you for this fix

packages/dmn-editor/src/normalization/normalize.ts
ljmotta: The `/** */` looks like a leftover. The `//` forces a line break.

ljmotta (approved):

PR #2352, opened 2024-05-22
NO-ISSUE: DMN Editor: Overlay modal can't be scrolled

danielzhe:
Before:
image

After:
BEFORE

When it has enough area, no scroll is shown:
image

### DIFF
--- a/packages/dmn-editor/src/diagram/Diagram.tsx
+++ b/packages/dmn-editor/src/diagram/Diagram.tsx
@@ -129,6 +129,8 @@ export const DEFAULT_VIEWPORT = { x: 100, y: 100, zoom: 1 };
 
 const DELETE_NODE_KEY_CODES = ["Backspace", "Delete"];
 
+const AREA_ABOVE_OVERLAYS_PANEL = 120;
+
 const nodeTypes: Record<NodeType, any> = {
   [NODE_TYPES.decisionService]: DecisionServiceNode,
   [NODE_TYPES.group]: GroupNode,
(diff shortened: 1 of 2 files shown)


tiagobento (approved):

yesamer (approved):

Josephblt (changes requested):
Hello, nice improvement here, but I found some issues. They are related to the fact that size changes in the parent panel are not being considered.

The following images illustrate that.
Screenshot from 2024-05-22 09-54-17
Here I close the problems tab after opening the overlay modal.
Screenshot from 2024-05-22 09-55-27
Here I open the overlay modal and than the problems tab.

I request changes here because this behavior can lead to more serious issues. Like this:
Screenshot from 2024-05-22 09-56-49
To reproduce it:

1. Add a decision node.
2. Open the Problems tab
3. Open the overlay modal.
4. Close the problems tab.
5. Select the decision node.

danielzhe:
@Josephblt I think this is a minor issue since it doesn't compromise usability like the issue that this PR is fixing, it is just ugly. We can add "live resize" for the panel that adapts the screen as its available size changes but in my opinion this is another ticket. What do you think, @tiagobento ?

Josephblt:
Hello @danielzhe, I completely agree. The improvement is clear.  This change gives access to all the options, and the problems are merely aesthetical. Plus they are easily solvable with a few clicks. So I don´t see any problens in addressing this in another issue.

Josephblt (approved):
Hello @danielzhe, I completely agree. The improvement is clear. This fix gives access to all the options, and the problems are merely aesthetical. Plus they are easily solvable with a few clicks. So I don´t see any problems in addressing this in another issue. I'm changing the review status to approved.

tiagobento (approved):

PR #2353, opened 2024-05-22
NO-ISSUE: DMN Editor: Prevent from 'Include model' click in case of empty model

jomarko:
### Before
Imagine there is folder with two DMN model and both are empty - there is no node or custom data type defined in them. That is a similar to situation to early phase of any project.

Now please follow these steps.
- lets say models have names **a.dmn** and **b.dmn**
- open the model **a**
- click the included models tab
- try to import the model **b**
- you will be in situation like shown in the attached screenshot
- then press **Include model** button - **Here is the issue, nothing happens**

and there will be an error in js console

### Now
User is prevented from clicking 'Include model' in case the model can not be included.
Screenshot 2024-05-22 142715


Josephblt (approved):
Working great, both for empty models and valid ones.
Screenshot from 2024-05-22 10-46-38
Screenshot from 2024-05-22 10-46-48

packages/dmn-editor/src/includedModels/IncludedModels.css
```diff
@@ -0,0 +1,3 @@
+.selected-model-to-include-error {
```
tiagobento: Please use the `kie-dmn-editor--` prefix

tiagobento (changes requested):

packages/dmn-editor/src/includedModels/IncludedModels.css
jomarko: Sure, good point, added in new commit.

yesamer (approved):

tiagobento (approved):

PR #2358, opened 2024-05-23
kie-issues#119: VS Code Extensions should keep their Editors in sync even when files change externally

### LINKED ISSUE #3734: VS Code Extensions should keep their Editors in sync even when files change externally
When using VS Code, if you have any file, like `my-dmn.dmn` open, and you do a `git pull` that changes this file, the Editor does not update, basically ignoring the `git pull` you just did. That's just an example, any external change you make to files that are open in VS Code will not update them.

That's really bad, because if you make a change, you'll override what you just pulled.

The expected behavior is that the Editor will reload itself if its associated file changed externally, thus displaying the updated version of the file.

Josephblt:
Closes https://github.com/apache/incubator-kie-tools/issues/3734

Done:
- Created file watcher to track changes in file sources.

Next steps:
- Identify source sync status.
- Create sync control dialog.
- Sync accordingly.


tiagobento:
I merged apache/incubator-kie-issues#2165, can you please update this PR?

Josephblt:
Hello @tiagobento. Yes, sure thing. I still need to make some changes to this, but I'll update it as soon as possible.

tiagobento:
Closing due to inactivity.

PR #2360, opened 2024-05-23
NO-ISSUE: DMN Editor: fix external node underline (input data)

jomarko:
### Before
External `Input Data` node isn't with the underlined name. The bug is present both for standard and alternative `Input Data` shape.

### After
Screenshot 2024-05-23 125818
Screenshot 2024-05-23 125810


ljmotta (approved):

Josephblt (approved):
Working nicely.

PR #2368, opened 2024-05-24
NO-ISSUE: Holding alt and navigating away from the page locks the editor

klbowers:
Adding alt to pan was done to simulate the "Pan" action of the legacy DMN Editor, but it isn't ideal when pressing alt and navigating away from the page locks the editor, with e.g. alt+tab, ctrl+alt+delete, holding alt and clicking to a new tab.

Removing this shortcut as using the middle mouse button or the right mouse button allows one to pan around easily.


ljmotta (approved):
I'm in favor to remove the keyboard shortcut if we already have another way to pan.

tiagobento (approved):

PR #2371, opened 2024-05-24
NO-ISSUE: DMN Editor: Unable to "unhide" a hidden Decision Service Node

klbowers:
Temporarily disabling the ability to hide a decision service in the case where there isn't any other depiction of that Decision Service in another DRD.


jomarko (commented):
I think this PR works, just not sure why it is needed only for the Decision Service node, to keep at least one Decision Service depiction visible, while for other node types we do not enforce this rule.

Also found one interesting scenario for the decision service node when we compare using single file vs included model.

### Single model
- Create new dmn file
- add Decision Service node to the default DRD
- add new DRD
- add the Decision Service node also to the new DRD
- So we have One Decision Service node depiction in two DRDs
- Switch to the default DRD and hide, using X key, the decision service, we will have this state [1]

### Included models
- Create new dmn file
- add Decision Service node to it
- create another dmn file in the same workspace and include the file from the steps above
- now in the default DRD add to the diagram the included decision service node
- add new DRD to the diagram
- add the included Decision Service node also to the new DRD
- So we have one included Decision Service node depiction in two DRDs
- Switch to the default DRD and hide, using X key, the decision service, we will have this state [2]

[1]
Screenshot 2024-05-28 170021

[2]
Screenshot 2024-05-28 170052

I think it is strange the states [1] and [2] are different, as the state of the models is basically the same. Or maybe it is only my inapprop[riate understanding?

klbowers:
@jomarko 

> just not sure why it is needed only for the Decision Service node, to keep at least one Decision Service depiction visible, while for other node types we do not enforce this rule.

This is a temporary workaround, as the logic currently isn't implemented to unhide a hidden decision service, when there's no other depiction of that decision service anywhere. This resulted in being able to hide a decision service, but not being able to unhide it, so this seemed to be the better temporary solution until all Decision Service logic is fully implemented

klbowers:
Not sure if I'm missing something about your scenario, but as far as I can tell that's the way all types of nodes work. The node will be present in the external nodes panel, and not in the drg elements panel

jomarko:
I see new label, 'DO NOT MERGE', please let me know once this is ready for a review.

ljmotta:
Still missing a case where the external mode doesn't have a depict of a Decision Service. The Decision Service should be auto generated

porcelli (approved):

tiagobento (approved):

Josephblt (approved):

danielzhe (approved):

PR #2373, opened 2024-05-27
kie-issues#1269: kn-plugin-workflow: Improve gen-manifests

### LINKED ISSUE #1269: Improve gen-manifests
Can we add two more flags to gen-manifests:
 
 --image - Specify an image to be used as the workflow's container image ( sets a value for .spec.podTemplate.container.image)
 --profile - Specify the value for the sonataflow.org/profile annotation (any of prod/dev/nil or future gitops when/if be added).
 --skip-namespace

treblereel:
Closes: https://github.com/apache/incubator-kie-issues/issues/1269


mareknovotny:
cc here @LuboTerifaj @domhanak @rgdoliveira

tomasdavidorg:
> cc here @LuboTerifaj @domhanak @rgdoliveira

I am looking at this PR. I hope I will finish today or on Monday.

tomasdavidorg (approved):
Looks good to me.

Just one note:

- it is possible to use following command where `namespace` and `skip-namespace` which are mutually exclusive:
`kn workflow gen-manifest --namespace "mynamespace" --skip-namespace` 
- is it the same thing when `namespace` is empty?
`kn workflow gen-manifest --namespace="" --skip-namespace` 

These are non blocking notes. Just thinking if it is ok to have it like this.

packages/kn-plugin-workflow/pkg/command/deploy_undeploy_common.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/common"
 	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/metadata"
 	"github.com/apache/incubator-kie-tools/packages/sonataflow-operator/workflowproj"
+	apiMetadata "github.com/apache/incubator-kie-tools/packages/sonataflow-operator/api/metadata"
```
ricardozanini: nit; avoid camel case when aliasing imports.

packages/kn-plugin-workflow/pkg/command/deploy_undeploy_common.go
```diff
@@ -203,11 +206,19 @@ func generateManifests(cfg *DeployUndeployCmdConfig) error {
 		handler.AddResourceAt(filepath.Base(dashboardFile), metadata.DashboardsDefaultDirName, specIO)
 	}
 
+	if cfg.Profile != "" {
```
ricardozanini: For ubiquity, please use `len(cfg.Profile) == 0` instead.

ricardozanini:
@tomasdavidorg has excellent points here. I believe when these conditions are met, the CLI should halt and do nothing to fail fast and avoid misunderstandings.

treblereel:
> Looks good to me.
> 
> Just one note:
> 
> * it is possible to use following command where `namespace` and `skip-namespace` which are mutually exclusive:
>   `kn workflow gen-manifest --namespace "mynamespace" --skip-namespace`
> * is it the same thing when `namespace` is empty?
>   `kn workflow gen-manifest --namespace="" --skip-namespace`
> 
> These are non blocking notes. Just thinking if it is ok to have it like this.

1) in case of `--namespace "mynamespace" --skip-namespace` --skip-namespace would win

2) yeap, as I understand correctly from the issue, --skip-namespace should ack as ` --namespace=""`. Correct me, if I'm wrong please

ricardozanini:
> > Looks good to me.
> > Just one note:
> > 
> > * it is possible to use following command where `namespace` and `skip-namespace` which are mutually exclusive:
> >   `kn workflow gen-manifest --namespace "mynamespace" --skip-namespace`
> > * is it the same thing when `namespace` is empty?
> >   `kn workflow gen-manifest --namespace="" --skip-namespace`
> > 
> > These are non blocking notes. Just thinking if it is ok to have it like this.
> 
> 1. in case of `--namespace "mynamespace" --skip-namespace` --skip-namespace would win
> 2. yeap, as I understand correctly from the issue, --skip-namespace should ack as ` --namespace=""`. Correct me, if I'm wrong please

A slightly difference here.

`namespace = ""` is different than stripping the `namespace` attribute from the CR. Not sure how `kubectl` would interpret it, we can test it. But I believe we can remove the `namespace` from the CR when `--skip-namespace` and add the attribute being blank when `-namespace=""`.

treblereel:
> > > Looks good to me.
> > > Just one note:
> > > 
> > > * it is possible to use following command where `namespace` and `skip-namespace` which are mutually exclusive:
> > >   `kn workflow gen-manifest --namespace "mynamespace" --skip-namespace`
> > > * is it the same thing when `namespace` is empty?
> > >   `kn workflow gen-manifest --namespace="" --skip-namespace`
> > > 
> > > These are non blocking notes. Just thinking if it is ok to have it like this.
> > 
> > 
> > 
> > 1. in case of `--namespace "mynamespace" --skip-namespace` --skip-namespace would win
> > 2. yeap, as I understand correctly from the issue, --skip-namespace should ack as ` --namespace=""`. Correct me, if I'm wrong please
> 
> A slightly difference here.
> 
> `namespace = ""` is different than stripping the `namespace` attribute from the CR. Not sure how `kubectl` would interpret it, we can test it. But I believe we can remove the `namespace` from the CR when `--skip-namespace` and add the attribute being blank when `-namespace=""`.

thank you for the clarification, I'll fix it

treblereel:
> > > Looks good to me.
> > > Just one note:
> > > 
> > > * it is possible to use following command where `namespace` and `skip-namespace` which are mutually exclusive:
> > >   `kn workflow gen-manifest --namespace "mynamespace" --skip-namespace`
> > > * is it the same thing when `namespace` is empty?
> > >   `kn workflow gen-manifest --namespace="" --skip-namespace`
> > > 
> > > These are non blocking notes. Just thinking if it is ok to have it like this.
> > 
> > 
> > 
> > 1. in case of `--namespace "mynamespace" --skip-namespace` --skip-namespace would win
> > 2. yeap, as I understand correctly from the issue, --skip-namespace should ack as ` --namespace=""`. Correct me, if I'm wrong please
> 
> A slightly difference here.
> 
> `namespace = ""` is different than stripping the `namespace` attribute from the CR. Not sure how `kubectl` would interpret it, we can test it. But I believe we can remove the `namespace` from the CR when `--skip-namespace` and add the attribute being blank when `-namespace=""`.

@ricardozanini, I'm not sure how to force `sigs.k8s.io/yaml` to write empty strings atm :)

In case this issue needs to be investigated more (I mean, how does kubectl work with empty namespace decl), maybe we can have one more separate issue for it?

packages/kn-plugin-workflow/pkg/command/gen_manifest.go
```diff
@@ -197,3 +233,23 @@ func resolveManifestDir(folderName string) (string, error) {
 
 	return absPath, nil
 }
+
+func isValidProfile(profile string) error {
```
ricardozanini, line 237: I'll add this to the operator's API package. Can you please open an issue here? https://github.com/apache/incubator-kie-kogito-serverless-operator/

packages/sonataflow-operator/workflowproj/workflowproj.go
```diff
@@ -64,6 +64,8 @@ type WorkflowProjectHandler interface {
 	SaveAsKubernetesManifests(path string) error
 	// AsObjects returns a reference to the WorkflowProject holding the Kubernetes Manifests based on your files.
 	AsObjects() (*WorkflowProject, error)
+	// Image overrides the default image in the generated SonataFlow manifest
+	Image(image string) WorkflowProjectHandler
```
ricardozanini, line 68: Can you please also backport this to the operator's repo? https://github.com/apache/incubator-kie-kogito-serverless-operator/

treblereel:
@tomasdavidorg @ricardozanini A separate issue has been created, so I think we can skip --skip-namespace for now.

So if there are no more notes, maybe this issue can be merged?

mareknovotny:
can you guys merge this? @ricardozanini @tomasdavidorg

@treblereel can't merge it by himself

PR #2379, opened 2024-05-28
NO-ISSUE: DMN Editor: Blank screen with DMN models 1.1

danielzhe:
The new editor does not support opening old DMN 1.1 models. 

We now show an error/warning message to the user that models from 1.1 are not supported.

image


tiagobento (approved):

thiagoelg (approved):

ljmotta (approved):

porcelli (approved):

jomarko:
sorry for a silly question, do we use that 'sad smile' as icon also on other places of tooling? just trying to be sure we are consistent and do not use many different icons

PR #2433, opened 2024-06-17
NO-ISSUE: Adjust CI for all branches

rodrigonull:
In this PR:

- Adjust CI to be able to run the Jenkins pipelines for any branch.
- Fix some wrong image names
- Downgrade CI base image to Ubuntu 22.04 because Playwright does not support Ubuntu 24.04 yet. (https://github.com/microsoft/playwright/issues/30368)

### DIFF
--- a/.ci/incubator-kie-tools-ci-build.Dockerfile
+++ b/.ci/incubator-kie-tools-ci-build.Dockerfile
@@ -1,4 +1,4 @@
-FROM cruizba/ubuntu-dind:noble-26.1.3
+FROM cruizba/ubuntu-dind:jammy-26.1.3
 
 SHELL ["/bin/bash", "-c"]
 
@@ -45,7 +45,7 @@ xvfb \
 fluxbox \
 subversion && \
 apt-get clean autoclean && apt-get autoremove --yes && \
-rm -rf /var/lib/{apt,dpkg,cache,log}/
+rm -rf /var/lib/{apt,cache,log}/
 
 # Install firefox
 RUN wget -O /tmp/firefox-latest.tar.bz2 "https://download.mozilla.org/?product=firefox-latest&os=linux64&lang=en-US" && \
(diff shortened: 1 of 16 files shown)


.ci/incubator-kie-tools-ci-build.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM cruizba/ubuntu-dind:noble-26.1.3
+FROM cruizba/ubuntu-dind:jammy-26.1.3
```
tiagobento, line 1: What's the difference between "noble" and "jammy"? :D

.ci/jenkins/Jenkinsfile.ci-main
```diff
@@ -54,8 +53,7 @@ pipeline {
                             "https://github.com/${pipelineVars.githubRepositoryOrg}/${pipelineVars.githubRepositoryName}.git",
                             "${pipelineVars.kieToolsBotGithubCredentialsId}"
                         )
-
-                        BUILD_IMAGE_REQUIRED = githubUtils.fileIsInChangeset('incubator-kie-tools-ci-build.Dockerfile')
+                        BUILD_IMAGE_REQUIRED = (BRANCH_NAME != 'main' || githubUtils.fileIsInChangeset('incubator-kie-tools-ci-build.Dockerfile'))
```
tiagobento: Is `BRANCH_NAME != 'main'` for the case of new branches? If so, maybe we could have a way to check if a tag for that stream exists or not, right? Because otherwise, all PRs for non-`main` branches will re-build the image, no?

.ci/incubator-kie-tools-ci-build.Dockerfile
rodrigonull: noble is ubuntu 24.04 and jammy is ubuntu 22.04

.ci/jenkins/Jenkinsfile.ci-main
rodrigonull: Your assumption is correct. The way we filter the job run is by adding a regular expression: `(main|\d{1,2}\.\d{1,2}\.x)` in the job configuration triggers.

tiagobento: Right, so we need to change that... otherwise the new branches will unnecessarily build the image every time...

rodrigonull: As I mentioned in my previous comment we already have the regular expression filter applied in the jobs, so they will only be triggered for the main branch and for the branches that have their names following this pattern: `0.0.x`

tiagobento (approved):

PR #2444, opened 2024-06-21
NO-ISSUE: Set locale to "en-US" before generating `image-builder` and `run-script-if` READMEs

ljmotta:
Adding this prevent `yargs` of using the machine locale.

### DIFF
--- a/packages/image-builder/package.json
+++ b/packages/image-builder/package.json
@@ -25,7 +25,7 @@
   "scripts": {
     "install": "rimraf dist && tsc && pnpm write-readme",
     "write-readme": "run-script-os",
-    "write-readme:darwin:linux": "echo '```' > README.md && ./bin.js --help >> README.md && echo '```' >> README.md",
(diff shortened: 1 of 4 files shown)


thiagoelg (approved):

tiagobento:
What problem does this solve? I get that this pins the locale, but why do we need that?

ljmotta:
@tiagobento `yargs --help` is used to write the the `image-builder` and `run-script-if` README files. By default `yargs` use the machine locale, so in my case it was `en-BR`, which re-writed the files using Portuguese.

tiagobento:
Maybe instead of hardcoding it to all the invocations of `image-builder` and `run-script-if`, we could simply foce the locale in the invocations specific to building the READMEs, no?

tiagobento:
I mean, defining a locale is pretty out of scope for `run-script-if` and `image-builder`.

tiagobento (approved):

PR #2446, opened 2024-06-24
kie-issues#1337: kn-plugin-workflow should verify if the cluster has SonataFlow Operator installed

### LINKED ISSUE #1337: kn workflow should verify if the cluster has SonataFlow Operator installed
If you try to run kn workflow deploy in a cluster without Knative installed, the user might face a weird message from the Kube API complaining about the lack of the KService CRD.

It would be nice if this command checks for a Knative installation before attempting to deploy the application.

jira: https://issues.redhat.com/browse/KOGITO-7665

treblereel:
Closes: https://github.com/apache/incubator-kie-issues/issues/1337

### DIFF
--- a/packages/kn-plugin-workflow/pkg/command/deploy.go
+++ b/packages/kn-plugin-workflow/pkg/command/deploy.go
@@ -188,6 +188,14 @@ func runDeployCmdConfig(cmd *cobra.Command) (cfg DeployUndeployCmdConfig, err er
 		return cfg, fmt.Errorf("❌ ERROR: failed to get default dashboards files folder: %w", err)
 	}
 
+	// check if sonataflow operator CRDs are installed
+	crds := []string{"sonataflowbuilds.sonataflow.org", "sonataflowclusterplatforms.sonataflow.org", "sonataflowplatforms.sonataflow.org", "sonataflows.sonataflow.org"}
+	for _, crd := range crds {
+		if !common.CheckKubectlCrdExists(crd) {
+			return cfg, fmt.Errorf("❌ ERROR: the required CRDs are not installed.. Install the SonataFlow Operator CRD first")
+		}
+	}
+
 	//setup manifest path
 	if err := setupConfigManifestPath(&cfg); err != nil {
 		return cfg, err
--- a/packages/kn-plugin-workflow/pkg/command/deploy.go
+++ b/packages/kn-plugin-workflow/pkg/command/deploy.go
@@ -189,8 +189,7 @@ func runDeployCmdConfig(cmd *cobra.Command) (cfg DeployUndeployCmdConfig, err er
 	}
 
 	// check if sonataflow operator CRDs are installed
-	crds := []string{"sonataflowbuilds.sonataflow.org", "sonataflowclusterplatforms.sonataflow.org", "sonataflowplatforms.sonataflow.org", "sonataflows.sonataflow.org"}
-	for _, crd := range crds {
+	for _, crd := range metadata.SonataflowCRDs {
 		if !common.CheckKubectlCrdExists(crd) {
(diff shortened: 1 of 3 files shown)


treblereel:
@ricardozanini a few notes:
* As I understand, we should check if the sonataflow operator CRD is installed (I used to ), this is what you meant under Knative installed, right?
* looks like we didn't test the plugin against a real k8s cluster, so no tests (but I'm not sure if we need them)
* Under the hood, it still uses direct kubectl calls. It will be refactored as part of the migration to the go-cli issue.

ricardozanini:
@treblereel the implementation is correct, but the title of the PR is confusing.

Knative -> https://knative.dev, which is the platform we rely on for some use cases (like events and workflows as a function)
SonataFlow -> our platform that runs workflows based on the CNCF Serverless Workflow spec.

So having said that, we should aim for this issue to verify if the SonataFlow APIs are installed in the cluster, which is what you are doing.

For Knative, this is another check slightly different, which I don't think it's the responsibility of the CLI since the operator is handling that: https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/controllers/knative/knative.go#L67

treblereel:
> @treblereel the implementation is correct, but the title of the PR is confusing.
> 
> Knative -> https://knative.dev, which is the platform we rely on for some use cases (like events and workflows as a function) SonataFlow -> our platform that runs workflows based on the CNCF Serverless Workflow spec.
> 
> So having said that, we should aim for this issue to verify if the SonataFlow APIs are installed in the cluster, which is what you are doing.
> 
> For Knative, this is another check slightly different, which I don't think it's the responsibility of the CLI since the operator is handling that: https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/controllers/knative/knative.go#L67

without sonataflow operator installed, the error is:
```
❌ ERROR: applying deploy: ❌ ERROR: failed to deploy manifest /var/folders/df/r7mb2d9n1_s29v94c7zf9gw40000gn/T/manifests4164473872/01-sonataflow_hello.yaml,  ❌ ERROR: failed to execute kubectl apply command for /var/folders/df/r7mb2d9n1_s29v94c7zf9gw40000gn/T/manifests4164473872/01-sonataflow_hello.yaml: exit status 1
```

ricardozanini:
@treblereel but what is the error message from `kubectl` when you try to apply a CR that doesn't have a corresponding CRD in the cluster? (in this case, the `SonataFlow` resource).

packages/kn-plugin-workflow/pkg/command/deploy.go
```diff
@@ -188,6 +188,14 @@ func runDeployCmdConfig(cmd *cobra.Command) (cfg DeployUndeployCmdConfig, err er
 		return cfg, fmt.Errorf("❌ ERROR: failed to get default dashboards files folder: %w", err)
 	}
 
+	// check if sonataflow operator CRDs are installed
+	crds := []string{"sonataflowbuilds.sonataflow.org", "sonataflowclusterplatforms.sonataflow.org", "sonataflowplatforms.sonataflow.org", "sonataflows.sonataflow.org"}
+	for _, crd := range crds {
+		if !common.CheckKubectlCrdExists(crd) {
+			return cfg, fmt.Errorf("❌ ERROR: the required CRDs are not installed.. Install the SonataFlow Operator CRD first")
```
ricardozanini, line 194:
```suggestion
			return cfg, fmt.Errorf("❌ ERROR: the required CRDs are not installed. Install the SonataFlow Operator CRD first")
```

ricardozanini: nitpick: you may move these strings to a const since you might have to use them again in the future.

ricardozanini (approved):

wmedvede (approved):

domhanak (approved):

PR #2453, opened 2024-06-27
kie-issues#114: JavaCodeCompletion module fails to compile because of API change

### LINKED ISSUE #114: JavaCodeCompletion module fails to compile because of API change
It seems that an API change on JDT.LS project is incompatible with our codebase.

Screenshot 2023-05-11 at 12 49 09

To solve that, I'm going to split this into multiple sub-tasks:
- Using the tag version 1.23.0 instead of the snapshot one. That will solve the issue. 
- Unfortunately, tagged versions get stale after some months (see https://github.com/eclipse/eclipse.jdt.ls/issues/1970) So, we need to update our codebase accordingly with the API changes

```[tasklist]
### Acceptance criteria
- [ ] https://github.com/kiegroup/kie-tools/pull/1626
- [ ] https://github.com/kiegroup/kie-tools/pull/1663
- [ ] https://github.com/kiegroup/kie-tools/pull/1688
- [ ] Update our codebase accordingly with the API changes (Using TypeHierarchy message?)
```

yesamer:
Finally closes https://github.com/apache/incubator-kie-issues/issues/114

One year ago the JDT.LS API suddenly changed breaking the Java class "Fetching" mechanism we had in place.
We that change, it was no longer possible to retrieve a class FQCN, which is a required property for the "Fetching" mechanism to properly work. 
In the latest release, the API now supports the TypeHierarchy command: given a class name, that command can retrieve the class hierarchy and the first items represent the class itself in FQCN format. 
To restore the "Fetching" feature, we can then rely on this new command.

For example, if we use that command on a String type variable, this will return:
Screenshot 2024-06-12 at 10 46 23
The first item contains exactly what we need.

https://github.com/apache/incubator-kie-tools/assets/16005046/15f0c4bc-f772-4b9f-93af-6a427edf25c4


packages/vscode-java-code-completion-extension-plugin/vscode-java-code-completion-extension-plugin-core/src/main/resources/templates/templateClass.java
```diff
@@ -1,7 +1,7 @@
-package com.redhat;
+package org.kie;
```
tiagobento: I guess `org.apache.kie` would be more future-proof...

yesamer: Updated @tiagobento

tiagobento (approved):

tiagobento:
Thanks @yesamer! It's been a long long journey. I respect the patience and the resilience!!! 💪

jomarko (approved):
@yesamer thank you for the PR. I did not find an issue during my manual check.

PR #2454, opened 2024-06-27
kie-issues#947: Playwright expansion phase - part 2 - Decision Service node

jomarko:
This is another part of the the playwright testsuite expansion phase https://github.com/apache/incubator-kie-tools/issues/3905


packages/dmn-editor/src/draggable/Draggable.tsx
```diff
@@ -252,24 +252,27 @@ export function Draggable(props: {
...
-          onPointerLeave={() => setDraggable(false)}
-          style={props.handlerStyle}
-        >
-          <GripVerticalIcon
-            className={
-              hovered ? "kie-dmn-editor--draggable-icon-handler-hovered" : "kie-dmn-editor--draggable-icon-handler"
-            }
-          />
-        </Icon>
+        <div data-testid={"kie-dmn-editor--draggable-icon"}>
```
ljmotta: We should use `kie-tools--dmn-editor--` prefix for `data-testid`. Please review all `data-testids`

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
```diff
@@ -163,13 +163,15 @@ export function DecisionServiceProperties({
 
       <FormGroup label="Output decisions">
         <DecisionServiceElementList
+          elementListId={"kie-tools--dmn-editor--decision-service-output-decisions"}
```
ljmotta: Please, use `elementListDataTestId` to avoid ambiguity. The same applies for all occurences.

packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/decisionServicePropertiesPanel.ts
```diff
@@ -120,4 +120,71 @@ export class DecisionServicePropertiesPanel extends PropertiesPanelBase {
...
+      await this.panel()
+        .getByTestId("kie-tools--dmn-editor--decision-service-input-decisions")
+        .getByTestId("kie-dmn-editor--draggable-children")
+        .allTextContents()
+    ).map((content) => content.trim());
+  }
+
+  public async moveInputDecision(args: { nth: number; way: "up" | "down" }) {
+    // TODO
+  }
```
ljmotta: It's a leftover? The `moveInputDecision` will be left for another PR?

packages/dmn-editor/tests-e2e/drgElements/modelDecisionService.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+      await diagram.resetFocus();
+    });
+
+    test("Decision Service Signature - Output Decisions", async ({
+      decisionServicePropertiesPanel,
+      diagram,
+      nodes,
+      palette,
+    }) => {
+      // https://github.com/apache/incubator-kie-issues/issues/663
```
ljmotta: This kind of information is useful on the test report, and we can add Annotations to the report with `test.info().annotations.push(...)`. The same applies to the other comments.

packages/dmn-editor/tests-e2e/drgElements/modelDecisionService.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+        // !!! CURRENTLY PUSHED ONE OF MANY ATTEMPTS
+        await decisionServicePropertiesPanel.moveInputData({ nth: 0, way: "down" });
+        await decisionServicePropertiesPanel.moveInputDecision({ nth: 0, way: "down" });
+        expect(await decisionServicePropertiesPanel.getInvokingThisDecisionServiceInFeel()).toEqual(
+          "New Decision Service(A, B, AA, BB)"
+        );
+        expect(await decisionServicePropertiesPanel.getInputDecisions()).toEqual(["A", "B"]);
+        expect(await decisionServicePropertiesPanel.getInputData()).toEqual(["AA", "BB"]);
+      });
+    });
```
ljmotta: What isn't specifically working? The re-order by clicking and draging?

packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/decisionServicePropertiesPanel.ts
```diff
@@ -120,4 +120,70 @@ export class DecisionServicePropertiesPanel extends PropertiesPanelBase {
...
+  public async getInputDecisions() {
+    return (
+      await this.panel()
+        .getByTestId("kie-tools--dmn-editor--decision-service-input-decisions")
+        .getByTestId("kie-dmn-editor--draggable-children")
+        .allTextContents()
+    ).map((content) => content.trim());
+  }
+
+  public async moveInputDecision(args: { nth: number; way: "up" | "down" }) {
```
ljmotta: This type of reorder will enable to move only one up/down. Why not using `from/to`?
I mean, we can have a case like:
1. InputData-1
2. InputData-2
3. InputData-3
4. InputData-4

And reorder to:
1. InputData-1
2. InputData-4
3. InputData-2
4. InputData-3

ljmotta (changes requested):
@jomarko Thanks for this PR. The tests are looking very good. I've made some comments inline.

packages/dmn-editor/src/draggable/Draggable.tsx
jomarko: done

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
jomarko: sure, good suggestion, done

packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/decisionServicePropertiesPanel.ts
jomarko: yes, leftover, implemented now

packages/dmn-editor/tests-e2e/drgElements/modelDecisionService.spec.ts
jomarko: done 6f6aabd63f

packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/decisionServicePropertiesPanel.ts
jomarko: good suggestion, done 317ef019cf

jomarko:
Feedback incorporated

ljmotta (commented):
@jomarko Thanks for addressing the comments. I've just one regarding the annotation option. If the test ensures the problem was solved, it should be a `REGRESSION`, if the test is affected by a bug (e.g. not working as expected) it should be the `AFFECTED_BY`, and if the test has a workaround because of another issue (it can be a bug too, but we add a workaround step) it should be `WORKAROUND_DUE_TO`. For this case, I guess it should be "REGRESSION"?

packages/dmn-editor/tests-e2e/drgElements/modelDecisionService.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+        await page.keyboard.press("ArrowDown");
+      }
+
+      await nodes.select({ name: DefaultNodeName.DECISION_SERVICE, position: NodePosition.TOP });
+      await decisionServicePropertiesPanel.open();
+      expect(await decisionServicePropertiesPanel.getOutputDecisions()).toEqual("New Decision");
+
+      await expect(diagram.get()).toHaveScreenshot("move-decision-in-decision-service-by-keyboard.png");
+    });
+  });
```
ljmotta: No problem with the test name choice here. But please, stick with one name pattern for the entire suite (decision-service). You can change the others if you think it's better, but remember that the name should be unique and the `describe` blocks should help us filter a type of test. Revisit the previous suites if needed.

jomarko: Thank you, I tried to incorporate the feedback. Please let me know @ljmotta .

ljmotta (approved):
@jomarko Thanks a lot for your patience! PR is looking great.

packages/dmn-editor/tests-e2e/drgElements/resizeDecisionService.spec.ts
```diff
@@ -304,5 +304,28 @@ test.describe("Resize node - Decision Service", () => {
...
+      palette,
+    }) => {
+      test.info().annotations.push({
+        type: TestAnnotations.REGRESSION,
+        description: "https://github.com/apache/incubator-kie-issues/issues/881",
+      });
+      await diagram.resetFocus();
+      await palette.dragNewNode({ type: NodeType.DECISION, targetPosition: { x: 400, y: 100 } });
+      await diagram.resetFocus();
+      // move into Decision Service
```
klbowers, line 320: Is this comment needed? I feel like most of the tests you've added in this PR move a node into a Decision Service

packages/dmn-editor/tests-e2e/drgElements/resizeDecisionService.spec.ts
```diff
@@ -304,5 +304,28 @@ test.describe("Resize node - Decision Service", () => {
...
+      decisionServicePropertiesPanel,
+      diagram,
+      nodes,
+      palette,
+    }) => {
+      test.info().annotations.push({
+        type: TestAnnotations.REGRESSION,
+        description: "https://github.com/apache/incubator-kie-issues/issues/881",
+      });
+      await diagram.resetFocus();
```
klbowers: Do we need to reset at the start of a test?

packages/dmn-editor/tests-e2e/drgElements/modelDecisionService.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+        nodes,
+        palette,
+      }) => {
+        test.skip(true, "https://github.com/apache/incubator-kie-issues/issues/896");
+        test.info().annotations.push({
+          type: TestAnnotations.AFFECTED_BY,
+          description: "https://github.com/apache/incubator-kie-issues/issues/896",
+        });
+
+        await diagram.resetFocus();
```
klbowers: Do we need to reset at the start of a test?

klbowers (approved):
Thanks for the PR Jozef, LGTM! 
Added a few nitpick comments inline, just wanted to clarify if there was a reason for resetting the focus at the start of the test, as mostly you seem to drag the new node and then reset focus.

packages/dmn-editor/tests-e2e/drgElements/resizeDecisionService.spec.ts
jomarko: I would like to keep this comment for now, currently we have API that works more with coordinates. I mean, after a while, reading the test line, `await nodes.move({ name: DefaultNodeName.DECISION, targetPosition: { x: 200, y: 200 } });` may be not so clear the coordinates are inside a decision service node.

If we, in future introduce playwright fixture method like: `nodes.moveInto({name, targetNodeName})` then for sure we can remove this comment. What do you think @kbowers-ibm ?

packages/dmn-editor/tests-e2e/drgElements/resizeDecisionService.spec.ts
jomarko: well spotted, removing

packages/dmn-editor/tests-e2e/drgElements/modelDecisionService.spec.ts
jomarko: well spotted, removing

packages/dmn-editor/tests-e2e/drgElements/resizeDecisionService.spec.ts
danielzhe: I know you didn't ask for me, but since I'm here: I had the same "issue" with some tests I did. I think that `nodes.moveInto({name, targetNodeName})` would be a great idea but for now, since we have another test doing exactly the same thing, I think we should leave it as it is. Also, it can be very tricky to do a `nodes.moveInto({name, targetNodeName})`, because consider the scenarios when we have more than one node inside, different sizes, and so on. Probably the user of `nodes.moveInto({name, targetNodeName})` would be expected that the node is moved in a good position with no overlap.

klbowers: Yeah sure makes sense, twas just a minor thing!

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
```diff
@@ -193,6 +196,7 @@ export function DecisionServiceProperties({
       </FormGroup>
       <FormGroup label="Input data">
         <DraggableDecisionServiceElementList
+          elementListDataTestId={"kie-tools--dmn-editor--decision-service-input-data"}
```
danielzhe: I don't think it is a good practice to add the data-testid into the logic of the component by adding a new parameter, the `elementListDataTestId`.

What you can do is to add `data-testid` directly in the `FormGroup`:

`<FormGroup label="Input data" data-testid={"kie-tools--dmn-editor--decision-service-input-data"}>`

And then remove the `elementListDataTestId` from the `DraggableDecisionServiceElementList` and its nested `<ul>`.

If you need the list items inside the `"kie-tools--dmn-editor--decision-service-input-data"`, you can do something like this at the end:
`getByRole("listitem")`

Also, please, check the `DecisionServiceElementList` for the same thing.

danielzhe (changes requested):

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
jomarko: good suggestion @danielzhe , I like it, will incorporate it

jomarko: @danielzhe changes pushed d9500cfff1

tiagobento: Great catch @danielzhe and thanks for applying the suggestion @jomarko. Way to go!

danielzhe (approved):
Thank you! Great improvement to our test coverage!

PR #2455, opened 2024-06-27
kie-issues#1352: Enforce reproducible build on kie-tools

### LINKED ISSUE #1352: Enforce reproducible build on kie-tools
Enforce reproducible build on kie-tools

pefernan:
Closes https://github.com/apache/incubator-kie-issues/issues/1352


packages/extended-services-java/install.js
```diff
@@ -20,8 +20,11 @@
 const buildEnv = require("./env");
 const { setup } = require("@kie-tools/maven-config-setup-helper");
 
+console.log(buildEnv.env.mavenBase);
+
```
tiagobento: Oops

packages/maven-base/not-reproducible-plugins.properties
```diff
@@ -0,0 +1,81 @@
...
+
+org.codehaus.plexus+plexus-component-metadata=2.1.0
+# https://github.com/codehaus-plexus/plexus-containers/issues/27
+org.codehaus.mojo+jaxb2-maven-plugin=fail:https://github.com/mojohaus/jaxb2-maven-plugin
+#
+org.codehaus.mojo+properties-maven-plugin=1.1.0
+
+# DISABLING SINCE https://github.com/mojohaus/flatten-maven-plugin/issues/256 is completed with no change required
+# https://github.com/mojohaus/properties-maven-plugin/pull/75
+# org.codehaus.mojo+flatten-maven-plugin=fail:https://github.com/mojohaus/flatten-maven-plugin/issues/256
```
tiagobento: Would be great to make this link to an issue describing that we can safely remove this mechanism once we have a new version of maven-artifact-plugin supported by the Quarkus version we're using (next LTS after 3.9). Can you please create this issue and link it here?

tiagobento: Also, I think it's fair to put a disclaimer at the beginning saying where this file originated from, and what are the modifications we're making to it.

packages/maven-base/pom.xml
```diff
@@ -219,4 +219,43 @@
...
     </plugins>
   </build>
+
+  <profiles>
+    <profile>
+      <id>reproducible-build</id>
+      <activation>
+        <property>
+          <name>reproducible</name>
+        </property>
```
tiagobento, line 236: Can we change the activation to
```
<activation>
      <activeByDefault>false</activeByDefault>
</activation>
```
    
and use -P instead of -D to activate it on packages? I'm being extra careful here because managing profiles on a env-var-parameterized repo can become messy really quickly. I respect the extra care you had to only add it on `build:prod` scripts, but I'd love if we could do this extra step too.

packages/maven-base/pom.xml
```diff
@@ -219,4 +219,43 @@
...
+                <id>compare</id>
+                <goals>
+                  <goal>compare</goal>
+                </goals>
+                <!-- The execution's configuration is part of the pluginManagement. This piece here only makes sure the
+                     execution is enabled (by specifying a phase) for full profile builds. -->
+                <phase>install</phase>
+              </execution>
+            </executions>
+          </plugin>
```
tiagobento: @pefernan Wouldn't it be possible to configure `not-reproducible-plugins.properties` by default here? I mean, I get the flexibility that the build-env property gives, but maybe we need it to be centralized here, so that EVERY invocation of the `reproducible` profile on Maven-based packages will use that same exact file. WDYT?

Also, if for some reason a package needs to override that, packages can always setup <properties> on their pom.xml files, no?

If that's not possible, I think I'd simply rename the build-env prop to 
```ts
${buildEnv.env.mavenBase.reproducibleBuilds.notReproduciblePluginsPropertiesFilePathRelativeToInvocation
```

packages/maven-base/pom.xml
```diff
@@ -219,4 +219,43 @@
       </plugin>
     </plugins>
   </build>
+
+  <profiles>
+    <profile>
+      <id>reproducible-build</id>
```
tiagobento, line 231:
```suggestion
      <id>reproducible</id>
```

?

tiagobento: See my other comment on this file...

packages/maven-base/env/index.js
```diff
@@ -30,6 +31,9 @@ module.exports = composeEnv([require("@kie-tools/root-env/env")], {
     return {
       mavenBase: {
         version: require("../package.json").version,
+        reproducibleBuildIssues: path.resolve(
+          "./node_modules/@kie-tools/maven-base/not-reproducible-plugins.properties"
+        ),
```
tiagobento: I'm not 100% sold on this being here, especially with the `path.resolve` that makes it dependent on the caller's invocation location. See my other comments about it...

tiagobento (changes requested):
Hi @pefernan! Amazing solution. Glad to see you digging deep. I left some comments inline, especially trying to make it even more concise. Love the `README.md` notes and please don't forget to update them after incorporating changes.

tiagobento:
btw @gitgabrio this being in the `build:prod` script will make all builds reproducibility checked on all invocations of `build:prod` anywhere (including the CI and PR checks) 👍

packages/maven-base/not-reproducible-plugins.properties
gitgabrio: @tiagobento - about upgrading maven-artifact-plugin, pls keep in mind that it will be an extremely low priority, unless something else forces us to do so. Even with the new quarkus version, I see no reason to go over and change it risking to break something; again, unless we are forcfed to by something else.

packages/maven-base/pom.xml
gitgabrio: @tiagobento about -P and -D: I personally have no strong opinion (TBH, I also prefer the -P approach), but, as we agreed upon about other topic, we must consider all our codebase as a single unit, and strive for consistency. For some reason, the `-D`approach is used elsewhere, so I think it is better to enforce this kind of consistency. Needless to say that consistency and clear conventions is the key to avoid a manually crafted, long-time consuming maintenance (in that specific case, I think we must get to the point that all such cloned/copied stuff lives in one single place, and note duplicated over and over)

packages/maven-base/pom.xml
gitgabrio: @pefernan I agree with @tiago that somehow the `not-reproducible-plugins.properties` override should be go there: usually, in maven, whatever is a command line parameter can go inside the plugin configuration

packages/maven-base/pom.xml
gitgabrio: @tiagobento this is exactly the same config used all over other repositories, included the `reproducible-build` id. Again, it is not important one way or the other, but it is important tha we have the same configurations all over our codebase: I already stumbled on the fact that the `full` flag has a slightly different meaning between some repository, and I want to avoid to increase such kind of entropy. Moreover, if, ever, we'll get to an easy and simple CI configuration, having the same profile named same way all over our code would be surely beneficial.

packages/maven-base/pom.xml
pefernan: @gitgabrio @tiagobento yea, actually that was my first idea but couldn't make it properly work. Going to rethink it.

pefernan: @gitgabrio @tiagobento already found an alternative way of configuring it in the pom.xml, everything will be cleaner now :D

packages/maven-base/not-reproducible-plugins.properties
pefernan: @tiagobento I created https://github.com/apache/incubator-kie-issues/issues/1371 as a reminder to do the upgrade and removal of this setup.

pefernan:
@tiagobento in this case, just to keep the consistency with other repos, I'll +1 to @gitgabrio proposal to keep the profile as it is. But if there's a problem in CI I won't fight to change it

packages/maven-base/pom.xml
gitgabrio: @pefernan 
I think @tiagobento  is right about the

`<activeByDefault>false</activeByDefault>`

config

gitgabrio (changes requested):
Hi @pefernan Thanks for all your effort in this PR.
I agree with @tiagobento  to add the `activeByDefault` to false, even if it should be already implicitly so

gitgabrio (approved):
Great jog @pefernan !!!

tiagobento (approved):

PR #2456, opened 2024-06-28
kie-issues#1369: Create E2E tests API and some test cases for the DMN Boxed Expression Editor

### LINKED ISSUE #1369: Create E2E tests API and some test cases for the DMN Boxed Expression Editor
**The problem:**
In DMN Boxed Expression Editor, we are using Playwright to write E2E tests, which can be a bit verbose and hard to write new tests with a lot of copy-paste of code.

For example, consider the code below to add a new entry in the Context Expression:
```
    await this.locator.getByRole("cell", { name: "ContextEntry-1 (<Undefined>)" }).nth(0).hover();
    await this.locator.getByRole("cell", { name: "ContextEntry-1 (<Undefined>)" }).nth(0).locator("svg").click();
```

Also, consider that in the example above we can change the name of the Context, so the example will not work without changing the `ContextEntry-1` to the new name.

This can become very tricky and annoying in more complex scenarios.

**The solution:**

We need an API to encapsulate the Playwright details and make it easier and clearer to perform tests in BEE.

The API should follow the fluent pattern, with the current BEE editor being the entry point.

The entry point should be a Playwright's fixture.

Some API examples:

```
// Select the Decision Table in an empty Boxed Expression Editor
await bee.selectExpressionMenu.selectDecisionTable();

// Adds an output column at the start (the rightmost)
await bee.expression.asDecisionTable().addOutputAtStart();

// Adds a row in the decision table
await bee.expression.asDecisionTable().addRow();

// Fills the cell at row 2 column 1 with the text "this is my text"
await bee.expression.asDecisionTable().cellAt({ row: 2, column: 1 }).fill("this is my text"); 
```
On-screen Result:
image

```
// Select the Context in an empty Boxed Expression Editor
await bee.selectExpressionMenu.selectContext();

// In the "Select Expression Menu" of the first entry, select Context to add a new nested context
await bee.expression.asContext().entry(0).selectExpressionMenu.selectContext();

// In the firs entry of the nested Context in the first entry of the top-level Context, select Literal
await bee.expression.asContext().entry(0).expression.asContext().entry(0).selectExpressionMenu.selectLiteral()
```
On-screen Result:
image

**Tests:**

We must cover the common and basic functionalities for every expression in the BEE.
Here is a list of the tests that should cover those features:

**1. TO ALL EXPRESSIONS:**

1. Copy, Paste, Cut:
	- [x] a. Copy and Past top-level expression
	- [x] b. Cut and Past top-level expressions

- [x] 2. Change name and type

**2. SPECIFIC TESTS**

**1. Literal**
- [x] Can fill and change the content

**2. Relation**
- [x] Can add/delete a column
- [x] Can add/delete a column at a specific position
- [x] Can add/delete a row
- [x] Context menu

**3. Context**
- [x] Can add/remove entry
- [x] Context menu

**4. Decision Table**
- [x] Can add/delete an input column
- [x] Can add/delete an input column at a specific position
- [x] Can add/delete an output column
- [x] Can add/delete an output column at a specific position
- [x] Can add/delete an annotation column
- [x] Can add/delete an annotation column at a specific position
- [x] Can change the hitpolicy (check every case)
- [x] Can add/delete a row
- [x] Context menu

**5. List**
- [x] Can add/delete a row
- [x] Context menu

**6. Invocation**
- [ ] Can change the Function Name
- [ ] Can copy/paste/reset the Function Name
- [x] Can add/change/delete a parameter
- [ ] Can select/reset parameter expression

**7. Conditional**
- [ ] Can select/change/reset "if"
- [ ] Can select/change/reset "then"
- [ ] Can select/change/reset "else"

**8. For**
- [ ] Can select/change/reset "for" variable
- [ ] Can select/change/reset "in"
- [ ] Can select/change/reset "return"

**9. Every**
- [ ] Can select/change/reset "every" variable
- [ ] Can select/change/reset "in"
- [ ] Can select/change/reset "satisfies"

**10. Some**
- [ ] Can select/change/reset "some" variable
- [ ] Can select/change/reset "in"
- [ ] Can select/change/reset "satisfies"

**11. Filter**
- [ ] Can select/change/reset the "in" expression
- [ ] Can select/change/reset the "match" expression

**3. TO EXPRESSION WITH NESTED EXPRESSIONS**
Need some more complex tests to mix expressions that allow nested expressions (Filter, Some, Every, For, Conditional, List, Context and Invocation)

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1369

This PR introduces a new Test API for BEE and some new tests.

### DIFF
--- a/packages/boxed-expression-component/src/expressions/LiteralExpression/LiteralExpression.tsx
+++ b/packages/boxed-expression-component/src/expressions/LiteralExpression/LiteralExpression.tsx
@@ -209,7 +209,7 @@ export function LiteralExpression({
   }, []);
 
   return (
-    <div className={`literal-expression`}>
+    <div className={`literal-expression`} data-testid={`literal-expression-${id}`}>
       <div className={"literal-expression-body-container"}>
         <div className={"equals-sign"}>{`=`}</div>
         <BeeTable<ROWTYPE>
--- a/packages/boxed-expression-component/src/expressions/LiteralExpression/LiteralExpression.tsx
+++ b/packages/boxed-expression-component/src/expressions/LiteralExpression/LiteralExpression.tsx
@@ -209,7 +209,7 @@ export function LiteralExpression({
   }, []);
 
   return (
-    <div className={`literal-expression`} data-testid={`literal-expression-${id}`}>
+    <div className={`literal-expression`} data-testid={`kie-tools--bee--literal-expression-${id}`}>
       <div className={"literal-expression-body-container"}>
         <div className={"equals-sign"}>{`=`}</div>
         <BeeTable<ROWTYPE>
--- a/packages/boxed-expression-component/stories/getDefaultBoxedExpressionForStories.ts
+++ b/packages/boxed-expression-component/stories/getDefaultBoxedExpressionForStories.ts
@@ -263,15 +263,15 @@ export function getDefaultBoxedExpressionForStories({
       "@_id": generateUuid(),
       if: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.
       },
       then: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.,
       },
       else: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.
       },
     };
     return conditionalExpression;
@@ -281,11 +281,11 @@ export function getDefaultBoxedExpressionForStories({
       "@_id": generateUuid(),
       return: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.
       },
       in: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.
       },
     };
     return forExpression;
@@ -295,11 +295,11 @@ export function getDefaultBoxedExpressionForStories({
       "@_id": generateUuid(),
       satisfies: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.
       },
       in: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.
       },
     };
     return someExpression;
@@ -309,11 +309,11 @@ export function getDefaultBoxedExpressionForStories({
       "@_id": generateUuid(),
       satisfies: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.
       },
       in: {
         "@_id": generateUuid(),
-        expression: undefined as any,
+        expression: undefined!, // SPEC DISCREPANCY: Starting without an expression gives users the ability to select the expression type.
       },
     };
     return everyExpression;
--- a/packages/boxed-expression-component/stories/getDefaultBoxedExpressionForStories.ts
+++ b/packages/boxed-expression-component/stories/getDefaultBoxedExpressionForStories.ts
@@ -22,15 +22,19 @@ import {
   DMN15__tOutputClause,
 } from "@kie-tools/dmn-marshaller/dist/schemas/dmn-1_5/ts-gen/types";
 import {
+  BoxedConditional,
   BoxedContext,
   BoxedDecisionTable,
+  BoxedEvery,
   BoxedExpression,
   BoxedFilter,
+  BoxedFor,
   BoxedFunction,
   BoxedInvocation,
   BoxedList,
   BoxedLiteral,
   BoxedRelation,
+  BoxedSome,
   generateUuid,
 } from "../src/api";
 import {
@@ -253,6 +257,66 @@ export function getDefaultBoxedExpressionForStories({
     };
     widthsById.set(filterExpression["@_id"]!, [FILTER_EXPRESSION_MIN_WIDTH]);
     return filterExpression;
+  } else if (logicType === "conditional") {
+    const conditionalExpression: BoxedConditional = {
+      __$$element: "conditional",
+      "@_id": generateUuid(),
+      if: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+      then: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+      else: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+    };
+    return conditionalExpression;
+  } else if (logicType === "for") {
+    const forExpression: BoxedFor = {
+      __$$element: "for",
+      "@_id": generateUuid(),
+      return: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+      in: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+    };
+    return forExpression;
+  } else if (logicType == "some") {
+    const someExpression: BoxedSome = {
+      __$$element: "some",
+      "@_id": generateUuid(),
+      satisfies: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+      in: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+    };
+    return someExpression;
+  } else if (logicType === "every") {
+    const everyExpression: BoxedEvery = {
+      __$$element: "every",
+      "@_id": generateUuid(),
+      satisfies: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+      in: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
+      },
+    };
+    return everyExpression;
   } else {
     throw new Error(`No default expression available for ${logicType}.`);
   }
--- a/packages/boxed-expression-component/tests-e2e/__fixtures__/boxedExpression.ts
+++ b/packages/boxed-expression-component/tests-e2e/__fixtures__/boxedExpression.ts
@@ -40,7 +40,7 @@ export class BoxedExpressionEditor {
   }
 
   get expression() {
-    return new ExpressionContainer(this.page.getByTestId("expression-container"), this.monaco);
+    return new ExpressionContainer(this.page.getByTestId("expression-container").nth(0), this.monaco);
   }
 
   public async pasteToSelectExpression(nth?: number) {
--- a/packages/boxed-expression-component/tests-e2e/__fixtures__/boxedExpression.ts
+++ b/packages/boxed-expression-component/tests-e2e/__fixtures__/boxedExpression.ts
@@ -40,7 +40,7 @@ export class BoxedExpressionEditor {
   }
 
   get expression() {
-    return new ExpressionContainer(this.page.getByTestId("expression-container").nth(0), this.monaco);
+    return new ExpressionContainer(this.page.getByTestId("kie-tools--bee--expression-container").nth(0), this.monaco);
   }
 
   public async pasteToSelectExpression(nth?: number) {
--- a/packages/boxed-expression-component/tests-e2e/__fixtures__/boxedExpression.ts
+++ b/packages/boxed-expression-component/tests-e2e/__fixtures__/boxedExpression.ts
@@ -17,20 +17,30 @@
  * under the License.
  */
 
-import { Locator, Page } from "@playwright/test";
+import { Page } from "@playwright/test";
 import { Monaco } from "./monaco";
 
+import { ExpressionContainer } from "../api/expressionContainer";
+import { SelectExpressionMenu } from "../api/expressions/selectExpressionMenu";
+
 export class BoxedExpressionEditor {
+  private readonly _selectExpressionMenu: SelectExpressionMenu;
+
   constructor(
     public page: Page,
     private monaco: Monaco,
     public baseURL?: string
   ) {
     this.page = page;
+    this._selectExpressionMenu = new SelectExpressionMenu(page);
   }
 
-  public async select(from: Page | Locator = this.page) {
-    await from.getByText("Select expression").click();
+  get selectExpressionMenu(): SelectExpressionMenu {
+    return this._selectExpressionMenu;
+  }
+
+  get expression() {
+    return new ExpressionContainer(this.page.getByTestId("expression-container"), this.monaco);
   }
 
   public async pasteToSelectExpression(nth?: number) {
@@ -39,96 +49,6 @@ export class BoxedExpressionEditor {
       .nth(nth ?? 0)
       .click();
     await this.page.getByRole("menuitem", { name: "Paste" }).click();
-  }
-
-  public async selectBoxedLiteral(from: Page | Locator = this.page) {
-    this.select(from);
-    await this.page.getByRole("menuitem", { name: "Literal" }).click();
-  }
-
-  public async selectBoxedContext(from: Page | Locator = this.page) {
-    this.select(from);
-    await this.page.getByRole("menuitem", { name: "Context" }).click();
-  }
-
-  public async selectDecisionTable(from: Page | Locator = this.page) {
-    this.select(from);
-    await this.page.getByRole("menuitem", { name: "Decision" }).click();
-  }
-
-  public async fillDecisionTable(args: { startAtCell: number; tableData: any[][] }) {
-    let cellNumber = args.startAtCell;
-    for (const row of args.tableData) {
-      for (const cellData of row) {
-        if (cellData === "-") {
-          cellNumber++;
-          continue;
-        }
-        await this.monaco.fill({ monacoParentLocator: this.page, content: cellData, nth: cellNumber });
-        cellNumber++;
-      }
-      cellNumber++;
-    }
-  }
-
-  public async selectRelation(from: Page | Locator = this.page) {
-    this.select(from);
-    await this.page.getByRole("menuitem", { name: "Relation" }).click();
-  }
-
-  public async fillRelation(args: { startAtCell: number; relationData: any[][] }) {
-    let cellNumber = args.startAtCell;
-    for (const row of args.relationData) {
-      for (const cellData of row) {
-        await this.monaco.fill({ monacoParentLocator: this.page, content: cellData, nth: cellNumber });
-        cellNumber++;
-      }
-    }
-  }
-
-  public async selectBoxedInvocation(from: Page | Locator = this.page) {
-    this.select(from);
-    await this.page.getByRole("menuitem", { name: "Invocation" }).click();
-  }
-
-  public async selectBoxedList(from: Page | Locator = this.page) {
-    this.select(from);
-    await this.page.getByRole("menuitem", { name: "List" }).click();
-  }
-
-  public async selectBoxedFunction(from: Page | Locator = this.page) {
-    this.select(from);
-    await this.page.getByRole("menuitem", { name: "Function" }).click();
-  }
-
-  public async selectBoxedFilter(from: Page | Locator = this.page) {
-    this.select(from);
-    await this.page.getByRole("menuitem", { name: "Filter" }).click();
-  }
-
-  public async copyFilter(from: Page | Locator = this.page) {
-    await this.page.getByTestId("logic-type-button-test-id").click();
-    await this.page.getByRole("menuitem", { name: "Copy" }).click();
-  }
-
-  public async resetFilter(from: Page | Locator = this.page) {
-    await from.getByTestId("logic-type-button-test-id").click();
-    await this.page.getByRole("menuitem", { name: "Reset" }).click();
-  }
-
-  public async fillFilter(args: { collectionIn: any[]; collectionMatch: any }) {
-    for (let i = 0; i < args.collectionIn.length; i++) {
-      await this.monaco.fill({
-        monacoParentLocator: this.page.getByTestId("kie-tools--boxed-expression-component--filter-collection-in"),
-        content: args.collectionIn[i],
-        nth: i,
-      });
-    }
-
-    await this.monaco.fill({
-      monacoParentLocator: this.page.getByTestId("kie-tools--boxed-expression-component--filter-collection-match"),
-      content: args.collectionMatch,
-    });
   }
 
   public async goto() {
--- a/packages/boxed-expression-component/tests-e2e/api/expressionContainer.ts
+++ b/packages/boxed-expression-component/tests-e2e/api/expressionContainer.ts
@@ -44,51 +44,51 @@ export class ExpressionContainer {
     return new ExpressionHeader(this.locator);
   }
 
-  asLiteral() {
+  public asLiteral() {
     return new LiteralExpressionElement(this.locator, this.monaco);
   }
 
-  asRelation() {
+  public asRelation() {
     return new RelationExpressionElement(this.locator, this.monaco);
   }
 
-  asContext() {
+  public asContext() {
     return new ContextExpressionElement(this.locator, this.monaco);
   }
 
-  asDecisionTable() {
+  public asDecisionTable() {
     return new DecisionTableExpressionElement(this.locator, this.monaco);
   }
 
-  asList() {
+  public asList() {
     return new ListExpressionElement(this.locator, this.monaco);
   }
 
-  asInvocation() {
+  public asInvocation() {
     return new InvocationExpressionElement(this.locator, this.monaco);
   }
 
-  asFunction() {
+  public asFunction() {
     return new FunctionExpressionElement(this.locator, this.monaco);
   }
 
-  asConditional() {
+  public asConditional() {
     return new ConditionalExpressionElement(this.locator, this.monaco);
   }
 
-  asFor() {
+  public asFor() {
     return new ForExpressionElement(this.locator, this.monaco);
   }
 
-  asEvery() {
+  public asEvery() {
     return new EveryExpressionElement(this.locator, this.monaco);
   }
 
-  asSome() {
+  public asSome() {
     return new SomeExpressionElement(this.locator, this.monaco);
   }
 
-  asFilter() {
+  public asFilter() {
     return new FilterExpressionElement(this.locator, this.monaco);
   }
 
@@ -96,7 +96,7 @@ export class ExpressionContainer {
     return new ContextMenu(this.locator.nth(0));
   }
 
-  async isEmpty() {
+  public async isEmpty() {
     return (await this.locator.nth(0).getByTestId("kie-tools--bee--logic-type-selected-header").count()) === 0;
   }
 }
@@ -107,15 +107,15 @@ export class ExpressionCell {
     private monaco: Monaco
   ) {}
 
-  async fill(content: string) {
+  public async fill(content: string) {
     return await this.monaco.fill({ monacoParentLocator: this.locator, content: content });
   }
 
-  get content() {
+  public get content() {
     return this.locator.nth(0);
   }
 
-  get contextMenu() {
+  public get contextMenu() {
     return new ContextMenu(this.locator);
   }
 }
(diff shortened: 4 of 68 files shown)


packages/boxed-expression-component/tests-e2e/boxedExpressions/filter/filterExpression.spec.ts
```diff
@@ -41,53 +41,51 @@ test.describe("Create Boxed Filter", () => {
...
 
-    await boxedExpressionEditor.resetFilter(page.getByTestId("expression-row-0"));
+    await bee.expression.asContext().entry(0).expression.header.reset();
 
-    await expect(boxedExpressionEditor.getContainer()).toHaveScreenshot("boxed-filter-nested-reset.png");
+    await expect(bee.getContainer()).toHaveScreenshot("boxed-filter-nested-reset.png");
   });
 
-  test("should correctly create a nested filter", async ({ boxedExpressionEditor, page, stories }) => {
+  test("should correctly create a nested filter", async ({ bee, page, stories }) => {
```
danielzhe: Your tests "translated" to the new API, @jomarko

packages/boxed-expression-component/tests-e2e/api/contextExpressionElement.ts
```diff
@@ -0,0 +1,51 @@
...
+      .getByRole("cell", { name: "ContextEntry-1 (<Undefined>)" })
+      .nth(0)
+      .hover({
+        position: {
+          x: 0,
+          y: 0,
+        },
+      });
+    await this.locator.getByRole("cell", { name: "ContextEntry-1 (<Undefined>)" }).nth(0).locator("svg").click();
+  }
```
jomarko: What if the first context entry is renamed or has non default data type already?

packages/boxed-expression-component/tests-e2e/api/decisionTableExpressionElement.ts
```diff
@@ -0,0 +1,260 @@
...
+      .hover({
+        position: {
+          x: 0,
+          y: 0,
+        },
+      });
+    await this.locator.locator(".output").nth(index).locator("svg").click();
+  }
+
+  private async addOutputFromHeaderGroupElementAtIndex(index: number) {
```
jomarko: Reading code of this class, am I correct the new API always uses the 'plus' icon for adding rows or columns? Do we use somewhere the context menu -> Insert row / Insert column .... ?

packages/boxed-expression-component/tests-e2e/api/functionExpressionElement.ts
```diff
@@ -0,0 +1,42 @@
...
+export class FunctionExpressionElement {
+  constructor(
+    private locator: Locator,
+    private monaco: Monaco
+  ) {}
+
+  async addEntry() {
+    await this.locator.getByRole("cell", { name: "1" }).nth(0).hover();
+    await this.locator.getByRole("cell", { name: "1" }).nth(0).locator("svg").click();
+  }
```
jomarko: not sure if this is correct, we can add parameters to the function, but I do not think entries

packages/boxed-expression-component/tests-e2e/api/invocationExpressionElement.ts
```diff
@@ -0,0 +1,49 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Locator } from "@playwright/test";
+import { Monaco } from "../__fixtures__/monaco";
+import { ExpressionElementEntry } from "./expressionContainer";
+
+export class InvocationExpressionElement {
```
jomarko: how to fill in `function name` 
image

packages/boxed-expression-component/tests-e2e/api/listExpressionElement.ts
```diff
@@ -0,0 +1,31 @@
+import { Locator } from "@playwright/test";
```
jomarko: missing header

packages/boxed-expression-component/tests-e2e/boxedExpressions/boxedExpressionHeader.spec.ts
```diff
@@ -36,19 +36,19 @@ test.describe("Boxed expression header", () => {
...
 
-  test("should copy, reset and paste expression", async ({ boxedExpressionEditor, page, clipboard }) => {
+  test("should copy, reset and paste expression", async ({ bee, page, clipboard }) => {
     await page.getByTestId("logic-type-button-test-id").click();
     clipboard.use();
     await page.getByRole("menuitem", { name: "copy" }).click();
     await page.getByTestId("logic-type-button-test-id").click();
     await page.getByRole("menuitem", { name: "Reset" }).click();
-    await boxedExpressionEditor.pasteToSelectExpression();
+    await bee.pasteToSelectExpression();
```
jomarko, line 45: is the rewriting of the tests using new API still in progress? asking because lines above seems also be available in the new bee API

packages/boxed-expression-component/tests-e2e/boxedExpressions/filter/filterExpression.spec.ts
jomarko: thank you, generally, is the rewriting using new API still in progress?

for example
```
  test("should rename a filter", async ({ page, stories }) => {
    await stories.openBoxedFilter("base");

    await page.getByRole("columnheader", { name: "Expression Name (<Undefined>)" }).click();
    await page.getByPlaceholder("Expression Name").fill("Filter Expression Name");
    await page.keyboard.press("Enter");

    await expect(page.getByRole("columnheader", { name: "Filter Expression Name (<Undefined>)" })).toBeVisible();
  });
```

^ I would expect this PR change some line in this code to be like `bee.asFilter()`

packages/boxed-expression-component/tests-e2e/api/filterExpressionElement.ts
```diff
@@ -0,0 +1,58 @@
...
+        content: args.collectionIn[i],
+        nth: i,
+      });
+    }
+
+    await this.monaco.fill({
+      monacoParentLocator: this.locator.getByTestId("kie-tools--boxed-expression-component--filter-collection-match"),
+      content: args.collectionMatch,
+    });
+  }
```
jomarko: Looking on one of the tests:
```
    await bee.expression.asFilter().in.selectExpressionMenu.selectList();
    await bee.expression.asFilter().in.expression.asList().addEntry();
    await bee.expression.asFilter().in.expression.asList().entry(0).selectExpressionMenu.selectLiteral();
    await bee.expression.asFilter().in.expression.asList().entry(1).selectExpressionMenu.selectLiteral();
    await bee.expression.asFilter().in.expression.asList().entry(0).expression.asLiteral().fill("Passenger One");
    await bee.expression.asFilter().in.expression.asList().entry(1).expression.asLiteral().fill("Passenger Two");

    await bee.expression.asFilter().match.selectExpressionMenu.selectLiteral();
    await bee.expression.asFilter().match.expression.asLiteral().fill("item.Flight Number = Flight.Flight Number");
```

I think this `fill` (commented using this comment) should be removed. the `fill` should be probably only for cells we are sure they are implemented using monaco, in other words, do not contains nested expressions .

jomarko (commented):
Thank you for the PR @danielzhe , I haven't go through the tests yet, rather had a look on the API itself.

packages/boxed-expression-component/tests-e2e/api/contextExpressionElement.ts
danielzhe: This is an WIP, not a finished work.

packages/boxed-expression-component/tests-e2e/api/filterExpressionElement.ts
danielzhe: We are using `fill` to fill Literal Expressions, which are, by definition, single cells where the user can write feel expressions. Can you explain again what is the issue? Sorry, I don't understand this comment.  @jomarko

packages/boxed-expression-component/tests-e2e/boxedExpressions/filter/filterExpression.spec.ts
danielzhe: Yes. I'm going to finish the API first, then write new tests.
I was just showing you an example.

packages/boxed-expression-component/tests-e2e/api/decisionTableExpressionElement.ts
danielzhe: Yes, we don't have nothing to context menu yet.

packages/boxed-expression-component/tests-e2e/api/functionExpressionElement.ts
danielzhe: We can rename it.

packages/boxed-expression-component/tests-e2e/api/invocationExpressionElement.ts
danielzhe: WIP. 
Check the ticket, there are a lot of tests that will be done and the API will evolve as they are getting implemented.
https://github.com/apache/incubator-kie-issues/issues/1369

packages/boxed-expression-component/tests-e2e/boxedExpressions/boxedExpressionHeader.spec.ts
danielzhe: Yes. It is not done. See the ticket:
https://github.com/apache/incubator-kie-issues/issues/1369

packages/boxed-expression-component/tests-e2e/api/listExpressionElement.ts
danielzhe: Thank you! Good catch!

packages/boxed-expression-component/tests-e2e/api/filterExpressionElement.ts
jomarko: yes, I do not understand why we have `fill` method in the `filterExpressionElement.ts`

packages/boxed-expression-component/src/expressions/LiteralExpression/LiteralExpression.tsx
```diff
@@ -209,7 +209,7 @@ export function LiteralExpression({
   }, []);
 
   return (
-    <div className={`literal-expression`}>
+    <div className={`literal-expression`} data-testid={`literal-expression-${id}`}>
```
ljmotta: Please, norrow the `data-testid` scope. On the new DMN Editor we're using the `kie-tools--dmn-editor--`. I think we could use `kie-tools--bee--`. WDYT? Either way, this applies to all `data-testid` from the Boxed Expression Editor.

packages/boxed-expression-component/stories/getDefaultBoxedExpressionForStories.ts
```diff
@@ -253,6 +257,66 @@ export function getDefaultBoxedExpressionForStories({
     };
     widthsById.set(filterExpression["@_id"]!, [FILTER_EXPRESSION_MIN_WIDTH]);
     return filterExpression;
+  } else if (logicType === "conditional") {
+    const conditionalExpression: BoxedConditional = {
+      __$$element: "conditional",
+      "@_id": generateUuid(),
+      if: {
+        "@_id": generateUuid(),
+        expression: undefined as any,
```
ljmotta: No problem using the `as any`, but on another locations we use the `!`. I think we must stick with one. Either way, we need to add the comment regarding the spec discrepancy.

packages/boxed-expression-component/tests-e2e/api/functionExpressionElement.ts
tiagobento: @danielzhe I think what Jozef means is that "adding entries" is not something that makes sense to the Function expression. WDYT about `addImplementationExpression`, or `addImplementation`, or even `selectImplementationExpression`?

ljmotta (commented):
Thanks for addressing this. I think tests will look very good after this PR.
I know it's a WIP, but I've added a review of some things that I think must be addressed at some time, so I'll leave it here as a reminder 😄 

### Expression
I didn't like the `ExpressionElementEntry` name. Entry is a very specific name for the Context expression (Context Entry). And this class is being used by other expressions, like the `Conditional`, `Every`, `Filter`, etc. For some of those boxed expressions (I didn't verify all of them), the spec calls the nested expression `ChildExpression`, so I think this would be a more appropriate name to avoid ambiguity. WDYT?
I don't think the `ExpressionVariable` on the Iterator expressions is good for the same reason. The spec uses `iteratorVariable` to avoid ambiguity, but at the end is a `ChildExpression` as well. Not sure which name would be more appropriate.

### Boxed Context
In the Boxed Context, we have a pair of variable/expression which together make an entry. I couldn't see a clear distinction between them on the current API. Could you provide an example?

### Decision Table
As covered in the best practices and on the [Playwright docs](https://playwright.dev/docs/other-locators), the CSS locators are very discouraged. 
I'm unsure if it was intentional, but I think the `AtRight`/`AtLeft` used on the Output columns is better than the `AtStart` and `AtEnd` used on the Input!

### Boxed Function
As mentioned by @jomarko and @tiagobento  the `addEntry` does not make much sense for the Boxed Function.

### Boxed List
Like the Boxed Function, an `addEntry` doesn't make sense for it, I guess it should be `addRow`, WDYT?

### General (WIP section)
We must improve the `data-testid` of the entire test codebase.
It's missing the `addEntry`/`addRow` from `atTop`/`atBottom` on some expressions.
I know it's a personal preference, I see some `public` methods, but none are marked as `private`. Even though TS/JS allows this I personally don't think it's considered good practice as it's not explicitly telling what the variable/method is. As I said, feel free to leave like this though.

Overall the new API looks great. I think you've done a great job putting it together! As you can see, most of the comments are regarding the name choice!

danielzhe:
@ljmotta Thank you for your review! I'm going to address the proposed changes as soon as I find out why some tests are failing only in CI 😢

danielzhe:
> I'm unsure if it was intentional, but I think the AtRight/AtLeft used on the Output columns is better than the AtStart and AtEnd used on the Input!

The question here is: at "left" or "right" of what? I think "atStart" and "atEnd" are more clear. Maybe in some cultures, the start is at the right, but still, I don't think that "atLeft" or "atRight" is better. But those who will use this API also have the option to use the `addInputAtIndex` or `addOutputAtIndex` and so on. @ljmotta

danielzhe:
> I know it's a personal preference, I see some public methods, but none are marked as private. Even though TS/JS allows this I personally don't think it's considered good practice as it's not explicitly telling what the variable/method is. As I said, feel free to leave like this though.

If we want to force some standardization that is just a "personal preference", we should force it in ESLint. Also, yes, it has private methods

ljmotta:
@danielzhe  Well, I can use the same question for `atStart`/`atEnd`. `atStart` and `atEnd` of what? 😄. Both questions are answered by the API, as for now we have `addInputAtStart`, `addInputAtEnd`, `addOutputAtRightOfIndex`, and`addOutputAtLeftOfIndex`. I just mentioned the one I preferred because I saw the Output column had a different set of methods. I wasn't sure if you were experimenting or not.

> If we want to force some standardization that is just a "personal preference", we should force it in ESLint. Also, yes, it has private methods.

EDIT:
Actually, the methods are public by default. You're not using the `public` keyword on some methods. But as I said, for this case, leave the code as you wish. It's just something that I noticed. I've made a confusion thinking the default behavior was `private` when I first made the review.

For example on the `contextExpressionElement.ts`:
```ts
  public entry(index: number) {
    return new ExpressionElementEntry(this.locator.getByTestId(`expression-row-${index}`).nth(0), this.monaco);
  }

  get resultEntry() {
    return new ExpressionElementEntry(this.locator.locator(`.additional-row`).nth(0), this.monaco);
  }

  async entriesCount() {
    return (await this.locator.getByRole("row").count()) - 2;
  }
```

danielzhe:
@tiagobento GREEN!🕺

danielzhe:
Can we get this merged or is there any pending change that I forgot? 😄 
 @tiagobento @ljmotta

packages/boxed-expression-component/src/expressions/LiteralExpression/LiteralExpression.tsx
ljmotta: This is still missing the narrowed `data-testid` scope.

packages/boxed-expression-component/stories/getDefaultBoxedExpressionForStories.ts
ljmotta: Here is still with `as any`, making it different than the other expressions that use the `undefined!`

packages/boxed-expression-component/tests-e2e/__fixtures__/boxedExpression.ts
```diff
@@ -17,20 +17,30 @@
...
+  }
+
+  get selectExpressionMenu(): SelectExpressionMenu {
+    return this._selectExpressionMenu;
   }
 
-  public async select(from: Page | Locator = this.page) {
-    await from.getByText("Select expression").click();
+  get expression() {
+    return new ExpressionContainer(this.page.getByTestId("expression-container").nth(0), this.monaco);
```
ljmotta: It's using a `data-testid` without the narrowed scope.

packages/boxed-expression-component/tests-e2e/api/expressions/contextExpressionElement.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  public entry(index: number) {
+    return new ContextExpressionEntry(
+      this.locator.getByTestId(`kie-tools--bee--expression-row-${index}`).nth(0),
+      this.monaco
+    );
+  }
+
+  get result() {
+    return new ChildExpression(this.locator.locator(`.additional-row`).nth(0), this.monaco);
```
ljmotta: As mentioned on the previous review we should avoid class locators. Can't we use a `data-testid` for this case?

packages/boxed-expression-component/tests-e2e/api/expressions/decisionTableExpressionElement.ts
ljmotta, line 1: This file has several `class` locators. The previous comment applies for this entire file. Why can't we use `data-testid` locators?

packages/boxed-expression-component/tests-e2e/api/expressions/filterExpressionElement.ts
ljmotta, line 1: The `data-testid` here uses another pattern. It uses `kie-tools--boxed-expression-component--` instead of `kie-tools--bee--` of another files. Please stick with one.

packages/boxed-expression-component/tests-e2e/api/expressions/decisionTableExpressionElement.ts
```diff
@@ -0,0 +1,327 @@
...
+          position: {
+            x: 0,
+            y: 0,
+          },
+        });
+      await this.locator.locator(".input").nth(index).locator("svg").click();
+    }
+  }
+
+  async addRowAtTop() {
```
ljmotta: Why we don't have a `addRowAtTopOfIndex`?

packages/boxed-expression-component/tests-e2e/api/expressionContainer.ts
```diff
@@ -0,0 +1,182 @@
...
+  asFilter() {
+    return new FilterExpressionElement(this.locator, this.monaco);
+  }
+
+  get contextMenu() {
+    return new ContextMenu(this.locator.nth(0));
+  }
+
+  async isEmpty() {
+    return (await this.locator.nth(0).locator(">.logic-type-not-present").count()) === 1;
```
ljmotta: `class` locator

packages/boxed-expression-component/tests-e2e/api/expressionContainer.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  button(option: string) {
+    return this.locator.page().locator(".context-menu-container").getByRole("button", { name: option, exact: true });
+  }
+
+  radio(option: string) {
+    return this.locator.page().locator(".context-menu-container").getByRole("radio", { name: option, exact: true });
+  }
+}
```
ljmotta: same.

packages/boxed-expression-component/tests-e2e/api/expressionContainer.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  get expression() {
+    return this._expression;
+  }
+
+  get selectExpressionMenu() {
+    return new SelectExpressionMenu(this.locator.getByTestId("expression-container").nth(0));
+  }
+}
```
ljmotta: missing narrowed scope

packages/boxed-expression-component/tests-e2e/api/nameAndDataTypeCell.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  async getName() {
+    return await this.locator.locator(".expression-info-name").nth(0).innerText();
+  }
+
+  async getDataType() {
+    return await this.locator.locator(".expression-info-data-type").nth(0).innerText();
+  }
```
ljmotta: `class` locators.

packages/boxed-expression-component/tests-e2e/boxedExpressions/context/contextExpression.spec.ts
```diff
@@ -17,17 +17,17 @@
...
     await expect(page.getByRole("columnheader", { name: "Expression Name (<Undefined>)" })).toBeAttached();
     await expect(page.getByRole("cell", { name: "ContextEntry-1 (<Undefined>)" })).toBeAttached();
     await expect(page.getByRole("cell", { name: "<result>" })).toBeAttached();
     await expect(page.getByText("Select expression")).toHaveCount(2);
     await expect(page.getByRole("columnheader")).toHaveCount(1);
     await expect(page.getByRole("cell")).toHaveCount(4);
-    await expect(boxedExpressionEditor.getContainer()).toHaveScreenshot("boxed-context.png");
+    await expect(bee.getContainer()).toHaveScreenshot("boxed-context.png");
   });
 });
```
ljmotta, line 33: It's not using the new API.

packages/boxed-expression-component/tests-e2e/boxedExpressions/context/contextMenu.spec.ts
```diff
@@ -17,122 +17,130 @@
...
 
-      await page.getByTestId("monaco-container").click({ button: "right" });
-      await expect(page.getByRole("heading", { name: "CONTEXT ENTRY" })).not.toBeAttached();
-      await expect(page.getByRole("heading", { name: "SELECTION" })).toBeAttached();
-      await expect(page.getByRole("heading", { name: "COLUMNS" })).not.toBeAttached();
+      await expect(expressionHeaderCell.contextMenu.heading("CONTEXT ENTRY")).not.toBeAttached();
+      await expect(expressionHeaderCell.contextMenu.heading("SELECTION")).toBeAttached();
+      await expect(expressionHeaderCell.contextMenu.heading("COLUMNS")).not.toBeAttached();
+
+      const entry = bee.expression.asContext().entry(0);
```
ljmotta: We couldn't create only one variable for this case? On the start of the test `const contextExpression = bee.expression.asContext()`.

packages/boxed-expression-component/tests-e2e/boxedExpressions/context/populate.spec.ts
```diff
@@ -17,14 +17,14 @@
  * under the License.
  */
 
-import { test, expect } from "../../__fixtures__/base";
+import { expect, test } from "../../__fixtures__/base";
 import { TestAnnotations } from "@kie-tools/playwright-base/annotations";
 
 test.describe("Populate Boxed Context", () => {
   test("should correctly create pre-bureau risk category boxed context", async ({
```
ljmotta, line 24: It's not using the new API.

packages/boxed-expression-component/tests-e2e/boxedExpressions/decisionTable/decisionTableExpression.spec.ts
```diff
@@ -17,28 +17,24 @@
...
-import { test, expect } from "../../__fixtures__/base";
+import { expect, test } from "../../__fixtures__/base";
 
 test.describe("Create Decision table", () => {
   test.beforeEach(async ({ stories }) => {
     await stories.openDecisionTable();
   });
 
-  test("should render expression correctly", async ({ boxedExpressionEditor, page }) => {
+  test("should render expression correctly", async ({ bee, page }) => {
```
ljmotta, line 27: It's not using the new API.

packages/boxed-expression-component/tests-e2e/boxedExpressions/decisionTable/contextMenu.spec.ts
ljmotta, line 1: Each test creates a few different variables. Why not use `const decisionTable = bee.expression.asDecisionTable()` at the top of the test, and the operations and assertions use this `decisionTable`?

packages/boxed-expression-component/tests-e2e/boxedExpressions/filter/populate.spec.ts
```diff
@@ -20,7 +20,7 @@
 import { test, expect } from "../../__fixtures__/base";
 
 test.describe("Populate Boxed Filter", () => {
-  test("should correctly create a Rebooked Flights filter", async ({ stories, page, boxedExpressionEditor }) => {
+  test("should correctly create a Rebooked Flights filter", async ({ stories, page, bee }) => {
```
ljmotta, line 23: It's not using the new API.

packages/boxed-expression-component/tests-e2e/boxedExpressions/filter/filterExpression.spec.ts
```diff
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-import { test, expect } from "../../__fixtures__/base";
+import { expect, test } from "../../__fixtures__/base";
 
 test.describe("Create Boxed Filter", () => {
   test("should rename a filter", async ({ page, stories }) => {
```
ljmotta, line 23: It's not using the new API.

ljmotta (changes requested):
I've made my second round of reviews. Some of them are inline, and I've some general comments.

### Fixtures
Starting with the expressions. One general thing that I've noticed, that I didn't pay attention to at the first time, is the usage of the `locator` to refer to the `expression` itself, which I don't think is the best option, as `locator` is a way to identify any page element, and reading the code can cause ambiguity. For instance, we have a few `this.locator.locator`, which could be `this.expression.locator` or `this.expressionLocator`. But I'll let for @tiagobento to decide.

Almost all expressions have a getter `nameAndDataTypeCell`. The context expression is the only exception, which uses the `expressionHeaderCell` which in my opinion is a better suited name. The function expression is missing this getter, I'm not sure why.

Maybe I've misunderstood, but I thought we'd agreed on chat that it is best to have the `public` identifiers. Some methods still lack them, are we proceeding without them?

### Tests
I could see that some tests were not converted to the new API. Additionally, I could see that some tests create a very specific variable, and some tests don't create the variable at all. Looking at it, I think the idea of creating an expression variable on top of each test can help the legibility, for example, `decisionTable`, `context`, etc. But some time ago we decided for not to have any variable at all, so I'll leave this for @tiagobento to decide.

ljmotta (changes requested):
EDIT: oops, I've made a duplicated request change.

tiagobento:
@ljmotta The idea of not having variables was to prevent tests from becoming complicated. We wanted to have testes that could be read from start to finish without having to jump around too much to understand what the test case is. I see on this PR we have some variables already, but I think I'd go for a single variable at the top and then inlining stuff at the individual lines.. Example:

```typescript
    test("should open decision rules context menu and insert multiples rows below", async ({ bee }) => {
      const decisionTable = bee.expression.asDecisionTable()
      await decisionTable.cellAt({ row: 1, column: 1 }).fill("test");
      await decisionTable.cellAt({ row: 1, column: 0 }).contextMenu.open();
      await decisionTable.cellAt({ row: 1, column: 0 }).contextMenu.option("Insert").click();
      await decisionTable.cellAt({ row: 1, column: 0 }).contextMenu.button("minus").click();
      await decisionTable.cellAt({ row: 1, column: 0 }).contextMenu.radio("Below").click();
      await decisionTable.cellAt({ row: 1, column: 0 }).contextMenu.button("Insert").click();

      await expect(decisionTable.cellAt({ row: 1, column: 1 }).content).toContainText("test");
      await expect(decisionTable.cellAt({ row: 2, column: 0 }).content).toBeAttached();
    });
```

tiagobento:
I'll take a closer look at the `locator` thing in a bit...

packages/boxed-expression-component/tests-e2e/api/expressions/decisionTableExpressionElement.ts
danielzhe: Because it was not used in any test and I didn't notice I didn't do it.

packages/boxed-expression-component/tests-e2e/api/expressions/filterExpressionElement.ts
danielzhe: It was a copy-paste from a previous test, but I'll change it to `kie-tools--bee--`

danielzhe:
@ljmotta Changes applied.

ljmotta (approved):
@danielzhe Thanks for your patience. LGTM!

danielzhe:
> @danielzhe Thanks for your patience. LGTM!

Thank YOU for your patience, @ljmotta 😄

danielzhe:
@tiagobento Ready!

tiagobento (approved):

PR #2457, opened 2024-07-01
kie-issues#1285: Remove unnecessary base image from quarkus examples (quay.io/kiegroup/kogito-runtime-jvm:latest)

fantonangeli:
**Closes:** https://github.com/apache/incubator-kie-issues/issues/1285

This PR is related to:
- https://github.com/apache/incubator-kie-kogito-examples/pull/1978

### DIFF
--- a/packages/sonataflow-operator/bddframework/pkg/framework/util.go
+++ b/packages/sonataflow-operator/bddframework/pkg/framework/util.go
@@ -51,7 +51,7 @@ const (
 	// defaultBuilderImage Builder Image for Kogito
 	defaultBuilderImage = "kogito-s2i-builder"
 	// defaultRuntimeJVM Runtime Image for Kogito with  JRE
-	defaultRuntimeJVM = "kogito-runtime-jvm"
+	defaultRuntimeJVM = "registry.access.redhat.com/ubi9/openjdk-17"
 	//defaultRuntimeNative Runtime Image for Kogito for Native Quarkus Application
 	defaultRuntimeNative = "kogito-runtime-native"
 	// imageRegistryEnvVar ...
@@ -292,7 +292,11 @@ func AppendImageDefaultValues(image *api.Image) {
 	}
 
 	if len(image.Tag) == 0 {
-		image.Tag = GetKogitoImageVersion(version.GetTagVersion())
+		if image.Name == defaultRuntimeJVM {
+			image.Tag = GetKogitoImageVersion(version.GetOpenJDKImageVersion())
+		} else {
+			image.Tag = GetKogitoImageVersion(version.GetTagVersion())
+		}
 	}
 }
 
--- a/packages/sonataflow-operator/bddframework/pkg/framework/util.go
+++ b/packages/sonataflow-operator/bddframework/pkg/framework/util.go
@@ -293,7 +293,7 @@ func AppendImageDefaultValues(image *api.Image) {
 
 	if len(image.Tag) == 0 {
 		if image.Name == defaultRuntimeJVM {
-			image.Tag = GetKogitoImageVersion(version.GetOpenJDKImageVersion())
+			image.Tag = GetKogitoImageVersion(version.GetOpenJDKImageTagVersion())
 		} else {
 			image.Tag = GetKogitoImageVersion(version.GetTagVersion())
 		}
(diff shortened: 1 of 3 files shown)


packages/sonataflow-operator/bddframework/pkg/framework/util.go
```diff
@@ -292,7 +292,11 @@ func AppendImageDefaultValues(image *api.Image) {
 	}
 
 	if len(image.Tag) == 0 {
-		image.Tag = GetKogitoImageVersion(version.GetTagVersion())
+		if image.Name == defaultRuntimeJVM {
+			image.Tag = GetKogitoImageVersion(version.GetOpenJDKImageVersion())
+		} else {
+			image.Tag = GetKogitoImageVersion(version.GetTagVersion())
+		}
```
tiagobento: `defaultRuntimeJVM` is a constant, so the `else` branch of this `if` statement is basically dead code...

packages/sonataflow-operator/version/version.go
```diff
@@ -33,8 +33,15 @@ const (
 	tagVersion = "main"
 	// Kogito images tag version. Used for data-index and jobs-service images.
 	kogitoImagesTagVersion = "999-20240623"
+	// OpenJDK image version
+	openJDKImageVersion = "1.20"
```
tiagobento:
```suggestion
	openJDKImageTagVersion = "1.20"
```

I learned from @ricardozanini that `tag` is the entire thing, like `docker.io/apache/something:1.1.1`. In this case, tagVersion would be `1.1.1`.

packages/sonataflow-operator/version/version.go
```diff
@@ -33,8 +33,15 @@ const (
...
 )
 
+// GetOpenJDKImageVersion gets the current OpenJDK image version.
+func GetOpenJDKImageVersion() string {
```
tiagobento:
```suggestion
func GetOpenJDKImageTagVersion() string {
```

tiagobento (changes requested):
Thanks for addressing this @fantonangeli !

packages/sonataflow-operator/bddframework/pkg/framework/util.go
fantonangeli: @tiagobento are you sure the `else` is dead code? 
I think `AppendImageDefaultValues()` is called with a different `image.Name` here:
https://github.com/apache/incubator-kie-tools/blob/main/packages/sonataflow-operator/bddframework/pkg/framework/kogitoserviceutils.go#L131

packages/sonataflow-operator/version/version.go
fantonangeli: @tiagobento I used `1.20` as a value following other places in kie-tools and in the page below I can see it is a real tag:
https://catalog.redhat.com/software/containers/ubi9/openjdk-17/61ee7c26ed74b2ffb22b07f6/history
Anyway I can change it to `1.20-2` or `1.20-2.1719294794` of course.

packages/sonataflow-operator/bddframework/pkg/framework/util.go
tiagobento: You're right, I apologize. Didn't realize this was part of a more generic method.

tiagobento (approved):

domhanak (approved):

domhanak (changes requested):
In the referenced PR from examples I see;
`registry.access.redhat.com/ubi9/openjdk-17:1.18`

but in this PR I see:
`registry.access.redhat.com/ubi9/openjdk-17:1.20`

Please check, imho they should match

fantonangeli:
@domhanak 
I used the "1.20" to keep consistency with other places in `kie-tools`, which uses "1.20", but I can change this PR to "1.18".

Otherwise, we need to update the version in `kogito-examples` where it is used in 18 places.
WDYT?

tiagobento:
If we can upgrade to a newer version, then I think we should... @fantonangeli @domhanak... +1 for consistency too.

domhanak:
@fantonangeli imho we update also examples if no issues, we should aim having this consistent

fantonangeli:
@tiagobento @domhanak good, I will prepare the kogito-example update PR

tiagobento (approved):

fantonangeli:
`kogito-examples` update PR: https://github.com/apache/incubator-kie-kogito-examples/pull/1980

domhanak (approved):

PR #2462, opened 2024-07-08
kie-issues#947: Playwright expansion phase - part 3 - DRDs

jomarko:
This PR introduces tests for manipulation of DRDs:
- create DRD
- rename DRD
- delete DRD

This PR introduces also test for interaction with a DRD, and test interactions with nodes that are placed in multiple DRDs at the same time.

This PR furthermore closes also: https://github.com/apache/incubator-kie-issues/issues/1174

This PR contains a preparation for **part 4**, where we will focus on Decision Services nodes being used in DRDs. So please ignore changes in the file `modelDecisionServicesInDrds.spec.ts`


packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
```diff
@@ -163,15 +163,15 @@ export class Nodes {
...
+  public async select(args: { name: string; position?: NodePosition; dblClick?: boolean }) {
     const node = this.get({ name: args.name });
 
     const position =
       args.position !== undefined
         ? await this.getPositionalNodeHandleCoordinates({ node, position: args.position })
         : undefined;
 
-    await node.click({ position });
+    args.dblClick ? await node.dblclick({ position }) : await node.click({ position });
```
ljmotta: To select a node only one click is required. Why adding the option to select with double click?

packages/dmn-editor/tests-e2e/drds/modelDecisionServicesInDrds.spec.ts
ljmotta, line 1: I guess this file is still a WIP?

packages/dmn-editor/src/diagram/Palette.tsx
```diff
@@ -221,11 +225,16 @@ export function Palette({ pulse }: { pulse: boolean }) {
...
+            <div
+              data-testid={"kie-tools--dmn-editor--palette-nodes-popover"}
+              className={"kie-dmn-editor--palette-nodes-popover"}
+              style={{ maxHeight }}
+            >
               <DrgNodesPanel />
             </div>
           )}
           <button
+            data-testid={"kie-tools--dmn-editor--drg-panel-toggle-button"}
```
ljmotta: This `data-testid` is required? Even with the `DRG nodes` title?

ljmotta (commented):
@jomarko The tests are looking great, I guess the only thing needed for them is to revisit the name choice as I can see some different patterns, but they're very easy to read and understand. I've added a few comments inline, and I have a question regarding the locator choice. Looking at the `drds` fixture, all methods are using `data-testid` locators, and I'm not sure if they are required for all cases. Remember, `data-testid` should be used for cases where we can't select the element using the more accessible locators.

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
jomarko: yes, well spotted. the reason for adding is actually the `nodes.rename`. Until now, we used `nodes.rename` always just after a node was dragged into the diagram (it was focused and ready to be renamed, i.e. the text box had focus and all text was selected so could be replaced with new value)

This PR add a test, where we want to rename a node, that is already in diagram. For the current implementation of `nodes.rename` it is not enough to do `node.click()` before renaming is started, but the `node.dblClick` is needed to start the renaming.

So I decided to keep the responsibility on the test author side. It means test author needs to know `dblCkick` is needed before node rename starts so test author needs to call select with `dblClick` option. In other words, I keep single responsibility for the fixture methods we have. `select` is only for selecting node (by click or dblClick) and `rename` only renames the node.

I could probably adapt the `nodes.rename` to contain also this `dblCkick` but then probably we need to change the method name because it will not only rename, what do you think?

packages/dmn-editor/tests-e2e/drds/modelDecisionServicesInDrds.spec.ts
jomarko: yes, you could see my conversation with @tiagobento about this and my plan to work on this in a separate ticket because I think this will be a lot of tests. However, it seems, other tickets have higher priority now. I see options:
- Keep this file as placeholder, what makes kind of visible we are going to test it in future
- Remove this file to seed up (very slightly) the builds and make easier potential refactoring like moving whole folders when doing some test structure change or doing a global change in file licenses , etc


so please let me know if I should keep the file or remove it

packages/dmn-editor/src/diagram/Palette.tsx
jomarko: I will remove it

jomarko:
@ljmotta updates pushed

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
ljmotta: Alright, so we have a problem here. The `select` should select the node, which is accomplished with one click. Adding more functionalities to it will hide what we're actually doing here, which is enter in edit mode. Now, if we have another way of selecting the node, we can add to this method. But for this case we want to enter on edit mode for the node name, and we can do that with some alternative ways, which should be encapsulated on another method. I mean, we can do that with `dbclick` and `click+enter` (select + enter).

packages/dmn-editor/tests-e2e/drds/modelDecisionServicesInDrds.spec.ts
ljmotta: Ok, I think it's ok to left as a placeholder, but could you add the `test.skip` + `annotation` for each one of the cases? I guess the numbers are issues?

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
ljmotta: Taking a better look at the current fixtures, I could find a `selectLabel` on the `nodes` fixture which selects the node label by using the `dbclick`.

packages/dmn-editor/tests-e2e/drds/modelDecisionServicesInDrds.spec.ts
jomarko: @ljmotta added

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
jomarko: @ljmotta thank you, changed

danielzhe (approved):
I think it is looking great and it is following the patterns we are using in the other tests.

klbowers (approved):
The changes look good, thanks for that @jomarko!

PR #2465, opened 2024-07-11
kie-issues#1362: Migrate kn-workflow-plugin to kubernetes/client-go

### LINKED ISSUE #1362: Migrate kn-workflow-plugin to kubernetes/client-go

treblereel:
Closes https://github.com/apache/incubator-kie-issues/issues/1362

### DIFF
--- a/packages/kn-plugin-workflow/pkg/command/deploy.go
+++ b/packages/kn-plugin-workflow/pkg/command/deploy.go
@@ -20,7 +20,10 @@
 package command
 
 import (
+	"errors"
 	"fmt"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"path"
 
@@ -138,7 +141,7 @@ func deploy(cfg *DeployUndeployCmdConfig) error {
 		return fmt.Errorf("❌ ERROR: failed to get manifest directory and files: %w", err)
 	}
 	for _, file := range files {
-		if err = common.ExecuteKubectlApply(file, cfg.NameSpace); err != nil {
+		if err = common.ExecuteApply(file, cfg.NameSpace); err != nil {
 			return fmt.Errorf("❌ ERROR: failed to deploy manifest %s,  %w", file, err)
 		}
 		fmt.Printf(" - ✅ Manifest %s successfully deployed in namespace %s\n", path.Base(file), cfg.NameSpace)
@@ -188,11 +191,9 @@ func runDeployCmdConfig(cmd *cobra.Command) (cfg DeployUndeployCmdConfig, err er
 		return cfg, fmt.Errorf("❌ ERROR: failed to get default dashboards files folder: %w", err)
 	}
 
-	// check if sonataflow operator CRDs are installed
-	for _, crd := range metadata.SonataflowCRDs {
-		if !common.CheckKubectlCrdExists(crd) {
-			return cfg, fmt.Errorf("❌ ERROR: the required CRDs are not installed.. Install the SonataFlow Operator CRD first")
-		}
+	// check if sonataflow operator and knative CRDs are installed
+	if err := CheckCRDs(metadata.SonataflowCRDs, "SonataFlow Operator"); err != nil {
+		return cfg, err
 	}
 
 	//setup manifest path
@@ -201,4 +202,19 @@ func runDeployCmdConfig(cmd *cobra.Command) (cfg DeployUndeployCmdConfig, err er
 	}
 
 	return cfg, nil
+}
+
+func CheckCRDs(crds []string, typeName string) error {
+	for _, crd := range crds {
+		err := common.CheckCrdExists(crd)
+		if err != nil {
+			var statusErr *apierrors.StatusError
+			if errors.As(err, &statusErr) && statusErr.ErrStatus.Reason == metav1.StatusReasonNotFound {
+				return fmt.Errorf("❌ ERROR: the required CRDs are not installed.. Install the %s CRD first", typeName)
+			} else {
+				return fmt.Errorf("❌ ERROR: failed to check if CRD %s exists: %w", crd, err)
+			}
+		}
+	}
+	return nil
 }
--- a/packages/kn-plugin-workflow/pkg/command/quarkus/deploy.go
+++ b/packages/kn-plugin-workflow/pkg/command/quarkus/deploy.go
@@ -21,9 +21,12 @@ package quarkus
 
 import (
 	"fmt"
+	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/command"
 	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/common"
+	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/metadata"
 	"github.com/ory/viper"
 	"github.com/spf13/cobra"
+	"path/filepath"
 )
 
 type DeployCmdConfig struct {
@@ -71,6 +74,11 @@ func NewDeployCommand() *cobra.Command {
 	return cmd
 }
 
+var crds = map[string][]string{
+	"SonataFlow Operator":                  metadata.SonataflowCRDs,
+	"Knative Serving and Knative Eventing": metadata.KnativeCoreServingCRDs,
+}
+
 func runDeploy(cmd *cobra.Command, args []string) error {
 	fmt.Println("🛠️  Deploying your Quarkus SonataFlow project...")
 
@@ -79,8 +87,11 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("initializing deploy config: %w", err)
 	}
 
-	if err = common.CheckKubectl(); err != nil {
-		return err
+	// check necessary CRDs are installed
+	for name, crds := range crds {
+		if err := command.CheckCRDs(crds, name); err != nil {
+			return err
+		}
 	}
 
 	if _, err = deployKnativeServiceAndEventingBindings(cfg); err != nil {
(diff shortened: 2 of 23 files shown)


packages/kn-plugin-workflow/pkg/common/k8sclient/goapi.go
```diff
@@ -0,0 +1,192 @@
...
+func (m GoAPI) ExecuteKubectlApply(path, namespace string) error {
+	config, err := KubeRestConfig()
+	if err != nil {
+		return fmt.Errorf("❌ ERROR: Failed to create rest config for Kubernetes client: %v", err)
+	}
+
+	client, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return fmt.Errorf("❌ ERROR: Failed to create dynamic Kubernetes client: %v", err)
+	}
```
ricardozanini: Maybe we should isolate this part of the codebase, it seems very similar to the other methods.

packages/kn-plugin-workflow/pkg/common/k8sclient/goapi.go
```diff
@@ -0,0 +1,192 @@
...
+
+	_, err = res.Create(context.Background(), unstructuredMap, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("❌ ERROR: Failed to apply CRD: %w", err)
+	}
+	return nil
+}
+
+func (m GoAPI) ExecuteKubectlDelete(path, namespace string) error {
```
ricardozanini:
```suggestion
func (m GoAPI) ExecuteKubectlDelete(path, namespace string) error {
```

You can remove the `kubectl` prefix from these operations now that you are using the Rest client directly. The only portion of the code that we should link to `kubectl` is when we look for the cluster rest configuration.

packages/kn-plugin-workflow/pkg/common/k8sclient/kubectl.go
```diff
@@ -0,0 +1,182 @@
...
+		return false, err
+	}
+	return true, nil
+}
+
+func CheckOperatorInstalled() error {
+	if err := CheckKubectl(); err != nil {
+		return err
+	}
+	cmd := exec.Command("kubectl", "get", "pods", "-n", metadata.OperatorName)
```
ricardozanini: You should use the Pod API instead, see an example here: https://github.com/kubernetes/client-go/blob/master/examples/out-of-cluster-client-configuration/main.go#L74

Any references to `exec.Command` must be gone.

packages/kn-plugin-workflow/pkg/common/k8sclient/kubectl.go
```diff
@@ -0,0 +1,182 @@
...
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("❌ ERROR: failed to execute kubectl delete command for %s: %s", crd, err)
+	}
+
+	return nil
+}
+
+func (m KubeCtl) CheckKubectlCrdExists(crd string) (bool, error) {
+	cmd := exec.Command("kubectl", "get", "crd", crd)
```
ricardozanini: Same here regarding the usage of `kubectl`.

packages/kn-plugin-workflow/pkg/common/k8sclient/kubectl.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (m KubeCtl) ExecuteKubectlDelete(crd, namespace string) error {
+	if err := CheckKubectl(); err != nil {
+		return err
+	}
+	cmd := exec.Command("kubectl",
+		"delete",
+		"-f", crd,
+		"-n", namespace)
```
ricardozanini: Here we also should refactor to delete the objects directly from the REST Api.

packages/kn-plugin-workflow/pkg/common/k8sclient/kubectl.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (m KubeCtl) ExecuteKubectlApply(crd, namespace string) error {
+	if err := CheckKubectl(); err != nil {
+		return err
+	}
+	cmd := exec.Command("kubectl",
+		"apply",
+		"-f", crd,
+		"-n", namespace,
+		"--validate=false")
```
ricardozanini: Handling the kubectl `last-applied-configuration` is unnecessary if we use the Rest clients instead.

packages/kn-plugin-workflow/pkg/common/k8sclient/kubectl.go
```diff
@@ -0,0 +1,182 @@
...
+	if context == "" {
+		return "", fmt.Errorf("❌ ERROR: No current kubectl context found")
+	}
+	fmt.Printf(" - ✅ kubectl current context: %s \n", context)
+	return context, nil
+}
+
+func CheckKubectl() error {
+	fmt.Println("🔎 Checking if kubectl is available...")
+	_, kubectlCheck := exec.LookPath("kubectl")
```
ricardozanini: Why do we need kubectl installed? To read the cluster's configuration?

ricardozanini (commented):
Thanks! But I believe we are still using the `kubectl` command in a few places.

treblereel:
> Thanks! But I believe we are still using the `kubectl` command in a few places.

nope, there three impls of K8sApi: goapi -- uses go api, kubectl -- k8s cli and fake.go -- for tests. By default, goapi is active. I left kubectl in case we want to have an option to failback, if you don't like I can remove it any moment.

ricardozanini:
@treblereel thanks for the clarification. It's not a matter of my personal taste, but if we are not using `kubectl`, it's better to remove anything from the code base.

packages/kn-plugin-workflow/pkg/common/k8sclient/goapi.go
```diff
@@ -53,42 +53,37 @@ func KubeRestConfig() (*rest.Config, error) {
...
 
-func (m GoAPI) ExecuteKubectlApply(path, namespace string) error {
-	config, err := KubeRestConfig()
-	if err != nil {
-		return fmt.Errorf("❌ ERROR: Failed to create rest config for Kubernetes client: %v", err)
-	}
-
-	client, err := dynamic.NewForConfig(config)
+func (m GoAPI) ExecuteApply(path, namespace string) error {
+	client, err := getDynamicClient()
```
ricardozanini: There's a catch here you might need to check. The file can contain many objects, so the deserialization from the file below might not work as expected.

packages/kn-plugin-workflow/pkg/command/deploy.go
```diff
@@ -202,3 +203,18 @@ func runDeployCmdConfig(cmd *cobra.Command) (cfg DeployUndeployCmdConfig, err er
...
 	return cfg, nil
 }
+
+func CheckCRDs(crds []string, typeName string) error {
+	for _, crd := range crds {
+		err := common.CheckCrdExists(crd)
+		if err != nil {
+			var statusErr *apierrors.StatusError
+			if errors.As(err, &statusErr) && statusErr.ErrStatus.Reason == metav1.StatusReasonNotFound {
+				return fmt.Errorf("❌ ERROR: the required CRDs are not installed.. Install the %s CRD first", typeName)
```
ricardozanini, line 213:
```suggestion
				return fmt.Errorf("❌ ERROR: the required CRDs are not installed.. Install the %s CRD first. Usually this is done by installing the SonataFlow Operator", typeName)
```

packages/kn-plugin-workflow/pkg/command/quarkus/deploy.go
```diff
@@ -71,6 +74,11 @@ func NewDeployCommand() *cobra.Command {
 	return cmd
 }
 
+var crds = map[string][]string{
+	"SonataFlow Operator":                  metadata.SonataflowCRDs,
+	"Knative Serving and Knative Eventing": metadata.KnativeCoreServingCRDs,
```
ricardozanini, line 79: These are not mandatory—only the SonataFlow one. The Operator can manage the cluster without Knative and deploy the objects accordingly.

ricardozanini: We can warn the users about it: "If Knative is not installed, you may not be able to deploy serverless services or orchestrate events."

packages/kn-plugin-workflow/pkg/command/quarkus/deploy.go
```diff
@@ -94,36 +105,19 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
 func deployKnativeServiceAndEventingBindings(cfg DeployCmdConfig) (bool, error) {
 	isKnativeEventingBindingsCreated := false
-	createService := common.ExecCommand("kubectl", "apply", "-f", fmt.Sprintf("%s/knative.yml", cfg.Path), fmt.Sprintf("--namespace=%s", cfg.Namespace))
-	if err := common.RunCommand(
-		createService,
-		"deploy",
-	); err != nil {
-		fmt.Println("❌ ERROR: Deploy failed, Knative service was not created.")
+
+	err := common.ExecuteApply(filepath.Join(cfg.Path, "knative.yml"), cfg.Namespace)
```
ricardozanini, line 110: If we go to this branch, it means that the user is generating the artifacts from Quarkus (which we will review later), which makes Knative mandatory in this specific unique case.

packages/kn-plugin-workflow/pkg/command/quarkus/testdata/kogito-complex3-custom-namespace.yml
```diff
@@ -0,0 +1,66 @@
...
+      restartPolicy: Never
+  backoffLimit: 4
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: example-cronjob
+  namespace: mycustom
+spec:
+  schedule: "*/5 * * * *" # Запуск каждые 5 минут
```
ricardozanini: Nice Russian comment :D

packages/kn-plugin-workflow/pkg/common/k8sclient/goapi.go
```diff
@@ -0,0 +1,242 @@
...
+	"k8s.io/apimachinery/pkg/util/yaml"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+type GoAPI struct{}
```
ricardozanini, line 24: It is a very nice adapter implementation. 👍

ricardozanini (commented):
Seems really clean now! I just left minor comments if you don't mind!

packages/kn-plugin-workflow/pkg/command/quarkus/deploy.go
treblereel: @ricardozanini we need it in case we deploy the current quarkus created/converted deployment. It fails without serving-crds.yaml and serving-core.yaml.

packages/kn-plugin-workflow/pkg/command/deploy.go
treblereel: Depending on the case, there are `Install the SonataFlow Operator CRD first` or  `Install the Knative Serving and Knative Eventing CRD first`

packages/kn-plugin-workflow/pkg/command/quarkus/deploy.go
ricardozanini: It makes perfect sense!

ricardozanini: So if we are not checking for Knative in any other use case, we are good to go. Thanks for confirming it!

ricardozanini (approved):
@wmedvede, can you also please take a quick look? Thanks!

domhanak (approved):

PR #2468, opened 2024-07-16
kie-issues#1386: Improve pom.xml generated by kn-workflow-plugin

### LINKED ISSUE #1386: Improve pom.xml generated by kn-workflow-plugin
It looks like, while publishing, the kn-workflow-plugin still uses `0.0.0` version of sonataflow-quarkus-devui.

It must be investigated.
Plus, the apache repository must be added to the generated pom.xml

```
<repositories>
    <repository>
        <id>apache-public-repository-group</id>
        <name>Apache Public Repository Group</name>
        <url>https://repository.apache.org/content/groups/public/</url>
    </repository>
</repositories>
```

treblereel:
Closes https://github.com/apache/incubator-kie-issues/issues/1386

### DIFF
--- a/packages/kn-plugin-workflow/e2e-tests/quarkus_build_test.go
+++ b/packages/kn-plugin-workflow/e2e-tests/quarkus_build_test.go
@@ -123,8 +123,7 @@ func RunQuarkusBuildTest(t *testing.T, cfgTestInputQuarkusCreate CfgTestInputQua
 	require.NoErrorf(t, err, "Expected nil error, got %v", err)
 
 	// Run `quarkus build` command
-	args := transformQuarkusBuildCmdCfgToArgs(test.input);
-	AddSnapshotRepositoryDeclarationToPom(t, projectDir)
+	args := transformQuarkusBuildCmdCfgToArgs(test.input)
 	_, err = ExecuteKnWorkflowQuarkus(args...)
 	require.NoErrorf(t, err, "Expected nil error, got %v", err)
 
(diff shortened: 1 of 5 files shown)


wmedvede (changes requested):
Hi @treblereel looks good, many thanks.

Just a couple o nitpicks.

- In the generated pom.xml, I can see that the maven central repo is being added, I think we don't need to add, since it's always present.

- Then, I was thinking that maybe we can rename -> `sonataflow.quarkus.devui.version` to `kogito.tooling.version`

- And similar to what you did for the property above, for the kogito runtimes extensions, we can introduce a property `kogito.version`.  This will make the generated project much more flexible, wdyt?

- Finally, if the generated project is oriented to execute the dev mode, similar to what we do with the data-index in memmory we must add: `org.kie:kie-addons-quarkus-process-management` and  `org.kie:kogito-addons-quarkus-jobs-service-embedded`

treblereel:
@wmedvede done

wmedvede (approved):

tiagobento:
If I may, `Kogito Tooling` is the old name we used to have for `KIE Tools`. I'd suggest referring to it as `kie.tools.version`, even though we don't have different versions on releases anymore, but in development we still do.

wmedvede:
> If I may, `Kogito Tooling` is the old name we used to have for `KIE Tools`. I'd suggest referring to it as `kie.tools.version`, even though we don't have different versions on releases anymore, but in development we still do.

good catch, to introduce the old name was my fault :angry:

wmedvede:
> If I may, `Kogito Tooling` is the old name we used to have for `KIE Tools`. I'd suggest referring to it as `kie.tools.version`, even though we don't have different versions on releases anymore, but in development we still do.

Hi @treblereel , would you mind apply the suggestion

wmedvede:
> > If I may, `Kogito Tooling` is the old name we used to have for `KIE Tools`. I'd suggest referring to it as `kie.tools.version`, even though we don't have different versions on releases anymore, but in development we still do.
> 
> Hi @treblereel , would you mind apply the suggestion

Many Thanks

domhanak (approved):
Thanks!

PR #2472, opened 2024-07-17
kie-issues#1040: Remove packages/sonataflow-images and packages/sonataflow-operator from kie-tools

thiagoelg:
Part of https://github.com/apache/incubator-kie-issues/issues/1040

- Removed sonataflow-operator
- Removed sonataflow-devmode-image
- Removed sonataflow-builder-image
- Removed references from Jenkinsfiles
- Removed references from GitHub CI partitions
- Fixed some Dockerfiles `ENV` definitions with deprecated syntaxes (unrelated, but noticed some warnings about the deprecation)


porcelli (approved):
Thank you @thiagoelg for your prompt action following thru the agreed 10 release proposal [1]

[1] https://lists.apache.org/thread/pw327lkpmy9gxklq4t5zbwzxjo2y3c0w

wmedvede (changes requested):
@porcelli ,  @thiagoelg Guys, this removal must be coordinated with the move of this bits to their new destination in apache/incubator-kie-kogito-serverless-operator. We can't remove this as is.

@ricardozanini will execute this move to apache/incubator-kie-kogito-serverless-operator when he's back in two weeks. Until this, we must hold-on this PR.


fyi @krisv @domhanak @fjtirado

porcelli:
@wmedvede I'm not sure what needs to be coordinated... this PR will only accomplish what has been long time decided.

AFAIK, the original repos from where the content had to be moved never stop the development... so I can't see why hold this change

wmedvede:
> @wmedvede I'm not sure what needs to be coordinated... this PR will only accomplish what has been long time decided.
> 
> AFAIK, the original repos from where the content had to be moved never stop the development... so I can't see why hold this change

@porcelli 
It must be coordinated because, for example the build of  incubator-kie-sonataflow-builder image will disappear from one repo (incubator/kie-tools), and a new build of the same image, will appear, but form a different repo (incubator/incubator-kie-kogito-serverless-operator)

But, for example incubator-kie-sonataflow-builder, is based on bits from kie-kogito-runtimes.

If we stop producing  incubator-kie-sonataflow-builder from one day to the other, we'll have that image outdated with reference to kie-kogito-runtimes. This represent something no good for the us.  (example, if we need to do a fix in kogito-runtimes, we won't have this fix reflected in incubator-kie-sonataflow-builder)
 
On the other hand, considering that the CI for incubator-kie-sonataflow-builder is already set, and this image is already being being produced regularly, I don't think this represents a problem nor any extra work.
We just keep all the CI as is, that is already working, and we remove the production for these images "coordinated".
This will help us.

tiagobento:
@wmedvede We have the discussions going on the Mailing List (https://lists.apache.org/thread/ldb45nljq2q2d8pln3bxmh4g0m6k7o6l), but it was agreed that we would do this even if that broke stuff. Anyway, please let's continue the conversations on the email thread.

tiagobento:
Closing. See https://lists.apache.org/thread/85tdj40gngmkl9fgovy6qg9d6wyfx8oy

PR #2474, opened 2024-07-21
kie-issues#1335: kn workflow build output log should be saved in the target dir

### LINKED ISSUE #1335: kn workflow build output log should be saved in the target dir
When running kn workflow build the user is clueless about what is happening unless the verbose flag is on. The problem is that they might have to run the build twice to figure out the problem.

It would be easier if the tool would output the build log to the /target/build.log file each time a build is performed. The CLI ought to override the file each time.

jira: https://issues.redhat.com/browse/KOGITO-7669

treblereel:
Closes: https://github.com/apache/incubator-kie-issues/issues/1335

### DIFF
--- a/packages/kn-plugin-workflow/pkg/command/quarkus/build.go
+++ b/packages/kn-plugin-workflow/pkg/command/quarkus/build.go
@@ -21,6 +21,10 @@ package quarkus
 
 import (
 	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


treblereel:
@ricardozanini 

At the moment, I log only quarkus build output (where magic happens), because in order to save all messages that are output to the console, we need to replace fmt.println with logging, which clearly falls outside the scope of this issue. If this is necessary, I suggest creating another issue.

ricardozanini (approved):

domhanak (approved):

Issue #2475, opened 2024-07-22
Include the bump of variable tagVersion in version.go via the bump-version.sh script


wmedvede:
We need to incorporate the variable tagVersion in version.go to the  bump.

tiagobento:
https://github.com/apache/incubator-kie-tools/pull/2476 is merged. Will this be forward-ported to `main` or not? Please update the issue accordingly...

wmedvede:
Counterpart PR for main: #2522

PR #2476, opened 2024-07-22
[10.0.x] incubator-kie-tools#2475: Include the bump of variable tagVersion in version.go via the bump-version.sh script

wmedvede:
Closes https://github.com/apache/incubator-kie-tools/issues/2475


wmedvede:
Hi @tiagobento , @rodrigonull ,  @jstastny-cz I have noted that https://github.com/apache/incubator-kie-tools/blob/10.0.x/packages/sonataflow-operator/version/version.go#L33  was not being picked up by the bump.

This PR updates the script, and for the 10.0.x I have already added the value that we expect to have there. 

Could you please take a look and prioritize?

tiagobento (approved):

tiagobento:
Great catch @wmedvede. Thanks for the PR. Would you mind sending one to `main` as well? Even if we don't have the definitions yet on whether or not we'll delete those modules from `main`, would be good to keep it in good shape regardless. Thanks!

thiagoelg (approved):

PR #2477, opened 2024-07-22
kie-issues#1182: Performing the `undo` operation after resetting the expressions crashes the new DMN Editor

### LINKED ISSUE #1182: Performing the `undo` operation after resetting the expressions crashes the new DMN Editor
Performing a `undo` operation after reseting an expression crashs the editor. It's necessary to investigate the root cause, and the reproducer. 

File:
https://gist.github.com/ljmotta/144abe93b458247414cc4e84d0e78d0a

Video:
https://github.com/apache/incubator-kie-issues/assets/24302289/d251dce0-df0a-421a-acdd-4e17a1c6befe

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1182

The issue was that BEE was providing expressions without IDs, which was not expected by the DMN Editor.

This change guarantees that every expression and its nested components (like variables, context entries, and so on) that have IDs will have its ID defined (no more undefined IDs if they are optional in the XML definitions).

### DIFF
--- a/packages/boxed-expression-component/src/expressions/ExpressionDefinitionRoot/ExpressionDefinitionLogicTypeSelector.tsx
+++ b/packages/boxed-expression-component/src/expressions/ExpressionDefinitionRoot/ExpressionDefinitionLogicTypeSelector.tsx
@@ -34,7 +34,7 @@ import { ResourcesAlmostEmptyIcon } from "@patternfly/react-icons/dist/js/icons/
 import { ResourcesFullIcon } from "@patternfly/react-icons/dist/js/icons/resources-full-icon";
 import * as React from "react";
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
-import { BoxedExpression } from "../../api";
+import { BoxedExpression, Normalized } from "../../api";
 import { useCustomContextMenuHandler } from "../../contextMenu";
 import { MenuItemWithHelp } from "../../contextMenu/MenuWithHelp";
 import { useBoxedExpressionEditorI18n } from "../../i18n";
@@ -62,7 +62,7 @@ import FilterIcon from "@patternfly/react-icons/dist/esm/icons/filter-icon";
 
 export interface ExpressionDefinitionLogicTypeSelectorProps {
   /** Expression properties */
-  expression?: BoxedExpression;
+  expression?: Normalized<BoxedExpression>;
   /** Function to be invoked when logic type changes */
   onLogicTypeSelected: (logicType: BoxedExpression["__$$element"] | undefined) => void;
   /** Function to be invoked when logic type is reset */
@@ -147,9 +147,7 @@ export function ExpressionDefinitionLogicTypeSelector({
       case "context":
         return <ContextExpression expression={expression} isNested={isNested} parentElementId={parentElementId} />;
       case "decisionTable":
-        return (
-          <DecisionTableExpression expression={expression} isNested={isNested} parentElementId={parentElementId} />
-        );
+        return <DecisionTableExpression expression={expression} isNested={isNested} />;
       case "list":
         return <ListExpression expression={expression} isNested={isNested} parentElementId={parentElementId} />;
       case "invocation":
@@ -321,8 +319,8 @@ export function ExpressionDefinitionLogicTypeSelector({
 
       const newIdsByOriginalId = mutateExpressionRandomizingIds(clipboard.expression);
 
-      let oldExpression: BoxedExpression | undefined;
-      setExpression((prev: BoxedExpression) => {
+      let oldExpression: Normalized<BoxedExpression> | undefined;
+      setExpression((prev: Normalized<BoxedExpression>) => {
         oldExpression = prev;
         return clipboard.expression;
       }); // This is mutated to have new IDs by the ID randomizer above.
--- a/packages/boxed-expression-component/tests-e2e/__fixtures__/jsonModel.ts
+++ b/packages/boxed-expression-component/tests-e2e/__fixtures__/jsonModel.ts
@@ -18,6 +18,7 @@
  */
 
 import { Page } from "@playwright/test";
+import { BoxedContext, BoxedEvery, BoxedFilter, BoxedFor, BoxedSome, Normalized } from "../../src/api";
 
 interface BoxedExpressionComponent {
   expression: string;
@@ -53,6 +54,31 @@ export class JsonModel {
     }
     const widthsById = jsonObject.widthsById || {};
     return widthsById[decisionTableId] || [];
+  }
+
+  public async getContextExpression(): Promise<Normalized<BoxedContext>> {
+    const content = await this.getBoxedExpressionContent();
+    return (content?.expression || {}) as Normalized<BoxedContext>;
+  }
+
+  public async getFilterExpression(): Promise<Normalized<BoxedFilter>> {
+    const content = await this.getBoxedExpressionContent();
+    return (content?.expression || {}) as Normalized<BoxedFilter>;
+  }
+
+  public async getEveryExpression(): Promise<Normalized<BoxedEvery>> {
+    const content = await this.getBoxedExpressionContent();
+    return (content?.expression || {}) as Normalized<BoxedEvery>;
+  }
+
+  public async getForExpression(): Promise<Normalized<BoxedFor>> {
+    const content = await this.getBoxedExpressionContent();
+    return (content?.expression || {}) as Normalized<BoxedFor>;
+  }
+
+  public async getSomeExpression(): Promise<Normalized<BoxedSome>> {
+    const content = await this.getBoxedExpressionContent();
+    return (content?.expression || {}) as Normalized<BoxedSome>;
   }
 
   private async getBoxedExpressionContent(): Promise<BoxedExpressionComponent | undefined> {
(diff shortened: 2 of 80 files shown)


packages/dmn-editor/src/mutations/updateExpression.ts
```diff
@@ -64,4 +65,12 @@ export function updateExpression({
...
     throw new Error("DMN MUTATION: Can't update expression for drgElement that is not a Decision or a BKM.");
   }
+
+  getNewDmnIdRandomizer()
+    .ack({
+      json: definitions.drgElement,
+      type: "DMN15__tDefinitions",
+      attr: "drgElement",
+    })
+    .randomize();
```
tiagobento: This is not necessary with the ids being added in all places we need, right?

packages/boxed-expression-component/src/expressions/ExpressionDefinitionRoot/ExpressionDefinitionLogicTypeSelector.tsx
```diff
@@ -147,9 +147,7 @@ export function ExpressionDefinitionLogicTypeSelector({
       case "context":
         return <ContextExpression expression={expression} isNested={isNested} parentElementId={parentElementId} />;
       case "decisionTable":
-        return (
-          <DecisionTableExpression expression={expression} isNested={isNested} parentElementId={parentElementId} />
-        );
+        return <DecisionTableExpression expression={expression} isNested={isNested} />;
```
tiagobento, line 150: Why remove the parentElementId?

tiagobento (changes requested):

tiagobento:
Thanks for the PR @danielzhe! Definitely a much needed improvement. Can you please add a regression test too?

packages/boxed-expression-component/src/expressions/ExpressionDefinitionRoot/ExpressionDefinitionLogicTypeSelector.tsx
danielzhe: It is not used anywhere. ESLint warned me about so I removed it. 😄

packages/dmn-editor/src/mutations/updateExpression.ts
danielzhe: Yes, it is probably a leftover from some tests we were doing.

danielzhe:
I added some new tests and new stories to expressions that we don't have stories yet. We need more work on that, but we have the base now. I need your review here, @ljmotta .

danielzhe:
@tiagobento Changes applied! I added tests to every expression known to break due to the lack of the ID.

P.S.: sorry for the rebase. New PRs are coming only with merge. 😄

packages/stunner-editors-dmn-loader/src/gwtToBee.ts
ljmotta, line 1: 🤩

packages/dmn-editor/src/mutations/updateExpression.ts
ljmotta: This will be removed?

packages/dmn-editor/src/boxedExpressions/BoxedExpressionScreen.tsx
```diff
@@ -81,6 +81,7 @@ import { useDmnEditorStore, useDmnEditorStoreApi } from "../store/StoreContext";
 import { getDefaultColumnWidth } from "./getDefaultColumnWidth";
 import { getDefaultBoxedExpression } from "./getDefaultBoxedExpression";
 import { Normalized } from "../normalization/normalize";
+import { getNewDmnIdRandomizer } from "../idRandomizer/dmnIdRandomizer";
```
ljmotta: leftover?

packages/boxed-expression-component/tests-e2e/boxedExpressions/some/someExpression.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { expect, test } from "../../__fixtures__/base";
+import { TestAnnotations } from "../../../../playwright-base/annotations";
```
ljmotta: The import should point to `@kie-tools/playright-base/annotations`. The same applies for the other files as well.

packages/boxed-expression-component/tests-e2e/boxedExpressions/some/someExpression.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+test.describe("Create Boxed Some", () => {
+  test.beforeEach(async ({ stories }) => {
+    await stories.openBoxedSome("base");
+  });
+  test("should keep IDs after resetting entries", async ({ bee, jsonModel, page }) => {
```
ljmotta: Not sure if understand the purpose of the test based on the name. The test should check if the ID was kept? I mean, the ID should be the same before and after the reset? Currently, the test is checking if the expression has an ID after the reset. The same applies for the other tests as well.

packages/boxed-expression-component/tests-e2e/__fixtures__/stories.ts
```diff
@@ -65,4 +65,16 @@ export class Stories {
...
+    await this.page.goto(`${this.baseURL}/${this.getIframeURL(`boxed-expressions-every--${type}`)}` ?? "");
+  }
+
+  public async openBoxedSome(type: BoxedExpressionTypes) {
+    await this.page.goto(`${this.baseURL}/${this.getIframeURL(`boxed-expressions-some--${type}`)}` ?? "");
+  }
+
+  public async openBoxedFor(type: BoxedExpressionTypes) {
+    await this.page.goto(`${this.baseURL}/${this.getIframeURL(`boxed-expressions-for--${type}`)}` ?? "");
+  }
```
ljmotta: Could you please add the `base` as the default `type`, following the other expressions?

packages/boxed-expression-component/tests-e2e/__fixtures__/jsonModel.ts
```diff
@@ -18,6 +18,7 @@
  */
 
 import { Page } from "@playwright/test";
+import { BoxedContext, BoxedEvery, BoxedFilter, BoxedFor, BoxedSome, Normalized } from "../../src/api";
```
ljmotta: We mustn't couple the `src` code with the `test` code, as the test should know it. For this case, we duplicate the code to test folder. I'm not sure where we should put them, maybe with the new API?

ljmotta (changes requested):
@danielzhe Thanks for this PR. This will be a huge improvement. You got many missing `ids`! I've made some comments inline, more related to the tests. I guess we still have a leftover from Tiago's review.

packages/boxed-expression-component/tests-e2e/boxedExpressions/some/someExpression.spec.ts
danielzhe: Thank you for spotting this. IDE trolled me again. I shouldn't trust IDE to add references anymore.

packages/boxed-expression-component/tests-e2e/boxedExpressions/some/someExpression.spec.ts
danielzhe: It doesn't need to be the same, just needs to be there. 
The issue was that the ID was removed when the expression was reset and its ID was not set again when the new expression was set.

Maybe the test name is misleading. What about "should set IDs after resetting entries"?

packages/boxed-expression-component/tests-e2e/__fixtures__/jsonModel.ts
danielzhe: Yeah, I think the best place is the new API. I'll push the changes and let me know what you think.

packages/boxed-expression-component/tests-e2e/boxedExpressions/some/someExpression.spec.ts
danielzhe: Also, maybe "entries" is not the right name, only for "context". I'm reviewing it and pushing with new suggestions, let me know what you think!

danielzhe:
@ljmotta I removed the leftovers now. I removed it before when @tiagobento spotted it but I guess I did bad things with rebase lol.
I also moved the types to "types.ts" inside the API and changed the test names. I think they are more clear now reflecting what was the real issue and what we're testing.

packages/dmn-editor/src/mutations/updateExpression.ts
```diff
@@ -24,6 +24,7 @@ import {
 } from "@kie-tools/dmn-marshaller/dist/schemas/dmn-1_5/ts-gen/types";
 import { renameDrgElement } from "./renameNode";
 import { Normalized } from "../normalization/normalize";
+import { getNewDmnIdRandomizer } from "../idRandomizer/dmnIdRandomizer";
```
ljmotta: Leftover?

ljmotta (commented):
@danielzhe Thanks for addressing the change request. On my first review I missed one thing that I guess it's a leftover.

packages/dmn-editor/src/mutations/updateExpression.ts
danielzhe: You didn't miss it, I missed it :shyface 
It was a new leftover after I removed the `getNewDmnIdRandomizer`.

ljmotta (approved):

tiagobento (approved):

PR #2485, opened 2024-07-25
kie-issues#371: [DMN Editor] Enable modification of constraints on DT columns

### LINKED ISSUE #371: [DMN Editor] Enable modification of constraints on DT columns
As @jomarko  described in [this blog](https://blog.kie.org/2023/01/automatically-generating-decision-tables-in-dmn.html), it is possible to add restrictions to an input row of a decision table. If a common type is being used during the generation of the DT, the values of that type are copied into the definition of that column. 
However, as a user of the editor, I have no control of these values. If the list of values in the type is modified, there is no way to `sync` those values. There is no way to see the restrictions, let alone to modify them. Currently the only way to do so is to hack the xml in a text editor. 
Wat I would like to have:

- A way to view and edit the restrictions of a column
- A button to sync the values with the given data type
- Some visual indication that tells whether or not the given values are in sync with the values of the given data type.

(This was first discussed in [this zulip thread](https://kie.zulipchat.com/#narrow/stream/232676-kogito/topic/Input.20values.20of.20an.20input.20column.20of.20a.20decision.20table.3F))

ljmotta:
Closes https://github.com/apache/incubator-kie-issues/issues/371

## Description
Add 76 tests to the `dmn-editor` checking the attribute types and the respective constraint for the Decision Table header and rules. The tests change the column type using the properties panel, check if it's correctly attributed, and check if the constraint, in case of the type is an Item Definition (e.g. Custom Data Type), is in `readonly` mode. It also checks if the type selection is in `readonly` mode in the rule cells. All tests are running on `chromium`, `webkit`, and `Google Chrome`.

During the test suite creation, some additional changes were made:
 - A Decision Table with one output column merges the column and the expression header in one cell. Previously, it was possible to attribute a different type for the column and expression, even though this doesn't make sense. This PR fixes this, by making the column type `readonly` and following the expression type. If the user has a model where the column has a different type than the expression, the column type will not be in `readonly`, allowing the user to fix this.
 - Adds the `dataTypes` fixture. This fixture adds the possibility of interacting with the Data Type tab. As the issue is not specific to the data types feature, it is still not complete.
 - Adds the `beePropertiesPanel` fixture. This fixture is still incomplete, as this PR only adds the Decision Table properties panel.
 - Type improvement on the `Constraint` components.


packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/parts/dataTypeProperties.ts
```diff
@@ -28,6 +28,56 @@ export class DataTypeProperties {
...
+  public getConstraintButton(args: { type: ConstraintType }) {
+    switch (args.type) {
+      case ConstraintType.ENUMERATION:
+        return this.panel.getByRole("button", { name: ConstraintType.ENUMERATION, exact: true });
+      case ConstraintType.EXPRESSION:
+        return this.panel.getByRole("button", { name: ConstraintType.EXPRESSION, exact: true });
+      case ConstraintType.NONE:
+        return this.panel.getByRole("button", { name: ConstraintType.NONE, exact: true });
+      case ConstraintType.RANGE:
+        return this.panel.getByRole("button", { name: ConstraintType.RANGE, exact: true });
```
klbowers: Is there any need for this switch statement? Couldn't it be simplified to be a one liner, e.g.
return this.panel.getByRole("button", {name: args.type, exact : true });

ljmotta: I'm not sure why I've done a `switch`/`case` for this 😅 . Thanks for sppoting it.

tiagobento: Great catch @kbowers-ibm. Well spotted.

tiagobento (approved):

klbowers (approved):
Many thanks for the tests @ljmotta, LGTM!

PR #2491, opened 2024-07-27
kie-issues#1398: Create Standalone Editor for the new DMN Editor

### LINKED ISSUE #1398: Create Standalone Editor for the new DMN Editor
On `kie-tools`, we need a new package called "dmn-editor-standalone" that will implement the same API as the  Standalone DMN Editor (classic), with the exception of the `canvas` attribute of the returned object by the `open` function.

This package will provide a JavaScript library (with embedded TypeScript typings) so that customers can embed the new DMN Editor in their applications.

It is expected that the "react" dependency will be a `peerDependency` (see https://github.com/apache/incubator-kie-issues/issues/1365).

The new DMN Editor is expected to be rendered inside an iframe and take advantage of the Multiplying Architecture (Envelope & Channel) to isolate the context and prevent CSS and global object (window) clashes with the host application.

thiagoelg:
Closes https://github.com/apache/incubator-kie-issues/issues/1398

- Created new package `dmn-editor-standalone`:
	- Implements the EditorStandaloneApi for the New DMN Editor;
	- Builds to a single .js file that can be imported and used anywhere, including fonts, images, and css;
	- Adds e2e tests to test the standalone editor's exposed API;
	- Adds Storybook story to develop the API and serve as an example.
- Adds the New DMN Editor Standalone to the `examples/webapp` package.


packages/kie-editors-standalone/README.md
```diff
@@ -15,14 +15,16 @@
    under the License.
 -->
 
-## BPMN and DMN Standalone Editors
+## BPMN and Legacy DMN Standalone Editors
```
tiagobento:
```suggestion
## BPMN and DMN Standalone Editors (classic)
```

packages/kie-editors-standalone/README.md
```diff
@@ -15,14 +15,16 @@
...
 
 ### Description
 
 This library provides standalone DMN and BPMN Editors (one all-in-one JavaScript file each) that can be embedded into any web application.
 
 A comprehensive API is also provided for setup and interaction with the Editor.
 
+For the updated DMN Editor, check the [`@kie-tools/dmn-editor-standalone`](../dmn-editor-standalone/) package.
```
tiagobento:
```suggestion
For the new DMN Editor, check the [`@kie-tools/dmn-editor-standalone`](../dmn-editor-standalone/) package.
```

packages/dmn-editor-standalone/stories/DevWebApp.stories.tsx
```diff
@@ -0,0 +1,231 @@
...
+      editor.close();
+      setEditCount(0);
+    };
+  }, []);
+
+  const onUndo = useCallback(() => {
+    setEditCount((currentCount) => {
+      if (currentCount > 0) {
+        editorRef.current?.undo();
+        // -2 because undoing will generateg a contentChange notification.
```
tiagobento:
```suggestion
        // -2 because undoing will generate a contentChange notification.
```

packages/dmn-editor-standalone/stories/DevWebApp.stories.tsx
```diff
@@ -0,0 +1,231 @@
...
+      readOnly: false,
+      resources: new Map(),
+      origin: "*",
+    });
+
+    editor.subscribeToContentChanges(() => setEditCount((currentCount) => currentCount + 1));
+
+    (editorRef as any).current = editor;
+
+    console.log(editor);
```
tiagobento, line 94: Lefover?

packages/dmn-editor-standalone/tests-e2e/api/getContent.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+import { loanPreQualificationDmn } from "../__fixtures__/externalModels";
+import prettier from "prettier";
+
+test.describe("Dmn Editor - API", () => {
+  test.describe("getContent", () => {
+    test.beforeEach(async ({ editor }) => {
+      await editor.open();
+    });
+
+    test("should get DMN contents of inputted DMN file", async ({ page, editor }) => {
```
tiagobento:
```suggestion
    test("should get DMN contents of input DMN file", async ({ page, editor }) => {
```

tiagobento (changes requested):
Looking great. Requested some minor changes inline.
Also, I guess `test.describe()` should contain `"DMN Editor - Standalone - API"`... for our test reports to be more clear.

@ljmotta Please take a look at the code of the tests, as it seems like we're heavily relying on CSS classes etc. I'll defer the review of those tests to you... Thanks!

packages/dmn-editor-standalone/stories/DevWebApp.stories.tsx
thiagoelg: Since this is the demo webapp I wanted whoever is using it to access the `editor` object easily via the console, but maybe using the Storybook args is a better idea to control the parameters 🤔

tiagobento: Right, agreed! So maybe adding a console.log("..."); with some explanation? Like "Access the "editor" instance by doing xyz" or whatever :D

tiagobento (approved):

tiagobento:
Thanks @thiagoelg!

packages/dmn-editor-standalone/env/index.js
```diff
@@ -0,0 +1,36 @@
...
+const { varsWithName, composeEnv } = require("@kie-tools-scripts/build-env");
+
+module.exports = composeEnv([require("@kie-tools/root-env/env"), require("@kie-tools-core/webpack-base/env")], {
+  vars: varsWithName({}),
+  get env() {
+    return {
+      dmnEditorStandalone: {
+        dev: {
+          port: "9006",
+        },
```
ljmotta, line 29: I guess we don't use this.

packages/dmn-editor-standalone/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import dmnEnvelopeJs from "../dist/envelope.js";
```
ljmotta, line 20: This will not make necessary to build first?

packages/dmn-editor-standalone/stories/DevWebApp.stories.tsx
```diff
@@ -0,0 +1,260 @@
...
+ * under the License.
+ */
+
+import * as React from "react";
+import type { Meta, StoryObj } from "@storybook/react";
+import { useRef, useState, useEffect, useCallback } from "react";
+import "@patternfly/react-core/dist/styles/base.css";
+import { Page, PageSection } from "@patternfly/react-core/dist/js/components/Page";
+import * as DmnEditor from "../dist/index";
+import { DmnEditorStandaloneApi } from "../dist/DmnEditorStandaloneApi";
```
ljmotta, line 26: Not sure that we want it depending on the `dist` folder as the storybook is like a dev webapp.

packages/dmn-editor-standalone/tests-e2e/__fixtures__/editor.ts
```diff
@@ -0,0 +1,138 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Page, expect } from "@playwright/test";
+import type { open } from "../../src/index";
+import { ContentType } from "@kie-tools-core/workspace/dist/api";
+import { DmnEditorStandaloneApi } from "../../src/DmnEditorStandaloneApi";
```
ljmotta: We don't couple the `src` code with the `tests`. Currently, we're duplicating the code on the test side.

packages/dmn-editor-standalone/tests-e2e/__fixtures__/editor.ts
```diff
@@ -0,0 +1,138 @@
...
+      window.currentEditor.close();
+    });
+  }
+
+  public getEditorIframe() {
+    return this.page.frameLocator("#dmn-editor-standalone");
+  }
+
+  public getEditorDiagram() {
+    return this.getEditorIframe().getByTestId("kie-dmn-editor--diagram-container");
```
ljmotta: Could you please update it to `kie-tools--dmn-editor--diagram-container`?

packages/patternfly-base/index.js
ljmotta, line 1: 👍

packages/dmn-editor-standalone/tests-e2e/api/getContent.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+        parser: "xml",
+      });
+
+      const formattedLoanPreQualificationDmn = await prettier.format(loanPreQualificationDmn, {
+        ...(await prettier.resolveConfig(".")),
+        parser: "xml",
+      });
+
+      await expect(formattedContent).toEqual(formattedLoanPreQualificationDmn);
+    });
```
ljmotta: Currently, we are avoiding creating variables and additional logic on the E2E test to improve readbility and reduce cogntive load. The idea is each line of the test should be one action on the screen. To achieve this, you could:

Add the `externalModels`  on a `files` fixture, to be added to the `base`. Additionally, the `prettier.format`, could be called on the same file on a different method, for example: `getLoanPreQualification` and `getFormattedLoadPreQualification`. Also, the editor fixture could already have a `getFormattedContent` method. This way we would simplify the test code as follow:

```ts
test("should get DMN contents of input DMN file", async ({ editor, files }) => {
     await editor.setContent("loanPreQualification.dmn", await files.getLoanPreQualification());

     await expect(editor.getEditorIframe().getByText("Loan Pre-Qualification", { exact: true })).toBeAttached();
     await expect(await editor.getFormattedContent()).toEqual(await files.getFormattedLoanPreQualification());
})

File: packages/dmn-editor-standalone/tests-e2e/api/getContent.spec.ts [hunk 12]
```diff
@@ -0,0 +1,101 @@
...
+
+      await expect(async () => {
+        const dmnContentWith4Inputs2Decisions = await editor.getContent();
+        await expect((dmnContentWith4Inputs2Decisions.match(/<inputData/gm) || []).length).toBe(3);
+        await expect((dmnContentWith4Inputs2Decisions.match(/<decision/gm) || []).length).toBe(2);
+      }).toPass({
+        intervals: [100, 250, 500, 1000],
+        timeout: 10_000,
+      });
+    });
```
ljmotta: The same will apply for this test, to avoid creating variables and the adding the logic, you could import the `palette` and `nodes` fixtures from `dmn-editor`. It would be necessary to change both fixtures `page` type to `Page | Frame` and to initialize both you pass `editor.getFrame() ?? page`. This should do the trick. Here is how it should look like:

```ts
test("should get current DMN contents via getContent", async ({ palette, editor, nodes }) => {
  await palette.dragNewNode({ type: NodeType.INPUT_DATA, targetPosition: { x: 100, y: 100 }, thenRenameTo: "Input-A"})
  await palette.dragNewNode({ type: NodeType.INPUT_DATA, targetPosition: { x: 300, y: 100 }, thenRenameTo: "Input-B"})
  await palette.dragNewNode({ type: NodeType.INPUT_DATA, targetPosition: { x: 500, y: 100 }, thenRenameTo: "Input-C"})
  await palette.dragNewNode({ type: NodeType.INPUT_DATA, targetPosition: { x: 700, y: 100 }, thenRenameTo: "Input-D"})

  await expect(nodes.get({ name: "Input-A" })).toBeAttached();
  await expect(nodes.get({ name: "Input-B" })).toBeAttached();
  await expect(nodes.get({ name: "Input-C" })).toBeAttached();
  await expect(nodes.get({ name: "Input-D" })).toBeAttached();

  await palette.dragNewNode({ type: NodeType.DECISION, targetPosition: { x: 100, y: 300 }, thenRenameTo: "Decision-A"})
  await palette.dragNewNode({ type: NodeType.DECISION, targetPosition: { x: 300, y: 300 }, thenRenameTo: "Decision-B"})

  await expect(nodes.get({ name: "Decision-A" })).toBeAttached();
  await expect(nodes.get({ name: "Decision-B" })).toBeAttached();
  await expect(async () => {
    expect(await editor.getInputDataNodesQtt()).toBe(4);
    expect(await editor.getDecisionNodesQtt()).toBe(2);
  }).toPass({
    intervals: [100, 250, 500, 1000],
    timeout: 10_000,
  });

  await nodes.delete({ name: "Input-A"})

  await expect(nodes.get({ name: "Input-A" })).toBeAttached();
  await expect(nodes.get({ name: "Input-B" })).toBeAttached();
  await expect(nodes.get({ name: "Input-C" })).toBeAttached();
  await expect(async () => {
    expect(await editor.getInputDataNodesQtt()).toBe(3);
    expect(await editor.getDecisionNodesQtt()).toBe(2);
  }).toPass({
    intervals: [100, 250, 500, 1000],
    timeout: 10_000,
  });
});
```

As you can see of the above example, I've also hided the `regex` logic inside the `editor` fixture.

packages/dmn-editor-standalone/tests-e2e/api/openAndClose.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+      await editor.open();
+    });
+
+    test("should open and close the editor with a blank DMN", async ({ page, editor }) => {
+      await expect(page).toHaveScreenshot("open-editor.png");
+
+      await editor.close();
+
+      await expect(editor.getEditorIframe().getByText("This DMN's Diagram is empty")).not.toBeAttached();
+      await expect(page).toHaveScreenshot("close-editor.png");
```
ljmotta: The `close-editor` screenshot is a blank screenshot, maybe we could just assert if the `editor.getEditorIframe()` has anything inside it? I'm not sure if the `close` de attach everything.

packages/dmn-editor-standalone/tests-e2e/api/getPreview.spec.ts
ljmotta, line 1: Here you could inline all variables as they are used once.

packages/dmn-editor-standalone/tests-e2e/api/markAsSaved.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      const inputSelector = editorIFrame.getByTitle("Input Data", { exact: true });
+      const editorDiagram = editor.getEditorDiagram();
+
+      // Add 4 Input Data nodes and check if dirty count increases to 4
+      for (let i = 0; i < 4; i++) {
+        await inputSelector.dragTo(editorDiagram, { targetPosition: { x: 100 + i * 200, y: 100 } });
+        await editor.resetFocus();
+      }
+      await expect(await page.locator("#edit-counter")).toHaveText("4");
+      await expect(await page.locator("#is-dirty")).toHaveText("true");
```
ljmotta: I think it's missing a fixture to encapsulate all interaction with the hidden divs. I'm not sure which name it would be good for it, but when we use such locators it's better to put them in just one place, so we can have a better control over it.

packages/dmn-editor-standalone/tests-e2e/api/subscribeToContentChanges.spec.ts
ljmotta, line 1: The same regarding the fixtures applies to these tests.

packages/dmn-editor-standalone/tests-e2e/api/undoAndRedo.spec.ts
ljmotta, line 1: Same

ljmotta (changes requested):
@thiagoelg The PR is looking great! The majority of my comments relies on the tests. Generally speaking, we try to avoid creating variables on the tests as I've add on one of my comments. Also, all the necessary logic to perform an action, and very specific locators we try to hide it behind fixtures so we can improve the readability of the test.

packages/dmn-editor-standalone/tests-e2e/api/getContent.spec.ts
ljmotta: Actually, you can't change the `page` property to `Page | Frame`, as the Frame doesn't have a few methods. It will be necessary to add a new property:

```ts
export class Nodes {
  constructor(
    public page: Page,
    public diagram: Diagram,
    public browserName: string,
    public nodesLocation: Page | Frame
  ) {}

  public get(args: { name: string }) {
    return this.nodesLocation.locator(`div[data-nodelabel="${args.name}"]`)!;
  }
```

packages/dmn-editor-standalone/src/index.ts
thiagoelg: This library is built in two stages: first, the envelope, then the library itself.

The reason for this is so that we can inline the envelop JS into the iframe created by the library when opening the editor.

If you check the `pacakge.json`, you can see that the build is defined as `webpack --config webpack.envelope-config.js --env dev && webpack --env dev`.

packages/dmn-editor-standalone/stories/DevWebApp.stories.tsx
thiagoelg: Mainly because of the two-stage build.
The DMN Editor Standalone is a webapp contained in a JS library, and unfortunately, storybook can't just import its source code and build it.

What I'm trying right now is using `nodemon` + `concurrently` to watch the source code and rebuild it while the storybook is running (we already do something similar on runtime-tools web apps).

packages/dmn-editor-standalone/tests-e2e/__fixtures__/editor.ts
thiagoelg: Even types? Why?

Imagine replicating all the types required to make the `DmnEditorStandaloneApi` > `EditorApi`...

Is it ok to replicate only the types of the current package?

packages/dmn-editor-standalone/env/index.js
thiagoelg: It's used in webpack.config.js if anyone wants to run `pnpm exec webpack` one day

packages/dmn-editor-standalone/tests-e2e/__fixtures__/editor.ts
thiagoelg: In the `dmn-editor` package?

ljmotta: Yes, in all places.

packages/dmn-editor-standalone/tests-e2e/__fixtures__/editor.ts
ljmotta: The idea behind this is end to end tests should perform tests without knowing what is on the code. But, looking at the other tests on other packages, we import types from other packages, even though we're going to fall at the same problem. 🤔 

I guess we could import from the `dist`. So it would be `@kie-tools/dmn-editor-standalone/dist/...`.

FYI @tiagobento

thiagoelg: In this case, it's a JS library, just like the `dmn-editor`; everyone who uses it will have access to its types, so it makes sense that the test can also use them.

ljmotta: Yeah.. You've conviced me 😅. After thinking about it, I don't think it's a problem for types. But please, import from the `dist`. The end to end tests should run after the app was built.

ljmotta (approved):

PR #2493, opened 2024-07-29
NO-ISSUE: Upgrade Angular to v18 and use local BPMN editor on `examples/webapp`

thiagoelg:
- Upgraded Angular to `18.1.2`;
- Simplified the `ping-pong-view-angular` implementation;
- Fixed `zone.js` polyfill messing up with Promise prototype;
- Changed BPMN Editor on `examples/webapp` to use the local version instead of the latest published release.


tiagobento:
@thiagoelg Can we update syncpack.json to remove the exception now that everything is upgraded?

tiagobento (approved):

ljmotta (approved):

PR #2508, opened 2024-08-02
kie-issues#1298: Decision Services & multiple DRDs: make it possible o add external Decisions to a Decision Service

### LINKED ISSUE #1298: Decision Services & multiple DRDs: make it possible to add external Decisions to a Decision Service
Decision Service (DS) cases that should be handled:

 - Add an external output Decision to a DS;
 - Add an external encapsulated Decision to a DS;
 - Add an external Decision to a DS which is used on an external DS. Add the external DS to the DRD (Conflict);
 - Add an external DS, and add a external Decision to a DS which is used on the external DS to the DRD (Conflict); 
 - Add an external DS which uses a Decision from another external model. (3 DMNs required);

For all those cases, the Decision and DS should respect the current features.
 - Move
 - Resize
 - Rename
 - Delete
 - Hide (shouldn't be possible to hide a Decision node that is from a DS)

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1298

https://github.com/user-attachments/assets/2f14961c-28ea-4d38-a5c9-98f8580c9026


danielzhe:
More testes for you, @ljmotta 😄

packages/dmn-editor/stories/dev/DevWebApp.stories.tsx
```diff
@@ -18,20 +18,19 @@
  */
 
 import * as React from "react";
-import type { Meta, StoryObj } from "@storybook/react";
 import { useCallback, useMemo, useRef, useState } from "react";
+import type { Meta, StoryObj } from "@storybook/react";
```
danielzhe, line 22: Those changes are made automatically by the Prettier

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
```diff
@@ -25,25 +25,42 @@ import { SnapGrid } from "../store/Store";
...
 }: {
   definitions: Normalized<DMN15__tDefinitions>;
-  decisionId: string;
+  drgElement: DrgElement;
   decisionServiceId: string;
   drdIndex: number;
   snapGrid: SnapGrid;
+  decisionShape: Normalized<DMNDI15__DMNShape>;
+  elementId: string;
 }) {
```
tiagobento: `elementId` --> `decisionId`

I understand `drgElement` doesn't necessarily is inside `definitions`, so please make this action dependent on external models too. There's no way to know whether or not `drgElement` and `elementId` are aligned...

packages/dmn-editor/src/diagram/Palette.tsx
```diff
@@ -245,7 +245,11 @@ export function Palette({ pulse }: { pulse: boolean }) {
         <br />
         <aside className={"kie-dmn-editor--external-nodes-panel-toggle"}>
           {diagram.openLhsPanel === DiagramLhsPanel.EXTERNAL_NODES && (
-            <div className={"kie-dmn-editor--palette-nodes-popover"} style={{ maxHeight }}>
+            <div
+              className={"kie-dmn-editor--palette-nodes-popover"}
+              style={{ maxHeight }}
+              data-testid={"kie-tools--dmn-editor--external-nodes-container"}
+            >
```
tiagobento: The "DRG nodes" panel doesn't have a `data-testid`, why don't we use the same strategy we're using for it?

packages/dmn-editor/src/diagram/Diagram.tsx
```diff
@@ -1009,12 +1009,14 @@ export const Diagram = React.forwardRef<DiagramRef, { container: React.RefObject
...
                   drdIndex: state.computed(state).getDrdIndex(),
-                  decisionId: selectedNodes[i].data.dmnObject!["@_id"]!, // We can assume that all selected nodes are Decisions because the contaiment was validated above.
+                  drgElement: selectedNodes[i].data.dmnObject as DrgElement,
                   decisionServiceId: state
                     .computed(state)
                     .getDiagramData(externalModelsByNamespace)
                     .nodesById.get(dropTargetNode.id)!.data.dmnObject!["@_id"]!,
                   snapGrid: state.diagram.snapGrid,
+                  decisionShape: selectedNodes[i].data.shape,
+                  elementId: selectedNodes[i].id, // The "real" id is here, which can be the local id or an imported node id (uri + external element id).
```
tiagobento: Please see my comment about the mutation itself...

packages/dmn-editor/tests-e2e/drgElements/addIncludedNodes.spec.ts
tiagobento, line 1: Can we rename this to `externalNodes`? I mean, this is the first test we have on that "domain".

packages/dmn-editor/tests-e2e/drgElements/addIncludedNodes.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    await includedModels.getIncludeModelButton().click();
+    await includedModels.fillModelToInclude({ modelName: "sumDiffDs.dmn" });
+    await includedModels.selectModel({ modelName: "sumDiffDs.dmn" });
+    await includedModels.fillModelName({ modelName: "MY_MODEL" });
+    await includedModels.includeModel();
+
+    await editor.changeTab({ tab: TabName.EDITOR });
+    await palette.dragNewNode({ type: NodeType.DECISION_SERVICE, targetPosition: { x: 400, y: 200 } });
+
+    await palette.getExternalNodesButton().click();
```
tiagobento: I think this is leaking internal details... I would expect something like `palette.openExternalNodesPanel()`.

packages/dmn-editor/tests-e2e/drgElements/addIncludedNodes.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+test.beforeEach(async ({ stories }) => {
+  await stories.openEmptyWithAvailableExternalModels();
+});
+
+test.describe("Add included node - Decision", () => {
+  test("add to a Decision Service", async ({ editor, page, palette, diagram, nodes, includedModels }) => {
+    await editor.changeTab({ tab: TabName.INCLUDED_MODELS });
+
+    await includedModels.getIncludeModelButton().click();
```
tiagobento: I think this is leaking internal details... I would expect something like includedModels.addNew(). Not sure how this is being done in other parts of our test suite, though.

packages/dmn-editor/tests-e2e/drgElements/addIncludedNodes.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      includedModelName: "MY_MODEL",
+      nodeName: "Diff",
+      targetPosition: {
+        x: 300,
+        y: 300,
+      },
+    });
+
+    // We click on the button again to close the external nodes panel.
+    await palette.getExternalNodesButton().click();
```
tiagobento: Same? Maybe instead of `openExternalNodesPanel` you could call it `toggleExternalNodesPanel`

packages/dmn-editor/tests-e2e/drgElements/addIncludedNodes.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    await diagram.resetFocus();
+    await nodes.move({ name: "Sum", targetPosition: { x: 500, y: 300 } });
+    await nodes.move({ name: "Diff", targetPosition: { x: 500, y: 430 } });
+
+    await diagram.resetFocus();
+
+    // We move it to make it sure that the nodes are added inside the Decision Services
+    // and are really attached to it, not only "visually over it".
+    await nodes.move({ name: DefaultNodeName.DECISION_SERVICE, targetPosition: { x: 120, y: 120 } });
```
tiagobento: Great comment.

packages/dmn-editor/tests-e2e/drgElements/addIncludedNodes.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    await expect(diagram.get()).toHaveScreenshot("add-included-node-inside-decision-service.png");
+  });
+});
```
tiagobento: Please add more tests to the basic operations, like changing from one DS to another. Removing an external Decision from a DS, things like that.

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
```diff
@@ -140,9 +140,19 @@ export class Nodes {
...
-      targetPosition: args.targetPosition,
-    });
+    if (DefaultNodeName.DECISION_SERVICE === args.name) {
+      // Decision Services only have some draggable areas near the borders.
+      // If you drag it to the center, you'll drag the divide line.
+      // Also, neither the entire upper area nor the entire downer area is draggable.
+      await this.get({ name: args.name }).dragTo(this.diagram.get(), {
+        targetPosition: args.targetPosition,
+        sourcePosition: { x: 20, y: 20 },
+      });
```
tiagobento, line 150: Great comment also.

tiagobento (changes requested):
Thanks for the PR @danielzhe! Made some change requests inline, please consider them.

jomarko:
I will review this tomorrow.

packages/dmn-editor/tests-e2e/drgElements/addIncludedNodes.spec.ts
danielzhe: I did as you suggested at first, but then I found in other tests that we were using this way, getting the button and then calling the click(), like [here](https://github.com/apache/incubator-kie-tools/blob/b7cdf3caa49dde144cd545e8bb4ba68905bf438d/packages/dmn-editor/tests-e2e/checkDecisionTableCellsDataType.spec.ts#L238), for example, and changed to this "clicky" approach. I will change it back. 😄

packages/dmn-editor/src/diagram/Palette.tsx
danielzhe: Do we have tests for "DRG Nodes" panel? I only found for drag nodes from the palette not from the "DRG Nodes" panel. 🤔

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
danielzhe: In the previous code we was getting the ID, in `Diagram.tsx` using the following code:

```
decisionId: selectedNodes[i].data.dmnObject!["@_id"]!, // We can assume that all selected nodes are Decisions because the contaiment was validated above.
```

So I think it is safe to assume that `selectedNodes[i].data.dmnObject` is always a `Decision` which is the `DrgElement` that we looking for. Maybe I should change the type of `DrgElement` I defined [here](https://github.com/apache/incubator-kie-tools/pull/2508/files#diff-b544ce6235db5e0b538561f2aac58e54da6162dac8a58fcb39cee682e8420f63R36) 
to:

`export type DecisionDrgElement = Normalized<{ __$$element: "decision" } & DMN15__tDecision>`

About the ID, I can't see any case where the ID of the selected node will be different than the DRG, except for the imported nodes the ID in the DRG is not full, it is only the local ID (the Guid), but from `selectedNodes[I].["@_id"]` is the real ID used in the current diagram (namespace + id).

danielzhe: In fact, we don't have DRG element always in the definitions: the external nodes do not have DRG element, only DMNDI. The DRG we have in `selectedNodes[i].data.dmnObject` is a loaded DRG from the external model, that's why its ID is different to the internal one, the one in href. 
See below:

```
  <decisionService name="New Decision Service" id="_A0908B2B-DBE7-41C7-B055-0C0DAD17CE54">
    <variable name="New Decision Service" id="_145CD7A7-B14D-4D63-B9D3-85601FC915B1" />
    <outputDecision href="https://kie.apache.org/dmn/_D19B0015-2CBD-4BA8-84A9-5F554D84A9E1#_05621ED4-9236-47F1-B93A-164A4527B136" />
    <encapsulatedDecision href="https://kie.apache.org/dmn/_D19B0015-2CBD-4BA8-84A9-5F554D84A9E1#_1991FB34-1253-4A54-AD3D-89697938DDFA" />
    <encapsulatedDecision href="#_8E89DDF6-7ABC-4FBF-A1B6-B68BC32A52AE" />
  </decisionService>
  <decision name="New Decision" id="_8E89DDF6-7ABC-4FBF-A1B6-B68BC32A52AE">
    <variable name="New Decision" id="_BFB8A43B-5ADA-47D8-815E-8A0505814AA1" />
  </decision>
```

Notice only the `8E89DDF6-7ABC-4FBF-A1B6-B68BC32A52AE` is present in the file. 
The other nodes are not, only as a `dmndi`.

packages/dmn-editor/src/diagram/Palette.tsx
jomarko: @tiagobento actually we are about to add `data-testid` also for the `DRG Nodes` panel, see the PR https://github.com/apache/incubator-kie-tools/pull/2462

One thing we could unify is the `data-testid` suffix. @danielzhe is using `-container` while me is using `-popover`, what do you prefer? I chose `-popover` due to related `className`.

packages/dmn-editor/src/externalNodes/ExternalNodesPanel.tsx
```diff
@@ -159,6 +159,7 @@ export function ExternalNodesPanel() {
                           externalDrgElementId: drgElement["@_id"]!,
                         })
                       }
+                      data-testid={`kie-tools--dmn-editor--external-node-${_import["@_name"]}-${drgElement["@_name"]}`}
```
jomarko: Is ${_import["@_name"]} always non empty? I mean containing some non white characters?

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
jomarko: Unfortunately, we have one more problem, that `node.move()` can be invoked also for a decision service node, that is already renamed and is completely different from `DefaultNodeName.DECISION_SERVICE`

I think we should also document that the decision service move will work only for 'default name' decision service nodes.

packages/dmn-editor/tests-e2e/__fixtures__/palette.ts
```diff
@@ -40,6 +40,21 @@ export class Palette {
...
+  }) {
+    await this.page
+      .getByTestId("kie-tools--dmn-editor--external-nodes-container")
+      .getByTestId(`kie-tools--dmn-editor--external-node-${args.includedModelName}-${args.nodeName}`)
+      .dragTo(this.diagram.get(), { targetPosition: args.targetPosition });
+  }
+
+  public async toggleExternalNodesPanel() {
+    await this.page.getByRole("button", { name: "External nodes" }).click();
+  }
```
jomarko: In my PR https://github.com/apache/incubator-kie-tools/pull/2462, I added similar as separate fixture `drgNodes.ts`. Should I also add add it into `palette.ts` instead of a separate fixture file?

packages/dmn-editor/tests-e2e/__screenshots__/Google-Chrome/drgElements/move-external-decision-out-from-decision-service-after.png
jomarko, line 1: Maybe it is not so clear , if the `MY_MODEL_Diff` is really out of the decision service node. Could we maybe move it a little bit more to make it clear? Or is this something like corner case test?

packages/dmn-editor/tests-e2e/drgElements/addExternalNode.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  await stories.openEmptyWithAvailableExternalModels();
+});
+
+test.describe("Add external node - Decision", () => {
+  test("add to a Decision Service", async ({ editor, page, palette, diagram, nodes, includedModels }) => {
+    await editor.changeTab({ tab: TabName.INCLUDED_MODELS });
+
+    await includedModels.getIncludeModelButton().click();
+    await includedModels.fillModelToInclude({ modelName: "sumDiffDs.dmn" });
+    await includedModels.selectModel({ modelName: "sumDiffDs.dmn" });
```
jomarko, line 34: Why both `fillModelToInclude` and `selectModel` are needed please?

jomarko (commented):
Some comments inline, however manual check passed and the feature works well!

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
tiagobento: We only need the ID, but we need to know that this ID represents a Decision. The only way to know that is having the externalModelsByNamespace passed as parameter to the "addDecisionToDecisionService" mutation, so we can make that validation. You're mentioning the arguments you're passing when calling this mutation, but the mutation itself, after your changes, can be called with 

```typescript
{ 
  elementId: "bar"
  drgElement: { 
    "@_id": "foo" 
    ...
  }
}
```

which has the potential of being wrong. What I'm saying is that we rollback to the way it was before, and we include the `externalModelsByNamespace` parameters so we can look if that href (`nmspc#id`) represents a Decision on the model with namespace `nmspc`

packages/dmn-editor/src/diagram/Palette.tsx
tiagobento: All good then! Thanks for the insights. I have no preference for the suffix, just please coordinate that it is consistent. @jomarko @danielzhe

packages/dmn-editor/tests-e2e/drgElements/addIncludedNodes.spec.ts
tiagobento: Yeah, I like it better when the actual screen elements are not exposed, to be honest, since that's how we're doing for the other parts of our tests.

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
tiagobento: Good point @jomarko. @danielzhe I think we can update this to have a flag "grabAtBorders" that we use as `true` when moving Decision Services around.

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
danielzhe: I think got your point! I guarantee in the `Diagram.tsx` that the ID represents a decision, but I am not doing the same inside the mutation, in other words, the mutation is not "safe" anymore, right?

packages/dmn-editor/src/externalNodes/ExternalNodesPanel.tsx
tiagobento: No, it can be an empty string. Great catch.

packages/dmn-editor/tests-e2e/__screenshots__/Google-Chrome/drgElements/move-external-decision-out-from-decision-service-after.png
tiagobento: Agree, this screenshot is kind of ambiguous...

packages/dmn-editor/tests-e2e/drgElements/addExternalNode.spec.ts
tiagobento: One "types" the string, and the other picks it from the list.

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
danielzhe: I made a recent change where the name must now contain `DefaultNodeName.DECISION_SERVICE`, but not exactly it. But I agree, it isn't good to rely on the name. What we can do? A separate "move" for Decision Service or add a doc in the code? 

We also can just change the `move` to always get the node from a corner not from the middle. Then it will work for every node. Can add an optional parameter where the user can add the `sourcePosition` also. I think this is the best option. What do you guys think? @jomarko @tiagobento

packages/dmn-editor/src/diagram/Palette.tsx
danielzhe: I'll change it to `-popover` since it is related to the `className` and makes more sense! Thank you, @jomarko

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
tiagobento: Yes, you can guarantee that you're calling it correctly now, but you can't guarantee that this will be true forever :P

packages/dmn-editor/tests-e2e/__fixtures__/nodes.ts
tiagobento:
```typescript
  public async move(args: { grabAtBorder?: boolean; targetPosition: { x: number; y: number } }) {
    if (args.grabAtBorder ?? false) {
      // Decision Services only have some draggable areas near the borders.
      // If you drag it to the center, you'll drag the divide line.
      // Also, neither the entire upper area nor the entire downer area is draggable.
      await this.get({ name: args.name }).dragTo(this.diagram.get(), {
        targetPosition: args.targetPosition,
        sourcePosition: { x: 20, y: 20 },
      });
    } else {
      await this.get({ name: args.name }).dragTo(this.diagram.get(), {
        targetPosition: args.targetPosition,
      });
    }
  }
```

packages/dmn-editor/src/diagram/Diagram.tsx
```diff
@@ -992,8 +992,9 @@ export const Diagram = React.forwardRef<DiagramRef, { container: React.RefObject
                 for (let i = 0; i < selectedNodes.length; i++) {
                   deleteDecisionFromDecisionService({
                     definitions: state.dmn.model.definitions,
-                    decisionId: selectedNodes[i].data.dmnObject!["@_id"]!, // We can assume that all selected nodes are Decisions because the contaiment was validated above.
+                    decisionId: selectedNodes[i].id, // We can assume that all selected nodes are Decisions because the contaiment was validated above.
```
danielzhe: This is the full id. The ID of the ` selectedNodes[i].data.dmnObject!["@_id"]!` does not contain the href to the external model, only the "internal" id, so inside the mutation we would have to look everywhere looking for where is the `Decision`.

packages/dmn-editor/src/externalNodes/ExternalNodesPanel.tsx
```diff
@@ -136,7 +136,7 @@ export function ExternalNodesPanel() {
               const _import = importsByNamespace.get(namespace);
               if (!_import) {
                 console.debug(
-                  `DMN EDITOR: Couldn't find import for namespace '${namespace}', although there's an external DMN referncing it.`
+                  `DMN EDITOR: Couldn't find import for namespace '${namespace}', although there's an external DMN referencing it.`
```
danielzhe, line 139: Just a typo

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
```diff
@@ -25,25 +25,48 @@ import { SnapGrid } from "../store/Store";
...
   drdIndex: number;
   snapGrid: SnapGrid;
+  externalModelsByNamespace: ExternalModelsIndex | undefined;
 }) {
   console.debug(`DMN MUTATION: Adding Decision '${decisionId}' to Decision Service '${decisionServiceId}'`);
 
-  const decision = definitions.drgElement?.find((s) => s["@_id"] === decisionId);
-  if (decision?.__$$element !== "decision") {
-    throw new Error(`DMN MUTATION: DRG Element with id '${decisionId}' is either not a Decision or doesn't exist.`);
+  const href = parseXmlHref(decisionId);
```
danielzhe: So `decisionId` can be in any format here:
`#id`
`uri#id`
`id`

But, if it is an external node, we only be able to find it if the full id is passed: `uri#id`

danielzhe:
@tiagobento @jomarko Changes applied! Ready for you to continue.

packages/dmn-editor/src/diagram/Diagram.tsx
tiagobento: - "Full ID" = href
- href = `namespace#xsd:ID`
- xsd:ID = typically the `"@_id"` attribute of objects
- namespace = `definitions["@_namespace"]`

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
tiagobento: Good point. I think we better rename `decisionId` to `decisionHref`.

tiagobento: ...and the `id` of a node is an href.... please see https://github.com/apache/incubator-kie-issues/issues/983.... :x

packages/dmn-editor/src/externalNodes/ExternalNodesPanel.tsx
```diff
@@ -159,6 +159,7 @@ export function ExternalNodesPanel() {
                           externalDrgElementId: drgElement["@_id"]!,
                         })
                       }
+                      data-testid={`kie-tools--dmn-editor--external-node-${_import["@_name"] === "" ? _import["@_id"] : _import["@_name"]}-${drgElement["@_name"]}`}
```
tiagobento, line 162: Why dont' we use the FEEL name of the node? If it's imported without a name we just use `{nodeName}`, if its import has a name, we do `{importName}.{nodeName}`?

packages/dmn-editor/src/diagram/Diagram.tsx
```diff
@@ -1009,12 +1010,13 @@ export const Diagram = React.forwardRef<DiagramRef, { container: React.RefObject
...
                   definitions: state.dmn.model.definitions,
                   drdIndex: state.computed(state).getDrdIndex(),
-                  decisionId: selectedNodes[i].data.dmnObject!["@_id"]!, // We can assume that all selected nodes are Decisions because the contaiment was validated above.
+                  decisionHref: selectedNodes[i].id, // We can assume that all selected nodes are Decisions because the contaiment was validated above.
                   decisionServiceId: state
                     .computed(state)
                     .getDiagramData(externalModelsByNamespace)
                     .nodesById.get(dropTargetNode.id)!.data.dmnObject!["@_id"]!,
                   snapGrid: state.diagram.snapGrid,
+                  externalModelsByNamespace: externalModelsByNamespace,
```
tiagobento: Just for conciseness... 
```suggestion
                  externalModelsByNamespace,
```

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
```diff
@@ -65,10 +91,10 @@ export function addDecisionToDecisionService({
   const section = getSectionForDecisionInsideDecisionService({ decisionShape, decisionServiceShape, snapGrid });
   if (section === "encapsulated") {
     decisionService.encapsulatedDecision ??= [];
-    decisionService.encapsulatedDecision.push({ "@_href": `#${decisionId}` });
+    decisionService.encapsulatedDecision.push({ "@_href": `${hrefString}` });
   } else if (section === "output") {
     decisionService.outputDecision ??= [];
-    decisionService.outputDecision.push({ "@_href": `#${decisionId}` });
+    decisionService.outputDecision.push({ "@_href": `${hrefString}` });
```
tiagobento: Nice

packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts
```diff
@@ -44,11 +66,10 @@ export function deleteDecisionFromDecisionService({
...
 
-  decisionService.outputDecision = (decisionService.outputDecision ?? []).filter(
-    (s) => s["@_href"] !== `#${decisionId}`
-  );
+  const xmlHref = buildXmlHref({ namespace: href.namespace, id: href.id });
+  decisionService.outputDecision = (decisionService.outputDecision ?? []).filter((s) => s["@_href"] !== `${xmlHref}`);
   decisionService.encapsulatedDecision = (decisionService.encapsulatedDecision ?? []).filter(
-    (s) => s["@_href"] !== `#${decisionId}`
+    (s) => s["@_href"] !== `${xmlHref}`
   );
```
tiagobento: Nice

packages/dmn-editor/stories/dev/DevWebApp.stories.tsx
tiagobento: I think Prettier doesn't change the imports... Maybe it's your VS Code "Organize imports"?

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
```diff
@@ -25,25 +25,48 @@ import { SnapGrid } from "../store/Store";
...
-  if (decision?.__$$element !== "decision") {
-    throw new Error(`DMN MUTATION: DRG Element with id '${decisionId}' is either not a Decision or doesn't exist.`);
+  const href = parseXmlHref(decisionHref);
+
+  const externalModel = externalModelsByNamespace?.[href.namespace ?? ""];
+  if (href.namespace && !externalModel) {
+    throw new Error(`DMN MUTATION: Namespace '${href.namespace}' not found.`);
+  }
+
+  if (externalModel && (externalModel.model as Normalized<DmnLatestModel>).definitions) {
```
tiagobento:
```suggestion
  if (href.namespace) {
```

?

I mean, we only want to enter this branch if it is an external node... Right?

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
```diff
@@ -25,25 +25,48 @@ import { SnapGrid } from "../store/Store";
...
+    const externalDrgs = (externalModel.model as Normalized<DmnLatestModel>).definitions.drgElement;
+    const decision = externalDrgs?.find((drgElement) => drgElement["@_id"] === href.id);
```
tiagobento: Personal preference, but I always think we should avoid intermediate variables to avoid giving it a name. It's one more thing to hold onto my head when reading code...

packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts
```diff
@@ -20,21 +20,43 @@
...
-  if (decision?.__$$element !== "decision") {
-    throw new Error(`DMN MUTATION: DRG Element with id '${decisionId}' is either not a Decision or doesn't exist.`);
+  const href = parseXmlHref(decisionHref);
+
+  const externalModel = externalModelsByNamespace?.[href.namespace ?? ""];
+  if (href.namespace && !externalModel) {
+    throw new Error(`DMN MUTATION: Namespace '${href.namespace}' not found.`);
+  }
+
+  if (externalModel && (externalModel.model as Normalized<DmnLatestModel>).definitions) {
```
tiagobento:
```suggestion
  if (href.namespace) {
```

Same as above.

packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts
```diff
@@ -20,21 +20,43 @@
...
+    const externalDrgs = (externalModel.model as Normalized<DmnLatestModel>).definitions.drgElement;
+    const decision = externalDrgs?.find((drgElement) => drgElement["@_id"] === href.id);
```
tiagobento: Also same as above.

tiagobento (changes requested):
Thanks @danielzhe! This is looking great. Few last comments from my side.

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
tiagobento: (Not that your code is wrong, it's just longer)

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
danielzhe: This is a very good topic.
I like to split in variables because I find it easier to "think" step-by-step and have less complexity per line.

Also, in this case, I'm giving "meaning" to what `(externalModel.model as Normalized<DmnLatestModel>).definitions.drgElement`: it is the "external drgs". So I don't have to care where it comes from in line 59 or what it is. It is the "external drgs that come from somewhere". Ok, "somewhere" is just a line above, but in my mind, looking at line 59, it is a "problem solved" that I don't have to care about anymore.

But the main point is that if someone new looks at the code it can easily get what `(externalModel.model as Normalized<DmnLatestModel>).definitions.drgElement` means.

danielzhe:
> I think Prettier doesn't change the imports... Maybe it's your VS Code "Organize imports"?

It could be. I didn't notice that change in others PR that I did. I'll check it.

packages/dmn-editor/src/externalNodes/ExternalNodesPanel.tsx
danielzhe: Because the `importName` is a long URI it is annoying to test that.

tiagobento (approved):

danielzhe:
Waiting for you @jomarko when you can re-check it. Thank you! 😄

packages/dmn-editor/tests-e2e/__fixtures__/palette.ts
```diff
@@ -40,6 +40,21 @@ export class Palette {
...
   }
 
+  public async dragExternalNode(args: {
+    includedModelName: string;
+    nodeName: string;
+    targetPosition: { x: number; y: number };
+  }) {
+    await this.page
+      .getByTestId("kie-tools--dmn-editor--external-nodes-popover")
+      .getByTestId(`kie-tools--dmn-editor--external-node-${args.includedModelName}-${args.nodeName}`)
```
jomarko, line 50: Do we want somehow reflect the logic we have in the implementation?
```
data-testid={`kie-tools--dmn-editor--external-node-${_import["@_name"] === "" ? _import["@_id"] : _import["@_name"]}-${drgElement["@_name"]}`}
```

I mean sometime included mdoel may be empty

jomarko: Also a reminder about my previous question, if this is the code placement for similar methods that I should use in my PR

https://github.com/apache/incubator-kie-tools/pull/2508#discussion_r1704999440

packages/dmn-editor/tests-e2e/drgElements/moveExternalNode.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    palette,
+    diagram,
+    nodes,
+    includedModels,
+  }) => {
+    await editor.changeTab({ tab: TabName.INCLUDED_MODELS });
+
+    await includedModels.getIncludeModelButton().click();
+    await includedModels.fillModelToInclude({ modelName: "sumDiffDs.dmn" });
+    await includedModels.selectModel({ modelName: "sumDiffDs.dmn" });
```
jomarko, line 40: sharing my opinion here, but will not block PR if you decide to not apply it.

My understanding is we will always call these methods as a pair, that needs to be executed to include a model. As both are related to the same application logic and we invoke both with the same parameter value, to me it would make sense to simplify the `includedModels` fixture API to have a single method `selectModel` that would internally also `fillModelToInclude`.

jomarko (commented):
@danielzhe , thank you for the updates. I think the PR is in a fantastic shape, I put some minor comments to the code inline. Feel free to incorporate them or not.

There are two more findings that I have from the manual review, that I am not sure how much relevant to this PR are or we should handle them in a separate ticket or they are not issues, only my wrong understanding.

### 01
The string we print in the properties panel to demonstrate how to invoke the Decision Service node seems to ignore external Input Data nodes. See the attached screenshot where `log.A` and `log.B` are external Input Data nodes.
Screenshot 2024-08-09 090144

### 02
There seems to be a problem with execution of the model, that we started to support. I mean, if user want to execute a model, that contains single decision service (not external) and this external service is implemented using external Decision node and external Input data nodes, user is not able to execute such model using the DMN Runner. User is able to execute model with the same graph structure if all nodes are not external.

Screenshot 2024-08-09 090723

danielzhe:
Thank you, @jomarko ! Good points here.


Yep, this is an issue, but it seems to be a runner issue. It only works if the included decision does not have an input node.

Check this. The mmmmm.Pi is an included node and it works fine.

The mmm.Sum is an included node and it fails inside the DS **OR outside it**.


https://github.com/user-attachments/assets/75f9ad33-1fad-4798-9fd5-b7a4c7d8b26c

packages/dmn-editor/tests-e2e/__fixtures__/palette.ts
danielzhe: If the included model is empty, then nothing will happen, but if the included model exists, we will use its name (which is optional, as you spotted) or its id if the name is not present.

I'm replying the other question there.

packages/dmn-editor/tests-e2e/__fixtures__/palette.ts
danielzhe: I moved to palette because we understand the "palette" as all those buttons on the left of the screen. So I think the toggle should be done in the palette.

packages/dmn-editor/tests-e2e/drgElements/moveExternalNode.spec.ts
danielzhe: This is a good question. I understand your point and it makes sense. My concern is that when we'll do tests to the included models we may have to test those two steps separately (fill the textbox and select the model). 
We may want to select a model without filling the textbox or fill the textbox and then check if the model is present (or not) in the dropdown. That's why I'm trying to avoid to add "too many" steps in a single method of the API.

tiagobento: Not necessarily. Two separate actions, two separate methods.

tiagobento: Not necessarily. Two separate actions, two separate methods.

tiagobento:
Good point about the invocation string @jomarko!

tiagobento:
Another test I think we need, and sorry if I mentioned it already, but is to check Decision Services in mutiple DRDs, with the external Decisions....

Two examples:
1. Create a DS and add an external Decision to it. Then create a new DRD and drag the same DS to it. The depiction should be exactly the same.
2. Same as example above, but move the external Decision but keeping it inside the DS and see if this is reflected on the other DRD too.

danielzhe:
I'm checking if the error about the execution that is not related to this PR is easy to fix. If it is easy, I think we should include it in this PR because it's part of the feature.

danielzhe:
https://github.com/user-attachments/assets/58bfce14-b10a-4af2-bd00-dd6652016abe

@tiagobento The test.

danielzhe:
The issue found: https://github.com/apache/incubator-kie-issues/issues/1435

tiagobento:
> 
> 
> https://github.com/user-attachments/assets/58bfce14-b10a-4af2-bd00-dd6652016abe
> 
> 
> 
> @tiagobento The test. 


Thanks! Yeah, moving a Decision around while keeping it inside a DS is indeed not expected to be working yet... See

- https://github.com/apache/incubator-kie-issues/issues/889

danielzhe:
About the invocation string, should we handle it in this PR or in another one? It looks like the issue is in the `repopulateInputDataAndDecisionsOnAllDecisionServices` mutation. @tiagobento

jomarko (approved):
Approving as my understanding is the missing bits will be addressed in a new ticket. Thank you.

danielzhe:
Yes. The issue about decision service string is fixed. The runner issue is in another ticket, @jomarko 
It's up to @tiagobento now.

tiagobento:
@danielzhe Still reviewing, but please make sure you don't introduce ESLint warnings of missing hook dependencies.

danielzhe:
I'll check it. Thank you, @tiagobento

ljmotta (commented):
@danielzhe I've done some manual testing with the DS and external DS, and I could find one critical bug regarding the node conflict. Creating a DS with the external Decision nodes that are part of an external DS creates a conflict, as we have two DS that uses the same nodes. The latest DS that is added to the DRD should be depicted on the collapsed form, avoiding the conflict. On the following video we can see that both DS move the Decision nodes.

The bug with two DS, one external DS and one DS:

https://github.com/user-attachments/assets/32bc397f-6c8e-4037-b1ea-4bab0013c3e3

Working example with two DS (not external):

https://github.com/user-attachments/assets/195d54fb-4a14-4062-b2db-49ac3fcb9ae2

danielzhe:
Interesting issue you found, @ljmotta ! I'm working on that.

tiagobento:
As we discussed yesterday:
- Please incorporate `else` branches and throw errors on exceptional cases when repopulating the inputDecisions and inputData of Decision Services.
- Change the code to accommodate for the fact that `ir.requiredDecision["@_href"]` and `ir.requiredInput["@_href"]` may already contain a namespace, thus prefixing it with the external model's namespace may lead to invalid hrefs.

danielzhe:
I've found another related to the test that @ljmotta did:
1. Insert a local Decision Service with an "Included Decision-X"
2. Insert an included Decision Service that encapsulated the "Included Decision-X"
3. You'll get the same odd behavior that Luiz found.


https://github.com/user-attachments/assets/e80023f2-3205-473c-bb90-9ad1820ebe33


The both issues are related and it starts in the `[computeContainingDecisionServiceHrefsByDecisionHrefs](https://github.com/apache/incubator-kie-tools/blob/main/packages/dmn-editor/src/store/computed/computeContainingDecisionServiceHrefsByDecisionHrefs.ts.ts)`

When we insert an external Decision Service, the `decisionServiceHrefsByDecisionHrefs` we return from `computeContainingDecisionServiceHrefsByDecisionHrefs` the map is built with the wrong href for local nodes: we're adding the external namespace URI for local nodes, when we shouldn't.

danielzhe:
I need a re-review here, @jomarko @ljmotta for this new feature.
1. This should fix the issue @ljmotta found [here ](https://github.com/apache/incubator-kie-tools/pull/2508#pullrequestreview-2250908365)

2. It also fixes the same issue, that I reported [here](https://github.com/apache/incubator-kie-tools/pull/2508#issuecomment-2305653986)
 
3. It fixes an issue with 3rd models not included in the current model. For example:
Local DS -> Decision in Model A -> Input in Model B.

Consider that user only added the "Model A" in the local (the current) model. Now we handle this in this way:

https://github.com/user-attachments/assets/47fc08c6-e56e-46b7-b387-393977df9b2a

packages/dmn-editor/src/mutations/addEdge.ts
```diff
@@ -188,7 +190,7 @@ export function addEdge({
   // Replace with the new one.
   diagramElements.push(newDmnEdge);
 
-  repopulateInputDataAndDecisionsOnAllDecisionServices({ definitions });
+  repopulateInputDataAndDecisionsOnAllDecisionServices({ definitions, externalModelsByNamespace: undefined });
```
tiagobento: This doesn't look right..

packages/dmn-editor/src/store/computed/computeContainingDecisionServiceHrefsByDecisionHrefs.ts.ts
```diff
@@ -23,45 +23,45 @@ import { State } from "../Store";
 
 export function computeContainingDecisionServiceHrefsByDecisionHrefs({
   thisDmnsNamespace,
-  drgElementsNamespace,
-  drgElements,
+  drgElementsNamespaceByNamespace,
```
tiagobento: This new name doesn't make much sense to me. Would you mind explaining it please?

packages/dmn-editor/src/store/computed/computeContainingDecisionServiceHrefsByDecisionHrefs.ts.ts
```diff
@@ -23,45 +23,45 @@ import { State } from "../Store";
...
 }: {
   thisDmnsNamespace: string;
-  drgElementsNamespace: string;
-  drgElements: State["dmn"]["model"]["definitions"]["drgElement"];
+  drgElementsNamespaceByNamespace: Map<string, State["dmn"]["model"]["definitions"]["drgElement"]>;
 }) {
-  drgElements ??= [];
+  drgElementsNamespaceByNamespace ??= new Map();
```
tiagobento: The type of this parameter says it can never be undefined, so this line doesn't really do anything.

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
```diff
@@ -70,10 +74,23 @@ export function DecisionServiceProperties({
...
+      console.warn(`DMN EDITOR: Could not find model with namespace '${namespace}'. Ignoring.`);
+      return acc;
+    } else {
+      if (externalModel.type === "dmn") {
+        acc.push(externalModel);
+      }
+
+      return acc;
+    }
+  }, new Array<Normalized<ExternalDmn>>());
```
tiagobento: Please check hook dependencies.

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
```diff
@@ -70,10 +74,23 @@ export function DecisionServiceProperties({
...
+    if (!externalModel) {
+      console.warn(`DMN EDITOR: Could not find model with namespace '${namespace}'. Ignoring.`);
+      return acc;
+    } else {
+      if (externalModel.type === "dmn") {
+        acc.push(externalModel);
+      }
+
+      return acc;
+    }
```
tiagobento:
```suggestion
    if (!externalModel) {
      console.warn(`DMN EDITOR: Could not find model with namespace '${namespace}'. Ignoring.`);
      return acc;
    }
    
    if (externalModel.type === "dmn") {
      acc.push(externalModel);
    }

    return acc;
```

packages/dmn-editor/src/diagram/Diagram.tsx
```diff
@@ -876,6 +879,7 @@ export const Diagram = React.forwardRef<DiagramRef, { container: React.RefObject
                   __readonly_externalModelTypesByNamespace: state
                     .computed(state)
                     .getExternalModelTypesByNamespace(externalModelsByNamespace),
+                  externalModelsByNamespace,
```
tiagobento: React hook dependency missing.

packages/dmn-editor-envelope/src/DmnEditorRoot.tsx
```diff
@@ -645,6 +734,19 @@ function ExternalModelsManager({
...
+          const content = dmn.content ?? "";
+          const model = normalize(getMarshaller(content, { upgradeTo: "latest" }).parser.parse());
+          loadDependentModels(
+            model,
+            externalModelsIndex,
+            dmnsByNamespace,
+            loadedDmnsByPathRelativeToTheWorkspaceRoot,
+            thisDmnsNormalizedPosixPathRelativeToTheWorkspaceRoot
+          );
+        }
```
tiagobento: Missing hook dependencies.

tiagobento (changes requested):
Did another round of review. Thanks @danielzhe!

Can you please check if your IDE is correctly displaying ESLint warnings? There are too many hook dependencies missing, and they're *really* important.

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
tiagobento: Just a tip for conciseness (removing one indentation level)

danielzhe:
> Did another round of review. Thanks @danielzhe!
> 
> Can you please check if your IDE is correctly displaying ESLint warnings? There are too many hook dependencies missing, and they're _really_ important.

Thank you! I think I lost some of those in the merge. I got some conflicts, because before it was zero hooks errors, but I'm checking again.

packages/dmn-editor/src/store/computed/computeContainingDecisionServiceHrefsByDecisionHrefs.ts.ts
danielzhe: It doesn't make sense. I'm renaming it to `drgElementsByNamespace` now. I blame my glasses.

danielzhe:
@tiagobento Changes applied!

jomarko:
Sorry, I get lost in the conversation, since my approval it seems more code landed in. Should I do a re-reiew?

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
```diff
@@ -54,8 +77,11 @@ export function addDecisionToDecisionService({
   }
 
   const diagram = addOrGetDrd({ definitions, drdIndex });
+  const hrefString = buildXmlHref({ namespace: href.namespace, id: href.id });
```
ljmotta: The `decisionHref` isn't the same as the `hrefString`?

packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts
```diff
@@ -44,12 +66,11 @@ export function deleteDecisionFromDecisionService({
     );
   }
 
-  decisionService.outputDecision = (decisionService.outputDecision ?? []).filter(
-    (s) => s["@_href"] !== `#${decisionId}`
-  );
+  const xmlHref = buildXmlHref({ namespace: href.namespace, id: href.id });
```
ljmotta: The same here. `xmlHref` will not have the same value as the `decisionHref`?

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
```diff
@@ -25,25 +25,48 @@ import { SnapGrid } from "../store/Store";
...
+  }
+
+  if (href.namespace) {
+    const externalDrgs = (externalModel?.model as Normalized<DmnLatestModel>).definitions.drgElement;
+    const decision = externalDrgs?.find((drgElement) => drgElement["@_id"] === href.id);
+    if (decision?.__$$element !== "decision") {
+      throw new Error(
+        `DMN MUTATION: DRG Element with id '${href.id}' is either not a Decision or doesn't exist in the external model '${href.namespace}'`
+      );
+    }
```
ljmotta, line 68: The error message is targeting two different cases. Not having an element and the element not being a Decision are two distinguish scenarios. I think this should be break into two errors. If `decision === undefined` the element doesn't exist message. And the `decision.__$$element !== "decision"` the error telling isn't a Decision. WDYT?

packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts
```diff
@@ -20,21 +20,43 @@
...
+  }
+
+  if (href.namespace) {
+    const externalDrgs = (externalModel?.model as Normalized<DmnLatestModel>).definitions.drgElement;
+    const decision = externalDrgs?.find((drgElement) => drgElement["@_id"] === href.id);
+    if (decision?.__$$element !== "decision") {
+      throw new Error(
+        `DMN MUTATION: DRG Element with id '${href.id}' is either not a Decision or doesn't exist in the external model '${href.namespace}'`
+      );
+    }
```
ljmotta, line 56: The same with this error.

packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts
```diff
@@ -76,6 +85,44 @@ export function repopulateInputDataAndDecisionsOnDecisionService({
...
+            }
+          } else {
+            throw new Error(
+              `DMN MUTATION: Invalid information requirement referenced by external DecisionService: '${externalDecision["@_id"]}'`
+            );
+          }
+        });
+      }
+    }
+  }
```
ljmotta: Adding just a suggestion here. The usage of guard clauses to reduce nesting:

```tsx
if (externalModel === undefined) {
    continue;
}
const externalDecision = (externalModel.model as Normalized<DmnLatestModel>).definitions.drgElement?.find((drgElement) => drgElement["@_id"] === href.id) as Normalized<DMN15__tDecision>;

if (externalDecision === undefined) {
    continue;
}

...
```

Also, why did you used the `flatMap` instead of `forEach`?

packages/dmn-editor/tests-e2e/__fixtures__/base.ts
```diff
@@ -134,6 +138,12 @@ export const test = base.extend<DmnEditorFixtures>({
   textAnnotationPropertiesPanel: async ({ diagram, page }, use) => {
     await use(new TextAnnotationPropertiesPanel(diagram, page));
   },
+  stories: async ({ baseURL, page }, use) => {
+    await use(new Stories(page, baseURL));
+  },
+  includedModels: async ({ baseURL, page }, use) => {
```
ljmotta: Unused `baseUrl`

ljmotta (commented):
I've made some comments inline. One tip regarding objects. When the object property and value has the same name you could short it just the property name:


```tsx
drgElementsByNamespace: drgElementsByNamespace,
externalModelsByNamespace: externalModelsByNamespace,
```

```tsx
drgElementsByNamespace,
externalModelsByNamespace,
```

Also, all test methods are with the `page` in their signatures, but none use it.

Example:
```async ({ editor, page, palette, diagram, nodes, includedModels })```

**Comment by danielzhe** [comment]
> I've made some comments inline. One tip regarding objects. When the object property and value has the same name you could short it just the property name:

I'm aware of this. If I didn't do it probably I just overlooked it or it was something that was renamed.  😄  @ljmotta 

I'm checking the other things you reported. Thank you.

File: packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts (hunk 34 above)
Comment (danielzhe) [reply packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts]: No. The `decisionHref` can be something like `uri#id`, `id`, or even `#id`.

The `buildXmlHref` makes sure that we have `uri#id` or `#id`

File: packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts (hunk 35 above)
Comment (danielzhe) [reply packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts]: Same answer as above.

File: packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts (hunk 38 above)
Comment (danielzhe) [reply packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts]: The code with `flatMap` is a copy-paste of the same code a few lines above, but it uses the `drgElement` and has a simpler logic. Do you want me to change both?

Comment (danielzhe) [reply packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts]: Good point (the nesting)!

File: packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts (hunk 36 above)
Comment (danielzhe) [reply packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts:68]: I'm following the existing error message patterns used in this mutation and in `deleteDecisionFromDecisionService`. In my opinion, we will only be adding more code to do something that we can do in fewer lines.

**Comment by danielzhe** [comment]
Changes, applied, @ljmotta

File: packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts (hunk 34 above)
Comment (ljmotta) [reply packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts]: A ` href`  will always have the the `#` on its structure. If it's possible to have a `href` variable without it (an `id` value), the variable is wrongly named. Now, looking at the source of this value, it comes from the `computeDiagramData.ts` file which creates the `selectedNodesById` map. The map `id` is a value made with the `buildXmlHref`, so all ids are `href`, meaning the `decisionHref` is correctly named.

I think this is a problem in our codebase, where we mix the `id` and the `href`, and at some point we need to address this. FYI @tiagobento

File: packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts (hunk 38 above)
Comment (ljmotta) [reply packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts]: Yes, please. It doesn't make sense to use `flatMap` if we're not retuning the value.

File: packages/dmn-editor/src/externalNodes/DmnObjectListItem.tsx [hunk 40]
```diff
@@ -50,14 +52,70 @@ export function DmnObjectListItem({
...
+  // Nodes: Decision-A
+  // Included Models: Model-B.dmn
+  //
+  // Model Name: Model-B.dmn
+  // Nodes: Input-B
+  // Included Models: [none]
+  //
+  // So, the "Local Model" only "knows" the nodes from "Model-A" and NOT from "Model-B".
+  // That's why we have different logic to build description for "known dmnObjects" and "unknown dmnObjects".
+  const isNamespaceIncluded = useMemo(
```
tiagobento: Maybe `isNamespaceDirectlyImported`?

packages/dmn-editor/src/externalNodes/DmnObjectListItem.tsx
```diff
@@ -50,14 +52,70 @@ export function DmnObjectListItem({
...
+    () => namespace === relativeToNamespace || importsByNamespace.has(namespace),
+    [importsByNamespace, namespace, relativeToNamespace]
+  );
+
+  const notIncludedNamespaceDescription = useMemo(() => {
```
tiagobento: Maybe `transitivelyImportedNamespaceDescription`?

packages/dmn-editor/src/includedModels/IncludedModels.tsx
```diff
@@ -471,6 +472,7 @@ function IncludedModelCard({
           definitions: state.dmn.model.definitions,
           __readonly_index: index,
           __readonly_externalModelTypesByNamespace: externalModelTypesByNamespace,
+          externalModelsByNamespace,
```
tiagobento: __readonly_externalModelsByNamespace

packages/dmn-editor/src/diagram/Diagram.tsx
tiagobento: __readonly_externalModelsByNamespace

packages/dmn-editor/src/diagram/Diagram.tsx
```diff
@@ -1125,6 +1132,7 @@ export const Diagram = React.forwardRef<DiagramRef, { container: React.RefObject
               shapeId: targetNode.data.shape["@_id"],
             },
             keepWaypoints: true,
+            externalModelsByNamespace: externalModelsByNamespace,
```
tiagobento: externalModelsByNamespace

packages/dmn-editor/src/mutations/addEdge.ts
```diff
@@ -188,7 +191,10 @@ export function addEdge({
   // Replace with the new one.
   diagramElements.push(newDmnEdge);
 
-  repopulateInputDataAndDecisionsOnAllDecisionServices({ definitions });
+  repopulateInputDataAndDecisionsOnAllDecisionServices({
+    definitions,
+    externalModelsByNamespace: externalModelsByNamespace,
```
tiagobento: externalModelsByNamespace

packages/dmn-editor/src/mutations/addExistingDecisionServiceToDrd.ts
```diff
@@ -92,11 +92,15 @@ export function getStrategyToAddExistingDecisionServiceToDrd({
...
+  __readonly_externalDmnsIndex.forEach((value, key) => {
+    drgElementsByNamespace.set(key, value.model.definitions.drgElement);
+  });
+
   const containingDecisionServiceHrefsByDecisionHrefsRelativeToThisDmn =
     computeContainingDecisionServiceHrefsByDecisionHrefs({
       thisDmnsNamespace: __readonly_namespace,
-      drgElementsNamespace: __readonly_decisionServiceNamespace,
-      drgElements: decisionServiceDmnDefinitions.drgElement,
+      drgElementsByNamespace: drgElementsByNamespace,
```
tiagobento: drgElementsByNamespace

packages/dmn-editor/src/mutations/addExistingDecisionServiceToDrd.ts
```diff
@@ -308,6 +312,7 @@ export async function addAutoGeneratedDecisionServiceToDrd({
     __readonly_drdIndex: dummyDrdIndex,
     __readonly_dmnObjectNamespace: __readonly_decisionServiceNamespace,
     __readonly_externalDmnsIndex: externalModelTypesByNamespace.dmns,
+    __readonly_externalModelsByNamespace: __readonly_externalModelsByNamespace,
```
tiagobento: :D

packages/dmn-editor/src/mutations/deleteNode.ts
```diff
@@ -208,16 +212,15 @@ export function canRemoveNodeFromDrdOnly({
...
+  __readonly_externalDmnsIndex.forEach((value, key) => {
+    drgElementsByNamespace.set(key, value.model.definitions.drgElement);
+  });
 
   const containingDecisionServiceHrefsByDecisionHrefsRelativeToThisDmn =
     computeContainingDecisionServiceHrefsByDecisionHrefs({
       thisDmnsNamespace: definitions["@_namespace"],
-      drgElementsNamespace: __readonly_dmnObjectNamespace,
-      drgElements,
+      drgElementsByNamespace: drgElementsByNamespace,
```
tiagobento: drgElementsByNamespace

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
```diff
@@ -410,14 +423,14 @@ function DecisionServiceEquivalentFunction({
 
       const dmnObject = allDrgElementsByHref.get(potentialExternalHref);
 
-      return dmnObject
+      return dmnObject && (importsByNamespace.has(resolvedNamespace) || resolvedNamespace === thisDmnsNamespace)
```
tiagobento: Maybe extract `(importsByNamespace.has(resolvedNamespace) || resolvedNamespace === thisDmnsNamespace)` to a variable and give it a descriptive name?

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
```diff
@@ -449,3 +462,10 @@ function DecisionServiceEquivalentFunction({
     </Alert>
   );
 }
+
+function buildDisplayName(
+  dmnObject: Unpacked<Normalized<DMN15__tDefinitions>["drgElement"]> | undefined,
+  namespace: string
+) {
+  return `${namespace.substring(0, 11)}...${namespace.substring(namespace.length - 4)}.${dmnObject?.["@_name"]}`;
+}
```
tiagobento: Maybe `buildDisplayNameForDmnObject`?

packages/dmn-editor/src/propertiesPanel/DecisionServiceProperties.tsx
```diff
@@ -66,14 +69,24 @@ export function DecisionServiceProperties({
 
   const thisDmn = useDmnEditorStore((s) => s.dmn);
   const { externalModelsByNamespace } = useExternalModels();
-  const externalDmnsByNamespace = useDmnEditorStore(
-    (s) => s.computed(s).getExternalModelTypesByNamespace(externalModelsByNamespace).dmns
-  );
+
+  const allExternalDmns = Object.entries(externalModelsByNamespace ?? {}).reduce((acc, [namespace, externalModel]) => {
```
tiagobento, line 73: Maybe `useDmnEditorStore((s) => s.computed(s).getExternalModelTypesByNamespace(externalModelsByNamespace).dmns)`? Not sure if it is exactly the same, though, but looks like it.

packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts
```diff
@@ -20,21 +20,43 @@
...
-    throw new Error(`DMN MUTATION: DRG Element with id '${decisionId}' is either not a Decision or doesn't exist.`);
+  const href = parseXmlHref(decisionHref);
+
+  const externalModel = externalModelsByNamespace?.[href.namespace ?? ""];
+  if (href.namespace && !externalModel) {
+    throw new Error(`DMN MUTATION: Namespace '${href.namespace}' not found.`);
+  }
+
+  if (href.namespace) {
+    const externalDrgs = (externalModel?.model as Normalized<DmnLatestModel>).definitions.drgElement;
```
tiagobento: Instead of casting, you can check if `externalMode.type === "dmn"`, no?

packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts
tiagobento: `xmlHref` is exactly the same as `href`... you're even building it with the same values. No need to have this `xmlHref` variable...

packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts
```diff
@@ -76,6 +85,47 @@ export function repopulateInputDataAndDecisionsOnDecisionService({
     });
   }
 
+  for (const hrefString of hrefsToDecisionsInsideDecisionService) {
```
tiagobento: This for is complementary to the for above, but it ALSO treats external nodes. You can have the for above be part of this, checking whether `href` has a namespace or not. No need to have all these "continue"s.

packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts
```diff
@@ -76,6 +85,47 @@ export function repopulateInputDataAndDecisionsOnDecisionService({
...
+    const externalModel = externalModelsByNamespace?.[href.namespace ?? ""];
+    if (!externalModel) {
+      continue;
+    }
+
+    const externalDecision = (externalModel?.model as Normalized<DmnLatestModel>).definitions.drgElement?.find(
+      (drgElement) => drgElement["@_id"] === href.id
+    ) as Normalized<DMN15__tDecision>;
+    if (!externalDecision) {
+      continue;
```
tiagobento: Same comment as the other place. You can check if `externalModel` is a DMN to avoid the casting. Also, if you don't find the `externalDecision` here, it means there's a problem, right?

packages/dmn-editor/src/mutations/repopulateInputDataAndDecisionsOnDecisionService.ts
tiagobento: See the for above, you have a string with `#` there, hard coded... That's something I think we can't have anymore...

packages/dmn-editor-envelope/src/DmnEditorRoot.tsx
```diff
@@ -563,6 +563,94 @@ function ExternalModelsManager({
     };
   }, [thisDmnsNormalizedPosixPathRelativeToTheWorkspaceRoot]);
 
+  const getIncludedNamespacesFromModel = useCallback((model: Normalized<DmnLatestModel>) => {
+    return (model.definitions.import ?? [])
+      .map((i) => getNamespaceOfDmnImport({ dmnImport: i }))
+      .join(NAMESPACES_EFFECT_SEPARATOR);
+  }, []);
```
tiagobento: Duplicated code, and there's no need for this to be a `useCallback`, it can be a function outside of the component. Please extract it and use to compute the `namespaces` variable too.

packages/dmn-editor-envelope/src/DmnEditorRoot.tsx
```diff
@@ -563,6 +563,94 @@ function ExternalModelsManager({
...
+
+          loadedDmnsByPathRelativeToTheWorkspaceRoot.add(resource.normalizedPosixPathRelativeToTheWorkspaceRoot);
+
+          loadDependentModels(
+            includedModel,
+            externalModelsIndex,
+            resourcesByNamespace,
+            loadedDmnsByPathRelativeToTheWorkspaceRoot,
+            thisDmnsNormalizedPosixPathRelativeToTheWorkspaceRoot
+          );
```
tiagobento: This `useCallback` depends on itself.. This is not very good I'd say. Not even sure how ESLint didn't complain about this, as you're inhenritenly depending on `loadDependentModels` from a previous tick (I.e., outdated `getIncludedNamespacesFromModel`).

tiagobento (changes requested):

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
ljmotta: @danielzhe @tiagobento how are we going to proceed here?

packages/dmn-editor/src/mutations/deleteDecisionFromDecisionService.ts
```diff
@@ -20,21 +20,45 @@
 import { DMN15__tDefinitions } from "@kie-tools/dmn-marshaller/dist/schemas/dmn-1_5/ts-gen/types";
 import { repopulateInputDataAndDecisionsOnDecisionService } from "./repopulateInputDataAndDecisionsOnDecisionService";
 import { Normalized } from "../normalization/normalize";
+import { buildXmlHref, parseXmlHref } from "../xml/xmlHrefs";
```
ljmotta, line 23: unused import

packages/dmn-editor/src/mutations/deleteImport.ts
```diff
@@ -67,8 +70,9 @@ export function deleteImport({
         __readonly_dmnObjectId: node.data.dmnObject?.["@_id"],
         __readonly_dmnObjectQName: node.data.dmnObjectQName,
         __readonly_dmnObjectNamespace: node.data.dmnObjectNamespace!,
-        __readonly_externalModelTypesByNamespace,
-        mode: NodeDeletionMode.FROM_DRG_AND_ALL_DRDS,
+        __readonly_externalDmnsIndex: __readonly_externalModelTypesByNamespace.dmns,
+        __readonly_mode: NodeDeletionMode.FROM_DRG_AND_ALL_DRDS,
+        __readonly_externalModelsByNamespace: __readonly_externalModelsByNamespace,
```
ljmotta:
```suggestion
        __readonly_externalModelsByNamespace,
```

packages/dmn-editor/src/externalNodes/DmnObjectListItem.tsx
```diff
@@ -50,14 +52,70 @@ export function DmnObjectListItem({
...
+            namespace,
+            relativeToNamespace,
+          }).full
+        : dmnObject?.["@_name"],
+    [dmnObject, importsByNamespace, isNamespaceDirectlyIncluded, namespace, relativeToNamespace]
+  );
+
+  const nodeTypeTooltipDescription = useMemo(() => {
+    if (dmnObject === undefined) {
+      throw new Error("nodeTypeDescription can't be defined without a DMN object");
```
ljmotta, line 97: The `nodeTypeDescription` and `DMN Object` doesn't make much sense in the error message. The `dmnObject` is a `drgElement`, I'm not sure about the context behind this name, maybe @tiagobento could help us here. "Can't define the tooltip description without a DRG element" WDYT?

packages/dmn-editor/src/externalNodes/DmnObjectListItem.tsx
```diff
@@ -50,14 +52,70 @@ export function DmnObjectListItem({
...
+    }
+    const nodeType = getNodeTypeFromDmnObject(dmnObject);
+    if (nodeType === undefined) {
+      throw new Error("Can't determine nodeTypeDescription with undefined node type");
```
ljmotta, line 101: The same here. "Can't define the tooltip description without the node type" WDYT?

ljmotta (commented):
Great changes, the rename helped a lot. I've made some comments inline regarding error messages and minor things.

packages/dmn-editor/src/mutations/addDecisionToDecisionService.ts
tiagobento: @danielzhe Please use `decisionHref` where `hrefString` is currently being used, since they are exactly the same string.

packages/dmn-editor/src/externalNodes/DmnObjectListItem.tsx
tiagobento: Yeah, the name `dmnObject` is definitely sub-optimal. That's on me, though. We just need to be careful to not crash the editor over an error like this. Let's define to the "Unknown" string and remove both `throw`s please. 

@danielzhe

packages/dmn-editor/src/externalNodes/DmnObjectListItem.tsx
tiagobento: Same as above. @danielzhe

ljmotta:
I found some interesting cases doing some manual checkings:

1. We have here three DRDs, in DRD3 we can observe that the ds2 is in collapsed form, and d3 isn't available to pick from the node palette, as the node was depict.

image

image

image


2. Dragging a node near the collapse DS will trigger the invalid operation in the DS.

image


3. It's possible to hide external Decisions that are part of a Decision Service. Adding them back to the DRD will not remove them from the DS, resulting some weird behaviors

https://github.com/user-attachments/assets/7050aaf7-d263-4e1f-ad2d-aa0182f1c1a1

4. A DRD that has Decision nodes that are part of a collapsed Decision Service can't be moved:

https://github.com/user-attachments/assets/40de991e-55dc-4268-968c-6ebfbb9de4b0

tiagobento:
@ljmotta Can you please discriminate between what's already on `main` and what's been introduced by this PR? Some of those bugs (E.g., `Dragging a node near the collapse DS will trigger the invalid operation in the DS.`) you reported seem not to be related to the changes this PR introduces.

tiagobento:
@danielzhe I would double-check those too, just to be sure.

ljmotta:
@tiagobento Both, 1 and 2 are present on `main`. I could reproduce using the `daily-dev`.

danielzhe:
https://github.com/user-attachments/assets/a79345b9-aac5-4f9f-8ad5-1f1938d55ab3

danielzhe:
https://github.com/user-attachments/assets/09eebb41-595b-4f46-adc3-be793730698f

tiagobento (approved):

ljmotta (approved):
@danielzhe  I've done some manual checks, and I couldn't find any new bug related to this PR. Thanks for your patience

PR #2527, opened 2024-08-08
kie-issues#1147: DMN Editor Filter expression resize handler is at wrong position

### LINKED ISSUE #1147: DMN Editor Filter expression resize handler is at wrong position
In a scenario when the filter expression uses more complex boxed expressions, not literal expressions, the resize handler seems to be positioned wrongly. The 'blank' space highlighted in the upper row with orange color seems to be equal to the space in the second row also highlighted with orange color.
Screenshot 2024-05-03 103552

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1147

This also fixes an issue regarding some "jumps" while resizing Filter expression.

**BEFORE:**

https://github.com/user-attachments/assets/1d95ae53-53aa-46df-941c-4b79cc307e63

**AFTER:**

https://github.com/user-attachments/assets/289fa719-a067-4c20-a591-c6ff2264d134

**BONUS: SMOOTH RESIZE WITHOUT "JUMPS"**

https://github.com/user-attachments/assets/a9cf22aa-363a-483a-bf3f-de15c886e317


danielzhe:
Don't panic, @tiagobento : there is only a single line of code that is changed. The rest of the files are screenshots and tests.

packages/boxed-expression-component/src/resizing/Hooks.tsx
```diff
@@ -68,7 +68,11 @@ export function useNestedExpressionResizingWidthValue(
 
     return Math.max(
       nestedExpressionContainerResizingWidthValue - fixedColumnResizingWidth.value - extraWidth,
-      ...nestedExpressions.map((e) => getExpressionResizingWidth(e, new Map(), widthsById)),
```
danielzhe, line 71: After spending hours it looks like it was in fact a simple issue hard to find: the extraWidth for nested expressions was not being considered here. This issue was only noticeable when the upper cell was LESS WIDE than the bottom one.

Also, this fixed another issue we had with resizing that was causing some "jumps" in this scenario. I'll share a video to show.

jomarko (approved):
I can confirm the case reported in the referenced ticket is fixed now. I also tried to model more 'complex' boxed expressions where I used 'nesting' expressions each to the other. I think I didn't find an issue. Approving this PR. Thank you @danielzhe .

klbowers (approved):
Nice catch @danielzhe! LGTM

tiagobento:
So there's this "containerElement" method you added to some of the expression types, but I'm not sure it's a good idea to expose it as a "public" getter, since this "containerElement" is not really a visual thing we can differentiate on-screen. Not sure how to proceed, but wanted to mention that to see if you can try and find an alternative... As reading the tests that contain calls to this new "containerElement" is not 100% clear to me. Thanks!

danielzhe:
> So there's this "containerElement" method you added to some of the expression types, but I'm not sure it's a good idea to expose it as a "public" getter, since this "containerElement" is not really a visual thing we can differentiate on-screen. Not sure how to proceed, but wanted to mention that to see if you can try and find an alternative... As reading the tests that contain calls to this new "containerElement" is not 100% clear to me. Thanks!

I made some changes to this. It is more clear now. We can still call the element which contains the expression, but now it is called `elementCell` which is the `cell` that contains the element. Also now we can call `hover()` directly in the expressions.

tiagobento (approved):

klbowers (approved):

tiagobento:
@danielzhe As we discussed, please take a look at the other "jump" we found on the last video of the description (input-1 column, at 0:11). I'm ok with a NO-ISSUE for this, so I'll merge this PR and you can fix it separately.

PR #2533, opened 2024-08-14
NO-ISSUE: Add option to disable the extended services wizard

klbowers:
This environmental variable is being added for users who don't use a local extended services. In this case it doesn't make sense for them to see the wizard, so instead they'll be shown a modal alerting them that extended services is not running.

Normal behaviour:

https://github.com/user-attachments/assets/a67406aa-06d9-4022-8eab-9a817e06644b

Disabled behaviour:

https://github.com/user-attachments/assets/5f584ea7-f093-4033-af7f-aa369cdf63e0


thiagoelg (approved):

jomarko (changes requested):
@kbowers-ibm hi, could you please explain more the strategy we have for disabling the wizard?

I am a little bit confused, after reading the PR title my understanding was that we completely hide the button, that would allow the user to display the wizard. But it seems we still keep displaying the wizard but we only change the wizard the text.

Maybe a video/screenshots to demonstrate an effect of this new property would help me in the understanding.

jomarko (approved):
@kbowers shared video in a private chat. I agree with the new behavior.  I think the new env variable names are somehow misleading but it may be related to my non-native speaker English level. As I do not want to block the PR, approving.

tiagobento:
Please add a good description here...

PR #2536, opened 2024-08-16
kie-issues#1443: Change JBPM DevUI to use current host origin to generate GraphQL URL to be consumed by the frontend

### LINKED ISSUE #1275: Remove references to the `openaApiPath` in the Quarkus Dev UI extensions in `kie-tools`
Both `jbpm-quarkus-devui` & `sonataflow-quarkus-devui` are using the Kogito application `openapi.json` in order to read the Process/Workflows definitions. Due to this, they require a specific `openapiPath` to help building the URL to the `openapi.json`

Recently both devui's started using a `GraphQL` query to read the definitions & endpoints list this configuration is no longer needed and should be removed.

### LINKED ISSUE #1443: Change JBPM DevUI to use current host origin to generate GraphQL URL to be consumed by the frontend
When running the JBPM DevUI in a host that is not localhost or 0.0.0.0 (e.g. in OpenShift) the frontend will use the `kogito.dataindex.http.url` property (in application.properties) to define the Data Index service GraphQL URL.

3 problems need to be tackled with this task:

- When running in a different hostname, with a different URL, the default `localhost` or `0.0.0.0` won't work, and sometimes there's no way to know the correct URL before making a deploying (e.g. in OpenShift it's only possible to know the specific URL after the Route resource is created).
- The `quarkus.http.root-path` property is ignored, so if the Quarkus app is running on a subpath (e.g. http://my-custom-url.com/my-subpath/<quarkus_app>) the Data Index URL and the envelope.js used by JBPM DevUI will still try to load from the root path (`/`).
- There's no way to specify the URL for the Forms API (It defaults to `/forms`).

thiagoelg:
JBPM DevUI will now use the current host origin for all URLs unless configured otherwise.

Fixes https://github.com/apache/incubator-kie-issues/issues/1443

This PR also fixes https://github.com/apache/incubator-kie-issues/issues/1275 for the JBPM DevUI (Sonataflow DevUI was not changed)

https://github.com/user-attachments/assets/9fa9d756-58ba-4a92-9cda-347d40b4e913


packages/runtime-tools-process-dev-ui-webapp/src/components/DevUI/RuntimeTools/RuntimeTools.tsx
```diff
@@ -59,7 +63,9 @@ const RuntimeTools: React.FC<IOwnProps> = ({
   const httpLink = new HttpLink({
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    uri: dataIndexUrl,
+    uri: shouldReplaceQuarkusAppOriginWithWebappOrigin
+      ? dataIndexUrl.replace(quarkusAppOrigin, devUIOrigin)
+      : dataIndexUrl,
```
tiagobento: Do we still need `@ts-ignore`?

packages/runtime-tools-process-dev-ui-webapp/src/components/contexts/DevUIAppContext.tsx
```diff
@@ -70,12 +74,20 @@ export class DevUIAppContextImpl implements DevUIAppContext {
...
-    return this.args.remoteKogitoAppUrl;
+  getQuarkusAppOrigin(): string {
+    return this.args.quarkusAppOrigin;
+  }
+
+  getQuarkusAppRootPath(): string {
+    return this.args.quarkusAppRootPath;
+  }
+
+  getShouldReplaceQuarkusAppOriginWithWebappOrigin(): boolean {
```
tiagobento:
```suggestion
  shouldReplaceQuarkusAppOriginWithWebappOrigin(): boolean {
```

?

packages/runtime-tools-process-dev-ui-webapp/src/envelope/RuntimeToolsDevUIEnvelopeApiImpl.ts
```diff
@@ -61,9 +61,14 @@ export class RuntimeToolsDevUIEnvelopeApiImpl implements RuntimeToolsDevUIEnvelo
...
-    this.view().setRemoteKogitoAppUrl && this.view().setRemoteKogitoAppUrl(initArgs.remoteKogitoAppUrl);
+    this.view().setQuarkusAppOrigin && this.view().setQuarkusAppOrigin(initArgs.quarkusAppOrigin);
+    this.view().setQuarkusAppRootPath && this.view().setQuarkusAppRootPath(initArgs.quarkusAppRootPath);
+    this.view().setShouldReplaceQuarkusAppOriginWithWebappOrigin &&
+      this.view().setShouldReplaceQuarkusAppOriginWithWebappOrigin(
+        initArgs.shouldReplaceQuarkusAppOriginWithWebappOrigin
+      );
     this.view().setAvailablePages && this.view().setAvailablePages(initArgs.availablePages);
     this.view().setCustomLabels && this.view().setCustomLabels(initArgs.customLabels);
     this.view().setOmittedProcessTimelineEvents &&
```
tiagobento: You can use optional chaining here, I think it will look much better... like

`this.view().setAvailablePages?.(initArgs.availablePages)`

no?

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-deployment/src/main/java/org/jbpm/quarkus/devui/deployment/DevConsoleProcessor.java
```diff
@@ -112,19 +114,21 @@ public CardPageBuildItem pages(
         String uiPath = nonApplicationRootPathBuildItem.resolveManagementPath(BASE_RELATIVE_URL,
                 managementInterfaceBuildTimeConfig, launchModeBuildItem, true);
 
-        String openapiPath = getProperty(configurationBuildItem, systemPropertyBuildItems, "quarkus.smallrye-openapi.path");
         String devUIUrl = getProperty(configurationBuildItem, systemPropertyBuildItems, "kogito.dev-ui.url");
-        String dataIndexUrl = getProperty(configurationBuildItem, systemPropertyBuildItems, "kogito.data-index.url");
+        String dataIndexUrl = getProperty(configurationBuildItem, systemPropertyBuildItems, "kogito.dataindex.http.url");
```
tiagobento, line 118: Not sure I understand this very well. We're changing our public API? What other places do we need to update? Maybe the examples?

thiagoelg: Both `kogito.dataindex.http.url` and `kogito-data-index.url` settings exist, idk why.
I had more success using `kogito.dataindex.http.url`, and from checking the source code in kogito-apps this should be the one we use.
@pefernan might know better.

tiagobento (approved):

porcelli (approved):
lgtm

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-deployment/src/main/java/org/jbpm/quarkus/devui/deployment/DevConsoleProcessor.java
pefernan: @thiagoelg @tiagobento sorry for being late here, I missed this while on PTO.

This System property (`kogito.data-index.url`) is the one that was set by default by the Runtime Dev Services if it starts the Embedded Data-Index and this same property is set by default by the Data Index Addon if present. So that's why I made the DevProcesor use it.

The `kogito.dataindex.http.url` as far as I know is a config property that is being used by certain components (So far I only found the SVG addon) to build WebClients to interact with Data Index or in IT Tests. IDK if it has any other internal usage.

In this particular case I'd keep using the original property (`kogito.data-index.url`) but I think it would be worth checking if we could consolidate the usage of the data-index url in a single system property instead of having two of them.

PR #2538, opened 2024-08-19
kie-issues#1445: New DMN Editor: Support read-only mode

### LINKED ISSUE #1445: New DMN Editor: Support `read-only` mode
To be able to display DMN models using the new DMN Editor, we need to make it support a `readOnly` flag. Navigation must still be possible (E.g., entering the Boxed Expression Editor for a Decision/BKM) or Data Types and Included models, changing the Zoom level etc also must work fine.

thiagoelg:
Closes https://github.com/apache/incubator-kie-issues/issues/1445

https://github.com/user-attachments/assets/3d3cff8d-ca70-4929-bf45-9b62cbf21490


tiagobento (commented):
Looking great guys, just missing the E2E tests for BEE, and good to go IMHO.

ljmotta (commented):
@thiagoelg Thanks for adding this feature! It must have taken a big effort to add the readonly flag in the codebase! I've done some manual testing, and I've found some bugs:

1. It's still possible to change the Decision Service divider line position in read-only mode.
2. It's still possible to change the Alternative Input Shape option in the DRD while in read-only mode.
3. It's still possible to resize Boxed Expressions while in read-only mode.
4. It's not possible to select some cells of the Boxed Expression Editor while in read-only mode. This impossibility the checking ofmakes it impossible to check some expression properties using the properties panel.
5. Now we have a X button on the overlay panel, which doesn't work.

tiagobento:
@danielzhe ^

thiagoelg:
- Bug: The Hit policy can be changed while on Read Only mode
image

danielzhe:
> * Bug: The Hit policy can be changed while on Read Only mode
> 
> image

This is already fixed in the PR I opened for you.

danielzhe:
> 3. It's still possible to resize Boxed Expressions while in read-only mode.
> 4. It's not possible to select some cells of the Boxed Expression Editor while in read-only mode. This impossibility the checking ofmakes it impossible to check some expression properties using the properties panel.

Checking those. I edited your comment by mistake, @ljmotta . Sorry

danielzhe:
Some manual tests in Storybook. I didn't test every expression because of the size of the video. *Notice that I'm checking some not-readonly expressions too just in case to be if it is not broken (our E2E tests do that too)*:

https://github.com/user-attachments/assets/bd25bbb3-78cf-459f-b67e-b4f0a9fc0eb4

ljmotta (approved):
@danielzhe @thiagoelg Thanks for the changes, it appears that everything is working!

tiagobento (approved):

packages/dmn-editor/tests-e2e/readOnly/readOnlyDiagram.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { expect, test } from "../__fixtures__/base";
+
+test.beforeEach(async ({ editor, nodes }) => {
+  await editor.openLoanPreQualification();
+  await expect(nodes.get({ name: "Loan Pre-Qualification" })).toBeVisible();
+  await editor.setIsReadOnly(true);
+});
+
+test.describe("Read Only mode - Diagram", () => {
```
jomarko, line 28: Could we maybe unify the naming convention, for the boxed expression component test we described the root of the test as `Readonly` while here we use `Read Only Mode - Diagram`

packages/dmn-editor-standalone/src/DmnEditorStandaloneEnvelopeServer.ts
```diff
@@ -41,7 +41,7 @@ export const createEnvelopeServer = (iframe: HTMLIFrameElement, readOnly?: boole
           resourcesPathPrefix: "",
           fileExtension: "dmn",
           initialLocale: "en-US",
-          isReadOnly: readOnly ?? true,
+          isReadOnly: readOnly ?? false,
```
jomarko, line 44: Could we please confirm this change has no effect for the:
- Embed feature of the online-editor [1]
- Opening a dmn diagram using a chrome-extension?

[1]
Screenshot 2024-08-21 092555

jomarko (commented):
Thank you for the PR. I put some comments inline. Furthermore I have findings

### 01 colors
It seems color options (font and shape) are not readonly.

image

### 02 Snapping
When snapping is changed in readonly mode, nodes are moved slightly
image

ljmotta:
@jomarko Regarding the two points you've raised, they don't change the XML, so I don't think it's a bad thing to give the user the option to tweak them.

EDIT: I thought you're saying about the overlay options! Indeed, the color tweaks are not read-only 😮

packages/dmn-editor-standalone/src/DmnEditorStandaloneEnvelopeServer.ts
thiagoelg: The embed editor is still the classic one, so this change won't affect it. The same goes for the chrome-extension.

But, once the new one becomes the default, we just need to set the readOnly flag to true when required.

thiagoelg:
@jomarko Thanks for the review!

The overlay options don't change the XML; it was a choice to leave them enabled. But the color pickers should be disabled. That was a nice catch!
I've fixed that in the last commit and the test name.

tiagobento:
@jomarko Just for clarity, the overlays panel works 'non-destructively'. They're options that modify what/how the user sees the model, but not what is stored in the DMN XML.

PR #2546, opened 2024-08-23
kie-issues#886: On the DMN Editor, adding waypoints to edges that don't have a corresponding DMNEdge associated with it should create the DMNEdge and add the waypoint normally

### LINKED ISSUE #886: On the DMN Editor, adding waypoints to edges that don't have a corresponding DMNEdge associated with it should create the DMNEdge and add the waypoint normally.
### Reproducer:

https://github.com/apache/incubator-kie-issues/assets/1584568/7bd1c049-45d6-4128-a683-ec1a7f2f90e9

---

The fix here is to add a new DMNEdge to the DRD when a waypoint is added to an 'automatic' edge like this. We can also understand it as "lazily" creating the DMNEdge...

jomarko:
Closes: https://github.com/apache/incubator-kie-issues/issues/886

### Checklist
- [x] Adding a waypoint on the edge between internal nodes
- [x] Adding a waypoint on the edge targeting an external node
- [x] Adding a waypoint on the edge from external node targeting internal node
- [x] In non default DRD,  Adding a waypoint on the edge between internal nodes
- [x] In non default DRD, Adding a waypoint on the edge targeting an external node
- [x] In non default DRD, Adding a waypoint on the edge from external node targeting internal node
- [x] Delete edge between internal nodes and recreate it
- [x] Delete edge targeting external node - should be recreated automatically
- [x] Delete edge from external node targeting internal node and recreate it
- [x] In non default DRD, delete edge between internal nodes and recreate it
- [x] In non default DRD, delete edge targeting external node - should be recreated automatically
- [x] In non default DRD, delete edge from external node targeting internal node and recreate it

### Current Behavior
In non default DRD, notice the edges waypoints
Screenshot 2024-08-28 154528


packages/dmn-editor/src/mutations/addEdge.ts
```diff
@@ -195,20 +194,22 @@ export function addEdge({
 
 function doesInformationRequirementsPointTo(a: Normalized<DMN15__tInformationRequirement>, nodeId: string) {
   return (
-    a.requiredInput?.["@_href"] === `${nodeId}` || //
-    a.requiredDecision?.["@_href"] === `${nodeId}`
+    // use endsWith because @_href is sometimes prefixed with '#' and sometimes it is not
```
tiagobento: I know this is in draft, but let me tell you already that using "endsWith" is not the way to go here. `hrefs` are structured strings, composed by a `namespace` and an `id`. They have this format:

- `#id`; or
- `namespace#id`; where `namespace` is an URL.

We have a special parsing method to treat those and separate `namespace` and `id`. See `parseXmlHref`.

tiagobento:
@jomarko I think this PR is fixing the problem in the wrong place. There's a mutation that adds waypoints... we only need to create an Edge at that moment, if the edge doesn't exist.

ljmotta:
@jomarko @tiagobento The code makes sense to me, but I'm not certain about the solution. I'm just joining the discussion on this issue, so I might have missed something. Why are we adding the edge lazily? Can't we add the DMNEdge to the DRD at the same time we add the Node?

tiagobento:
@ljmotta We could, but it's more places to handle, as there are multiple places that can add a node to a DRD. If a DMNEdge doesn't exist in the DRD, we render a default edge between the nodes, based on the DRG. Now when we want to add the waypoint, that's when the DMNEdge becomes necessary, otherwise there's no place to add the waypoint to.

ljmotta:
@tiagobento I understand, but we will have an inconsistency between the model and what is visualized in the editor. I mean, the DRD depicts something that isn't on it. Taking a look into another issue, like https://github.com/apache/incubator-kie-tools/issues/3934 we will fall into the same problem, as it wouldn't be possible to tweak the edge properties.

tiagobento:
Thanks @ljmotta. As we discussed, the DRG should be understood as the primary source of truth, and the DRD should be used to provide a customized depiction of it. DMNEdges, like all DMN DI elements, are optional, so we need to do the best we can to have the DRG correctly depicted even if the DRD doesn't contain visual information about it.

jomarko:
After private discussion with @tiagobento converted to draft and will adopt another solution for this issue.

packages/dmn-editor/src/mutations/addEdge.ts
```diff
@@ -74,14 +74,21 @@ export function addEdge({
     autoPositionedEdgeMarker: AutoPositionedEdgeMarker | undefined;
   };
   keepWaypoints: boolean;
+  extraArg?: {
+    requirementEdgeTargetingExternalNodeId: string | undefined;
+  };
```
jomarko: `Diagram`, `onConnect` is broken now due to this probably

packages/dmn-editor/src/diagram/connections/isValidConnection.ts
```diff
@@ -39,14 +39,18 @@ export function checkIsValidConnection(
...
 ) {
   if (!sourceNode?.type || !targetNode?.type || !edgeType) {
     return false;
   }
 
-  // External nodes cannot be targeted
-  if (targetNode.data.dmnObjectQName.prefix) {
+  // External nodes cannot be targeted by default
+  // However there are exceptions, for example adding a waypoint on the edge
+  if (!extraArg?.allowExternalTarget && targetNode.data.dmnObjectQName.prefix) {
```
danielzhe: In the future, will we have local nodes with a prefix, @tiagobento ? We were talking about that in pvt chat a few weeks ago. 
Anyway, I don't feel confident concluding that nodes with a `prefix` are always remote. Maybe if someone else in the future looks at this code doesn't understand why we're checking if there is a `prefix` or not.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,86 @@ export function usePotentialWaypointControls(
...
 
   const onDoubleClick = useCallback(() => {
-    if (!potentialWaypoint || !snappedPotentialWaypoint || edgeIndex === undefined) {
+    if (!potentialWaypoint || !snappedPotentialWaypoint) {
       return;
     }
 
+    if (edgeIndex === undefined) {
+      /**
+       * This means we are adding a first wayipoint to one of following edges:
```
danielzhe: Typo in "wayipoint" = "waypoint"

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,86 @@ export function usePotentialWaypointControls(
...
+        const nodesById = state.computed(state).getDiagramData(externalModelsByNamespace).nodesById;
+        const edge = state.computed(state).getDiagramData(externalModelsByNamespace).edgesById.get(edgeId);
+        if (
+          edge === undefined ||
+          edge.type === undefined ||
+          edge.data === undefined ||
+          edge.data?.dmnShapeSource === undefined ||
+          edge.data?.dmnShapeTarget === undefined
+        ) {
+          console.debug(`DMN MUTATION: We can not add DMNEdge for '${edgeId}' edge into diagram.`);
```
danielzhe: Maybe we can add more details about why we can not add the DMNEdge here. We don't know if we can't because the edge is not defined, type, data, shapeSource, or shapeTarget and we have the exactly same message bellow in other 2 places. So, just looking at the console, we can not determine what caused the error or if it was in the first, second or third place.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,86 @@ export function usePotentialWaypointControls(
...
+          return;
+        }
+        const edgeSourceBounds = edge.data?.dmnShapeSource["dc:Bounds"];
+        const edgeTargetBounds = edge.data?.dmnShapeTarget["dc:Bounds"];
+
+        if (edgeSourceBounds === undefined || edgeTargetBounds === undefined) {
```
danielzhe, line 106: I think you can discard the null checks in lines 100-101 and merge with those, adding `?.` before the `["dc:Bounds"]`

```
const edgeSourceBounds = edge.data?.dmnShapeSource?.["dc:Bounds"];
const edgeTargetBounds = edge.data?.dmnShapeTarget?.["dc:Bounds"];
```

But doing this, in a few lines bellow, this will fail because we didn't null check only for `dmnShapeSource`.
`edge.data?.dmnShapeSource["@_id"]`

So, maybe we can re-work those null checks, making more clear error messages about what failed and joining what makes sense to be together (for example, the null checks about `dmnShapeTarget`).

What do you think? Does it make sense?

packages/dmn-editor/src/mutations/addEdge.ts
```diff
@@ -74,14 +74,20 @@ export function addEdge({
     autoPositionedEdgeMarker: AutoPositionedEdgeMarker | undefined;
   };
   keepWaypoints: boolean;
+  extraArg?: {
```
danielzhe: I'm being annoying here: I don't like the `extraArgs`. Why they are "extra"?

If are optional args, I think we can use  `requirementEdgeTargetingExternalNodeId?: string` for example.

packages/dmn-editor/src/mutations/addEdgeWaypoint.ts
```diff
@@ -38,7 +38,7 @@ export function addEdgeWaypoint({
 
   const diagramElement = diagramElements[edgeIndex];
   if (diagramElement.__$$element !== "dmndi:DMNEdge") {
-    throw new Error("DMN MUTATION: Can't remove a waypoint from an element that is not a DMNEdge.");
+    throw new Error("DMN MUTATION: Can't add a waypoint for an element that is not a DMNEdge.");
```
danielzhe, line 41: 👍

packages/dmn-editor/src/mutations/addStandaloneNode.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import { switchExpression } from "@kie-tools-core/switch-expression-ts";
-import { DmnBuiltInDataType, generateUuid } from "@kie-tools/boxed-expression-component/dist/api";
```
danielzhe, line 21: I've been doing that kind of cleanup every time I see it. Nice!

packages/dmn-editor/tests-e2e/drds/modelDrd.spec.ts
```diff
@@ -136,13 +136,13 @@ test.describe("Model DRD", () => {
         await drds.toggle();
         await drds.navigateTo({ name: "Second DRD" });
         await drds.toggle();
-        await drgNodes.open();
```
danielzhe, line 139: 🎉

danielzhe (changes requested):
Nice feature! I leave some comments.

packages/dmn-editor/src/mutations/addEdge.ts
tiagobento: This is a trick for forcing named parameters

packages/dmn-editor/src/diagram/connections/isValidConnection.ts
jomarko: @danielzhe thank you for a comment.

the logic that marks node as external or not according to the prefix `targetNode.data.dmnObjectQName.prefix` was in codebase even before my PR. I would suggest the following:
- if there is an existing, more robust way, how to decide if node is or is not external, I can incorporate it as part of this PR
- if there is not, what I understand is the situation, we should probably change this behavior as part of a separate ticker ticket you mention - "In the future, will we have local nodes with a prefix" - I see this as a ticket that should handle what you suggest


please let me know what do you think

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
jomarko: well spotted, thank you, will address in a new commit

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
jomarko: I tried to do my best with the null checks simplification, lets have a look

packages/dmn-editor/src/mutations/addEdge.ts
jomarko: do you have other names suggestions @danielzhe ?

packages/dmn-editor/src/diagram/connections/isValidConnection.ts
danielzhe: Yeah, you're right! It doesn't make sense change it now since we're using that everywhere.

Maybe we can just extract to a variable:

`const isExternalNode = targetNode.data.dmnObjectQName.prefix != undefined`

packages/dmn-editor/src/mutations/addEdge.ts
danielzhe: Can't we use `requirementEdgeTargetingExternalNodeId` directly, @tiagobento ?

packages/dmn-editor/src/diagram/connections/isValidConnection.ts
jomarko: @danielzhe done

packages/dmn-editor/src/mutations/addEdge.ts
tiagobento: Sorry I overlook this a little bit. `requirementEdgeTargetingExternalNodeId` is already inside an object, meaning the "named arguments" trick is already in effect. Please lose the `extraArg` @jomarko. Thanks.

danielzhe:
I didn't find any issues in the manual tests. I tried some not-common scenarios too using external models and it seems everything is working.

danielzhe (approved):

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,81 @@ export function usePotentialWaypointControls(
...
+       * - an edge in a non default DRD
+       * - an edge targeting an external node
+       */
+      dmnEditorStoreApi.setState((state) => {
+        const nodesById = state.computed(state).getDiagramData(externalModelsByNamespace).nodesById;
+        const edge = state.computed(state).getDiagramData(externalModelsByNamespace).edgesById.get(edgeId);
+        if (edge === undefined || edge.data?.dmnShapeSource === undefined || edge.data?.dmnShapeTarget == undefined) {
+          console.debug(
+            `DMN MUTATION: We can not add DMNEdge for '${edgeId}' edge into diagram. There are missing data edge: ${edge}, edge.data: ${edge?.data}`
+          );
```
tiagobento: This is not a mutation error. Please adjust the logs.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,81 @@ export function usePotentialWaypointControls(
...
+          return;
+        }
+        const edgeSourceBounds = edge.data?.dmnShapeSource["dc:Bounds"];
+        const edgeTargetBounds = edge.data?.dmnShapeTarget["dc:Bounds"];
+
+        if (edgeSourceBounds === undefined || edgeTargetBounds === undefined) {
```
tiagobento:
```suggestion
        }
        
        const edgeSourceBounds = edge.data?.dmnShapeSource["dc:Bounds"];
        const edgeTargetBounds = edge.data?.dmnShapeTarget["dc:Bounds"];
        if (edgeSourceBounds === undefined || edgeTargetBounds === undefined) {
```

100% nit-picking, but I usually bring local variables used in guard clauses like this closer to the `if` itself, and give a little separation from the code that came before it.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,81 @@ export function usePotentialWaypointControls(
...
+          console.debug(
+            `DMN MUTATION: We can not add DMNEdge for '${edgeId}' edge into diagram. There are missing data edgeSourceBounds: ${edgeSourceBounds}, edgeTargetBounds: ${edgeTargetBounds}`
+          );
```
tiagobento: Not in a mutation. Please adjust the logs.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,81 @@ export function usePotentialWaypointControls(
...
+          return;
+        }
+
+        const sourceNode = nodesById.get(edge.source);
+        const targetNode = nodesById.get(edge.target);
+
+        if (sourceNode === undefined || targetNode === undefined) {
+          console.debug(
+            `DMN MUTATION: We can not add DMNEdge for '${edgeId}' edge into diagram. There are missing data sourceNode: ${sourceNode}, targetNode: ${targetNode}`
+          );
```
tiagobento: Not in a mutation. Please adjust the logs.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,81 @@ export function usePotentialWaypointControls(
...
+            data: targetNode.data,
+            bounds: edgeTargetBounds,
+            index: nodesById.get(edge.target)?.data.index ?? 0,
+            shapeId: edge.data?.dmnShapeTarget["@_id"],
+          },
+          keepWaypoints: false,
+          requirementEdgeTargetingExternalNodeId: isTargetExternalNode ? edgeId : undefined,
+        });
+
+        console.debug(`DMN MUTATION: DMNEdge for '${edgeId}' edge was added into diagram.`);
```
tiagobento: Not inside a mutation.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -74,10 +80,81 @@ export function usePotentialWaypointControls(
...
+        const targetNode = nodesById.get(edge.target);
+
+        if (sourceNode === undefined || targetNode === undefined) {
+          console.debug(
+            `DMN MUTATION: We can not add DMNEdge for '${edgeId}' edge into diagram. There are missing data sourceNode: ${sourceNode}, targetNode: ${targetNode}`
+          );
+          return;
+        }
+
+        const isTargetExternalNode = targetNode.data.dmnObjectQName.prefix !== undefined;
```
tiagobento:
```suggestion
        const isTargetingExternalNode = targetNode.data.dmnObjectQName.prefix !== undefined;
```

or 

```suggestion
        const isTargetExternal = targetNode.data.dmnObjectQName.prefix !== undefined;
```

or 

```suggestion
        const targetsExternalNode = targetNode.data.dmnObjectQName.prefix !== undefined;
```

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -96,18 +173,27 @@ export function usePotentialWaypointControls(
     }
 
     dmnEditorStoreApi.setState((state) => {
+      const dmnEdgeIndex = state.computed(state).indexedDrd().dmnEdgesByDmnElementRef.get(edgeId)?.index;
+      if (dmnEdgeIndex === undefined) {
+        console.debug(`DMN MUTATION: DMNEdge for '${edgeId}' edge has missing index.`);
+        return;
+      }
```
tiagobento: I think we should `throw` here, as at this point, this operation should never fail. If `edgeIndex` doesn't exist, you create a new edge, thus querying for the `edgeId` again should always give you an edge. If that doesn't happen, something exceptional happened, so we throw. Also not in a mutation.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -96,18 +173,27 @@ export function usePotentialWaypointControls(
     }
 
     dmnEditorStoreApi.setState((state) => {
+      const dmnEdgeIndex = state.computed(state).indexedDrd().dmnEdgesByDmnElementRef.get(edgeId)?.index;
```
tiagobento: I think it's worth to add a comment here explaining why you're getting `dmnEdgeIndex` when you already have `edgeIndex` in scope.

packages/dmn-editor/src/diagram/edges/usePotentialWaypointControls.ts
```diff
@@ -96,18 +173,27 @@ export function usePotentialWaypointControls(
...
+        return;
+      }
       addEdgeWaypoint({
         definitions: state.dmn.model.definitions,
         drdIndex,
         beforeIndex: i - 1,
-        edgeIndex,
+        edgeIndex: dmnEdgeIndex,
         waypoint: snappedPotentialWaypoint,
       });
```
tiagobento: Probably worth changing the name of this parameter to `dmnEdgeIndex`, since it represents a DMNEdge object.

packages/dmn-editor/src/diagram/connections/isValidConnection.ts
```diff
@@ -39,14 +39,19 @@ export function checkIsValidConnection(
...
   }
 
-  // External nodes cannot be targeted
-  if (targetNode.data.dmnObjectQName.prefix) {
+  // External nodes cannot be targeted by default
+  // However there are exceptions, for example adding a waypoint on the edge
+  const isTargetExternalNode = targetNode.data.dmnObjectQName.prefix != undefined;
+  if (!extraArg?.allowExternalTarget && isTargetExternalNode) {
     return false;
   }
```
tiagobento: This way, it is clear that `false` is the default for when `extraArg` is `undefined`.

```typescript
export function _checkIsValidConnection(
  sourceNode: { type?: string; data: DmnDiagramNodeData } | undefined,
  targetNode: { type?: string; data: DmnDiagramNodeData } | undefined,
  edgeType: string | null | undefined,
  extraArg?: { allowExternalTarget: boolean; }
) {
  if (!sourceNode?.type || !targetNode?.type || !edgeType) {
    return false;
  }

  // External nodes cannot be targeted by default. However there are exceptions,
  // for example adding a waypoint on an edge between two external nodes.
  const isTargetExternalNode = targetNode.data.dmnObjectQName.prefix != undefined;
  const allowExternalTarget = extraArg?.allowExternalTarget ?? false;
  if (isTargetExternalNode && !allowExternalTarget) {
    return false;
  }
}
```

packages/dmn-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -402,8 +410,13 @@ function ackRequirementEdges(
...
       (dmnObject.informationRequirement ?? []).forEach((ir, index) => {
         const irHref = parseXmlHref((ir.requiredDecision ?? ir.requiredInput)!["@_href"]);
+        // Search for definitions[`@_xmlns:included${includedIndex}`] that holds proper namespace
+        // and store the proper prefix: `included${includedIndex}` value
+        const namespaceIncludedPrefix = Object.entries(definitions)
+          .find(([key, val]) => val === namespace)?.[0]
+          ?.replace("@_xmlns:", "");
         ackEdge({
-          id: ir["@_id"]!,
+          id: (namespaceIncludedPrefix ? `${namespaceIncludedPrefix}:` : "") + ir["@_id"]!,
```
tiagobento: Why not

```suggestion
          id: (drgElementsNamespace === thisDmnsNamespace 
            ? ir["@_id"]
            : buildXmlHref({ namespace: drgElementsNamespace, id: ir["@_id"]  }),
```

?


----

External nodes are currently using an HREF (`namespace#id`) as ID, I think we should keep the same strategy for edges, instead of using a QName (`prefix:id`).

tiagobento (changes requested):
Getting there @jomarko. I like the intentionality of the code, especially where you added comments. Left some inline comments, please consider them. Thanks!

packages/dmn-editor/src/store/computed/computeDiagramData.ts
tiagobento: You can remove the `definitions` parameter you added too, if that ^ makes sense.

jomarko: @tiagobento thank you for this topic. actually, nodes/shapes, currently uses `prefix:id` , not `namespace#id`, that is why implemented that way for edges.

Applying your suggestion produces xml like attached:
```
 <dmndi:DMNDI>
    <dmndi:DMNDiagram id="_B90C2A89-A2E3-4610-AC66-A6DC0F9C1CEF" name="Default DRD" useAlternativeInputDataShape="false">
      <di:extension>
        <kie:ComponentsWidthsExtension>
        </kie:ComponentsWidthsExtension>
      </di:extension>
      <dmndi:DMNEdge id="_D763F5B1-FAF0-4790-A905-A5B1D63F0D2C" dmnElementRef="included2:_90FB4683-E1A0-4149-A1B8-444DEE06E749" sourceElement="_A9428417-FA8E-47B7-9F4A-10921FEA34BA" targetElement="_CAFE156E-4FB5-4285-9079-E3165FB92DDE">
        <di:waypoint x="1300" y="260" />
        <di:waypoint x="900" y="240" />
        <di:waypoint x="600" y="400" />
      </dmndi:DMNEdge>
      <dmndi:DMNShape id="_CD133A74-86D5-4DA7-81BF-F122A69A7927" dmnElementRef="included0:_2081EEE3-7B1B-48B3-B3B9-944F0A013CF1">
        <dc:Bounds x="960" y="160" width="160" height="80" />
      </dmndi:DMNShape>
      <dmndi:DMNShape id="_AA2DC9CB-C337-4E8F-8258-ECA14E0CDB10" dmnElementRef="included0:_92040BB1-AB7A-46D1-A4DE-A4FD1448ADF5">
        <dc:Bounds x="700" y="220" width="160" height="80" />
      </dmndi:DMNShape>
      <dmndi:DMNEdge id="_68653841-7F18-46EB-B3F0-37082CC0A3E8" dmnElementRef="https://kie.org/dmn/_AE5E9361-3580-4587-AE13-53A36057853B#_0A18CF15-F3E5-4B87-82A1-2851A246FDF4" sourceElement="_AA2DC9CB-C337-4E8F-8258-ECA14E0CDB10" targetElement="_CD133A74-86D5-4DA7-81BF-F122A69A7927">
        <di:waypoint x="780" y="260" />
        <di:waypoint x="940" y="260" />
        <di:waypoint x="860" y="200" />
        <di:waypoint x="1040" y="200" />
      </dmndi:DMNEdge>
    </dmndi:DMNDiagram>
  </dmndi:DMNDI>
  ```
  
  from this discussion my understanding is, I should update the code to use `namespace#id` also for nodes/shapes? Is that correct?

tiagobento: Ok, so let's take a step back. This `id` here is only the ID of the RF.Edge object we're telling ReactFlow exists. How it makes its way to the DMN XML is another story. I believe that ReactFlow objects (nodes and edges) should have their `id` in the HREF (I.e., `namespace#id`) format. That's one thing.

Now, you're correct to say that it is wrong to have an HREF in the `dmnElementRef` attribute, as, per the XSD, it's a QName (I.e., `prefix:id`). Now, let's try and see how the `id` of the RF.Edge is getting into the `dmnElementRef`....

The only place that adds DMNEdges to the XML is the `addEdge` mutation, so it must be an invocation to it. You added a new invocation to it on this PR... and there's this line:

A. `requirementEdgeTargetingExternalNodeId: targetsExternalNode ? edgeId : undefined`

See how the `edgeId` is now inside the `addEdge` mutation? Now let's look how this new `requirementEdgeTargetingExternalNodeId` parameter is used inside the `addEdge` mutation:

B. `let existingEdgeId: string | undefined = requirementEdgeTargetingExternalNodeId;`; and
C. `"@_dmnElementRef": existingEdgeId ?? newEdgeId,`

Hmm... something is not right :) 

---

Now, all this is really confusing, because there are multiple ""types"" of `id`, and there are multiple formats for `id`s. We have:

1. The id of the ReactFlow object (I.e., RF.Edge), which I'm saying should be in HREF format.
2. The id attribute (`@_id`) of the Information Requirement (which is a plain UUID).
3. The id attribute (`@_id`) of the DMNEdge (which is a plain UUID).
4. The id we need to write to dmnElementRef attribute (`@_dmnElementRef`), which is in QName format.

They're all "ids" in some way, so it's really hard to name things properly.

Ok, let's go back.

---

I'd say:
- A. should be:
  `dmnElementRefOfDmnEdge: targetsExternalNode ? informationRequirementQNameRelativeToThisDmn : undefined` (this renaming `requirementEdgeTargetingExternalNodeId`)
- B. should be:
  `let existingRequirementOrAssociationId: string | undefined;` (thus renaming `existingEdgeId`)
- C. should be:
  `"@_dmnElementRef": dmnElementRefOfDmnEdge ?? existingRequirementOrAssociationId ?? newEdgeId`
  
And I believe we should rename `newEdgeId` to `newRequirementOrAssociationId`.

Let me know if you have questions :)

tiagobento: Btw, we have `xmlHrefToQName` to help you on building `informationRequirementQNameRelativeToThisDmn` from `edgeId` (I.e., Information Requirement's HREF relative to this DMN`).

tiagobento (approved):

tiagobento:
@jomarko Conflicts...

PR #2550, opened 2024-09-02
NO-ISSUE: Sample DMN not compatible with classic DMN Editor

klbowers:
When opening the sample.dmn file using the classic DMN Editor, it's now incompatible due to the updated dmn:definitions. This PR Reverts them back to the 2018 versions, so that users can interact with the sample.dmn in both the new and classic DMN Editor

### DIFF
--- a/packages/online-editor/static/samples/Sample.dmn
+++ b/packages/online-editor/static/samples/Sample.dmn
@@ -18,17 +18,17 @@
   ~ under the License.
 -->
 <dmn:definitions
-  xmlns:dmn="https://www.omg.org/spec/DMN/20230324/MODEL/"
+  xmlns:dmn="http://www.omg.org/spec/DMN/20180521/MODEL/"
   xmlns="https://kie.apache.org/dmn/_857FE424-BEDA-4772-AB8E-2F4CDDB864AB"
   xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/"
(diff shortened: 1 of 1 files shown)


yesamer (approved):
@kbowers-ibm Good catch!

klbowers:
Ubuntu-latest 2 failing due to one test. The test in question is failing in multiple PRs, and passes when run locally

ljmotta:
@kbowers-ibm great catch. It looks like the branch you used wasn't from `main`, as it has many commits from other PRs.

klbowers:
It was originally, but when I did the 3rd commit I messed up the versioning, and couldn't figure out how to revert it. Best I could do was make it so that it only shows the files I changed!

ljmotta:
What I suggest is to rename this branch in your local to another name `git branch -m <new_name>`, get your commits SHA with `git log`, create a new branch from `main` with the same name as the one used in this PR `no-issue-rever-sample-dmn-definitions`. Apply the commits to this new branch using `git cherry-pick <commit_SHA>`, and make a push force. With the push force you will fix the commit history, but double/triple check if the commits are correct before you do it.

klbowers:
It worked, thank you very much @ljmotta! I've noted that process down for future reference :)

jomarko (approved):
I can confirm this can be opened both classic and new DMN Editor.

ljmotta (approved):

PR #2511, opened 2024-08-06
NO-ISSUE: temporary disable devservice to sonataflow-greeting-quarkus-example

fantonangeli:
This PR is related to: https://github.com/apache/incubator-kie-tools/pull/2510

During the work on https://github.com/apache/incubator-kie-issues/issues/1406 we found the error below in `sonataflow-greeting-quarkus-example`:
`com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"manifest for apache/incubator-kie-kogito-data-index-ephemeral:main not found: manifest unknown: manifest unknown"}`
This PR temporary disables the devservice in `sonataflow-greeting-quarkus-example` only to fix the error until the dev service issue is solved.

image

PR #2552, opened 2024-09-02
NO-ISSUE: Fix JBPM DevUI Quarkus HTTP Host and Port

thiagoelg:
- Use `0.0.0.0` instead of `localhost` for Quarkus Apps to support running dev mode on containers;
- Set default values for `quarkus.http.host` and `quarkus.http.port`.

### DIFF
--- a/examples/jbpm-compact-architecture-example/src/main/resources/application.properties
+++ b/examples/jbpm-compact-architecture-example/src/main/resources/application.properties
@@ -9,11 +9,11 @@ quarkus.kogito.data-index.graphql.ui.always-include=true
 quarkus.http.test-port=0
 
 # Kogito-service
-kogito.service.url=http://localhost:8080
+kogito.service.url=http://0.0.0.0:8080
 
 #Job-service
-kogito.jobs-service.url=http://localhost:8080
-kogito.dataindex.http.url=http://localhost:8080
+kogito.jobs-service.url=http://0.0.0.0:8080
+kogito.dataindex.http.url=http://0.0.0.0:8080
 
 # run create tables scripts
 quarkus.flyway.migrate-at-start=true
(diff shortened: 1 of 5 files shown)


pefernan (approved):

examples/jbpm-compact-architecture-example/src/main/resources/application.properties
```diff
@@ -9,11 +9,11 @@ quarkus.kogito.data-index.graphql.ui.always-include=true
 quarkus.http.test-port=0
 
 # Kogito-service
-kogito.service.url=http://localhost:8080
+kogito.service.url=http://0.0.0.0:8080
```
fantonangeli, line 12: @thiagoelg @pefernan Should I apply this change to:
`examples/sonataflow-greeting-quarkus-example/src/main/resources/application.properties`
in this PR: https://github.com/apache/incubator-kie-tools/pull/2511 ?

tiagobento (approved):

examples/jbpm-compact-architecture-example/src/main/resources/application.properties
thiagoelg: Maybe... we switched from `localhost` because `localhost` could have different meanings in different contexts (in a container, for example).

And our solution to run quarkus apps in dev mode on OpenShift was to replace `quarkus.http.host:quarkus.http.port` with `window.location.origin` basically.

ljmotta (approved):

examples/jbpm-compact-architecture-example/src/main/resources/application.properties
fantonangeli: thank you @thiagoelg

PR #2553, opened 2024-09-03
NO-ISSUE: Extended services wizard refer to non existing binaries

klbowers:
Relates to https://github.com/apache/incubator-kie-tools/pull/2533. This PR deals with the scenario when Extended Services stops running and the extended services wizard modal is displayed. This PR makes it so that, when the disableExtendedServicesWizard environmental variable is set to true, the disabled modal is displayed instead .

### DIFF
--- a/packages/online-editor/src/extendedServices/ExtendedServicesModal.tsx
+++ b/packages/online-editor/src/extendedServices/ExtendedServicesModal.tsx
@@ -612,6 +612,11 @@ export function ExtendedServicesModal() {
   useEffect(() => {
     if (extendedServices.status === ExtendedServicesStatus.NOT_RUNNING) {
       setModalPage(ModalPage.INITIAL);
+    } else if (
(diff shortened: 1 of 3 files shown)


thiagoelg (approved):

ljmotta (approved):

jomarko (changes requested):
@kbowers-ibm hi, I would like to ask before we merge the PR, is the old **Extended Services Setup** still displayed in some scenario? I am quite surprised we do not remove a code as part of similar PR.

Could you please explain this a little bit?

image

klbowers:
Hi @jomarko,
By default the disableExtendedServicesWizard environmental variable is set to false, so for most cases the default wizard that you showed above will be the one displayed.

jomarko (approved):
Thank you @kbowers-ibm , explained in private conversation.

yesamer (approved):

PR #2555, opened 2024-09-03
NO-ISSUE: Update screenshots for failing tests

danielzhe:
The test started to fail probably because of an update in Chrome that changed the behavior of the CSS since there was no PR with a failing test was merged.
Also, the others tests (Chromium and Webkit) didn't have any change.


tiagobento (approved):

ljmotta (approved):

yesamer:
@danielzhe It seems there's another screenshot to fix in Test Scenario Editor

danielzhe:
> @danielzhe It seems there's another screenshot to fix in Test Scenario Editor

Updated! It's weird how they were not failing before. I expect that maybe more can fail.

Btw, I checked each one of the failing tests just to be sure that there was nothing broken and it are just small details in the render.

klbowers (approved):
Thanks for sorting that out @danielzhe!

PR #2557, opened 2024-09-03
NO-ISSUE: Use correct properties on jbpm-compact-architecture-example

thiagoelg:
- Replace `kogito.dataindex.http.url` with `kogito.data-index.url`;
- Add CORS properties (to support any host)

### DIFF
--- a/examples/jbpm-compact-architecture-example/src/main/resources/application.properties
+++ b/examples/jbpm-compact-architecture-example/src/main/resources/application.properties
@@ -3,6 +3,8 @@
 
 #https://quarkus.io/guides/openapi-swaggerui
 quarkus.http.cors=true
+quarkus.http.cors.origins=*
+quarkus.dev-ui.cors.enabled=false
 quarkus.smallrye-openapi.path=/docs/openapi.json
 quarkus.swagger-ui.always-include=true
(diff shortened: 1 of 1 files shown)


tiagobento (approved):

examples/jbpm-compact-architecture-example/src/main/resources/application.properties
```diff
@@ -13,7 +15,7 @@ kogito.service.url=http://0.0.0.0:8080
 
 #Job-service
 kogito.jobs-service.url=http://0.0.0.0:8080
-kogito.dataindex.http.url=http://0.0.0.0:8080
+kogito.data-index.url=http://0.0.0.0:8080
```
pefernan, line 18: Ok! Looks like this prop (`kogito.data-index.url`) isn't needed in the compact setup for SVG addon since the communication between the data-index and the SVG addon works directly in VM via CDI .

Effectively, setting the `kogito.data-index.url` will prevent the Data Index Addon to auto generate this property.

pefernan (approved):
+1 Great! I tested all setups (dev mode + prod mod) and works fine, looks like the  `kogito.dataindex.http.url` is no longer needed in compact setup since the interactions will happen inside the VM :D

PR #2559, opened 2024-09-04
kie-issues#1365: Revisit all React-dependent packages on `kie-tools` declaring `react` as a `dependency` instead of a `peerDependency`

### LINKED ISSUE #1365: Revisit all React-dependent packages on `kie-tools` declaring `react` as a `dependency` instead of a `peerDependency`
React libraries shouldn't force a specific version of React, but rather declare its compatibility with a React version. This is most likely causing problems for people depending on React libraries/components published by `kie-tools`.

We need to check if Syncpack works for peerDependencies too, as we need alignment between all packages.

fantonangeli:
Closes https://github.com/apache/incubator-kie-issues/issues/1365

**Description:**
React libraries shouldn't force a specific version of React, but rather declare its compatibility with a React version. This is most likely causing problems for people depending on React libraries/components published by kie-tools.

We need to check if Syncpack works for peerDependencies too, as we need alignment between all packages.

**Notes:**
- I needed to upgrade syncpack to 13.0.0 to use sameRange policy
- @mareklibra I wanted to test this PR on the orchestrator plugin but maybe I need some help to do this test. Anyway I needed to use verdaccio server to publish locally the `kie-tools` packages to use in `backstage-plugins` 

Here is an analysis of the packages using react and which packages have been updated:
https://docs.google.com/spreadsheets/d/171HP_YReV1PSke71WE3gdxyrU8m4heTCpMKlEdMy-l0


packages/online-editor/package.json
```diff
@@ -123,5 +121,9 @@
     "webpack-cli": "^4.10.0",
     "webpack-dev-server": "^4.15.1",
     "webpack-merge": "^5.9.0"
+  },
+  "peerDependencies": {
+    "react": ">=17.0.2 <19.0.0",
+    "react-dom": ">=17.0.2 <19.0.0"
```
tiagobento: So `online-editor` is not a library... But it's also not something we publish as a package (hence `"private": true`). But I don't know... it doesn't feel right to put React and React DOM as peerDependencies? Also not as dependencies, since it really is just used by Webpack during bundling. Maybe `devDependencies` would be the correct place for all dependencies of private-application packages? Kind of a philosophical debate I know 😄 but I'd love to hear your thoughts!

tiagobento (approved):
Left a question inline... Thanks for the new syncpack setup, looks great!

packages/online-editor/package.json
fantonangeli: Thanks a lot for your point. I should not have modified the online-editor.  
I'm reverting and testing that package.json now as I think React should be just in `dependencies` section for a consumer package.
I modified the description with a spreadsheet listing the packages that have been updated in this PR.

thiagoelg (approved):

examples/base64png-editor/package.json
```diff
@@ -28,9 +28,7 @@
     "@kie-tools-core/patternfly-base": "workspace:*",
     "@kie-tools-core/workspace": "workspace:*",
     "@patternfly/react-core": "^4.276.6",
-    "@patternfly/react-icons": "^4.93.6",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
```
lordrip, line 33: [curious-question] Shouldn't we move this `react` and `react-dom` to the `devDependencies` object so we can still work with this editor in dev mode without accounting for another package in this workspace to bring them? :thinking:

fantonangeli: @lordrip Shouldn't be the final package, importing `@kie-tools-examples/base64png-editor`, to provide react in `dependencies` of `devDependencies`?

lordrip: @fantonangeli , yes, the final package (consumer) should definitively provide react. I was referring to another case though.

Let's ignore for a moment that `@kie-tools-examples/base64png-editor` belongs to this repository, so the package json for it would look like the following:

```json
{
  "name": "@kie-tools-examples/base64png-editor",
  ...
  "dependencies": {},
  "devDependencies": {
    "typescript": "5.0.0"
  },
  "peerDependencies": {
    "react": ">=17.0.2 <19.0.0",
    "react-dom": ">=17.0.2 <19.0.0"
  }
}
```

Now, performing a `yarn install` won't bring `react` to the project, so the project will fail when trying to run it in standalone mode as there's no react dependency in the `node_modules`.

Now, this repository is slightly different as it's very likely that some other package is bringing `react` to the `node_modules`, hence making it work, but if by any chance I bootstrap just the `@kie-tools-examples/base64png-editor` maybe it will fail, as some dependencies of other packages won't be brought?

I'm not familiar with how `pnpm` works in terms of bootstraping packages and whether if it will download dependencies for unresolved packages, but if it does, then I think we're good, otherwise, :boom: 

_Maybe :boom: is a bit too dramatic_

PR #2602, opened 2024-09-19
kie-issues#1376: The width information are not being deleted from DMN when the expression is deleted in the new Boxed Expression Editor

### LINKED ISSUE #1376: The width information are not being deleted from DMN when the expression is deleted in the new Boxed Expression Editor
In the new BEE, when the user deletes an expression the expression widths are not deleted from the model. Those widths should be removed.

See the video on how to reproduce. Notice in that case I already had a previous expression (not shown in the video), so at the end, we have 3 ComponentWidths nodes in the DMN file, while we should have only 1.

https://github.com/apache/incubator-kie-issues/assets/7305741/a12ad2c1-cfa6-4af4-9267-dec9579ba44a

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1376


packages/dmn-editor/src/boxedExpressions/BoxedExpressionScreen.tsx
```diff
@@ -89,7 +89,6 @@ export function BoxedExpressionScreen({ container }: { container: React.RefObjec
   const dmnEditorStoreApi = useDmnEditorStoreApi();
 
   const thisDmn = useDmnEditorStore((s) => s.dmn);
-  const diagram = useDmnEditorStore((s) => s.diagram);
```
danielzhe, line 92: Cleaning. This constant wast not being used.

packages/dmn-editor/src/mutations/updateExpressionWidths.ts
```diff
@@ -30,20 +30,9 @@ export function updateExpressionWidths({
   drdIndex: number;
   widthsById: Map<string, number[]>;
 }): void {
-  const { widthsExtension, widths } = addOrGetDrd({ definitions, drdIndex });
-  const componentWidthsMap = widths.reduce(
```
danielzhe, line 34: The issue was that we were always reloading all the widths from the `definitions` into the `widthsById`.

They are loaded the first time here: https://github.com/apache/incubator-kie-tools/blob/e21bbd8b7bc00bd36868af4b48ac3109048b4f3f/packages/dmn-editor/src/boxedExpressions/BoxedExpressionScreen.tsx#L146

And then the Boxed Expression Editor is who changes the `widthsById`.

The `updateExpressionWidth` should only update the `definitions` from `widthsById`, not fill the `widthsById`.

tiagobento:
`ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/dmn-editor/package.json`

jomarko (approved):
Seem to work properly to me. Thank you @danielzhe

tiagobento (approved):

PR #2607, opened 2024-09-20
kie-issues#1488: Importing `drools-build-parent` in Jbang script to enforce usage of correct depencendies version

### LINKED ISSUE #1488: Importing `drools-build-parent` in Jbang script to enforce usage of correct depencendies version.
In some conditions, we faced troubles building kie-tools  due to transitive SNAPSHOT dependency coming from org.jpmml:pmml-model:
`[ERROR] Could not resolve dependencies: Failed to collect dependencies at org.kie:kie-dmn-core:jar:9.1.1-ibm-0002 -> org.jpmml:pmml-model:jar:1.6.4 -> com.fasterxml.jackson.core:jackson-annotations:jar:2.12.6-SNAPSHOT`
This comes from a jbang command using jbang: -Dkogito-runtime.version=9.1.1-ibm-0002 /workspace/app/kie-tools/packages/dmn-marshaller-backend-compatibility-tester/src/DmnMarshallerBackendCompatibilityTesterScript.java
Looking at [the version definition in pmml-model](https://github.com/jpmml/jpmml-model/blob/1.6.4/pom.xml#L140) , it's defined as a range, but SNAPSHOTs should be (according to the internet) not resolved as matching the range criteria

To fix that, it's enough to directly import the `drools-build-parent` as a direct dependency in the script to point to the correct versions.

yesamer:
Closes: https://github.com/apache/incubator-kie-issues/issues/1488

In this PR:
- Import the `drools-build-parent` as a direct dependency in the script to point to the correct versions, to fix the issue described in the ticket
- Update Jbang to the latest 0.23.0 version.

### DIFF
--- a/packages/dmn-marshaller-backend-compatibility-tester/package.json
+++ b/packages/dmn-marshaller-backend-compatibility-tester/package.json
@@ -25,7 +25,7 @@
     "prefetch": "node dist/dependenciesFetch.js"
   },
   "dependencies": {
-    "@jbangdev/jbang": "0.2.0"
+    "@jbangdev/jbang": "0.2.3"
   },
   "devDependencies": {
     "@kie-tools/root-env": "workspace:*",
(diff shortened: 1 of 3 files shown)


gitgabrio (approved):

jstastny-cz (approved):
LGTM. Thank you.

tiagobento (approved):

PR #2611, opened 2024-09-23
kie-issues#1158: On the Boxed Expression Component, changing the ExpressionVariableCell text content will fail to commit if the focus is changed to another cell

### LINKED ISSUE #1158: On the Boxed Expression Component, changing the ExpressionVariableCell text content will fail to commit if the focus is changed to another cell
For boxed expressions `some`, `every`, `invocation` and `decision table annotation` user can not commit the text of specific cell, see the attached screenshot by click into cell that is `Undefined`, clicking outside the boxed expression most parent works and changes are committed. 

Screenshot 2024-05-06 175436

### Steps to reproduce
- Start to type a text into the red cell
- Click into the blue cell - **text is lost**
- vs click outside the boxed expression root - **text is saved**

jomarko:
Closes: https://github.com/apache/incubator-kie-issues/issues/1158

Screenshot 2024-09-25 090940

In the attached screenshot we can notice:
- rectangle containing "edited name here", that has currently a focus, it is a html `input` element
- dark blue rectangle wrapping this `input`, basically a table cell

The issue is about committing the "edited name here" text using mouse click. Currently user can commit changes by clicking:
- outside whole "Invocation" expression, it means outside table
- or somewhere in dark blue rectangle but outside "edited name here" rectangle

Otherwise, the [`onBlur` handler](https://github.com/apache/incubator-kie-tools/blob/main/packages/boxed-expression-component/src/table/BeeTable/InlineEditableTextInput.tsx#L54) of the "edited name here" is not invoked. So the current draft solution is to force blur event when some table cell is clicked.


tiagobento:
@jomarko Can you please explain this solution?

jomarko:
I have removed the flags, as the PR is basically done, however screenshots comparison tests are failing due to the difference of the GitHub Ubuntu machine and my local Ubuntu machine. Figuring out how to generate proper screenshots.

tiagobento (approved):

klbowers (approved):

yesamer (approved):

PR #2639, opened 2024-10-03
NO-ISSUE: Increase Playwright toHaveScreenshot assertion threshold to 10%

ljmotta:
This is a temporary solution.

We don't have a reproducible environment to locally create the screenshots. This causes minimal differences in the comparison of local screenshots with the CI screenshots. Increasing the threshold will avoid failing tests due to these minimal differences, but broken screenshots will make the test pass as well.

### DIFF
--- a/packages/boxed-expression-component/playwright.config.ts
+++ b/packages/boxed-expression-component/playwright.config.ts
@@ -40,7 +40,7 @@ const customConfig = defineConfig({
     toHaveScreenshot: {
       // An acceptable ratio of pixels that are different to the
       // total amount of pixels, between 0 and 1.
-      maxDiffPixelRatio: 0.001,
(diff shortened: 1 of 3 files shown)


jomarko (commented):
my tests (boxed-expression-component) are failing and reporting 0.01 of pixels to differ, not sure about other team members, I mean maybe we do not need 0.1, maybe sth like 0.08, 0.05 ...

otherwise I agree with this as temporal solution

ljmotta:
@jomarko If we increase past 1~2% we're already allowing errors to be asserted as they don't exist. This was the reason behind the previous 0.1% threshold. So I don't think changing to a smaller number will make any difference.

yesamer (approved):

tiagobento (approved):

PR #2655, opened 2024-10-10
NO-ISSUE: Add license header

jomarko:


packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-marshalling/src/main/java/org/kie/workbench/common/stunner/bpmn/client/marshall/converters/tostunner/properties/util/CorrelationPropertyReaderUtils.java
```diff
@@ -1,5 +1,25 @@
 package org.kie.workbench.common.stunner.bpmn.client.marshall.converters.tostunner.properties.util;
 
+/*
```
yesamer, line 3: Why after "package" ?

packages/stunner-editors/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-marshalling/src/main/java/org/kie/workbench/common/stunner/bpmn/client/marshall/converters/tostunner/properties/util/CorrelationReaderData.java
```diff
@@ -1,5 +1,25 @@
 package org.kie.workbench.common.stunner.bpmn.client.marshall.converters.tostunner.properties.util;
 
+/*
```
yesamer, line 3: Why after "package" ?

packages/stunner-editors/errai-common/src/main/java/org/jboss/errai/common/compat/javax/annotation/processing/Generated.java
```diff
@@ -40,6 +40,26 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+
```
yesamer, line 62: Why after "package" ?

yesamer: @jomarko This is a copied file from javax.annotation module.... not sure we can mark it with our package, please double check

packages/stunner-editors/errai-common/src/main/java/org/jboss/errai/common/client/util/TimeUnit.java
```diff
@@ -6,6 +6,26 @@
 
 package org.jboss.errai.common.client.util;
 
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
```
yesamer, line 13: Why after "package" ?

jomarko:
@yesamer hi, thank you for comments. In this message, lets discuss `Why after "package"`.

I used https://creadur.apache.org/rat017/apache-rat/ tool for adding the headers, especially the `--edit-license` option. I and @jstastny-cz used this tool for licenses presence verification, so I thought using it for license addition will be not problematic.

I see your point, that it is not consistent with the rest of similar files. Now it is question, if we have other automatic way of adding headers that would be:
- compatible with the kie build
- compatible with the existing headers placement
- compatible with the `apache-rat` automatic checks or if we have other tool for apache licenses header checks

Situation is even more complicated, for some yaml files related to helm-chart that I thought they need to license, running `pnpm format` will remove headers from them.

yesamer:
@jomarko Thank you for the explanation. It makes sense, and ultimately those Java files are part of the old editors... which should be dismissed in the future. I still believe that the Generated class shouldn't be marked with the Apache header, as copied from javax package.

jomarko:
This will be addressed via https://github.com/apache/incubator-kie-issues/issues/1613

PR #2657, opened 2024-10-11
kie-issues#1527: DMN Editor: Create new overlay option 'Highlight evaluation results'

### LINKED ISSUE #1527: DMN Editor: Create new overlay option “Highlight evaluation results”
Add a new option in the overlay menu to enable/disable the evaluation highlights as shown in [1]. This option must be available in the editor state as the other overlay options, and must be present only when the DMN Editor has a specific prop **"evaluationHighlights: boolean"**. After enabled, a badge will appear in the DMN Editor [2]. Clicking on the badge will disable the overlay and will hide the badge. The badge will be always present in the Boxed Expression Editor, either enabled or disabled as illustrated in [3].

[1]
image-2024-10-09-19-16-23-450

[2]
image-2024-10-09-19-16-32-095

[3]
image-2024-10-09-19-16-37-632

jomarko:
Closes: https://github.com/apache/incubator-kie-issues/issues/1527


packages/dmn-editor/src/overlaysPanel/OverlaysPanel.tsx
```diff
@@ -159,6 +161,27 @@ export function OverlaysPanel({ availableHeight }: OverlaysPanelProps) {
             }
           />
         </FormGroup>
+        <FormGroup
+          label={"Enable evaluation highlights"}
+          labelIcon={
+            <Tooltip
+              content={"Enable evaluation highlights on nodes(decisions?), decision tables and conditional expressions"}
```
jomarko: I will appreciate any hints, how to update this tooltip text. Not sure how much information it should provide.

packages/dmn-editor/src/store/Store.ts
```diff
@@ -110,6 +110,7 @@ export interface State {
       enableExecutionHitsHighlights: boolean;
       enableDataTypesToolbarOnNodes: boolean;
       enableCustomNodeStyles: boolean;
+      evaluationHighlights: boolean;
```
jomarko: there is existing **enableExecutionHitsHighlights**, however not sure if we should:
- keep it
- replace it with **evaluationHighlights**
- other action

tiagobento: `enableExecutionHitsHighlights` was supposed to be exactly what `evaluationHighlights` is going to be. Please remove `enableExecutionHitsHighlights` and continue with `evaluationHighlights`. Note that `enableEvaluationHighlights` is a better name.

packages/dmn-editor/src/overlaysPanel/OverlaysPanel.tsx
tiagobento: "Enable highlighting Decision Table rules and Boxed Conditional Expression branches based on evaluation results, also showing success/error status badges on Decision nodes."

jomarko: thank you

packages/dmn-editor/src/store/Store.ts
jomarko: thank you, applied

packages/dmn-editor/src/evaluationHighlights/EvaluationHighlightsBadge.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import * as React from "react";
+import { useDmnEditorStore, useDmnEditorStoreApi } from "../store/StoreContext";
+
+export function EvaluationHighlightsBadge() {
+  const dmnEditorStoreApi = useDmnEditorStoreApi();
+  const isEvaluationHighlights = useDmnEditorStore((s) => s.diagram.overlays.enableEvaluationHighlights);
```
tiagobento:
```suggestion
  const isEvaluationHighlightsEnabled = useDmnEditorStore((s) => s.diagram.overlays.enableEvaluationHighlights);
```

tiagobento (dismissed):

tiagobento:
Other than that ^ looks good to go! :)

packages/dmn-editor/src/evaluationHighlights/EvaluationHighlightsBadge.tsx
jomarko: thank you @tiagobento, I agree, it is a better name, changes pushed

tiagobento:
image


Border looks a little off, and the tooltip says (beta), but it isn't.

Also, I think the idea of having this as a https://www.patternfly.org/components/label/ instead of a "Button" is important to give it a different look compared to the other buttons, since this this is not really a button, but more like an indicator that the Evaluation highlights are turned on.

Same is true for the toggle on BEE.

We also need to be mindful of the case we choose when writing text on-screen. I always prefer to capitalize only the first word, not all words, so the label would read "Evaluation highlights: on/off", instead of "Evaluation Highlights: On/Off".

Another thing that caught my eye was the little displacement that happened when the I toggled between on and off on BEE, because of difference in text size. Being mindful of layout shifts is also very important for a good experience when using a web app.

Please consider my feedback, and let me know if you have questions!

jomarko:
@tiagobento good feedback, I will try to incorporate it.

> Border looks a little off, and the tooltip says (beta), but it isn't.

I will remove the "(beta)". Oh, you are correct, I haven't noticed the border off placement. It is an example that author eyes are focused on other things than reviewer eyes. Thank you for spotting.

> Also, I think the idea of having this as a https://www.patternfly.org/components/label/ instead of a "Button" is important to give it a different look compared to the other buttons, since this this is not really a button, but more like an indicator that the Evaluation highlights are turned on.

Ok, probably this I didn't understand from mockup. There it seemed to me as UI element with same semantic as other elements in that area, that is why I used the button. I will incorporate your proposal.

> Same is true for the toggle on BEE.

Sure.

> We also need to be mindful of the case we choose when writing text on-screen. I always prefer to capitalize only the first word, not all words, so the label would read "Evaluation highlights: on/off", instead of "Evaluation Highlights: On/Off".

I do not have preference about this. I don't mind follow your suggestion.

> Another thing that caught my eye was the little displacement that happened when the I toggled between on and off on BEE, because of difference in text size. Being mindful of layout shifts is also very important for a good experience when using a web app.

WIll try to fix the size of the new toggle/label. However, that will probably won't work with potential i18n in future.

jomarko:
Changes pushed:
Screenshot 2024-10-18 121225
Screenshot 2024-10-18 121220
Screenshot 2024-10-18 121209

ljmotta:
@jomarko The code looks great, but I found an issue in the UI. Using the toggle makes the overflow Y appear. This only happens when the Overlay panel is close to the bottom of the Editor, which is the case for me in the default zoom (100%).


https://github.com/user-attachments/assets/2c0a0c2d-5106-48c6-b19b-11216cc57eac


BUG (100%):

image

image


OK (90%): 

image

ljmotta (approved):
@jomarko I took a look into `main` and I could reproduce the bug by zooming in. I'll file a new issue for the behavior I described in my comment.

EDIT: https://github.com/apache/incubator-kie-issues/issues/1679

yesamer (approved):

PR #2675, opened 2024-10-16
[Incubator-kie-issues#1314] @kie-tools/jbpm-quarkus-devui: Wrong behaviors on the Error notification in Start Process Form component 

### LINKED ISSUE #1314: `@kie-tools/jbpm-quarkus-devui`: Wrong behaviours on the Error notification in `Start Process Form` component
If we try to start a process instance in the `jbpm-quarkus-devui` and something goes wrong after submitting the form, it shows an error dialog providing some feedback to the user. 

Image

However there are two issues identified:

1. **View Details** link should expand the dialog showing some information regarding the error, instead it breaks showing a blank schreen
Image
In console the following error appears:
```
Uncaught Error: Objects are not valid as a React child (found: object with keys {failedNodeId, id, message}). If you meant to render a collection of children, use an array instead.
    at throwOnInvalidObjectType (react-dom.development.js:13231:1)
    at reconcileChildFibers (react-dom.development.js:14133:1)
    at reconcileChildren (react-dom.development.js:16990:1)
    at updateHostComponent (react-dom.development.js:17632:1)
    at beginWork (react-dom.development.js:19080:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)
    at invokeGuardedCallback (react-dom.development.js:4056:1)
    at beginWork$1 (react-dom.development.js:23964:1)
    at performUnitOfWork (react-dom.development.js:22779:1)
```

2. The **Go to Process Details** open the `ProcessDetails` screen but passing an `undefined` **processId** making the `ProcessDetails` break. (In this case a ProcessInstance has been started in a error state)

This issues make me think that the component isn't getting the right information from the backend when an error happens.

Accpetance Criteria:

- [ ] Verify that the component gets the right information from the channel if an error happened after submit.
- [ ] Make **View Details** show extra information if available, if not possible just hide it.
- [ ] Make **Go to Process Details** only visible if there's a **processId**
- [ ] Make  **ProcessDetails** page more robust and don't break if no **processId** is provided in the URL

> **NOTE:** This issue has been found as a side effect to https://github.com/apache/incubator-kie-issues/issues/1313, maybe the errors are related, but I think it is worth investigating them separately

bncriju:
Fixes https://github.com/apache/incubator-kie-issues/issues/1314

Fixed code so as to meet the following acceptance criteria mentioned in the issue.
 Verify that the component gets the right information from the channel if an error happened after submit.
 Make View Details show extra information if available, if not possible just hide it.
 Make Go to Process Details only visible if there's a processId
 Make ProcessDetails page more robust and don't break if no processId is provided in the URL

### DIFF
--- a/packages/runtime-tools-process-dev-ui-webapp/src/components/containers/ProcessFormContainer/ProcessFormContainer.tsx
+++ b/packages/runtime-tools-process-dev-ui-webapp/src/components/containers/ProcessFormContainer/ProcessFormContainer.tsx
@@ -57,7 +57,9 @@ const ProcessFormContainer: React.FC<ProcessFormContainerProps & OUIAProps> = ({
               onSubmitSuccess(id);
             })
             .catch((error) => {
-              const message = error.response ? error.response.data : error.message;
+              const message = error.response
+                ? ` ${error.message}   ' ${error.response.request.statusText} '`
+                : error.message;
               onSubmitError(message);
             });
         },
--- a/packages/runtime-tools-process-dev-ui-webapp/src/components/containers/ProcessFormContainer/ProcessFormContainer.tsx
+++ b/packages/runtime-tools-process-dev-ui-webapp/src/components/containers/ProcessFormContainer/ProcessFormContainer.tsx
@@ -57,9 +57,7 @@ const ProcessFormContainer: React.FC<ProcessFormContainerProps & OUIAProps> = ({
               onSubmitSuccess(id);
             })
             .catch((error) => {
-              const message = error.response
-                ? ` ${error.message}   ' ${error.response.request.statusText} '`
-                : error.message;
+              const message = error.response ? `${error.response.statusText} : ${error.message}` : error.message;
               onSubmitError(message);
             });
         },
--- a/packages/runtime-tools-process-dev-ui-webapp/src/components/pages/ProcessFormPage/ProcessFormPage.tsx
+++ b/packages/runtime-tools-process-dev-ui-webapp/src/components/pages/ProcessFormPage/ProcessFormPage.tsx
@@ -72,18 +72,21 @@ const ProcessFormPage: React.FC<OUIAProps> = ({ ouiaId, ouiaSafe }) => {
             goToProcessDefinition();
           },
         },
-        {
-          label: "Go to Process details",
-          onClick: () => {
-            setNotification(null);
-            goToProcessDetails();
-          },
-        },
       ],
       close: () => {
         setNotification(null);
       },
     });
+    return (
+      <>
+        {processId && (
+          <div>
+            <label>Go to Process details</label>
+            <button onClick={handleClick}>Click Here</button>
+          </div>
+        )}
+      </>
+    );
   };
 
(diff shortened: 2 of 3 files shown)


packages/runtime-tools-process-dev-ui-webapp/src/components/pages/ProcessFormPage/ProcessFormPage.tsx
```diff
@@ -97,6 +100,11 @@ const ProcessFormPage: React.FC<OUIAProps> = ({ ouiaId, ouiaSafe }) => {
     showNotification("error", message, details);
   };
 
+  const handleClick = () => {
+    setNotification(null);
+    goToProcessDetails();
+  };
```
tiagobento: Please memoize this by wrapping it in a `useCallback`.

bncriju: Thanks for the comment. It is done. Please review the same.

packages/runtime-tools-process-dev-ui-webapp/src/components/containers/ProcessFormContainer/ProcessFormContainer.tsx
```diff
@@ -57,7 +57,9 @@ const ProcessFormContainer: React.FC<ProcessFormContainerProps & OUIAProps> = ({
               onSubmitSuccess(id);
             })
             .catch((error) => {
-              const message = error.response ? error.response.data : error.message;
+              const message = error.response
+                ? ` ${error.message}   ' ${error.response.request.statusText} '`
```
pefernan: Nip picking here. In this case I'd like to have a separator between both messages. I'd use `error.response.statusText` instead too.

Something like:


```suggestion
                ? `${error.response.statusText}': ${error.message}`
```

Notice I removed some blank spacess from the string too.

pefernan (changes requested):
Great work @bncriju! The solution works well. I just added a suggestion regarding the error message format. Otherwise everything well :D

danielzhe:
I did some manual tests and I think it is working great! I wish we could have a better way to inform the user exactly what happened, but due to the nature of the `error.response` being `any`, I think what you did is what we can do for now since it can be anything. Just address what @pefernan suggested and from my point of view, we're good to merge.

About the tests you mentioned in the private chat, we don't have any automated tests at all in this project so the manual tests you and the reviews did is what we have for now. Good work, @bncriju .

danielzhe (approved):

packages/runtime-tools-process-dev-ui-webapp/src/components/containers/ProcessFormContainer/ProcessFormContainer.tsx
bncriju: Thank you for the comment and suggestion. It's done. Please have a look now.

danielzhe (approved):

pefernan (approved):

PR #2681, opened 2024-10-18
kie-issues#1547: DMN Editor: Render evaluation highlights in the Boxed Expression Editor

### LINKED ISSUE #1547: DMN Editor: Render evaluation highlights in the Boxed Expression Editor (Decision Tables + Boxed Conditional Expressions)
This issue is another part related to https://github.com/apache/incubator-kie-issues/issues/1527

We should extend the functionality of the boxed-expression-component to be able to highlight evaluation hit count in the Decision Table and Boxed Conditional Expression using the evaluation results of DMN Runner. Disabling the "Evaluation Highlights" (#1527) must normally render the Boxed Expression Editor.
 
Figure 1 shows a Decision Table that had multiple rule hits and is highlighted in a purple color (rgb(215 201 255)). The number of hits is represented in the top left corner of the index row with a triangular badge. Rules that didn't have a hit must appear dimmed, with a grayscale and reduced opacity. In the case of the DMN Runner table, hovering the mouse on top of the badge must render the responsible input rows as illustrated.

image-2024-10-09-19-36-47-350

Figure 2 shows a Boxed Conditional Expression with the "then" expression highlighted, and the "else" expression dimmed. This example follows the same changes of the Decision Table highlights. 

image-2024-10-09-19-40-48-918

Figure 3 shows an example with nested expressions and a high number of hits. When the number of hits gets higher, the font size gets lower to accommodate the value inside the badge. The expression header inside the condition is highlighted to show the user that the entire expression is part of the condition.
image-2024-10-09-19-44-01-162

jomarko:
Closes: apache/incubator-kie-issues/issues/1547

### checklist
- [x] decision table
- [x] conditional
- [x] context additional row
- [x] nested expressions

### testing
New `Evaluation Hits` stories were added into `Conditional` and `Decision Table` stories.


packages/boxed-expression-component/src/table/BeeTable/BeeTable.css
```diff
@@ -203,6 +203,46 @@
   border: 0;
 }
 
+.expression-container .table-component tr.evaluation-highlights-row-overlay > td:first-child::before {
```
jomarko: `first-child` here was my decision to use, otherwise there is problem, when `then`/ `else` branch has nested decision table, but not all decision table rules were evaluated. see:
Screenshot 2024-10-30 132337

packages/boxed-expression-component/src/api/BeeTable.ts
```diff
@@ -97,6 +97,10 @@ export interface BeeTableProps<R extends object> {
   shouldShowColumnsInlineControls: boolean;
   resizerStopBehavior: ResizerStopBehavior;
   lastColumnMinWidth?: number;
+  /** Index of the column, where the evaluation hits count should be displayed. If not set, evaluation hits count number is not shown. */
+  evaluationHitsCountColumnIndex?: number;
+  /** Method should return true for table rows, that can display evaluation hits count, false otherwise. If not set, BeeTableBody defaults to false. */
+  getEvaluationHitsCountSupportedByRow?: (row: ReactTable.Row<R>) => boolean;
```
ljmotta: The name is a little strange for me. The idea is to the get if a row supports (or not) the "EvaluationHitsCount". I don't think the "ByRow" makes sense, so:
- `isEvaluationHitCountSupported(row: Row)`: The `is` tells the method will be boolean
- `supportsEvaluationHitCount(row: Row)`

Another idea is to use the `args: { row: Row }`, but I don't think is necessary, as we just have one parameter.

packages/boxed-expression-component/src/expressions/ConditionalExpression/ConditionalExpression.tsx
```diff
@@ -247,6 +256,8 @@ export function ConditionalExpression({
           shouldRenderRowIndexColumn={false}
           shouldShowRowsInlineControls={false}
           shouldShowColumnsInlineControls={false}
+          evaluationHitsCountColumnIndex={1}
```
ljmotta: Why `1`? Could you please add a comment here? Or make it a variable with a very good name.

packages/boxed-expression-component/src/expressions/DecisionTableExpression/DecisionTableExpression.tsx
```diff
@@ -1073,6 +1077,8 @@ export function DecisionTableExpression({
         shouldRenderRowIndexColumn={true}
         shouldShowRowsInlineControls={true}
         shouldShowColumnsInlineControls={true}
+        evaluationHitsCountColumnIndex={0}
```
ljmotta: I get the `0` here, but it would be good to add a comment too.

packages/boxed-expression-component/src/BoxedExpressionEditorContext.tsx
```diff
@@ -34,6 +34,7 @@ export interface BoxedExpressionEditorContextType {
   pmmlDocuments?: PmmlDocument[];
   dataTypes: DmnDataType[];
   isReadOnly?: boolean;
+  evaluationHitsCountPerId?: Map<string, number>;
```
ljmotta:
```suggestion
  evaluationHitsCountById?: Map<string, number>;
```

To keep consistency with other variables in the codebase.

packages/boxed-expression-component/src/table/BeeTable/BeeTableBody.tsx
```diff
@@ -72,15 +77,29 @@ export function BeeTableBody<R extends object>({
...
       reactTableInstance.prepareRow(row);
 
+      const rowKey = getRowKey(row);
+      const isEvalationHitsCountSupportedByRow: boolean = getEvaluationHitsCountSupportedByRow?.(row) ?? false;
+      const rowEvaluationHitsCount = evaluationHitsCountPerId ? evaluationHitsCountPerId?.get(rowKey) ?? 0 : undefined;
+      const canDisplayEvaluationHitsCountRowOverlay =
+        rowEvaluationHitsCount !== undefined && isEvalationHitsCountSupportedByRow === true;
+      const rowClassName = canDisplayEvaluationHitsCountRowOverlay
+        ? rowKey + (rowEvaluationHitsCount > 0 ? " evaluation-hits-count-row-overlay" : "")
+        : rowKey;
```
ljmotta: This part is a little dense, a lot of information here.
 - Could you please inline the `isEvalationHitsCountSupportedByRow` as it's used only one time?
 - I'm not sure about this one, but can you force the ternary to be written in multiple lines?
 - Could you please add a line break between? As multiple ternaries can be very difficult to read.

packages/boxed-expression-component/src/table/BeeTable/BeeTableTd.tsx
```diff
@@ -225,6 +231,12 @@ export function BeeTableTd<R extends object>({
     return onDataCellKeyUp?.(column.id);
   }, [column.id, onDataCellKeyUp]);
 
+  const evaluationHitsCountBadgeClassName = canDisplayEvaluationHitsCountBadge
+    ? (evaluationHitsCount ?? 0) > 0
+      ? "evaluation-hits-count-badge-colored"
+      : "evaluation-hits-count-badge-non-colored"
+    : "";
```
ljmotta: Please, use an `useMemo` hook here.

ljmotta (changes requested):
@jomarko Thanks a lot for your PR. I didn't have time to manually test it, but I will. Until then, I've made some comments inline.

packages/boxed-expression-component/src/table/BeeTable/BeeTableTd.tsx
jomarko: thank you, used

packages/boxed-expression-component/src/api/BeeTable.ts
jomarko: understand, done

packages/boxed-expression-component/src/table/BeeTable/BeeTableBody.tsx
jomarko: the ternary operator is tricky

there is an eslint rule, that would force what you propose:
```
https://eslint.style/rules/js/multiline-ternary
```

However once I used it in project and formatted ternary operators as multiple lines, the current prettier configuration, the pre-commit hook for formatting we have reverted those changes, or adapted them in a way, that caused eslint again complaining..

So we have currently a prettier configuration that is not compatible with this eslint rule.

ljmotta:
@jomarko Thanks for applying the changes.

After manually testing it, I could find a CSS that will highlight as first discussed:

```css
.expression-container-box .conditional-expression .table-component tr.evaluation-hits-count-row-overlay > td:first-child::before {
  content: "";
  position: absolute;
  background-color: rgb(215, 201, 255, 0.5);
  left: 0;
  top: 0;
  width: 100%;
  height: 100%;
}

.expression-container-box .conditional-expression .table-component tr.evaluation-hits-count-row-overlay > td > div > div > div > .logic-type-selected-header::before {
  content: "";
  position: absolute;
  background-color: rgb(215, 201, 255, 0.5);
  left: 0;
  top: 0;
  width: 100%;
  height: 100%;
}

.expression-container-box .decision-table-expression .table-component tr.evaluation-hits-count-row-overlay > td::before {
  content: "";
  position: absolute;
  background-color: rgb(215, 201, 255, 0.5);
  left: 0;
  top: 0;
  width: 100%;
  height: 100%;
}
```

You will need to add a `conditional-expression` class to the `ConditionalExpression.tsx` file:

```tsx
      <div className={"conditional-expression"} data-testid={"kie-tools--boxed-expression-component---conditional"}>
        <BeeTable<ROWTYPE>
           ...
```

image

I guess it can be simplified, but I broke into the `conditional-expression`, `decision-table-expression` and `logic-type-selected-header` CSS.

packages/boxed-expression-component/src/expressions/ConditionalExpression/ConditionalExpression.tsx
jomarko: removed from the codebase

packages/boxed-expression-component/src/expressions/DecisionTableExpression/DecisionTableExpression.tsx
jomarko: removed from the codebase

klbowers (approved):
Thanks @jomarko everything looks great to me!

packages/boxed-expression-component/stories/boxedExpressionStoriesWrapper.tsx
ljmotta, line 1: It's not necessary to modify the `BoxedExpressionEditorStoryArgs` type, as the `BoxedExpressionEditorProps` already contains the `evaluationHitsCountById` property.

Also, the `useState` isn't required.

The evaluationHitsCountById should have the same value as other props:

```tsx
evaluationHitsCountById={props?.evaluationHitsCountById ?? args?.evaluationHitsCountById}
```

ljmotta (changes requested):
@jomarko I've just one last comment regarding the `boxedExpressionStoriesWrapper`.

packages/boxed-expression-component/stories/boxedExpressionStoriesWrapper.tsx
jomarko: thank you, I am happy we can keep it simple

ljmotta (approved):

PR #2685, opened 2024-10-19
kie-issues#1549: kn-workflow-plugin check for the presence of an image in the local Docker image does not cover all cases

### LINKED ISSUE #1549: kn-workflow-plugin check for the presence of an image in the local Docker image does not cover all cases
We should add better checks. because --filter=reference="*/image:main" doesn't work for all cases

treblereel:
Closes https://github.com/apache/incubator-kie-issues/issues/1549

### DIFF
--- a/packages/kn-plugin-workflow/pkg/common/containers.go
+++ b/packages/kn-plugin-workflow/pkg/common/containers.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/docker/distribution/reference"
 	"io"
 	"os"
 	"os/exec"
@@ -35,7 +36,6 @@ import (
 	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/metadata"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
-	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/go-connections/nat"
@@ -49,6 +49,10 @@ const (
 type DockerLogMessage struct {
 	Status string `json:"status,omitempty"`
 	ID     string `json:"id,omitempty"`
+}
+
+type DockerClient interface {
+	ImageList(ctx context.Context, options types.ImageListOptions) ([]types.ImageSummary, error)
 }
 
 func getDockerClient() (*client.Client, error) {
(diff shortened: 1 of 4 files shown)


packages/kn-plugin-workflow/pkg/common/containers.go
```diff
@@ -51,6 +51,10 @@ type DockerLogMessage struct {
 	ID     string `json:"id,omitempty"`
 }
 
+type DockerClient interface {
+	ImageList(ctx context.Context, options types.ImageListOptions) ([]types.ImageSummary, error)
```
ricardozanini, line 58: Does it work for podman too? 🤔

treblereel: @ricardozanini you can take a look at `func RunContainerCommand` at containers.go,  only docker uses it.

I plan to refactor it a little in the future, this func can be improved

ricardozanini: oh so this problem won't happen with podman?

treblereel: @ricardozanini podman is awesome!

ricardozanini (approved):
@domhanak can you verify it locally?

domhanak:
@ricardozanini looking

packages/kn-plugin-workflow/pkg/common/containers_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+		{"docker.io/apache/incubator-kie-sonataflow-devmode", []string{"docker.io/apache/incubator-kie-sonataflow-devmode:latest"}, true},
+		{"docker.io/apache/incubator-kie-sonataflow-devmode", []string{"apache/incubator-kie-sonataflow-devmode:latest"}, true},
+
+		{"apache/incubator-kie-sonataflow-devmode:main", []string{"docker.io/apache/incubator-kie-sonataflow-devmode:main"}, true},
+		{"apache/incubator-kie-sonataflow-devmode:main", []string{"apache/incubator-kie-sonataflow-devmode:main"}, true},
+
+		{"docker.io/apache/incubator-kie-sonataflow-devmode:main", []string{"incubator-kie-sonataflow-devmode:main"}, false},
+		{"docker.io/apache/incubator-kie-sonataflow-devmode", []string{"incubator-kie-sonataflow-devmode:latest"}, false},
+		{"apache/incubator-kie-sonataflow-devmode:main", []string{"incubator-kie-sonataflow-devmode:main"}, false},
```
tiagobento: Are these fictitious, test-only names? If so, please don't use the same names we're using for real artifacts. People without context might think this is some sort of "indirect dependency" on these images.

ricardozanini (approved):

tiagobento (approved):

domhanak (approved):

PR #2687, opened 2024-10-21
kie-issues#1481: Test Scenario Editor: Create `@kie-tools/scesim-editor-envelope`

### LINKED ISSUE #1481: Test Scenario Editor: Create `@kie-tools/scesim-editor-envelope`
To make the new TestScenario Editor available on our platform through Multiplying Architecture, we first need to wrap <TestScenarioEditor> inside an Envelope — An application that is prepared to run inside an iframe and listens to messages coming from the hosting application (Channels) and sends messages back to it. Each hosting application implements an API that Envelopes can consume, and each Envelope implements an API that Channels can consume. Communication between Envelopes and Channels is two-way.
This will be achieved by creating a new package, **@kie-tools/scesim-editor-envelope**, containing the EditorFactory for the new Test Scenario Editor.

yesamer:
Closes: https://github.com/apache/incubator-kie-issues/issues/1481

In this PR:
- Creating `@kie-tools/scesim-editor-envelope`, module. The module functionality is still not properly tested. This will be done during https://github.com/apache/incubator-kie-issues/issues/970 and https://github.com/apache/incubator-kie-issues/issues/1552
- Preliminary work for https://github.com/apache/incubator-kie-issues/issues/970

### DIFF
--- a/packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
+++ b/packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
@@ -329,7 +329,6 @@ export class TestScenarioEditorRoot extends React.Component<TestScenarioEditorRo
                 this.state.openFilenormalizedPosixPathRelativeToTheWorkspaceRoot
               }
             />
-            /
             {
               <ExternalModelsManager
                 workspaceRootAbsolutePosixPath={this.props.workspaceRootAbsolutePosixPath}
--- a/packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
+++ b/packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
@@ -325,7 +325,7 @@ export class TestScenarioEditorRoot extends React.Component<TestScenarioEditorRo
               onRequestExternalModelByPath={this.onRequestExternalModelByPathsRelativeToTheOpenFile}
               onRequestToJumpToPath={this.onOpenFileFromPathRelativeToTheOpenFile}
               onRequestToResolvePath={this.onRequestToResolvePathRelativeToTheOpenFile}
-              openFilenormalizedPosixPathRelativeToTheWorkspaceRoot={
+              openFileNormalizedPosixPathRelativeToTheWorkspaceRoot={
                 this.state.openFilenormalizedPosixPathRelativeToTheWorkspaceRoot
               }
             />
--- a/packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
+++ b/packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
@@ -1,0 +1,518 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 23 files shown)


packages/scesim-editor/src/TestScenarioEditor.tsx
```diff
@@ -101,10 +125,33 @@ export type TestScenarioEditorProps = {
...
    */
   onModelDebounceStateChanged?: (changed: boolean) => void;
+  /**
+   * Called when the contents of a specific available model is necessary. Used by the "Included models" tab.
+   */
+  onRequestExternalModelByPath?: OnRequestExternalModelByPath;
+  /**
+   * Called when the list of paths of available models to be included is needed. Used by the "Included models" tab.
+   */
+  onRequestExternalModelsAvailableToInclude?: OnRequestExternalModelsAvailableToInclude;
```
jomarko: Does the SceSim editor has 'Included Models' tab? I thought it is only in the DMN Editor.

packages/scesim-editor/src/TestScenarioEditor.tsx
```diff
@@ -101,10 +125,33 @@ export type TestScenarioEditorProps = {
...
+  /**
+   * When users want to jump to another file, this method is called, allowing the controller of this component decide what to do.
+   * Links are only rendered if this is provided. Otherwise, paths will be rendered as text.
+   */
+  onRequestToJumpToPath?: OnRequestToJumpToPath;
+  /**
+   * All paths inside the DMN Editor are relative. To be able to resolve them and display them as absolute paths, this function is called.
+   * If undefined, the relative paths will be displayed.
+   */
+  onRequestToResolvePath?: OnRequestToResolvePath;
```
jomarko: similar question, the comment mention DMN Editor while it does not mention SceSim what makes me unsure, if this is related to SceSim.

packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
```diff
@@ -0,0 +1,517 @@
...
+    const ext = __path.extname(normalizedPosixPathRelativeToTheOpenFile);
+    if (ext === ".dmn") {
+      return {
+        normalizedPosixPathRelativeToTheOpenFile,
+        type: "dmn",
+        model: normalize(getDmnMarshaller(resource?.content ?? "", { upgradeTo: "latest" }).parser.parse()),
+        svg: "",
+      };
+    } else {
+      throw new Error(`Unknown extension '${ext}'.`);
```
jomarko, line 262: maybe is more unsupported than unknown, or will this new editor has single option - DMN? previously we had DMN and DRL

jomarko (commented):
Thank you @yesamer . Using the vsix you shared privately, I was not able to open empty scesim file, is that expected? I didn't try to open non empty scesim file.



image

packages/scesim-editor/src/TestScenarioEditor.tsx
yesamer: Fixed.

packages/scesim-editor/src/TestScenarioEditor.tsx
yesamer: Fixed

packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
yesamer: @jomarko Only DMN files are directly imported. In case of RULE-based Test Scenario, we don't import the DRL file, but the Java Classes present in the user project.

yesamer:
@jomarko Thank you for testing the Extension. Indeed the issue reproduces to me as well, but with new scesim files only. If you open an existing *.scesim file, it works correctly
Screenshot 2024-10-22 at 15 11 09

I'll investigate more this behavior when working on https://github.com/apache/incubator-kie-issues/issues/1552

jomarko (approved):
Thank you for updates @yesamer

tiagobento (approved):

PR #2708, opened 2024-10-24
NO-ISSUE: Bump express to ^4.21.1

jomarko:


jomarko:
Already done as #2707

PR #2711, opened 2024-10-27
kie-issues#2601: [kn-plugin-workflow] Minify the openAPI spec files to trim operations only used by the workflows in the current project

### LINKED ISSUE #2601: [kn-plugin-workflow] Minify the openAPI spec files to trim operations only used by the workflows in the current project
When users run the command kn workflow create, they can add any OpenAPI specification files to the specs folder. However, large files may exceed Kubernetes' [object size limit](https://github.com/kubernetes/kubernetes/blob/db1990f48b92d603f469c1c89e2ad36da1b74846/test/integration/master/synthetic_master_test.go#L315) during deployment. Reference: [Kubernetes object size limitations](https://stackoverflow.com/questions/60468110/kubernetes-object-size-limitations).

To address this, we will introduce a new CLI command:

```shell
# Explore additional handlers if needed
kn workflow specs minify openapi
```

## Command Behavior:
- This command minifies OpenAPI spec files in the `specs` folder, keeping only the operations and definitions used in the workflows (and sub flows) within the project context.
-  Minified files will be saved with the pattern `<original-file-name>.min.yaml|json`.
- All unused operations, types, and objects will be removed from the minified file.

## Deployment:
- The `ConfigMaps` will reference the minified files when deploying to Kubernetes. Although the `ConfigMaps` entries will use the original file names (to match references in `spec.flow.functions`), the content will be the minified version.

## Automation:
- If the user does not run the minify command manually, the process will be triggered automatically before creating `ConfigMaps` during deployment. This ensures that oversized files do not block deployment.

## File Size Validation
- The minified file must be validated to ensure it does not exceed Kubernetes' 3MB size limit. If it does, users will be prompted to store the OpenAPI files remotely (e.g., via CDN or a Git repository) and reference them using HTTP(S) URLs in the project.

## Debugging:
- The command allows users to manually minify and inspect the OpenAPI specs, offering better debugging and verification before deployment.

## Documentation:
- Ensure all relevant documentation is updated to reflect this new behavior.

## Additional Considerations:
**Editor Integration**: @handreyrc, please verify if this change impacts the editor's functionality. IntelliSense should remain unaffected, as it will continue working with the original, non-minified OpenAPI specs.

treblereel:
Fix https://github.com/apache/incubator-kie-tools/issues/2601

tests are missed atm, I'll add it in a couple of days


packages/kn-plugin-workflow/pkg/common/openapi_minifier.go
```diff
@@ -0,0 +1,278 @@
...
+
+	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/metadata"
+	"github.com/apache/incubator-kie-tools/packages/sonataflow-operator/api/v1alpha08"
+	"github.com/getkin/kin-openapi/openapi3"
+	"gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/util/sets"
+	yamlk8s "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+type OpenApiMinifier struct {
```
ricardozanini: Can you move this to the `pkg/specs` module?

packages/kn-plugin-workflow/pkg/command/deploy_undeploy_common.go
```diff
@@ -122,14 +122,14 @@ func generateManifests(cfg *DeployUndeployCmdConfig) error {
 	supportFileExtensions := []string{metadata.JSONExtension, metadata.YAMLExtension, metadata.YMLExtension}
 
 	fmt.Println("🔍 Looking for specs files...")
-	files, err = common.FindFilesWithExtensions(cfg.SpecsDir, supportFileExtensions)
+	minifiedfiles, err := common.NewMinifier(&common.OpenApiMinifierParams{
```
ricardozanini: Then if you accept my other suggestion we can have something like this:

```suggestion
	minifiedfiles, err := specs.NewMinifier(&common.OpenApiMinifierParams{
```

packages/kn-plugin-workflow/pkg/common/openapi_minifier.go
```diff
@@ -0,0 +1,278 @@
...
+	workflows  []string
+	params     *OpenApiMinifierParams
+	operations map[string]sets.Set[string]
+}
+
+type OpenApiMinifierParams struct {
```
ricardozanini:
```suggestion
type OpenApiMinifierOpts struct {
```

packages/kn-plugin-workflow/pkg/common/openapi_minifier.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	doc, err := openapi3.NewLoader().LoadFromData(data)
+	if err != nil {
+		return "", fmt.Errorf("❌ ERROR: failed to load OpenAPI document: %w", err)
+	}
+	if doc.Paths == nil {
+		return "", fmt.Errorf("OpenAPI document %s has no paths", specFileName)
+	}
+	for key, value := range doc.Paths.Map() {
```
ricardozanini: So we are removing `paths`, but we keep `components`? https://spec.openapis.org/oas/latest.html#components-object

Some components that were only referenced by these paths might also be removed.

ricardozanini (commented):
You can take a few openapi specs as an example for tests from here:
https://github.com/quarkiverse/quarkus-openapi-generator/tree/main/client/integration-tests

Things that should be included in the test cases:
- Full openapi spec
- Minified spec
- Workflow referencing the spec
- Adding another openapi operation to the workflow and recheck if the new operation has been added to the minified version.

packages/kn-plugin-workflow/pkg/common/openapi_minifier.go
treblereel: @ricardozanini https://github.com/apache/incubator-kie-tools/issues/2718

treblereel:
@ricardozanini I have added `--minify` option to the deploy, it allows us to skip minification if needed
By default it's true

packages/kn-plugin-workflow/pkg/specs/openapi_minifier.go
```diff
@@ -0,0 +1,282 @@
...
+
+	minifySpecsFiles, err := m.minifySpecsFiles()
+	if err != nil {
+		return nil, err
+	}
+
+	return minifySpecsFiles, nil
+}
+
+func (m *OpenApiMinifier) processFunctions() error {
```
ricardozanini:
```suggestion
func (m *OpenApiMinifier) fetchSpecFromFunctions() error {
```

packages/kn-plugin-workflow/pkg/specs/openapi_minifier.go
```diff
@@ -0,0 +1,282 @@
...
+	for _, workflowFile := range m.workflows {
+		err := m.processFunction(workflowFile)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (m *OpenApiMinifier) processFunction(workflowFile string) error {
```
ricardozanini:
```suggestion
func (m *OpenApiMinifier) fetchSpecFromFunction(workflowFile string) error {
```

ricardozanini: I believe this name gives a stronger meaning to what this func does.

ricardozanini (commented):
Well done! I left two minor suggestions.

ricardozanini:
@domhanak can you please verify this?

ricardozanini (approved):

domhanak:
> @domhanak can you please verify this?

Yup, working on it, investigating some issue with compilation after the minifications, Will let you know.

domhanak:
# Documentation of the command

- `kn workflow specs help`  - :heavy_check_mark:  
- `kn workflow specs minify` -> this does nothing, and when I do `kn workflow specs minify help` the result is the same. I think we can better explain the command to the user, in a sense he understands `openapi` needs to be added as argument to start the minification. For example adding more targeted HELP description here :construction: 


# Functional

- `kn workflow specs minify openapi`
  - [Input](https://gist.github.com/domhanak/dc00ed7547fa92f70bd9e000422c4914) results in [output](https://gist.github.com/domhanak/60d4df3c940f7a36d21f5c5afcf0c392) -> Is that correct? I can see a ConfigMap `01-hello-resources-specs.yaml` when deploying but the workflow is failing with compilation errors. 
:eyes: investigating why :eyes:
Cause: Typo in function definition of the workflow, once fixed workflow properly deploys without errors. :heavy_check_mark: 
   - `components` are part of the minified file, is that expected? :question: 
   - file size seems to be larger after minification, according to the message `:white_check_mark: Minified file specs/greetingAPI.min.yaml created with 7572 bytes (original size: 6212 bytes)`


More information about testing to follow, I will update the comment with latest info.
Double checking the scenarios with subflows.

domhanak (approved):

PR #2714, opened 2024-10-29
NO-ISSUE: Remove `@kie-tools/backend`

tiagobento:
This package was introduced when we were still figuring out the dynamics of Multiplying Architecture, and doesn't make sense anymore, since we always want to decouple micro-front-ends from backends, delegating the communication with backends to the hosting application as much as possible.

The API introduced by this package was not used except from one place on DashBuilder VS Code Extension, which was very easily adapted to work without it.


porcelli (approved):
thank you for the cleanup @tiagobento !!!

ljmotta (approved):

ljmotta:
@tiagobento Thanks! The PR looks good. Indeed, it looks like was very easy to adapt the funcionality.

PR #2717, opened 2024-10-30
NO-ISSUE: Upgrade `path-to-regexp@1.8.0` to `1.9.0`

tiagobento:
The strategy here was to manually remove the `path-to-regexp@.1.8.0` entry from the lockfile and run `pnpm install --fix-lockfile`.


pnpm-lock.yaml
```diff
@@ -14371,196 +14286,168 @@ packages:
   '@babel/plugin-proposal-async-generator-functions@7.16.8':
     resolution: {integrity: sha512-71YHIvMuiuqWJQkebWJtdhQTfd4Q4mF76q2IX37uZPkG9+olBxsX+rH1vkhFto4UeJZ9dPY2s+mDvhDm1u2BGQ==}
     engines: {node: '>=6.9.0'}
-    deprecated: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-async-generator-functions instead.
```
thiagoelg, line 14374: I never knew we had these `deprecated` notices here... and so many of them!
Maybe we should keep them and plan to tackle them soon.

I'm assuming most of these deprecated packages are transitive dependencies, but some of them (like [q](https://www.npmjs.com/package/q)) are really old!

thiagoelg (approved):
Just a comment about the `deprecated` notices, but other than that this looks good to me!

Good to see that the new logic for checking changes to the lockfile is working, even though it didn't help that much in this case.

pnpm-lock.yaml
tiagobento: Yeah.... I think dependabot added those when it updated express, I'm guessing those are from a newer version of pnpm? Need to investigate.

porcelli (approved):

Issue #2718, opened 2024-10-31
[kn-plugin-workflow] Minify the openAPI spec files to trim components/schemas


treblereel:
This issue is a continuation of the work on minifying OpenAPI specifications files.

The goal is to remove all unused components and schemas from the OpenAPI spec file.

Note: we don't support external refs in this issue

treblereel:
https://github.com/apache/incubator-kie-tools/issues/2601

PR #2749, opened 2024-11-18
kie-issues#2718: [kn-plugin-workflow] Minify the openAPI spec files to trim components/schemas

treblereel:
issue: https://github.com/apache/incubator-kie-tools/issues/2718

### DIFF
--- a/packages/kn-plugin-workflow/pkg/specs/openapi_minifier.go
+++ b/packages/kn-plugin-workflow/pkg/specs/openapi_minifier.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/common"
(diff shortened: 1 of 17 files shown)


treblereel:
Path declarations may contain $ref references, which in turn may refer to other declarations. Therefore, before removing them, it is necessary to identify all declarations that the desired operationId depends on. A special case is securitySchemes; if they are present in the document, they should be retained since the reference to them occurs implicitly.

packages/kn-plugin-workflow/pkg/specs/openapi_minifier.go
```diff
@@ -194,6 +182,66 @@ func (m *OpenApiMinifier) minifySpecsFile(specFileName string, operations sets.S
...
+	keep, err := collector.collect(operations)
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("❌ ERROR: failed to load OpenAPI document: %w", err)
+	}
+	if doc.Paths == nil {
+		return nil, fmt.Errorf("OpenAPI document %s has no paths", specFileName)
```
ricardozanini, line 202: Do we want to return an error here? Maybe just log silently to not spur the user's console with an error. If this is handled by the caller, just ignore my comment.

ricardozanini (approved):
Great work, man!

domhanak (approved):
Looks good to me, the file is minimized:
```dhanak@dhanak-thinkpadp1gen3 subflows]$ ./kn-workflow-linux-amd64 specs minify openapi
✅ Minified file /home/dhanak/Downloads/subflows/specs/greetingAPI.min.yaml created with 514 bytes (original size: 6248 bytes)
```

Components and schemas are trimmed as expected.

PR #2760, opened 2024-11-22
kie-issues#208: Renaming any "NamedElement" on the DMN Editor should update all references to the old name

### LINKED ISSUE #208: Renaming any "NamedElement" on the DMN Editor should update all references to the old name
When the user renames any "NamedElement" like, for instance, a DataType, an entry in a Context Expression, or a Decision Node, all places where referencing that element should be updated to the new name.

"All places" mean "all FEEL expressions".

**Confirmation Dialog** 
As discussed in the private chat, the rename could result in a refactoring or not, because it is not always that user may want to perform a refactor.

The user may not want to perform a refactor; they can just rename the named object to something else and create a new object with the older name to replace it.

So, for this scenario, we should ask for confirmation from the user, with the following dialog:

image

**The confirmation dialog should behave as the following:**

1. When renaming some node, it shows the dialog if the node is being referenced in an expression
2. When renaming some data type, it shows the dialog if the data type is being referenced in an expression
3. When renaming an included model (in the Included Models tab), **it does not show the dialog**, because doesn't make any sense in my opinion to not update the references in this case, since this already was the default behavior before this PR (when we renamed a included model, we already updated all the references except the expressions, now we update the expressions).

**Acceptance criteria:**

All the following renames should also perform a refactor if the user wants to by confirming in the confirmation dialog:

**_In DMN editor:_**
- [x] Rename the input nodes
- [x] Rename the nodes by the properties panel
- [x] Rename business knowledge nodes
- [x] Rename business knowledge nodes by properties panel
- [x] Rename decision nodes
- [x] Rename decision by properties panel
- [x] Rename data types

**_In Boxed Expression Editor_**
- [x] Rename context 
- [x] Rename the "some" variable in the "Some" expression
- [x] Rename the "every" variable in the "Every" expression
- [x] Rename the header of expressions
- [x] Rename the header of expressions by properties panel

The following refactor should be performed without asking confirmation:
- [x] Rename an included model

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/208

**Some comments about the changes:**
- We are tracking now all changes performed in BEE reported to DMN Editor. For now, we only have "special treatment" for the  `Action.VariableChanged`;
- A big part of the changes are related to this. Don't panic. It is just an indentation level that can make the PR bigger than it really is; 
- In the lower levels of the code, I refactored from "variable" to "identifier", because those objects are not only variables but identifiers;
- I added documentation to a few methods and classes.

**Renaming Data Types**

https://github.com/user-attachments/assets/865b600d-d15f-44c6-b870-f7e2172175b4

**Renaming BKMs and parameters**

https://github.com/user-attachments/assets/fb7bf2b6-58f9-4db4-ace6-a5d1e8e2ef0e

**General rename**

https://github.com/user-attachments/assets/e4485c5a-2285-4bac-9d1c-218edbe052a1

**Renaming Included Models name**
It does not show the confirmation dialog, because doesn't make sense to not update the references in this case, since this already was the default behavior before this PR (when we renamed a included model, we already updated all the references except the expressions, now we update the expressions).

https://github.com/user-attachments/assets/9b44114d-e302-4a09-bd33-81fce5254736


packages/boxed-expression-component/src/expressionVariable/ExpressionVariableMenu.tsx
```diff
@@ -100,17 +108,36 @@ export function ExpressionVariableMenu({
...
-  }, [expressionName, selectedExpressionName, dataType, selectedDataType, onVariableUpdated]);
+  }, [expressionName, selectedExpressionName, dataType, selectedDataType, variableUuid, onVariableUpdated]);
 
   const resetFormData = useCallback(() => {
     setExpressionName(selectedExpressionName);
     setDataType(selectedDataType);
   }, [selectedExpressionName, selectedDataType]);
 
-  // onCancel doens't prevent the onHide call
+  // onCancel doesn't prevent the onHide call
```
danielzhe, line 141: Just a typo.

tiagobento:
Thank you for the PR @danielzhe. It's going to take me a few days to go over everything with the care it deserves. Will keep you posted!

tiagobento:
@danielzhe Storybook is pointing to code using `../src` because it is used during development. The changes you introduced will make the development experience much worse, requiring us to rebuild stuff after a change, if I'm not mistaken... can you please check?

jomarko:
### 01
image
Using the feature, came to my mind, maybe we could have a third action button 'Cancel', like do nothing? During my testing I was in situation oh, I want to double-check what I will do, so I just wanted to cancel the operation.

jomarko:
### 02
The feature seems to not work in decision tables.
image
In the attached table there is input node `IDAnyStruct`. This node has a data type set to a custom structure with three fields: `A`, `B` and `C`. Once I renamed `A`, `B` or `C` on the `Data Types` tab, the occurrences in the decision table were not updated. 

I think that at least the row 2 should be updated. I slightly think also decision table header row should be, however that may be very tricky and if we decide to implement, maybe in a separate ticket? I am not sure about the second point.

The same seems to be reproducible if the user is renaming for example the `some` identifier while `in` or `satisfies` expression is a decision table and has an occurrence of the identifier `some`.

jomarko:
### 03
FYI, I didn't check this feature with the List expression because of https://github.com/apache/incubator-kie-issues/issues/1635

jomarko:
### 04
Question, should this feature work for BKM node and Invocation expression. Imagine this model:
Screenshot 2024-11-26 104246
Where internally the `Invocation` says it invokes `mybkm` node.

However if user renames the `mybkm` to `mybkm renamed` the `Invocation` remains the same (notice `mybkm` in last screenshot):
Screenshot 2024-11-26 104416
Screenshot 2024-11-26 104422

jomarko:
### 05
We have an issue editor allows to define two function parameters with the same name, then renaming does  not work.
Screenshot 2024-11-26 105859
Screenshot 2024-11-26 105912

jomarko:
### 06
Is there a way to disable this feature? Committing the rename action starts to be slower, so I expect some user may ask to disable this feature in their environment.

jomarko:
### 07
During included models review, I found this https://github.com/apache/incubator-kie-issues/issues/1636, however I think I didn't find issue related to changes of #2760.

danielzhe:
Thank you for your review, @jomarko ! 
About the points you commented:
1. I made a new screen with a X in the top corner, where we can cancel the change, like discussed in the private chat. Here is the print:
image
2. Change applied. Decision tables should be working now
3. ok
4. That field is not "expression aware", it is just a text box. We have two tickets to address that, so it is not covered in this:
a. https://github.com/apache/incubator-kie-tools/issues/3757
b. https://github.com/apache/incubator-kie-tools/issues/3772
5. As discussed in private chat, that is something that depends on user. We can't (shouldn't)  not allow the user to rename it if the name is duplicated. It is up to the user. Once the parameters have duplicated names, there is no way for us to know which one is which inside the expression.
6. No, we don't have. We had a discussion about this and we come to idea to add the confirmation dialog instead an option to disable/enable it, remember? I didn't found out any slow down using the feature. If you can provide some reproducible steps, we can check it.
7. ok

jomarko:
@danielzhe thank you for updates

I think 1. and 2. works now pretty well.

6. I used this file. Once I rename something there is ~1 second until new modal appears. Once I click rename in that modal, it also takes ~1. I am not saying it is a blocker, I understand that it may be complicated (long time) operation to compute indexes, find identifiers etc. Just want to avoid situation in past in 7.x or 8.x stream, where users working with large models (100+ nodes) were unable even open the model. That is why I asked about possibility to turn off the feature. However, my issue may be local, caused my by laptop, memory parameters and settings. I do not know.

[Untitled (2).dmn.txt](https://github.com/user-attachments/files/17956578/Untitled.2.dmn.txt)

packages/dmn-editor/src/refactor/RefactorConfirmationDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+        </Button>,
+      ]}
+    >
+      The identifier `{fromName ?? "<undefined>"}` was renamed to `{toName ?? "<undefined>"}`.
+      <br />
+      <br />
+      This identifier is used in one or more expressions.
+      <br />
+      <br />
+      Do you want also automatically update the expressions to the new name?
```
klbowers:
```suggestion
      Would you like to automatically replace all instances of `{fromName ?? "<undefined>"}` with `{toName ?? "<undefined>"}`?
```
Or if you prefer the original:
      Do you want _to_ also automatically update the expressions to the new name?

klbowers (commented):
Hi Daniel,
Thanks for the PR - it looks great! Just a minor comment from my side regarding the message:

danielzhe:
> Hi Daniel, Thanks for the PR - it looks great! Just a minor comment from my side regarding the message:

Thank you for your suggestion! It's always good to have a native English speaker to improve our messages. I'll update the text of the buttons too to match the new message.

packages/dmn-editor/stories/misc/empty/Empty.stories.tsx
```diff
@@ -23,7 +23,7 @@ import { ns as dmn15ns } from "@kie-tools/dmn-marshaller/dist/schemas/dmn-1_5/ts
 import { generateUuid } from "@kie-tools/boxed-expression-component/dist/api";
 import { DMN15_SPEC } from "@kie-tools/dmn-marshaller/dist/schemas/dmn-1_5/Dmn15Spec";
 import { DmnEditorWrapper, StorybookDmnEditorProps } from "../../dmnEditorStoriesWrapper";
-import { DmnEditor, DmnEditorProps } from "../../../src/DmnEditor";
+import { DmnEditor, DmnEditorProps } from "@kie-tools/dmn-editor/dist/DmnEditor";
```
ljmotta: Nice!

packages/dmn-language-service/src/refactor/IdentifiersRefactor.ts
ljmotta, line 1: It's not necessary to create a `refactor` folder for this file as the name already tells what the file is about.

packages/dmn-feel-antlr4-parser/src/Uuid.ts
```diff
@@ -0,0 +1,29 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { v4 as uuid } from "uuid";
+
+export {};
```
ljmotta: Why this is necessary?

packages/boxed-expression-component/src/expressionVariable/ExpressionVariableCell.tsx
```diff
@@ -50,21 +62,44 @@ export const ExpressionVariableCell: React.FunctionComponent<
...
-              ...expression,
-              "@_label": name,
-              "@_typeRef": typeRef,
-            }
-          : undefined!, // SPEC DISCREPANCY
-        variable: {
-          ...variable,
-          "@_name": name,
-          "@_typeRef": typeRef,
+      const variableChangedArgs: ExpressionChangedArgs = {
```
ljmotta: On other files is used `expressionChangedArgs`.

packages/boxed-expression-component/src/expressionVariable/ExpressionVariableMenu.tsx
```diff
@@ -100,17 +108,36 @@ export function ExpressionVariableMenu({
   }, [beeGwtService]);
 
   const saveExpression = useCallback(() => {
-    if (expressionName !== selectedExpressionName || dataType !== selectedDataType) {
-      onVariableUpdated({ name: expressionName, typeRef: dataType });
+    const changes: ExpressionChangedArgs = {
```
ljmotta: On other files is used `expressionChangedArgs`.

ljmotta (commented):
@danielzhe The code looks good. I still didn't finish reviewing all files and test the new feature, but I could see a few things. For the DMN Editor, I could see that many files have the same logic, so why not create a new hook that encapsulate the `refactor` logic? For instance, the refactor modal and newName states and the `applyRename` could be returned by this new hook. 

```
  const [isRefactorModalOpen, setIsRefactorModalOpen] = useState(false);
  const [newName, setNewName] = useState("");

// Not sure if this could be add to the hook or not
  const identifierId = useMemo(() => decision["@_id"], [decision]);
  const oldName = useMemo(() => decision["@_label"] ?? decision["@_name"], [decision]);
  const currentName = useMemo(() => {
    return newName === "" ? oldName : newName;
  }, [newName, oldName]);
// End

  const applyRename = useCallback(
    (args: {
      definitions: Normalized<DMN15__tDefinitions>;
      newName: string;
      shouldRenameReferencedExpressions: boolean;
    }) => {
      renameDrgElement({
        ...args,
        index,
        externalDmnModelsByNamespaceMap,
      });
    },
    [externalDmnModelsByNamespaceMap, index]
  );
```

Another thing that caught my attention was the `externalDmnModelsByNamespaceMap` creation in many places. Maybe this is the case to add it to the `store`?


Additionally, I've made some comments inline.

packages/dmn-language-service/src/refactor/IdentifiersRefactor.ts
danielzhe: The problem is that outside, where the code is called, it would be only `Identifiers` and I think `IdentifiersRefactor` is a better name because only `Identifiers` can lead to think that there are more functions inside of it and not only functions related to Identifiers Refactor. Wdyt?

packages/dmn-editor/stories/misc/empty/Empty.stories.tsx
tiagobento: Actually I think we need to revert that. Storybook is a development webapp, so it's supposed to work even without having to build the package itself. By pointing to the `src` folder directly we make sure it works like that and also that it live-reloads when you change the contents of the file.

packages/dmn-feel-antlr4-parser/src/Uuid.ts
danielzhe: 🤔 
It is a mystery. 
Removed.

packages/dmn-language-service/src/refactor/IdentifiersRefactor.ts
ljmotta: I agree and the file name is perfect. My point is the `refactor` folder. After removing the `refactor` folder, we would have `src/IdentifiersRefactor.ts`

tiagobento:
Conflicts :(

tiagobento (approved):
Amazing PR @danielzhe, thank you!

Minor feedback for a future change on this area would be to make the `useRefactor` hook a little bit less generic. `Refactor` is a general term that is associated with "making a change that will impact other places without changing the way things work", and although this is true for the renaming of a FEEL identifier, this hook could've been called `useFeelIdentifierNameRefactor` or something like that. Maybe when we add more refactors we could use the opportunity to rename it. Thank you!

ljmotta (approved):

ljmotta (approved):
@danielzhe Thanks for your patience. The last refactor made the code cleaner, approving!

PR #2763, opened 2024-11-25
kie-tools#2630: Extend kn-workflow CLI to Generate K8s secrets for workflows

### LINKED ISSUE #2630: Extend kn-workflow CLI to Generate K8s secrets for workflows
The current functionality of the kn-workflow CLI allows for the generation of Kubernetes resources, specifically SonataFlow CRs and a series of ConfigMap CRs, based on the application.properties file, schemas and spec folders. However, a common use case has emerged where users need to patch the generated SonataFlow CR to mount a **secret** into the pod, making environment variables available to the workflow's application.

To streamline this process, we propose extending the kn-workflow plugin with the ability to Generate Kubernetes Secret from **secret.properties files,** while ensuring proper resource ordering. We will also update the plugin documentation and will provide examples.

treblereel:
Closes: https://github.com/apache/incubator-kie-tools/issues/2630

### DIFF
--- a/packages/sonataflow-operator/workflowproj/workflowproj.go
+++ b/packages/sonataflow-operator/workflowproj/workflowproj.go
@@ -23,8 +23,11 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"regexp"
 	"sort"
 	"strings"
+
+	"github.com/magiconair/properties"
 
 	"github.com/pkg/errors"
 	"github.com/serverlessworkflow/sdk-go/v2/model"
@@ -56,6 +59,8 @@ type WorkflowProjectHandler interface {
 	WithWorkflow(reader io.Reader) WorkflowProjectHandler
 	// WithAppProperties reader for a file or the content stream of a workflow application properties.
 	WithAppProperties(reader io.Reader) WorkflowProjectHandler
+	// WithSecretProperties reader for a file or the content stream of a workflow application properties.
+	WithSecretProperties(reader io.Reader) WorkflowProjectHandler
 	// AddResource reader for a file or the content stream of any resource needed by the workflow. E.g. an OpenAPI specification file.
 	// Name is required, should match the workflow function definition.
 	AddResource(name string, reader io.Reader) WorkflowProjectHandler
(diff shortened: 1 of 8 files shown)


ricardozanini (approved):

ricardozanini:
@wmedvede @domhanak can you please take a look?

packages/sonataflow-operator/workflowproj/workflowproj.go
```diff
@@ -338,3 +406,18 @@ func isProfile(workflow *operatorapi.SonataFlow, profileType metadata.ProfileTyp
...
+func normalizeEnvName(name string) (string, error) {
+	name = strings.TrimSpace(name)
+
+	replacer := strings.NewReplacer(" ", "_", "-", "_", ".", "_")
+	name = replacer.Replace(name)
+	name = strings.ToUpper(name)
+
+	validName := regexp.MustCompile(`^[A-Z0-9_]+$`)
+	if !validName.MatchString(name) || name == "_" {
+		return "", fmt.Errorf("invalid environment variable name: %s (must only contain A-Z, 0-9, and _)", name)
```
wmedvede: I think it could be helpful for future error tracing, to print which was the original property name, and which was the resulting normalized env var name.

Also I believe there are property name cases not covered by the algorithm, example:

quarkus."myProperty".value 

should be normalized into:

QUARKUS__MYPROPERTY__VALUE

Double check please, I think it's a potential valid use cases with quarkus properties.

see: https://quarkus.io/guides/config-reference#environment-variables

wmedvede (changes requested):

packages/sonataflow-operator/workflowproj/workflowproj.go
treblereel: Thank you for the review. I missed the fact that application.properties can contain Quarkus-specific formats. I’ve fixed(and cases like  `%dev, prod.some.property=value`). By the way, we might consider adding support for application.yaml in the future.

wmedvede (approved):
LGTM, thank you!

ricardozanini:
Waiting for @domhanak review

domhanak (approved):
LGTM

ricardozanini:
@treblereel can you resolve the conflicts so I can merge?

treblereel:
@ricardozanini done, let's wait for the tests and merge. Once it is done, I will rebase the next one

PR #2788, opened 2024-12-04
NO-ISSUE: Update NOTICE Copyright year automatically on 1th of January

jomarko:
The added action should open a pull request on 1th of January and update the NOTICE Copyright year to be up to date.


tiagobento:
All PRs will become red on January 1st if we merge this, won't they? 🙈

jomarko:
@tiagobento thank you for the feedback, I changed the solution. It should open PR with needed update don the first of january. Please let me know your opinion. Or please propose what should be the solution.

I will then adapt PRs for other repositories.

jomarko:
The team conclusion is we will not merge this:

The use a 3rd party GitHub Action  is not allowed by Apache. Also automation of simple update, that is needed once a year is not worth, probably.

PR #2790, opened 2024-12-04
kie-issues#2789: [sonataflow-operator] Implement DB Migrator Changes into SonataFlow Operator

### LINKED ISSUE #2789: [sonataflow-operator] Implement DB Migrator Changes into SonataFlow Operator
This refers to migrating DB Migrator changes into SonataFlow operator from the old repo `https://github.com/apache/incubator-kie-kogito-serverless-operator` to the `https://github.com/apache/incubator-kie-tools` repo.

The old PR in the old repo can found [here](https://github.com/apache/incubator-kie-kogito-serverless-operator/pull/544)

rhkp:
Closes #2789 
Implement DB Migrator Changes into SonataFlow Operator.


packages/sonataflow-operator/api/v1alpha08/sonataflowplatform_services_types.go
```diff
@@ -23,6 +23,11 @@ import (
 
 // ServicesPlatformSpec describes the desired service configuration for workflows without the `sonataflow.org/profile: dev` annotation.
 type ServicesPlatformSpec struct {
+	// true = Use DB Migration Job with DB Migrator tool image
+	// false = Use built-in DB migration capability within services e.g. DI/JS
+	// +optional
+	// +default: false
+	JobBasedDbMigration bool `json:"jobBasedDbMigration,omitempty"`
```
ricardozanini: Why do we have a `bool` value here, whereas the type is an `enum`? Is this attribute to propagate `DBMigrationStrategyType`? In this case, we should use the same field.

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
```diff
@@ -0,0 +1,279 @@
+/*
```
ricardozanini: The file names do not use `camelCase`. Please rename it to `db_migrator_job.go` and all other files.

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
```diff
@@ -0,0 +1,279 @@
...
+	MigrateDBDataIndex    bool
+	DataIndexDataSource   *QuarkusDataSource
+	MigrateDBJobsService  bool
+	JobsServiceDataSource *QuarkusDataSource
+}
+
+const (
+	dbMigrationJobName       = "sonataflow-db-migrator-job"
+	dbMigrationContainerName = "db-migration-container"
+	dbMigratorToolImage      = "quay.io/rhkp/incubator-kie-kogito-service-db-migration-postgresql:latest"
```
ricardozanini: This image must come from the image package. So can you bring that here? Or we can merge the image first and then you review this const. I'm about to change this versioning/image naming in the operator package to read this info from ENV/Configuration.

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
```diff
@@ -0,0 +1,279 @@
...
+				}
+			} else {
+				return a
+			}
+		}
+	}
+	return defaultSchemaName
+}
+
+func getNewQuarkusDataSource(jdbcURL string, userName string, password string, schema string) *QuarkusDataSource {
```
ricardozanini: These parameters are always passed in blank. You can simplify this by removing this function and just instantiate the struct instead: 

```go
quarkusDataSource :=  &QuarkusDataSource{}
```

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
```diff
@@ -0,0 +1,279 @@
...
+
+func getQuarkusDataSourceFromPersistence(ctx context.Context, platform *operatorapi.SonataFlowPlatform, persistence *operatorapi.PersistenceOptionsSpec, defaultSchemaName string) *QuarkusDataSource {
+	quarkusDataSource := getNewQuarkusDataSource("", "", "", "")
+	if persistence != nil && persistence.PostgreSQL != nil {
+		quarkusDataSource.JdbcUrl = persistence.PostgreSQL.JdbcUrl
+		quarkusDataSource.Username, _ = services.GetSecretKeyValueString(ctx, persistence.PostgreSQL.SecretRef.Name, persistence.PostgreSQL.SecretRef.UserKey, platform.Namespace)
+		quarkusDataSource.Password, _ = services.GetSecretKeyValueString(ctx, persistence.PostgreSQL.SecretRef.Name, persistence.PostgreSQL.SecretRef.PasswordKey, platform.Namespace)
+		quarkusDataSource.Schema = getDBSchemaName(persistence.PostgreSQL, defaultSchemaName)
+	}
+	return quarkusDataSource
```
ricardozanini: Instead, you should return nil to make it easier for client code to handle this structure. It's easier to just check if a variable is nil.

```go 
func getQuarkusDataSourceFromPersistence(ctx context.Context, platform *operatorapi.SonataFlowPlatform, persistence *operatorapi.PersistenceOptionsSpec, defaultSchemaName string) *QuarkusDataSource {
	if persistence != nil && persistence.PostgreSQL != nil {
                quarkusDataSource := &QuarkusDataSource{}
		quarkusDataSource.JdbcUrl = persistence.PostgreSQL.JdbcUrl
		quarkusDataSource.Username, _ = services.GetSecretKeyValueString(ctx, persistence.PostgreSQL.SecretRef.Name, persistence.PostgreSQL.SecretRef.UserKey, platform.Namespace)
		quarkusDataSource.Password, _ = services.GetSecretKeyValueString(ctx, persistence.PostgreSQL.SecretRef.Name, persistence.PostgreSQL.SecretRef.PasswordKey, platform.Namespace)
		quarkusDataSource.Schema = getDBSchemaName(persistence.PostgreSQL, defaultSchemaName)
                return quarkusDataSource
	}
	return nil
}
```

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
```diff
@@ -0,0 +1,279 @@
...
+	if pshDI.IsPersistenceEnabledtInSpec() {
+		diJobsBasedDBMigration = services.IsJobsBasedDBMigration(platform.Spec.Services.DataIndex.Persistence)
+	}
+	if pshJS.IsPersistenceEnabledtInSpec() {
+		jsJobsBasedDBMigration = services.IsJobsBasedDBMigration(platform.Spec.Services.JobService.Persistence)
+	}
+
+	if (pshDI.IsServiceSetInSpec() && diJobsBasedDBMigration) || (pshJS.IsServiceSetInSpec() && jsJobsBasedDBMigration) {
+		quarkusDataSourceDataIndex := getNewQuarkusDataSource("", "", "", "")
+		quarkusDataSourceJobService := getNewQuarkusDataSource("", "", "", "")
```
ricardozanini: Same comment here regarding the usage of this function.

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
```diff
@@ -0,0 +1,279 @@
...
+	platform.Status.SonataFlowPlatformDBMigrationPhase = NewSonataFlowPlatformDBMigrationPhase(operatorapi.DBMigrationStatusStarted, operatorapi.MessageDBMigrationStatusStarted, operatorapi.ReasonDBMigrationStatusStarted)
+
+	dbMigratorJobStatus, err := dbmj.GetDBMigrationJobStatus(ctx, client, platform)
+	if err != nil {
+		return nil, err
+	}
+
+	klog.V(log.I).InfoS("Db migration job status: ", "active", dbMigratorJobStatus.Active, "ready", dbMigratorJobStatus.Ready, "failed", dbMigratorJobStatus.Failed, "success", dbMigratorJobStatus.Succeeded, "CompletedIndexes", dbMigratorJobStatus.CompletedIndexes, "terminatedPods", dbMigratorJobStatus.UncountedTerminatedPods)
+
+	if dbMigratorJobStatus.Failed == 1 {
```
ricardozanini: Can we use `bool` instead?

packages/sonataflow-operator/internal/controller/platform/services/secrets.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/apache/incubator-kie-tools/packages/sonataflow-operator/log"
+	"github.com/apache/incubator-kie-tools/packages/sonataflow-operator/utils"
+)
+
+func GetSecretKeyValueString(ctx context.Context, secretName string, secretKey string, nameSpace string) (string, error) {
+	secret := corev1.Secret{}
+	err := utils.GetClient().Get(ctx, ctrl.ObjectKey{Namespace: nameSpace, Name: secretName}, &secret)
+	if err != nil {
+		panic(err.Error())
```
ricardozanini: Please don't do this, or you will break the operator run.

packages/sonataflow-operator/internal/controller/platform/services/services.go
```diff
@@ -233,18 +249,51 @@ func (d *DataIndexHandler) hasPostgreSQLConfigured() bool {
...
+	dbMigrationStrategy := GetDBMigrationStrategy(persistence)
+	return dbMigrationStrategy == operatorapi.DBMigrationStrategyJob
+}
+
+func IsNoDBMigration(persistence *operatorapi.PersistenceOptionsSpec) bool {
+	dbMigrationStrategy := GetDBMigrationStrategy(persistence)
+	return dbMigrationStrategy == operatorapi.DBMigrationStrategyNone || dbMigrationStrategy == ""
+}
+
+func isDBMigrationStrategyService(persistence *v1alpha08.PersistenceOptionsSpec) string {
```
ricardozanini, line 277: Why do we need this? Sound confusing since we already have the `IsServiceBasedDBMigration`.

packages/sonataflow-operator/test/e2e/testdata/platform/persistence/service_based_db_migration/02-sonataflow_platform.yaml
```diff
@@ -0,0 +1,66 @@
...
+kind: SonataFlowPlatform
+metadata:
+  name: sonataflow-platform
+spec:
+  build:
+    config:
+      strategyOptions:
+        KanikoBuildCacheEnabled: "true"
+  services:
+    # jobBasedDbMigration: false
```
ricardozanini: Leftovers?

ricardozanini (changes requested):

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
rhkp: Updated to `"docker.io/apache/incubator-kie-kogito-db-migrator-tool:latest"` as named in DB Migrator Image package [here](https://github.com/rhkp/incubator-kie-tools/blob/flpath-1775-pr1/packages/kogito-db-migrator-tool-image/resources/incubator-kie-kogito-db-migrator-tool-image.yaml#L19).

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
rhkp: May be not: as dbMigratorJobStatus is of type *batchv1.JobStatus defined in batch/v1/types.go with int32 data type.

packages/sonataflow-operator/api/v1alpha08/sonataflowplatform_services_types.go
rhkp: I think this line is an invalid and has been removed as we are making use of dbMigrationStrategy and not jobBasedDbMigration.

packages/sonataflow-operator/internal/controller/platform/services/services.go
rhkp: Great point. The `isDBMigrationStrategyService` function returns `string` version of serviceBasedDBMigration boolean flag. The string value is later used as follows in `ConfigurePersistence()` function
`
// specific to DataIndex
		c.Env = append(c.Env, corev1.EnvVar{Name: quarkusHibernateORMDatabaseGeneration, Value: "update"}, corev1.EnvVar{Name: quarkusFlywayMigrateAtStart, Value: dbMigrationStrategyService})
`

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
rhkp: We need this structure in all cases for use CreateJobDBMigration() function and not checking for the structure to be nil.

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
ricardozanini: In this case you can create a const instead of comparing to a Magic Number.

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
ricardozanini: We should do a nil check when calling this function for all cases because persistence can be `nil` by reading your code.

packages/sonataflow-operator/internal/controller/platform/dbMigratorJob.go
rhkp: For now, unresolving this and reverting to old image for sake of seeing how CI behaves.

packages/sonataflow-operator/internal/controller/platform/db_migrator_Job.go
```diff
@@ -0,0 +1,301 @@
+/*
```
ricardozanini: The name of the file has a capitalized `J`.

rhkp: Great catch...

packages/sonataflow-operator/api/v1alpha08/sonataflow_persistence_types.go
```diff
@@ -17,6 +17,14 @@
 
 package v1alpha08
 
+type DBMigrationStrategyType string
+
+const (
+	DBMigrationStrategyService DBMigrationStrategyType = "service"
+	DBMigrationStrategyJob     DBMigrationStrategyType = "job"
```
wmedvede, line 24: @rhkp @ricardozanini wondering if we should call this strategy "operator" instead :thinking: , and we don't "expose" how it is performed.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,302 @@
...
+	MigrateDBDataIndex    bool
+	DataIndexDataSource   *QuarkusDataSource
+	MigrateDBJobsService  bool
+	JobsServiceDataSource *QuarkusDataSource
+}
+
+const (
+	dbMigrationJobName       = "sonataflow-db-migrator-job"
+	dbMigrationContainerName = "db-migration-container"
+	dbMigratorToolImage      = "quay.io/rhkp/incubator-kie-kogito-service-db-migration-postgresql:latest"
```
wmedvede: align with the new name apache/incubator-kie-kogito-db-migrator-tool and be sure that:

1. the tag is derived with a mechanism similar to what we have for the DataIndex and Jobs Serivce
2. we have a corresponding entry in the packages/sonataflow-operator/config/manager/controllers_cfg.yaml, and that we do an analogous processing to what we do for the DataIndex and Jobs Service images.
i.e., if we have an image configured in the controllers_cfg.yaml, then, this image wins. see also, packages/sonataflow-operator/internal/controller/cfg/controllers_cfg.go and the processing we do for DI and JS.
If not, we must still get the image name and mostly the tag from a common place, see: https://github.com/apache/incubator-kie-tools/blob/3933d90d67c8272813200b3c5615bf02a4ba49d0/packages/sonataflow-operator/internal/controller/platform/services/services.go#L141

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,302 @@
...
+	if pshJS.IsPersistenceEnabledtInSpec() {
+		jsJobsBasedDBMigration = services.IsJobsBasedDBMigration(platform.Spec.Services.JobService.Persistence)
+	}
+
+	if (pshDI.IsServiceSetInSpec() && diJobsBasedDBMigration) || (pshJS.IsServiceSetInSpec() && jsJobsBasedDBMigration) {
+		quarkusDataSourceDataIndex := &QuarkusDataSource{}
+		quarkusDataSourceJobService := &QuarkusDataSource{}
+
+		if diJobsBasedDBMigration {
+			quarkusDataSourceDataIndex = getQuarkusDataSourceFromPersistence(ctx, platform, platform.Spec.Services.DataIndex.Persistence, "defaultDi")
```
wmedvede: default schema names for the services must be consistent with what we do for each service, "defaultDI" is not good.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+
+		if jsJobsBasedDBMigration {
+			quarkusDataSourceJobService = getQuarkusDataSourceFromPersistence(ctx, platform, platform.Spec.Services.JobService.Persistence, "defaultJs")
```
wmedvede: same here, default schema name "defaultJs", not good.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,302 @@
...
+)
+
+type DBMigrationJobCfg struct {
+	JobName       string
+	ContainerName string
+	ToolImageName string
+	MigrationCmd  string
+}
+
+func getDBSchemaName(persistencePostgreSQL *operatorapi.PersistencePostgreSQL, defaultSchemaName string) string {
```
wmedvede: I think this method is not covering all cases well.

Follow the PersistencePostgreSQL struct, and see: PostgreSQLServiceOptions


type PostgreSQLServiceOptions struct {
	*SQLServiceOptions `json:",inline"`
	// Schema of postgresql database to be used. Defaults to "data-index-service"
	// +optional
	DatabaseSchema string `json:"databaseSchema,omitempty"`
}

Take a look at this also please: https://sonataflow.org/serverlessworkflow/main/cloud/operator/supporting-services.html#_configuring_the_supporting_services_persistence

The services persistence configurations supports various configuration alternatives.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,302 @@
...
+			MigrateDBDataIndex:    diJobsBasedDBMigration,
+			DataIndexDataSource:   quarkusDataSourceDataIndex,
+			MigrateDBJobsService:  jsJobsBasedDBMigration,
+			JobsServiceDataSource: quarkusDataSourceJobService,
+		}
+	}
+	return nil
+}
+
+func getNewQuarkusDataSource(jdbcURL string, userName string, password string, schema string) *QuarkusDataSource {
```
wmedvede: I'd prefer to call this method simply newQuarkusDataSource

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,302 @@
...
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  dbMigrationJobCfg.ContainerName,
+							Image: dbMigrationJobCfg.ToolImageName,
+							Env: []corev1.EnvVar{
+								{
+									Name:  "MIGRATE_DB_DATAINDEX",
```
wmedvede: I think we should use constants for this env var names.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,302 @@
...
+	if dbMigrationStatus != nil {
+		dbMigrationStatus.Status = status
+		dbMigrationStatus.Message = message
+		dbMigrationStatus.Reason = reason
+		return dbMigrationStatus
+	}
+	return nil
+}
+
+func getDBMigrationJobCfg() *DBMigrationJobCfg {
```
wmedvede: newDBMigrationJobCfg  instead?

wmedvede (commented):
Hi  @rhkp I have added some comments, but couldn't finish the review.
Will continue on this after Christmas.

Many thanks!

packages/sonataflow-operator/api/v1alpha08/sonataflow_persistence_types.go
rhkp: @wmedvede @ricardozanini I have updated the names as suggested, as follows
type DBMigrationOperatorType string

const (
	DBMigrationOperatorService DBMigrationOperatorType = "service"
	DBMigrationOperatorJob     DBMigrationOperatorType = "job"
	DBMigrationOperatorNone    DBMigrationOperatorType = "none"
)

rhkp: @wmedvede I guess we should keep it as it is. Renaming this with `operator` causes confusing with a go operator, which is not the case specifically for these vars and are anyway part of SonataFlow-Operator?

tiagobento:
Can someone please help me understand the relationship between this PR and https://github.com/apache/incubator-kie-tools/pull/2697? 

This PR's issue has no description, which makes it hard to follow what's the overall plan for the DB Migrator image.

Thanks!

rhkp:
> Can someone please help me understand the relationship between this PR and #2697?
> 
> This PR's issue has no description, which makes it hard to follow what's the overall plan for the DB Migrator image.
> 
> Thanks!

In short, the SonataFlow Operator [PR#2790](https://github.com/apache/incubator-kie-tools/pull/2790) will make use of the image created by [PR#2697](https://github.com/apache/incubator-kie-tools/pull/2697).

packages/sonataflow-operator/api/v1alpha08/sonataflow_persistence_types.go
rhkp: I am resolving this comment for now. Please un-resolve it, if my proposal above does not make sense.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
rhkp: Please check the changes now and let me know if meets your expectations.

packages/sonataflow-operator/operator.yaml
```diff
@@ -19403,9 +19422,14 @@ spec:
                     for the workflow
                   maxProperties: 2
                   properties:
-                    migrateDBOnStartUp:
-                      description: Whether to migrate database on service startup?
-                      type: boolean
+                    dbMigrationStrategy:
```
wmedvede, line 19425: I think that if we are not implementing the dbMigrationStrategy for the workflows, we shouldn't add it in the SonataFlow CRD.

packages/sonataflow-operator/config/manager/controllers_cfg.yaml
```diff
@@ -30,6 +30,8 @@ jobsServiceEphemeralImageTag: "docker.io/apache/incubator-kie-kogito-jobs-servic
 # The Data Index image to use, if empty the operator will use the default Apache Community one based on the current operator's version
 dataIndexPostgreSQLImageTag: "docker.io/apache/incubator-kie-kogito-data-index-postgresql:main"
 dataIndexEphemeralImageTag: "docker.io/apache/incubator-kie-kogito-data-index-ephemeral:main"
+# The Kogito PostgreSQL DB Migrator image to use (TBD: to replace with apache image)
+kogitoDBMigratorToolImageTag: "quay.io/rhkp/incubator-kie-kogito-service-db-migration-postgresql:latest"
```
wmedvede: For the other DI and JS imageTags, we just say, `jobsServicePostgreSQLImageTag`, `jobsServiceEphemeralImageTag`,     `dataIndexPostgreSQLImageTag`, `dataIndexEphemeralImageTag` and we don't prefix with kogito, even when they are.

I think we should just say:  `dbMigratorToolImageTag` for consistency.

And propagate the change along the controller config struct, etc.

wdyt? @ricardozanini @rhkp

packages/sonataflow-operator/config/manifests/bases/sonataflow-operator.clusterserviceversion.yaml
wmedvede, line 1: I think the dbMigrationStrategy  must be added to the DI and JS in this file too.

                          dbMigrationStrategy:
                            default: service
                            description: |-
                              DB Migration approach for service?
                              job: use job based approach
                              service: service itself shall migrate the db
                              none: no db migration needed

packages/sonataflow-operator/internal/controller/cfg/controllers_cfg.go
```diff
@@ -63,6 +63,7 @@ type ControllersCfg struct {
 	JobsServiceEphemeralImageTag    string            `yaml:"jobsServiceEphemeralImageTag,omitempty"`
 	DataIndexPostgreSQLImageTag     string            `yaml:"dataIndexPostgreSQLImageTag,omitempty"`
 	DataIndexEphemeralImageTag      string            `yaml:"dataIndexEphemeralImageTag,omitempty"`
+	KogitoDBMigratorToolImageTag    string            `yaml:"kogitoDBMigratorToolImageTag,omitempty"`
```
wmedvede: see comments on this field name

packages/sonataflow-operator/internal/controller/platform/services/secrets.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede: wrong licence header, we don't use the yearly format.

packages/sonataflow-operator/test/e2e/testdata/platform/persistence/job_based_db_migration/pg-service/01-postgres.yaml
```diff
@@ -0,0 +1,85 @@
+# Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede: wrong licence format

packages/sonataflow-operator/test/e2e/testdata/platform/persistence/job_based_db_migration/pg-service/kustomization.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede: wrong licence format

packages/sonataflow-operator/test/e2e/testdata/platform/persistence/job_based_db_migration/sonataflow-platform/01-sonataflow_platform.yaml
```diff
@@ -0,0 +1,43 @@
+# Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede: wrong license format

packages/sonataflow-operator/test/e2e/testdata/platform/persistence/job_based_db_migration/sonataflow-platform/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede: wrong license format

packages/sonataflow-operator/test/e2e/testdata/platform/persistence/service_based_db_migration/01-postgres.yaml
```diff
@@ -0,0 +1,85 @@
+# Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede: wrong license format

packages/sonataflow-operator/test/e2e/testdata/platform/persistence/service_based_db_migration/02-sonataflow_platform.yaml
```diff
@@ -0,0 +1,65 @@
+# Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede: wrong license format

packages/sonataflow-operator/test/e2e/testdata/platform/persistence/service_based_db_migration/kustomization.yaml
```diff
@@ -0,0 +1,31 @@
+# Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede: wrong license format

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+	MigrateDBDataIndex    bool
+	DataIndexDataSource   *QuarkusDataSource
+	MigrateDBJobsService  bool
+	JobsServiceDataSource *QuarkusDataSource
+}
+
+const (
+	dbMigrationJobName       = "sonataflow-db-migrator-job"
+	dbMigrationContainerName = "db-migration-container"
+	dbMigratorToolImage      = "quay.io/rhkp/incubator-kie-kogito-service-db-migration-postgresql:latest"
```
wmedvede: Unused constant

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+	if pshJS.IsPersistenceEnabledtInSpec() {
+		jsJobsBasedDBMigration = services.IsJobsBasedDBMigration(platform.Spec.Services.JobService.Persistence)
+	}
+
+	if (pshDI.IsServiceSetInSpec() && diJobsBasedDBMigration) || (pshJS.IsServiceSetInSpec() && jsJobsBasedDBMigration) {
+		quarkusDataSourceDataIndex := &QuarkusDataSource{}
+		quarkusDataSourceJobService := &QuarkusDataSource{}
+
+		if diJobsBasedDBMigration {
+			quarkusDataSourceDataIndex = getQuarkusDataSourceFromPersistence(ctx, platform, platform.Spec.Services.DataIndex.Persistence, "data-index-service")
```
wmedvede: When get the DataSource, we must consider the comments above, i.e., the potential case where the persistence is defined at the SPF level.

In the file packages/sonataflow-operator/internal/controller/profiles/common/persistence/postgresql.go  I think we have some methods that can be useful for these calculations.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+
+		if jsJobsBasedDBMigration {
+			quarkusDataSourceJobService = getQuarkusDataSourceFromPersistence(ctx, platform, platform.Spec.Services.JobService.Persistence, "jobs-service")
```
wmedvede: same as for the DI, see comment above.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+				}
+			} else {
+				return a
+			}
+		}
+	}
+	return defaultSchemaName
+}
+
+func getJdbcUrl(postgresql *operatorapi.PersistencePostgreSQL, defaultDBSchemaName string) string {
```
wmedvede: I believe this calculations are kind of duplicated here: https://github.com/apache/incubator-kie-tools/blob/c690d462602da4cf2c01b90279815b60a1959162/packages/sonataflow-operator/internal/controller/profiles/common/persistence/postgresql.go#L44

I think we should explore moving this calculation postgresql.go and reuse.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
wmedvede: Also, to get the default schema name, we do 

https://github.com/apache/incubator-kie-tools/blob/ab197e99ea446da684baa2a533a68413e2e23e22/packages/sonataflow-operator/internal/controller/platform/services/services.go#L291

I think we should do something similar here instead of setting "data-index-service"

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
wmedvede: and same comment for default schema name calculation, see: j.GetServiceName()

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
wmedvede, line 1: This file needs tests, the same as we have for other ones.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+)
+
+type DBMigrationJobCfg struct {
+	JobName       string
+	ContainerName string
+	ToolImageName string
+	MigrationCmd  string
+}
+
+func getDBSchemaName(persistencePostgreSQL *operatorapi.PersistencePostgreSQL, defaultSchemaName string) string {
```
wmedvede: tests please

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+	} else if len(postgresql.JdbcUrl) > 0 {
+		dataSourceURL = postgresql.JdbcUrl
+	} else {
+		dataSourceURL = fmt.Sprintf("jdbc:postgresql://%s:%d/%s?currentSchema=%s", postgresServiceName, dataSourcePort, databaseName, databaseSchema)
+	}
+
+	return dataSourceURL
+}
+
+func getQuarkusDataSourceFromPersistence(ctx context.Context, platform *operatorapi.SonataFlowPlatform, persistence *operatorapi.PersistenceOptionsSpec, defaultSchemaName string) *QuarkusDataSource {
```
wmedvede: Important:

Persistence configuration for a service can be taken not only from eg. spec.services.jobsService but also from the SPF global persistence configuration.

see:

https://sonataflow.org/serverlessworkflow/main/cloud/operator/supporting-services.html#common-persistence-configuration

https://sonataflow.org/serverlessworkflow/main/cloud/operator/using-persistence.html#configuring-persistence-using-the-sonataflowplatform-cr

Along this PR, we must consider that potential case too.

I.e, the service don't have "persistence", but the SPF do.

Examples:

```
	apiVersion: sonataflow.org/v1alpha08
	kind: SonataFlowPlatform
	metadata:
	  name: sonataflow-platform
	spec:
	  services:
	    # no persistence is defined for the DI, it will take it from the SFP if any.
	    # and apply the by default DBMigrationStrategy
	    dataIndex:
	      enabled: true

	    jobService:
	      # The only persistence setting we do is the DBMigrationStrategy
	      # So, the service grabs the persistence configuation from the SPF and the given DBMigrationStrategy
	      # from here.
	      
	      enabled: true
	      persistence:
		dbMigrationStrategy: "job"
	  
	  # global SPF persitence configuration.     
	  persistence:
	    postgresql:
	      secretRef:
		name: postgres-secrets
		userKey: POSTGRESQL_USER
		passwordKey: POSTGRESQL_PASSWORD
	      serviceRef:
		name: postgres
		databaseName: sonataflow

```

@rhkp  take look please and propagate in all places where needed.

@ricardozanini see my considerations for this particular case, where a service takes the persistence from the SPF, but we still can mark at the service level the dbMigrationStrategy

wmedvede (changes requested):

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	klog.V(log.I).InfoS("Db migration job status: ", "active", dbMigratorJobStatus.Active, "ready", dbMigratorJobStatus.Ready, "failed", dbMigratorJobStatus.Failed, "success", dbMigratorJobStatus.Succeeded, "CompletedIndexes", dbMigratorJobStatus.CompletedIndexes, "terminatedPods", dbMigratorJobStatus.UncountedTerminatedPods)
+
+	if dbMigratorJobStatus.Failed == dbMigrationJobFailed {
+		platform.Status.SonataFlowPlatformDBMigrationPhase = UpdateSonataFlowPlatformDBMigrationPhase(platform.Status.SonataFlowPlatformDBMigrationPhase, operatorapi.DBMigrationStatusFailed, operatorapi.MessageDBMigrationStatusFailed, operatorapi.ReasonDBMigrationStatusFailed)
+		klog.V(log.I).InfoS("DB migration job failed")
+		return dbMigratorJobStatus, errors.New("DB migration job failed")
```
wmedvede: Information about the job namespace and name would be nice here for future error tracing.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+	} else if dbMigratorJobStatus.Succeeded == dbMigrationJobSucceeded {
+		platform.Status.SonataFlowPlatformDBMigrationPhase = UpdateSonataFlowPlatformDBMigrationPhase(platform.Status.SonataFlowPlatformDBMigrationPhase, operatorapi.DBMigrationStatusSucceeded, operatorapi.MessageDBMigrationStatusSucceeded, operatorapi.ReasonDBMigrationStatusSucceeded)
+		klog.V(log.I).InfoS("DB migration job succeeded")
```
wmedvede: same here

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,356 @@
...
+	platform.Status.SonataFlowPlatformDBMigrationPhase = NewSonataFlowPlatformDBMigrationPhase(operatorapi.DBMigrationStatusStarted, operatorapi.MessageDBMigrationStatusStarted, operatorapi.ReasonDBMigrationStatusStarted)
+
+	dbMigratorJobStatus, err := dbmj.GetDBMigrationJobStatus(ctx, client, platform)
+	if err != nil {
+		return nil, err
+	}
+
+	klog.V(log.I).InfoS("Db migration job status: ", "active", dbMigratorJobStatus.Active, "ready", dbMigratorJobStatus.Ready, "failed", dbMigratorJobStatus.Failed, "success", dbMigratorJobStatus.Succeeded, "CompletedIndexes", dbMigratorJobStatus.CompletedIndexes, "terminatedPods", dbMigratorJobStatus.UncountedTerminatedPods)
+
+	if dbMigratorJobStatus.Failed == dbMigrationJobFailed {
```
wmedvede: I think this conditions to check the failure or sucesss fo the the job by using constants with the same value == 1 is a bit confusing when we read this method.

I'd recommend creating a function e.g.  hasFailed(status * JobStatus) and do this calculations there instead

packages/sonataflow-operator/internal/controller/platform/k8s.go
```diff
@@ -62,20 +63,86 @@ func (action *serviceAction) CanHandle(platform *operatorapi.SonataFlowPlatform)
...
+	}
+
+	return platform, nil
+}
+
+func (action *serviceAction) isJobsBasedDBMigration(platform *operatorapi.SonataFlowPlatform, pshDI services.PlatformServiceHandler, pshJS services.PlatformServiceHandler) bool {
+	diJobsBasedDBMigration := false
+	jsJobsBasedDBMigration := false
+
+	if pshDI.IsPersistenceEnabledtInSpec() {
```
wmedvede: Consider my comments regarding a potential persistence configuration at the SFP level

packages/sonataflow-operator/internal/controller/platform/k8s.go
```diff
@@ -62,20 +63,86 @@ func (action *serviceAction) CanHandle(platform *operatorapi.SonataFlowPlatform)
...
+		} else {
+			// DB migration is still running
+			return nil, nil
+		}
+	}
+
+	return platform, nil
+}
+
+func (action *serviceAction) isJobsBasedDBMigration(platform *operatorapi.SonataFlowPlatform, pshDI services.PlatformServiceHandler, pshJS services.PlatformServiceHandler) bool {
```
wmedvede: I believe this method is not particularly belonging to the serviceAction.

packages/sonataflow-operator/internal/controller/platform/k8s.go
```diff
@@ -62,20 +63,86 @@ func (action *serviceAction) CanHandle(platform *operatorapi.SonataFlowPlatform)
...
+		}); err != nil {
+			return dbMigratorJob, err
+		} else {
+			klog.V(log.I).InfoS("DB Migration Job successfully created on cluster", "operation", op)
+		}
+	}
+	return dbMigratorJob, nil
+}
+
+func (action *serviceAction) handleDBMigrationJob(ctx context.Context, platform *operatorapi.SonataFlowPlatform, psDI services.PlatformServiceHandler, psJS services.PlatformServiceHandler) (*operatorapi.SonataFlowPlatform, error) {
```
wmedvede: same for this method, I don't believe it belongs to the serviceAction struct

packages/sonataflow-operator/internal/controller/platform/k8s.go
```diff
@@ -62,20 +63,86 @@ func (action *serviceAction) CanHandle(platform *operatorapi.SonataFlowPlatform)
 	return platform.Status.IsReady()
 }
 
+func (action *serviceAction) createOrUpdateDBMigrationJob(ctx context.Context, client client.Client, platform *operatorapi.SonataFlowPlatform, pshDI services.PlatformServiceHandler, pshJS services.PlatformServiceHandler) (*DBMigratorJob, error) {
+	dbMigratorJob := NewDBMigratorJobData(ctx, action.client, platform, pshDI, pshJS)
+
+	// Invoke DB Migration only if both or either DI/JS services are requested, in addition to DBMigrationStrategyJob
+	if dbMigratorJob != nil {
+		klog.V(log.I).InfoS("Starting DB Migration Job: ")
```
wmedvede: printing the job namespace and name will be very helpful for future error tracing, etc.

packages/sonataflow-operator/internal/controller/platform/k8s.go
```diff
@@ -62,20 +63,86 @@ func (action *serviceAction) CanHandle(platform *operatorapi.SonataFlowPlatform)
...
+
+		job := dbMigratorJob.CreateJobDBMigration(platform)
+		if op, err := controllerutil.CreateOrUpdate(ctx, client, job, func() error {
+			return nil
+		}); err != nil {
+			return dbMigratorJob, err
+		} else {
+			klog.V(log.I).InfoS("DB Migration Job successfully created on cluster", "operation", op)
```
wmedvede: same here, namespace and job name is very welcome

packages/sonataflow-operator/internal/controller/platform/k8s.go
```diff
@@ -62,20 +63,86 @@ func (action *serviceAction) CanHandle(platform *operatorapi.SonataFlowPlatform)
...
 	if err := CreateOrUpdateWithDefaults(ctx, platform, false); err != nil {
 		return nil, nil, err
 	}
 
 	psDI := services.NewDataIndexHandler(platform)
+	psJS := services.NewJobServiceHandler(platform)
+
+	if action.isJobsBasedDBMigration(platform, psDI, psJS) {
+		p, err := action.handleDBMigrationJob(ctx, platform, psDI, psJS)
+		if p == nil {
```
wmedvede: if p == nill we return the err...., not good.

I suggest making this more clear as part of the suggestions for the isJobBasedDBMigration and handleDBMigratonJob functions.

wmedvede (changes requested):

packages/sonataflow-operator/internal/controller/platform/k8s.go
rhkp: Moved to db_migrator_job.go and now won't belong to serviceAction.

packages/sonataflow-operator/internal/controller/platform/k8s.go
rhkp: Moved to db_migrator_job.go and now won't belong to serviceAction.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
rhkp: Also added hasSucceeded method there.

packages/sonataflow-operator/config/manifests/bases/sonataflow-operator.clusterserviceversion.yaml
rhkp: @wmedvede Where are DI and JS in this file? Please advise.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
rhkp: Hi @wmedvede!
They are included here e.g.
* [Service Based Migration](https://github.com/rhkp/incubator-kie-tools/blob/ab197e99ea446da684baa2a533a68413e2e23e22/packages/sonataflow-operator/test/e2e/platform_test.go#L229)
* [Job Based Migration](https://github.com/rhkp/incubator-kie-tools/blob/ab197e99ea446da684baa2a533a68413e2e23e22/packages/sonataflow-operator/test/e2e/platform_test.go#L282)

Please let me know if you are referring to something else.

rhkp: Hi @wmedvede!
Somewhat inspired by https://github.com/apache/incubator-kie-tools/blob/main/packages/sonataflow-operator/internal/controller/platform/services/services_test.go I have added them now. 
Please let me know if that is not in-line with the expecatations.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
rhkp: Please see comment above. Added with inspiration from https://github.com/apache/incubator-kie-tools/blob/main/packages/sonataflow-operator/internal/controller/platform/services/services_test.go

packages/sonataflow-operator/operator.yaml
rhkp: @wmedvede 

- This is added by `make generate-all` command automatically.
- This is being added even though `DBMigrationStrategy` is annotated with `+optional`, operator-sdk adds it in `operator.yaml`. Please see [this](https://github.com/apache/incubator-kie-tools/blob/ab197e99ea446da684baa2a533a68413e2e23e22/packages/sonataflow-operator/api/v1alpha08/sonataflow_persistence_types.go#L69).
- Also, as the `PersistenceOptionsSpec` is included in [sonataflow_types.go](https://github.com/apache/incubator-kie-tools/blob/ab197e99ea446da684baa2a533a68413e2e23e22/packages/sonataflow-operator/api/v1alpha08/sonataflow_types.go#L147) by SonataFlowSpec struct.

packages/sonataflow-operator/config/manifests/bases/sonataflow-operator.clusterserviceversion.yaml
wmedvede: @rhkp your observation is good, and mine too :thinking: 

Rethinking here, It looks like we have a sort of outdated (and not necessarily outdated) specification of the CustomResourceDefinitions in this file... i.e., not all the fields were always added to the respective CRDs.

Some fields were added, and some other don't.

For example:
we have things like  `path: services.jobService.podTemplate`  but we don't have things like `services.jobService.persistence`

Which makes me think that not all the fields for the different CRDs owned by this operator where added to this file along the time. Question is, do we need them all added there? I think that not, see below.

Also, by looking at here: https://olm.operatorframework.io/docs/concepts/crds/clusterserviceversion I think the full specification of each CRD in that list is not needed (that information is not used for creating the respective CRDs), instead we could just only declare file the list of owned resources by this operator....


@ricardozanini  is my reasoning correct?

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
rhkp: I agree. Now we use JdbcURL calculations from postgresql.go.

packages/sonataflow-operator/config/manager/controllers_cfg.yaml
```diff
@@ -30,6 +30,8 @@ jobsServiceEphemeralImageTag: "docker.io/apache/incubator-kie-kogito-jobs-servic
 # The Data Index image to use, if empty the operator will use the default Apache Community one based on the current operator's version
 dataIndexPostgreSQLImageTag: "docker.io/apache/incubator-kie-kogito-data-index-postgresql:main"
 dataIndexEphemeralImageTag: "docker.io/apache/incubator-kie-kogito-data-index-ephemeral:main"
+# The Kogito PostgreSQL DB Migrator image to use (TBD: to replace with apache image)
+dbMigratorToolImageTag: "quay.io/rhkp/incubator-kie-kogito-service-db-migration-postgresql:latest"
```
jakubschwan: do not forget to replace with apache image

packages/sonataflow-operator/api/v1alpha08/sonataflow_persistence_types.go
```diff
@@ -54,10 +62,13 @@ type PersistenceOptionsSpec struct {
 	// +optional
 	PostgreSQL *PersistencePostgreSQL `json:"postgresql,omitempty"`
 
-	// Whether to migrate database on service startup?
+	// DB Migration approach for service?
```
jakubschwan: Please update the description to better describe the configuration for DB migration.  And remove the question mark from the description.

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+	if dbMigratorJob != nil {
+		klog.V(log.E).InfoS("Created DB migration job")
+		dbMigratorJobStatus, err := dbMigratorJob.ReconcileDBMigrationJob(ctx, client, platform)
+		if err != nil {
+			return nil, err
+		}
+		if dbMigratorJobStatus.BatchJobStatus.Failed == 1 {
+			return nil, errors.New("DB migration job " + dbMigratorJobStatus.Name + " failed in namespace: " + platform.Namespace)
+		} else if dbMigratorJobStatus.BatchJobStatus.Succeeded == 1 {
```
jakubschwan: As there are functions for checking these statuses, can you replace these conditions by functions `hasFailed`  and `hasSucceeded`?

packages/sonataflow-operator/config/manager/controllers_cfg.yaml
rhkp: Agreed as soon as the image PR is approved and image is available, we will do the same.

pkliczewski:
@ricardozanini @wmedvede  once https://github.com/apache/incubator-kie-tools/pull/2697, will this PR get approved or is there any more work needed?

wmedvede:
Hi @rhkp , I have built the db-migrator locally and pushed to my quay, etc, and also the latest changes in this RP.

Then, after installing everything in my OpenShift, etc, when deploying a SonataFlowPlatorm I go this error:

image


This is the platform I was working with:

```
apiVersion: sonataflow.org/v1alpha08
kind: SonataFlowPlatform
metadata:
  name: sonataflow-platform
spec:
  services:
    dataIndex:
      enabled: true
      persistence:
        dbMigrationStrategy: job
        postgresql:
          serviceRef:
            name: postgres
            databaseName: sonataflow
            databaseSchema: data-index-schema
          secretRef:
            name: postgres-secrets
            userKey: POSTGRESQL_USER
            passwordKey: POSTGRESQL_PASSWORD
      podTemplate:
        initContainers:
          - name: init-postgres
            image: registry.access.redhat.com/ubi9/ubi-minimal:latest
            imagePullPolicy: IfNotPresent
            command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
    jobService:
      enabled: true
      persistence:
        dbMigrationStrategy: job
        postgresql:
          serviceRef:
            name: postgres
            databaseName: sonataflow
            databaseSchema: jobs-service-schema
          secretRef:
            name: postgres-secrets
            userKey: POSTGRESQL_USER
            passwordKey: POSTGRESQL_PASSWORD
      podTemplate:
        initContainers:
          - name: init-postgres
            image: registry.access.redhat.com/ubi9/ubi-minimal:latest
            imagePullPolicy: IfNotPresent
            command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
```


However, the migrator tool image works well if I execute it locally.

rhkp:
> Hi @rhkp , I have built the db-migrator locally and pushed to my quay, etc, and also the latest changes in this RP.
> 
> Then, after installing everything in my OpenShift, etc, when deploying a SonataFlowPlatorm I go this error:
> 
> image
> 
> This is the platform I was working with:
> 
> ```
> apiVersion: sonataflow.org/v1alpha08
> kind: SonataFlowPlatform
> metadata:
>   name: sonataflow-platform
> spec:
>   services:
>     dataIndex:
>       enabled: true
>       persistence:
>         dbMigrationStrategy: job
>         postgresql:
>           serviceRef:
>             name: postgres
>             databaseName: sonataflow
>             databaseSchema: data-index-schema
>           secretRef:
>             name: postgres-secrets
>             userKey: POSTGRESQL_USER
>             passwordKey: POSTGRESQL_PASSWORD
>       podTemplate:
>         initContainers:
>           - name: init-postgres
>             image: registry.access.redhat.com/ubi9/ubi-minimal:latest
>             imagePullPolicy: IfNotPresent
>             command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
>     jobService:
>       enabled: true
>       persistence:
>         dbMigrationStrategy: job
>         postgresql:
>           serviceRef:
>             name: postgres
>             databaseName: sonataflow
>             databaseSchema: jobs-service-schema
>           secretRef:
>             name: postgres-secrets
>             userKey: POSTGRESQL_USER
>             passwordKey: POSTGRESQL_PASSWORD
>       podTemplate:
>         initContainers:
>           - name: init-postgres
>             image: registry.access.redhat.com/ubi9/ubi-minimal:latest
>             imagePullPolicy: IfNotPresent
>             command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
> ```
> 
> However, the migrator tool image works well if I execute it locally.

Hi @wmedvede ,
The image package has been updated and now works with OpenShift as well. Please do check and let me know if you face any problems. 
image
FYI only, prior to this, I had tried the following 3 approaches, which unfortunately did not work.
1. Use [Maven Archiver](https://maven.apache.org/shared/maven-archiver/examples/classpath.html) with classpath options.
2. Use [Quarkus Class Loading](https://quarkus.io/guides/class-loading-reference) by specifying loading quarkus runner first.
3. Use classpath option in java -jar command while executing the image.

wmedvede:
Hi @rhkp ,
while doing some testing I have found a small issue in the DB Migrator:

Let me explain:

A database produced by the DB Migrator looks like this:

DataBase-DB-Migrator


On the other hand, if we start the data index service (not using the db migrator) and we enable flyway for the data index, we can see a database like this:


DataBase-DataIndex

So we have a 1.5.0 and 1.5.1 files that are included by the data index service but not by the db migrator.

This is basically because the data index service has a dependency on the `persistence-commons` module that incorporates the missing files.

Your work is good! but we need to make the db migrator incorporate the .sql files provided by that module.

See here in the db migrator  build structure:

MissingSqlFiles

would you mind take a look ?

rhkp:
> Hi @rhkp , while doing some testing I have found a small issue in the DB Migrator:
> 
> Let me explain:
> 
> A database produced by the DB Migrator looks like this:
> 
> DataBase-DB-Migrator
> 
> On the other hand, if we start the data index service (not using the db migrator) and we enable flyway for the data index, we can see a database like this:
> 
> DataBase-DataIndex
> 
> So we have a 1.5.0 and 1.5.1 files that are included by the data index service but not by the db migrator.
> 
> This is basically because the data index service has a dependency on the `persistence-commons` module that incorporates the missing files.
> 
> Your work is good! but we need to make the db migrator incorporate the .sql files provided by that module.
> 
> See here in the db migrator build structure:
> 
> MissingSqlFiles
> 
> would you mind take a look ?

Hi @wmedvede / @ricardozanini,
The DB Migrator will now apply `persistence-commons` sql files too for data-index. Please let me know if this is in-line with your expectations?
Thanks.

pkliczewski:
@wmedvede @ricardozanini any news about this?

pkliczewski:
@ricardozanini any progress with this change?

ricardozanini:
@jakubschwan can you please verify?

packages/sonataflow-operator/config/manager/controllers_cfg.yaml
ricardozanini:
```suggestion
dbMigratorToolImageTag: "docker.io/apache/incubator-kie-kogito-db-migrator-tool:latest"
```
Can you please also review this image name in the DBMigrator Image package? https://github.com/apache/incubator-kie-tools/tree/main/packages/kogito-db-migrator-tool-image

I think it's wrong there.

ricardozanini: @rhkp ^

ricardozanini (approved):
@rhkp can you please double check the image tag?

Many thanks for this work, we really appreciate!

rhkp:
> @rhkp can you please double check the image tag?
> 
> Many thanks for this work, we really appreciate!

Hi @ricardozanini,
Thanks for approvals and comments. 
I do not have access to push to `apache` on docker registry. 
Last I had understood the `incubator-kie-tools CI` will push the image and not sure, I still do not see image on hub.docker.com?
Once the image is there I can change this line? Please advise. Thanks.

image

packages/sonataflow-operator/config/manager/controllers_cfg.yaml
rhkp: I agree but we need the built image to be available on docker registry before we change this line. That is why we still have this comment in unresolved state. Once the image is available on docker, we will update this line and resolve this comment.

ricardozanini:
@rhkp I meant to review this package: https://github.com/apache/incubator-kie-tools/tree/main/packages/kogito-db-migrator-tool-image to check if the image tag name is correct.

rhkp:
> @rhkp I meant to review this package: https://github.com/apache/incubator-kie-tools/tree/main/packages/kogito-db-migrator-tool-image to check if the image tag name is correct.

Hi @ricardozanini I understand what you are referring to. 
- Please note in the `kogito-db-migrator-tool-image` package currently we are making use of tag `main`. 
- Earlier we were using `latest` tag. 
- However based on [this review comment](https://github.com/apache/incubator-kie-tools/pull/2697#discussion_r1917118786) by @tiagobento we switched to using `main` there.
- So the `kogito-db-migrator-tool-image` package is correct and we shouldn't change it.

ricardozanini:
> > @rhkp I meant to review this package: https://github.com/apache/incubator-kie-tools/tree/main/packages/kogito-db-migrator-tool-image to check if the image tag name is correct.
> 
> Hi @ricardozanini I understand what you are referring to.
> 
> * Please note in the `kogito-db-migrator-tool-image` package currently we are making use of tag `main`.
> * Earlier we were using `latest` tag.
> * However based on [this review comment](https://github.com/apache/incubator-kie-tools/pull/2697#discussion_r1917118786) by @tiagobento we switched to using `main` there.
> * So the `kogito-db-migrator-tool-image` package is correct and we shouldn't change it.

So every place required the tag `docker.io/apache/incubator-kie-kogito-db-migrator-tool` is correct? I'm not referring to the tag version, but the whole naming.

rhkp:
> > > @rhkp I meant to review this package: https://github.com/apache/incubator-kie-tools/tree/main/packages/kogito-db-migrator-tool-image to check if the image tag name is correct.
> > 
> > 
> > Hi @ricardozanini I understand what you are referring to.
> > 
> > * Please note in the `kogito-db-migrator-tool-image` package currently we are making use of tag `main`.
> > * Earlier we were using `latest` tag.
> > * However based on [this review comment](https://github.com/apache/incubator-kie-tools/pull/2697#discussion_r1917118786) by @tiagobento we switched to using `main` there.
> > * So the `kogito-db-migrator-tool-image` package is correct and we shouldn't change it.
> 
> So every place required the tag `docker.io/apache/incubator-kie-kogito-db-migrator-tool` is correct? I'm not referring to the tag version, but the whole naming.

Hi @ricardozanini 
I totally agree we need to change the image names in `sonataflow-operator` package from `quay.io/rhkp/incubator-kie-kogito-service-db-migration-postgresql:latest` to `docker.io/apache/incubator-kie-kogito-db-migrator-tool:main`. 

But please refer to my earlier [comment to @jakubschwan ](https://github.com/apache/incubator-kie-tools/pull/2790/files#r1924217939) and [to you too](https://github.com/apache/incubator-kie-tools/pull/2790/files#r1962263527) that we do not have the `docker.io/apache` image yet available on hub.docker as shown in [screenshot](https://github.com/apache/incubator-kie-tools/pull/2790#issuecomment-2669579956) shared yesterday. So we cannot update this image right now in this package.

If we change the name at this moment to non-existent image, the operator will not work with DB migration job as the desired image `docker.io/apache/incubator-kie-kogito-db-migrator-tool:main` does not exist.

My recommendation is to keep the current name **only for now** so the testing can proceed. And as soon as the real `docker.io/apache` image for db migration is **available**, we will update this one.

Please let me know if we are on same understanding or still we have some difference in understanding?

ricardozanini:
@rhkp I'm not talking about changing the image name here, but in this package: https://github.com/apache/incubator-kie-tools/tree/main/packages/kogito-db-migrator-tool-image

Can you please double-check if all the naming there is correct and **have the expected** tag name?

If there's a place in that package that the name is not  `docker.io/apache/incubator-kie-kogito-db-migrator-tool`, we won't be able to run anything.

And I have a ticket to create the Dockerhub namespace: https://issues.apache.org/jira/browse/INFRA-26557

Once it's created we can run the Jenkins job to publish the first tag and you change it here if the PR is still opened.

rhkp:
> @rhkp I'm not talking about changing the image name here, but in this package: https://github.com/apache/incubator-kie-tools/tree/main/packages/kogito-db-migrator-tool-image
> 
> Can you please double-check if all the naming there is correct and **have the expected** tag name?
> 
> If there's a place in that package that the name is not `docker.io/apache/incubator-kie-kogito-db-migrator-tool`, we won't be able to run anything.
> 
> And I have a ticket to create the Dockerhub namespace: https://issues.apache.org/jira/browse/INFRA-26557
> 
> Once it's created we can run the Jenkins job to publish the first tag and you change it here if the PR is still opened.

Thanks @ricardozanini. This clarifies it and I will check it first thing start of the next week and submit change PR if needed.

pkliczewski:
@wmedvede any changes you would like to see being done as part of this PR?

rhkp:
> docker.io/apache/incubator-kie-kogito-db-migrator-tool

Hi @ricardozanini,
I can confirm all the following image tags in the `kogito-db-migrator-tool-image` package are `docker.io/apache/incubator-kie-kogito-db-migrator-tool` tags as expected...

- [ReadMe#L31](https://github.com/apache/incubator-kie-tools/blob/main/packages/kogito-db-migrator-tool-image/README.md?plain=1#L31)
- [ReadMe#L67](https://github.com/apache/incubator-kie-tools/blob/main/packages/kogito-db-migrator-tool-image/README.md?plain=1#L67)
- [incubator-kie-kogito-db-migrator-tool-image.yaml#L19](https://github.com/apache/incubator-kie-tools/blob/main/packages/kogito-db-migrator-tool-image/resources/incubator-kie-kogito-db-migrator-tool-image.yaml#L19).

Please let me know if this is as per your expectations. If not I can look further into it.

Thanks.

ricardozanini:
@rhkp the repository has been created: https://hub.docker.com/r/apache/kie-kogito-db-migrator-tool

As you can see, they removed the `incubator-` prefix. Can you update this PR and the packages merged accordingly? So that we can run Jenkins and check if the `:main` tag is pushed correctly?

rhkp:
> @rhkp the repository has been created: https://hub.docker.com/r/apache/kie-kogito-db-migrator-tool
> 
> As you can see, they removed the `incubator-` prefix. Can you update this PR and the packages merged accordingly? So that we can run Jenkins and check if the `:main` tag is pushed correctly?

@ricardozanini Please check and acknowledge/merge this new [PR](https://github.com/apache/incubator-kie-tools/pull/2928) or let me know of changes to be made.

wmedvede:
Hi @rhkp , I have configured the operator locally to use the image produced by @rgdoliveira and that locally works fine:

registry-proxy.engineering.redhat.com/rh-osbs/openshift-serverless-1-logic-db-migrator-tool-rhel8:openshift-serverless-nightly-rhel-8-containers-candidate-68139-20250226193811](http://registry-proxy.engineering.redhat.com/rh-osbs/openshift-serverless-1-logic-db-migrator-tool-rhel8:openshift-serverless-nightly-rhel-8-containers-candidate-68139-20250226193811

I have just pushed it to my quay here, to simplify: quay.io/wmedvede/openshift-serverless-1-logic-db-migrator-tool-rhel8:openshift-serverless-nightly-rhel-8-containers-candidate-68139-20250226193811

All good, image locally works wood.

But, when the image is used by the db-migrator Job in OpenShift, I can see the following error:

image

The produced image by  @rgdoliveira includes all the latest changes from main, since he has synchronized the latest changes from main.

would you mind take a look?

wmedvede:
@rhkp I have been working on the error above, and we have potential fix for the productized image, that is slightly different that the one we generate in community. Same stuff, but with some minor configs different.
The image in community is good

packages/sonataflow-operator/config/manager/controllers_cfg.yaml
rhkp: Resolving this as the image is now being pointed to `docker.io` instead of `quay.io/rhkp`.

rhkp: Resolving this as the image is now being pointed to `docker.io` instead of `quay.io/rhkp`.

wmedvede:
@rhkp 

**Migrator job labels**
Similar to what we have for the other DI / JS, we must add the following labels to the migrator job.
MigratorJobLabels

wmedvede:
@rhkp 

**Migrator job ownerReferences:**
Similar to what we have for the DI / JS we must configure the owner of the migrator job at creation time.
[see DI and JS as example]( https://github.com/apache/incubator-kie-tools/blob/aab8cc107809a519cf1650daf2c255cfd40a443a/packages/sonataflow-operator/internal/controller/platform/k8s.go#L197)

MigratorJobOwnerReferences

wmedvede:
@rhkp 
**Persistence configuration at the SonataFlowPlatform not working:**

When we configure a platform like below, the migrator job creation procedure is calculating wrong the namespaces for the ID and JS respectively.

```
apiVersion: sonataflow.org/v1alpha08
kind: SonataFlowPlatform
metadata:
  name: sonataflow-platform
spec:
  services:
    dataIndex:
      enabled: true
      persistence:
        dbMigrationStrategy: job
      podTemplate:
        initContainers:
          - name: init-postgres
            image: registry.access.redhat.com/ubi9/ubi-minimal:latest
            imagePullPolicy: IfNotPresent
            command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
    jobService:
      enabled: true
      persistence:
        dbMigrationStrategy: job
      podTemplate:
        initContainers:
          - name: init-postgres
            image: registry.access.redhat.com/ubi9/ubi-minimal:latest
            imagePullPolicy: IfNotPresent
            command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
  persistence:
    postgresql:
      secretRef:
        name: postgres-secrets
        userKey: POSTGRESQL_USER
        passwordKey: POSTGRESQL_PASSWORD
      serviceRef:
        name: postgres
        databaseName: sonataflow
```

Migrator job generator configures this:

image

The regular deployment procedure in that particular use case calculates this for each service:


image

image

Migrator job schema name calculation for this use case must be aligned with the calculation we do for DI and JS deployment.

wmedvede:
@rhkp
**Wrong default schema calculation when not set in the persistence config.**

When we configure a SPF like below, see that we are not providing the schema name.

```
apiVersion: sonataflow.org/v1alpha08
kind: SonataFlowPlatform
metadata:
  name: sonataflow-platform
spec:
  services:
    dataIndex:
      enabled: true
      persistence:
        postgresql:
          serviceRef:
            name: postgres
            databaseName: sonataflow
          secretRef:
            name: postgres-secrets
            userKey: POSTGRESQL_USER
            passwordKey: POSTGRESQL_PASSWORD
      podTemplate:
        initContainers:
          - name: init-postgres
            image: registry.access.redhat.com/ubi9/ubi-minimal:latest
            imagePullPolicy: IfNotPresent
            command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
    jobService:
      enabled: true
      persistence:
        postgresql:
          serviceRef:
            name: postgres
            databaseName: sonataflow
          secretRef:
            name: postgres-secrets
            userKey: POSTGRESQL_USER
            passwordKey: POSTGRESQL_PASSWORD
      podTemplate:
        initContainers:
          - name: init-postgres
            image: registry.access.redhat.com/ubi9/ubi-minimal:latest
            imagePullPolicy: IfNotPresent
            command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]

```

The operator produces the following default schema name calculations for DI and JS

DataIndexAndJobsServiceNoSchemaConfigured


But, if we enable the DB migrator on the same SFP:

```
apiVersion: sonataflow.org/v1alpha08
kind: SonataFlowPlatform
metadata:
  name: sonataflow-platform
spec:
  services:
    dataIndex:
      enabled: true
      persistence:
        dbMigrationStrategy: job
        postgresql:
          serviceRef:
            name: postgres
            databaseName: sonataflow
          secretRef:
            name: postgres-secrets
            userKey: POSTGRESQL_USER
            passwordKey: POSTGRESQL_PASSWORD
      podTemplate:
        initContainers:
          - name: init-postgres
            image: registry.access.redhat.com/ubi9/ubi-minimal:latest
            imagePullPolicy: IfNotPresent
            command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
    jobService:
      enabled: true
      persistence:
        dbMigrationStrategy: job
        postgresql:
          serviceRef:
            name: postgres
            databaseName: sonataflow
          secretRef:
            name: postgres-secrets
            userKey: POSTGRESQL_USER
            passwordKey: POSTGRESQL_PASSWORD
      podTemplate:
        initContainers:
          - name: init-postgres
            image: registry.access.redhat.com/ubi9/ubi-minimal:latest
            imagePullPolicy: IfNotPresent
            command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]

```

We get the following results:

Migrator job generated these by default wrong schema names:

image


DI deployment was generated with the correct by default schema name:

Screenshot from 2025-03-01 11-14-47

JS deployment was generated with the correct by default schema name:

image

Migrator job must generate the correct by default schema names

wmedvede:
@rhkp 
**DB migrator Job uses user and password directly instead of secretKeyRef**

The db migrator job must be configured to use secretKeyRef for both DI and JS DB parameters, similar to what we are doing for these services.


JobNotUsingSecretKeyRef3

packages/sonataflow-operator/internal/controller/platform/db_migrator_job.go
```diff
@@ -229,41 +236,61 @@ func HandleDBMigrationJob(ctx context.Context, client client.Client, platform *o
...
 			Name:      dbMigrationJobCfg.JobName,
 			Namespace: platform.Namespace,
+			Labels: map[string]string{
+				workflowproj.LabelApp:          platform.Name,
+				workflowproj.LabelAppNamespace: platform.Namespace,
+				workflowproj.LabelK8SComponent: fmt.Sprintf("%s-%s", platform.Name, dbMigrationJobCfg.JobName),
+				workflowproj.LabelK8SManagedBy: "sonataflow-operator",
+				workflowproj.LabelK8SName:      dbMigrationJobCfg.JobName,
+				workflowproj.LabelK8SPartOF:    platform.Name,
+				workflowproj.LabelService:      fmt.Sprintf("%s-%s", "sonataflow-db-job", dbMigrationJobCfg.JobName),
```
ricardozanini: Use this instead: https://github.com/apache/incubator-kie-tools/blob/main/packages/sonataflow-operator/workflowproj/operator.go#L102-L129

rhkp: Hi @ricardozanini ,
Please note GetDefaultLabels(workflow *operatorapi.SonataFlow) needs SonataFlow which we do not have  available in DB Migrator Job and thus we won't be able to use it. 
Unless we create a fake workflow struct just for sake of getting labels. I feel it will confuse us later. Please let me know if this is what should we do it?

ricardozanini: Why not creating a `GetDefaultLabelsMap` and refactor `GetDefaultLabels` to use the latter, so we reuse the method's body? We should not repeat ourselves in this case since labels must be manipulated in a single place.

wmedvede (approved):
LGTM, great work.

PR #2794, opened 2024-12-06
kie-tools#2751: Remove references to Quarkus and Kogito versions; Automatic change of platform images based on env context

### LINKED ISSUE #2751: [sonataflow-operator] Remove hardcoded version in SonataFlow Operator/CLI artifacts
Force versions to come from the `kogito-versions` modules. For example, the operator configuration files should not enable version overriding like:

```yaml
postgreSQLPersistenceExtensions:
  - groupId: io.quarkus
    artifactId: quarkus-jdbc-postgresql
    version: 3.8.6
  - groupId: io.quarkus
    artifactId: quarkus-agroal
    version: 3.8.6
  - groupId: org.kie
    artifactId: kie-addons-quarkus-persistence-jdbc
    version: 999-20241016-SNAPSHOT
```

Each images from Kogito/SonataFlow also must have the platform version defined. See Data Index for example:

```shell
GECOS=JBoss user
HOME=/home/kogito
HOSTNAME=9c853adaa11a
JAVA_DATA_DIR=/deployments/data
JAVA_HOME=/usr/lib/jvm/jre
JAVA_VENDOR=openjdk
JAVA_VERSION=17
JBOSS_CONTAINER_JAVA_JVM_MODULE=/opt/jboss/container/java/jvm
JBOSS_CONTAINER_JAVA_PROXY_MODULE=/opt/jboss/container/java/proxy
JBOSS_CONTAINER_JAVA_RUN_MODULE=/opt/jboss/container/java/run
JBOSS_CONTAINER_OPENJDK_JRE_MODULE=/opt/jboss/container/openjdk/jre
JBOSS_CONTAINER_UTIL_LOGGING_MODULE=/opt/jboss/container/util/logging/
JBOSS_IMAGE_NAME=ubi8/openjdk-17-runtime
JBOSS_IMAGE_VERSION=1.20
KOGITO_DATA_INDEX_QUARKUS_PROFILE=http-events-support
KOGITO_HOME=/home/kogito
LANG=C.utf8
PATH=/home/kogito/.local/bin:/home/kogito/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/home/kogito
SHLVL=1
TERM=xterm
UID=185
USER=kogito
USER_ID=1001
_=/usr/bin/env
container=oci
```

ricardozanini:
Fix #2751 

In this PR:

- Removal of hardcoded Quarkus and Kogito versions from the operator
- Add Kogito and Quarkus versions to all images in the platform
- When adding a new extension, use the Kogito and Quarkus versions from the platform
- Operator won't pass versions from the platform to the builder anymore; hence the builder will fetch the extension based on the platform version. It's still possible to pass the version if needed by users
- Removal of every docker.io mention in the code that's not generated
- Added a script to update the `controllers_cfg.yaml` and `SonataflowBuilder.dockerfile` files based on the env context from dependant images


packages/sonataflow-operator/hack/bump-version.sh
```diff
@@ -20,20 +20,10 @@
...
 set -e
 
-script_dir_path=$(dirname "${BASH_SOURCE[0]}")
-source "${script_dir_path}"/env.sh
-
 imageName=$(pnpm build-env sonataFlowOperator.registry)/$(pnpm build-env sonataFlowOperator.account)/$(pnpm build-env sonataFlowOperator.name)
 imageTag=$(pnpm build-env sonataFlowOperator.buildTag)
 version=$(pnpm build-env sonataFlowOperator.version)
 
-targetSonataflowBuilderImage=$(pnpm build-env sonataFlowOperator.sonataflowBuilderImage)
```
wmedvede, line 30: @ricardozanini If I don't remember wrong, this bump-version.sh script is executed as part of the automations executed during the release process, to adjust the image names to the version to be released etc.

If the image names setting is removed from this script to the make invoked python update_controllers_cfg.py, are you sure this is also executed as part of the automations?

Not saying this is bad, just double checking

packages/sonataflow-operator/internal/controller/cfg/controllers_cfg.go
```diff
@@ -45,36 +45,31 @@ var defaultControllersCfg = &ControllersCfg{
 	BuilderConfigMapName:          "sonataflow-operator-builder-config",
 }
 
-type GAV struct {
+type GroupArtifactId struct {
 	GroupId    string `yaml:"groupId,omitempty"`
 	ArtifactId string `yaml:"artifactId,omitempty"`
-	Version    string `yaml:"version,omitempty"`
```
wmedvede, line 51: OK, so the idea is that we remove the version from here, and at the time of executing run_quarkus_mvn_add_extension (bla, bla), the extension number will be taken from the kogito-version or the quarkus-version that corresponds to current product version (just built) accordingly, right?

And the mechanism works on the basis that, to have the version auto-filling, extensions must mandatory belong to gropups io.quarkus, org.kie, or "kogito".

This leaves out the possibility of configuring an extension that don't belong to any of these groups right? (at least in controllers_cfg.yaml)

Also, one thing that comes to my mind. With this new mechanism (in a given installation) it's not possible to configure the controllers config file to point to a different version, right?

Thinking about situations where you want to quickly do something like this:
1) something is not good
2) I quickly configure the controllers config ConfigMap with a different version ej 9.101.xxxxx, instead of 9.102.xxxxx 
3) restart the operator pod, And I can quickly see if things works with the previous version to test, etc.

wmedvede: Maybe we can just the data structure/processing as is, and keep the version field.
However, we don't set it anymore. So, if not set, the automatic mechanism will be applied. 
But we keep the trapdoor.

ricardozanini: @wmedvede I'd rather remove for good the possibility of adding the version since the platform will provide it. If we need to add the version to any other future library to the `controller_cfg`, we can then create a new data structure for it. But even in these cases, I'd rather have dependency control within the platform.

This hack you mentioned crossed my mind (change the version to check if the workaround works). In this case, one can use the build ARGS `QUARKUS_EXTENSIONS` directly to the `SonataFlowBuild`.

packages/sonataflow-operator/hack/bump-version.sh
ricardozanini: @wmedvede I've removed it here because we are bumping the new Python script.

This script runs during the build process, guaranteeing our versioning and image values. So, if the automation uses `pnpm build:prod` and follows the patterns we established in this repo, we should not have any problems. 

If we would, I'd blame @tiagobento 🤣

ricardozanini:
I'll rebase this PR after https://github.com/apache/incubator-kie-tools/pull/2818

jomarko:
FYI https://github.com/apache/incubator-kie-tools/pull/2813 was merged, probably causes conflicts to this PR.

ricardozanini:
@wmedvede @rgdoliveira can you please review?

rgdoliveira (approved):

wmedvede (approved):

Issue #2759, opened 2024-11-22
The configuration of Extended Services is not working when a path is included in the URL.


phungvannarich-kepler-aavn:
The error occurs when the port is missing, resulting in a redundant colon (':') in the URL:
![image](https://github.com/user-attachments/assets/220bb668-df43-4c82-8aeb-2ecd971f5f40)

The error occurs when the port is included twice in the URL, resulting in a duplicate entry:
![image](https://github.com/user-attachments/assets/73712c75-86e1-4d8e-831b-5fa63ffcaabb)

The version I am currently using is:
![image](https://github.com/user-attachments/assets/746218c6-8edf-458b-992f-893aadbe122c)

Could someone assist me in resolving this issue?

phungvannarich-kepler-aavn:
I made some adjustments to my Nginx configuration (for workaround), but it still returns an error:
![image](https://github.com/user-attachments/assets/05f52891-d0ef-4dde-a750-7cfe80901e2f)

I saw this message: 
![image](https://github.com/user-attachments/assets/d508c9ff-8095-44be-b709-498905341f3d)

The version I am currently using is: https://hub.docker.com/layers/apache/incubator-kie-sandbox-extended-services/main/images/sha256-42aaf500ed4c5396272f6de6f163ec74e54a028b5e4b439e37f2f858d33fa787?context=explore

jomarko:
@phungvannarich-kepler-aavn thank you for reporting this issue. I checked source code, and we didn't assume the case without the port, We will need to fix it.

jomarko:
@phungvannarich-kepler-aavn may I ask you to try as workaround, put the `host` value same as you did in screenshots, but add ending `/`, so for example in the fist screenshot put:
```
https://abc.com/xyz/
```

phungvannarich-kepler-aavn:
@jomarko Thanks for your response! The workaround worked perfectly. Will the fix be included in version 10.0.0?

jomarko:
Hi @phungvannarich-kepler-aavn @thiagoelg 

It seems we have three issues with the same root cause:
- https://github.com/apache/incubator-kie-tools/issues/2759
- https://github.com/apache/incubator-kie-tools/issues/2670
- https://github.com/apache/incubator-kie-tools/issues/3806

[...]

PR #2802, opened 2024-12-12
kie-issues#1694: Apache KIE Extended Services extension issues

### LINKED ISSUE #1684: Apache KIE Extended Services extension issues
Hello everyone,

After upgrading extension bundle to version 10.0.0 a new dependency extension named Apache KIE Extended Services introduced.

Since then, I have issues that occur on DMN file save warning me that DMN Validation failed as you can see below.

image

Could someone support on this please?

Thanks,
Vassilis

yesamer:
Fixes https://github.com/apache/incubator-kie-issues/issues/1684

In this PR:

The main issue is the "ambitious" management of the Extended Service Quarkus App lifecycle.
The original idea was to run the Quarkus application just after a DMN / BPMN file is opened in the editor and close it when all files are closed. Unfortunately, that implementation didn't consider the following scenarios:
- Validation checks are called when the Quarkus Application is starting up and not still ready
- Heartbeat requests are called in the same scenario described previously
- Close and opening a single DMN/BPMN file at the same moment leads the App to reach an inconsistent state, as the request of closing and starting are launched at the same moment

For that reason, I simplified the logic: the Extended Service Quarkus App is started when a DMN / BPMN file is opened for the first time and never stopped. The original idea was indeed better from a performance and resource consumption point of view, but it requires a more complex implementation.

- Updated node-fetch to 3.3.2
- `extendedServices.connectionHeartbeatIntervalinSecs` default value changed from 1 to 10. Having a heartbeat occurring every second contributed to the reported issue
- Introduced Debug logging
- Improved error message management
- Improved Configuration management
- Improved validation call body management, using the same terminology used in JIT (backend)


packages/extended-services-vscode-extension/src/configurations/Configuration.ts
```diff
@@ -19,9 +19,10 @@
 
 import * as vscode from "vscode";
 
-export const enableAutoRunID: string = "extendedServices.enableAutorun";
-export const connectionHeartbeatIntervalinSecsID: string = "extendedServices.connectionHeartbeatIntervalinSecs";
-export const extendedServicesURLID: string = "extendedServices.extendedServicesURL";
+export const enableAutoRunID = "extendedServices.enableAutorun";
+export const connectionHeartbeatIntervalinSecsID = "extendedServices.connectionHeartbeatIntervalinSecs";
+export const extendedServicesURLID = "extendedServices.extendedServicesURL";
+const defaultExtendedServicesURL = "http://localhost:21345";
```
ljmotta: We should use the `env` value instead of hardcoding. To do so, take a look into the `build/defaultEnvJson.ts` file in some packages (e.g. `online-editor`).

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
```diff
@@ -35,65 +36,30 @@ export class Configuration {
...
-    connectionHeartbeatIntervalinSecs = fetchConnectionHeartbeatIntervalinSecs();
-  } catch (error) {
-    errorMessages.push(error.message);
+const getConfigurationPropertyValue = <T>(property: string, defaultValue: T): T => {
+  const value = vscode.workspace.getConfiguration().get(property) as T;
+  if (!value) {
+    console.warn("Property: " + property + " is missing, using the default: " + defaultValue);
   }
+  return value || defaultValue;
+};
```
ljmotta: I'm not sure if I followed. `T` can be a `boolean`, `number` or `string`. If is a `boolean` it will always be `true`, as it is the default value. For `number`, can't be `0`, and `string` can't be `""`? Or `value` can't be `undefined`? The `(!value)` can be tricky, please, make it explicity.

packages/extended-services-vscode-extension/src/requests/ValidationRequests.ts
```diff
@@ -75,6 +74,14 @@ export async function validateBPMN(
...
     return validate(serviceURL, kieFile, "/jitbpmn/validate", validationresponse.parseBPMNValidationResponse);
   } catch (error) {
+    console.error(
+      "[Extended Services Extension] An error happened while trying to validate " +
+        kieFile +
+        " from serviceUrl " +
+        serviceURL +
+        ": " +
+        error
+    );
```
ljmotta: You could change to `bpmnFile` as you did in another file. :wink:

The same applies for the `validateDMN`.

ljmotta (commented):
@yesamer Thanks for your PR! I've made some comments inline.

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
yesamer: @ljmotta It's a good point. I would like to retrieve that default value directly from the configuration here https://github.com/apache/incubator-kie-tools/pull/2802/files#diff-153d43651b25ee3120020fae349c9e8ed99ccf90aa15995784702244decf9499R91 to avoid duplication and a single place for that property value, do you think is a valid alternative should I consider?

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
yesamer: > T can be a boolean, number or string. If is a boolean it will always be true, as it is the default value. For number, can't be 0, and string can't be ""

So, if the user explicitly removes the property the default value is always returned?

packages/extended-services-vscode-extension/src/requests/ValidationRequests.ts
yesamer: Done

packages/extended-services-vscode-extension/src/Validator.ts
```diff
@@ -49,26 +49,32 @@ function createDMNDiagnostics(validationResponses: validationResponse.DMNValidat
...
-      kieFile
+      bpmnFile
     );
     return createBPMNDiagnostics(validationResponses);
   } catch (error) {
-    throw new Error("VALIDATE BPMN ERROR - " + error.message);
+    console.error("An error happened while trying to validate BPMN file: " + kieFile + " with error: " + error.message);
+    throw new Error(
+      "An error happened while trying to validate BPMN file: " + kieFile + " with error: " + error.message
+    );
```
jomarko: shouldn't we use `bpmnFile` here?

packages/extended-services-vscode-extension/src/Validator.ts
```diff
@@ -49,26 +49,32 @@ function createDMNDiagnostics(validationResponses: validationResponse.DMNValidat
...
-      kieFile
+      dmnFile
     );
     return createDMNDiagnostics(validationResponses);
   } catch (error) {
-    throw new Error("VALIDATE DMN ERROR - " + error.message);
+    console.error("An error happened while trying to validate DMN file: " + kieFile + " with error: " + error.message);
+    throw new Error(
+      "An error happened while trying to validate DMN file: " + kieFile + " with error: " + error.message
+    );
```
jomarko: shouldn't we use `dmnFile` here?

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
jomarko: yes, it is important to have an option to configure the url of extended services, users report issues about it https://github.com/apache/incubator-kie-tools/issues/2759

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
```diff
@@ -35,65 +36,31 @@ export class Configuration {
...
-  }
-
-  try {
-    connectionHeartbeatIntervalinSecs = fetchConnectionHeartbeatIntervalinSecs();
-  } catch (error) {
-    errorMessages.push(error.message);
+const getConfigurationPropertyValue = <T>(property: string, defaultValue: T): T => {
+  const value = vscode.workspace.getConfiguration().get(property) as T;
+  if (value == null) {
+    console.warn("Property: " + property + " is missing, using the default: " + defaultValue);
```
jomarko: `warn` should support formatting placeholders:
```
console.warn("Property: %s is missing, using the default: %s", property, defaultValue);
```

maybe even:
```
console.warn("Property: '%s' is missing, using the default: '%s'", property, defaultValue);
```

packages/extended-services-vscode-extension/src/Validator.ts
```diff
@@ -49,26 +49,32 @@ function createDMNDiagnostics(validationResponses: validationResponse.DMNValidat
...
   try {
     const validationResponses: validationResponse.BPMNValidationResponse[] = await validationRequests.validateBPMN(
       serviceURL,
-      kieFile
+      bpmnFile
     );
     return createBPMNDiagnostics(validationResponses);
   } catch (error) {
-    throw new Error("VALIDATE BPMN ERROR - " + error.message);
+    console.error("An error happened while trying to validate BPMN file: " + kieFile + " with error: " + error.message);
```
jomarko: similar comment about formatting placeholders like `%s`

packages/extended-services-vscode-extension/src/Validator.ts
```diff
@@ -49,26 +49,32 @@ function createDMNDiagnostics(validationResponses: validationResponse.DMNValidat
...
   try {
     const validationResponses: validationResponse.DMNValidationResponse[] = await validationRequests.validateDMN(
       serviceURL,
-      kieFile
+      dmnFile
     );
     return createDMNDiagnostics(validationResponses);
   } catch (error) {
-    throw new Error("VALIDATE DMN ERROR - " + error.message);
+    console.error("An error happened while trying to validate DMN file: " + kieFile + " with error: " + error.message);
```
jomarko: similar comment about formatting placeholders `%s`

packages/extended-services-vscode-extension/src/extension/extension-browser.ts
```diff
@@ -67,43 +70,57 @@ function initializeVSCodeElements() {
...
 
 function startExtendedServices(): void {
+  console.debug("[Extended Services Extension] Starting Extended Service");
   try {
-    statusBarItem.show();
     configuration = fetchConfiguration();
   } catch (error) {
-    stopExtendedServices();
-    vscode.window.showErrorMessage("An error happened while trying to start the Extended Services: " + error.message);
+    console.error("[Extended Services Extension] Extension configuration is wrong: " + error.message);
```
jomarko: `%s` comment

jomarko: it is in whole PR, please let me know if you like that more, as it is usual shorter than manually concatenating string for the message

packages/extended-services-vscode-extension/src/extension/extension-main.ts
```diff
@@ -43,16 +43,19 @@ let configurationWatcher: ConfigurationWatcher;
 let connection: Connection;
 let localService: LocalExtendedServices;
 
-let userDisconnected: boolean = false;
+/* Determines if the extension is corrected with the Extended Services Backend */
```
jomarko:
```suggestion
/* Determines if the extension is connected with the Extended Services Backend */
```

jomarko (commented):
Thank you for the PR. I am happy we improve this extension.

### 01
Sometimes we use `extended service` (singletone) and sometimes we use `extended services` (plural).

### 02
The PR description says we never stop the extended services. Should then `extension-browser.ts` contain `const stopExtendedServicesCommandUID: string = "extended-services-vscode-extension.stopExtendedServices";` and related code? Or do we keep the option to manually stop the extended services?

No manual check done, waiting until code changes are finished.

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
ljmotta: @yesamer I just considered the calls to this method. We have just 3 on this file, each one of them with T equals to `boolean`, `number` or `string`. The `if` will be `true` for the cases I listed, and we will have the log. Additionally, the return will always go to the default value for the same cases:

```sh
> "" || "my_default_value"
'my_default_value'
> 0 || 25
25
> false || true
true
```

If the `value` (has the type `T` due your casting) can have a different type, please make it explicity. Something like: `T | null` or `T | undefined`.

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
ljmotta: I'm inclined to think it would be better to use template strings with the backtick:

```ts
`Property: ${property} is missing, using the default: ${defaultValue}` 
```

But this is a personal preference. As we don't have a standard for this any option is ok.

jomarko: I totally agree it is a personal preference. However, if there is an option, that has significant pros, like performance or security, maybe we should opt for it. Not saying there is such, just saying we have aspects that can help us to decide.

packages/extended-services-vscode-extension/src/Validator.ts
yesamer: done

packages/extended-services-vscode-extension/src/Validator.ts
yesamer: done

packages/extended-services-vscode-extension/src/extension/extension-main.ts
yesamer: Done

packages/extended-services-vscode-extension/src/Validator.ts
yesamer: Done

packages/extended-services-vscode-extension/src/Validator.ts
yesamer: Done

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
yesamer: Done, thank you both.

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
yesamer: @ljmotta In the examples I checked, the env variables are accessed using the _useEnv()_ hook, but in this case, we are not managing a React app. How can I access to env variable from "plain" typescript?

packages/extended-services-vscode-extension/src/extension/extension-browser.ts
yesamer: done

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
ljmotta: @yesamer That's true. the `defaultEnvJson` is used with the webpack transform + `fetch`, which isn't the case here. Now, you could use the webpack's `EnvironmentPlugin` to set env variables. The `online-editor` uses this approach to set some values inside the codebase. Take a look into the `online-editor/webpack.config.ts`, and search for `EnvironmentPlugin`. The `WEBPACK_REPLACE__*` values are the ones used in the codebase.

packages/extended-services-vscode-extension/env/index.js
ljmotta, line 1: @yesamer To set `env` variables that **must** have the same values of other package, please use the available composite feature. You can take a look into the `dmn-editor/env/index.js` file, where we load multiple env files.

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
yesamer: @ljmotta It should be done, there are two points to discuss:
- I added EnvironmentPlugin in both [extension-main](https://github.com/apache/incubator-kie-tools/pull/2802/files#diff-1768335dedd5cfa5e1b3830bb3b6c504f1d316dea21de55738df74f7f19c27a5R53) and [extension-browser](https://github.com/apache/incubator-kie-tools/pull/2802/files#diff-1768335dedd5cfa5e1b3830bb3b6c504f1d316dea21de55738df74f7f19c27a5R65). There's a way to define that once?
- Currently, the host and the post are parametrized, In my opinion, it would be better to have the entire URL as a parameter. E.g. http protocol is hardcoded, what if the user provides its own instance using https?

yesamer:
@jomarko 
01
I updated all the references to use the plural form

02
Yes, the user can close the connection using a button in the bottom bar of the editor.

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
ljmotta: @yesamer 
 - If you need to use in both, you can set the env variables in the `commonConfig` variable.
 - Good point. I'm not familiar on how the user set/use the Extended Services in the VS Code extension. Is it a embeded Extended Services or the user need to manually start the server? For the `extended-services-java` it will always use `http`. If the user wants to deploy it, the user will need to make the necessary adjustments to make it available through `https`.

yesamer: @ljmotta 
- Do you have an example for that? I mean, I added the plugin entry for both `web` and `node` https://github.com/apache/incubator-kie-tools/pull/2802/files#diff-1768335dedd5cfa5e1b3830bb3b6c504f1d316dea21de55738df74f7f19c27a5R52
To create a single entry for the plugin, which `target` and `entry` should I use?
- You're right, considering extended-services-java always uses `http`, we are safe here :)

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
```diff
@@ -19,81 +19,48 @@
...
-  } catch (error) {
-    errorMessages.push(error.message);
+const getConfigurationPropertyValue = <T>(property: string, defaultValue: T): T => {
+  const value: T | null = vscode.workspace.getConfiguration().get(property) as T;
+  if (value == null) {
+    console.warn(`Property: ${property} is missing, using the default: ${defaultValue}`);
+    value == defaultValue;
   }
+  return value;
+};
```
ljmotta: @yesamer I just have one comment related to this method, here is a suggestion of improvement. Also, I'm fixing two things here, the first double equals (`==`) to triple equals, and the second double equals should be an attribution.

```suggestion
// Place on top of the file or on another file
// I don't think this is a suitable name, it's just to exemplify
type Property = string | number | boolean

const getConfigurationPropertyValue = <T extends Property | null>(property: string, defaultValue: T): T => {
  const value = vscode.workspace.getConfiguration().get(property) as T;
  if (value === null) {
    console.warn(`Property: ${property} is missing, using the default: ${defaultValue}`);
    value = defaultValue;
  }
  return value;
};
```

jomarko:
### Closing the model file
Hi @yesamer or anyone, should closing the dmn/bpmn file to close also the message in problems tab? Currently it seems the problems tab is not cleared if model file is closed. Is that an expected behavior?
Screenshot 2025-01-03 132148

### Disable Autorun and use URL
Once I disabled the autorun and started extended services as `pnpm -F @kie-tools/extended-services-java start` I was not able to connect VSCode extended services extension to the extended services instance.
Screenshot 2025-01-03 133741

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
yesamer: @ljmotta Thank you, I committed your suggested change, I'll update the comment and the property name

yesamer:
@jomarko 
/cc @ljmotta @tiagobento 

I'm double-checking the behavior you reported. However, I'm not fully aware of the expected behavior of this feature, so my goal is to implement a minimal use-case scenario for this extension.

yesamer:
@jomarko I fixed the first use case you reported,  **Closing the model file**.
The second scenario, **Disable Autorun and use URL**, works correctly in my local

https://github.com/user-attachments/assets/4887c612-2ed4-4534-a32c-152f6efa45b6

ljmotta:
@jomarko I'm not sure if using `0.0.0.0` is correct here, please try with `localhost`.

@yesamer Please let me kwown when the PR is ready for another review.

yesamer:
@ljmotta Yes, please :)

ljmotta (approved):
@yesamer Thanks for your patience, LGTM!

tiagobento:
Deferring my review to @jomarko, feel free to merge when approved by him and green! Thanks for the great effort here.

packages/extended-services-vscode-extension/src/Validator.ts
```diff
@@ -49,26 +49,28 @@ function createDMNDiagnostics(validationResponses: validationResponse.DMNValidat
...
     const validationResponses: validationResponse.BPMNValidationResponse[] = await validationRequests.validateBPMN(
       serviceURL,
-      kieFile
+      bpmnFile
     );
     return createBPMNDiagnostics(validationResponses);
   } catch (error) {
-    throw new Error("VALIDATE BPMN ERROR - " + error.message);
+    console.error(`An error occured while trying to validate DMN file: ${bpmnFile} with error: ${error.message}`);
+    throw new Error(`An error occured while trying to validate DMN file: ${bpmnFile} with error: ${error.message}`);
```
jomarko:
```suggestion
    console.error(`An error occured while trying to validate BPMN file: ${bpmnFile} with error: ${error.message}`);
    throw new Error(`An error occured while trying to validate BPMN file: ${bpmnFile} with error: ${error.message}`);
```

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
```diff
@@ -19,81 +19,53 @@
...
-    errorMessages.push(error.message);
-  }
-
-  try {
-    connectionHeartbeatIntervalinSecs = fetchConnectionHeartbeatIntervalinSecs();
-  } catch (error) {
-    errorMessages.push(error.message);
+const getConfigurationPropertyValue = <T extends ConfigurationProperty | null>(
+  property: string,
+  defaultValue: T
```
jomarko, line 52: Why do we need to have also `defaultValue` parameter here? the three configurations related to this PR in `extended-services-vscode-extension/package.json`. Do we use `getConfigurationPropertyValue` also for loading configurations that do not have default value specified?

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
```diff
@@ -19,81 +19,53 @@
...
-    extendedServicesURL = fetchExtendedServicesURL();
-  } catch (error) {
-    errorMessages.push(error.message);
-  }
+export function fetchConfiguration(): Configuration {
+  const enableAutoRun = getConfigurationPropertyValue<boolean>(enableAutoRunID, true);
+  const connectionHeartbeatIntervalInSecs = getConfigurationPropertyValue<number>(
+    connectionHeartbeatIntervalInSecsID,
+    10
+  );
```
jomarko, line 67: related to question above, to me it seems that `true` and `10` and duplicated values that are already set in package.json what is vulnerable to get out of sync and then potentially make debugging and solving issues more difficult

packages/extended-services-vscode-extension/src/extension/extension-browser.ts
```diff
@@ -67,60 +70,71 @@ function initializeVSCodeElements() {
...
   const bpmnFiles: KieFile[] = await kiefilesfetcher.findActiveKieFiles([kiefilesfetcher.bpmnDocumentFilter]);
   const dmnFiles: KieFile[] = await kiefilesfetcher.findActiveKieFiles([kiefilesfetcher.dmnDocumentFilter]);
 
   for (const bpmnFile of bpmnFiles) {
     try {
-      const bpmnDiagnostics: vscode.Diagnostic[] = await validator.validateBPMN(
-        configuration.extendedServicesURL,
-        bpmnFile
-      );
+      console.debug(`[Extended Services Extension] Validating DMN file: ${bpmnFile.uri.path}`);
```
jomarko:
```suggestion
      console.debug(`[Extended Services Extension] Validating BPMN file: ${bpmnFile.uri.path}`);
```

packages/extended-services-vscode-extension/src/requests/PingRequest.ts
```diff
@@ -23,6 +23,7 @@ export async function ping(extendedServicesURL: URL): Promise<pingresponse.PingR
   const url = new URL("/ping", extendedServicesURL);
 
   try {
+    console.debug("[Extended Services Extension] Pinging: " + url.toString());
```
jomarko: maybe we could improve the variable name. what is your opinion about `extendedServicesPingURL`?

jomarko (commented):
Thank you @yesamer , I put some comments to code inline.

One question is also what is the difference of `extension-main.ts` and `extension-browser.ts` as their content seems to be very similar.

My biggest problem is there is no vsix file produced by the PR, or am I searching inappropriately? I tried to build the extension locally: `pnpm -F extended-services-vscode-extension... build:prod` however in that case it seems to not work properly, I would like to use vsix produced by the CI.
Screenshot 2025-01-16 075703

packages/extended-services-vscode-extension/src/requests/PingRequest.ts
yesamer: @jomarko Done

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
yesamer: @jomarko Those configuration parameters are loaded from the extension's setting, and these values can be accidentally erased by the user. Despite we could argue this is a low-probability scenario, I guess it's worth ensuring the extension can effectively work in most of the scenarios.

Screenshot 2025-01-16 at 08 13 40

Does make sense?

jomarko: ok, I understand now, thank you

packages/extended-services-vscode-extension/src/configurations/Configuration.ts
jomarko: resolved because of https://github.com/apache/incubator-kie-tools/pull/2802#discussion_r1917863297

yesamer: @jomarko  You're right, I had (and still have) the same concern, but honestly, I didn't find a solution to avoid this duplication. The ideal solution should be to retrieve the values from the package.json, I have to admit I didn't investigate if this is possible or not.

yesamer: @jomarko ^^ :)

jomarko (approved):
Thank you @yesamer. Using the VSIX produced by the CI I was able to:
- use extended-services extension in the auto run mode
- use extended-services extension connecting to services started in standalone terminal

Approving the PR, however opening two points that are not blocking from the merge.

### Extension info
Seems to show old information, probably it needs to be first pushed to marketplace? I do not know. Notice the ping interval value.
image

### Icon
Extended-services liveness icon indicator (bottom right corner) is shown from my point of view in 'difficult to understand' way. From my point of view, in separate ticket it would be worth to do a simplification. Here is my proposal, however not sure what is opinion of others.

Once extended-services extension is installed:
- we either show connection live icon
- or connection not live icon
- there should be no case icon is missing

yesamer:
@jomarko Thank you for the review, it's indeed strange the Extension info is not updated. I checked, and the default value is correct in the package.json.
Regarding your proposal with the ICON, I agree with you that should be revisited, together with other components.
We agreed with @tiagobento to invest more time on this extension in the future, based on our priorities, so it makes sense to keep track of all elements/behavior you already found during this PR review, and open a discussion of its requirements

PR #2804, opened 2024-12-12
NO-ISSUE: fix issue when using image digest in Operator Makefile

rgdoliveira:
When using a digest for the operator (USE_IMAGE_DIGESTS=true), the separator between the image name and the tag should be an `@` instead of a `:`


ricardozanini (approved):

domhanak (approved):
Thanks!

PR #2726, opened 2024-11-05
NO-ISSUE: Make examples better reflect their domain and contain more focused code

tiagobento:
In this PR I'm improving almost all examples to have a more descriptive README and package name. I also split the huge `webapp` example to make our builds faster, separating the `serverless-workflow-editor-on-webapp` example be built as part of `partition1.txt`.

In this PR:
- Split `examples/webapp` into several different examples, to make our builds faster and to have the examples have less boilerplate code, allowing users to focus on the content that matters for each individual component usage inside webapps:
  - `examples/bpmn-editor-classic-on-webapp`
  - `examples/dmn-editor-classic-on-webapp`
  - `examples/dmn-editor-on-webapp`
  - `examples/dmn-editor-standalone-on-webapp`
  - `examples/micro-frontends-multiplying-architecture-base64png-editor-on-webapp`
  - `examples/micro-frontends-multiplying-architecture-ping-pong-views-on-webapp`
  - `examples/micro-frontends-multiplying-architecture-todo-list-view-on-webapp`
  - `examples/serverless-workflow-editor-standalone-on-webapp`
- Renamed `examples/commit-message-validation-service` to `examples/kie-sandbox-commit-message-validation-service`
- Renamed `examples/jbpm-compact-architecture-example` to `examples/process-compact-architecture`
- Renamed `example/drools-process-usertasks-quarkus-example` to `examples/process-user-tasks-subsystem`
- Prefixed `examples/ping-pong*` and `examples/todo-list*` packages with `micro-frontends-multiplying-architecture`, to make it clear that those examples are related to Multiplying Architecture. The "Micro-frontends" part is to not have them be confused with Compact Architecture, which is a completely unrelated term used to signify the embedding of Data-Index, Jobs Service, and Data-Audit on Process applications.
- Revisited the READMEs of all examples
- Added a very short README to the `packages` and `examples` folder.

---

I know this is quite a lot, but there's not much difference in content. I recommend checking out this branch and looking at each of the changed examples individually, they're all quite small and simple, so shouldn't be a big effort to do it.

PR #2805, opened 2024-12-12
[NO-ISSUE] Adding some new dmn examples

LightGuard:
Adding some standalone dmn examples that can be used in either the web app or any of the standalone editors.

### DIFF
--- a/examples/dmn-editor-standalone-examples/flight-has-capacity.dmn
+++ b/examples/dmn-editor-standalone-examples/flight-has-capacity.dmn
@@ -1,4 +1,22 @@
 <?xml version="1.0" encoding="UTF-8" ?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~ 
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~ 
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.      
+-->
 <definitions
   xmlns="https://www.omg.org/spec/DMN/20230324/MODEL/"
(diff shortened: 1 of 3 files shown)


ljmotta:
@LightGuard Thanks for the PR. After #2726 we place the standalone example in the `examples/dmn-editor-standalone-on-webapp/static/models`. Them please update the `index.ts` of the designed type of example, as we have the `readOnly` (for read only examples), `stateControl` (for examples with the `undo/redo` buttons) and `withIncludedModels` (for dms that include other dms).

LightGuard:
I don't see anything in the index.ts that needs to be updated. I don't think we want to change which file is automatically loaded. Am I missing something?

ljmotta:
@LightGuard The `dmn-editor-standalone-on-webapp/src` has 3 folders: `readOnly`, `stateControl`, and `withIncludedModels`. Each one of them has an `index.ts` file that loads the respective DMNs. I don't know the purpose of these PR DMNs, but if they are DMNs that don't have included models, I'd put them in the `stateControl` category. At the end of the `src/stateControl/index.ts` file you will have an `initFileLoader` method with the DMNs that will be loaded, add the new DMNs there if that is the case. You will need to tweak the `index.html` to add the buttons for the new DMNs.

After your changes, please give it a try. The steps for building and starting are in the example README.

tiagobento:
This example's purpose is to showcase the DMN Editor Standalone API/usage. Having more models without showcasing features specific to the DMN Editor Standalone only adds noise/clutter to the example itself. We don't have a place for showcasing DMN features yet, but in any case, this example is not it.

LightGuard:
> @LightGuard The `dmn-editor-standalone-on-webapp/src` has 3 folders: `readOnly`, `stateControl`, and `withIncludedModels`. Each one of them has an `index.ts` file that loads the respective DMNs. I don't know the purpose of these PR DMNs, but if they are DMNs that don't have included models, I'd put them in the `stateControl` category. At the end of the `src/stateControl/index.ts` file you will have an `initFileLoader` method with the DMNs that will be loaded, add the new DMNs there if that is the case. You will need to tweak the `index.html` to add the buttons for the new DMNs.
> 
> After your changes, please give it a try. The steps for building and starting are in the example README.

Ah, got it. I was looking at the main index.ts file in src. That makes sense. Thanks!

LightGuard:
> This example's purpose is to showcase the DMN Editor Standalone API/usage. Having more models without showcasing features specific to the DMN Editor Standalone only adds noise/clutter to the example itself. We don't have a place for showcasing DMN features yet, but in any case, this example is not it.

Are you saying, this isn't where these examples should go?

ljmotta:
@tiagobento It makes sense.

@LightGuard Yes, it's not the correct place, and I incorrectly point it out to you. We don't have a place to put DMN files. It would be necessary to create a new example application. The only place I could think of is the `dmn-editor` Storybook's. We have the `useCases` section there.

PR #2814, opened 2024-12-16
kie-issues#1669: jBPM Quarkus DevUI seems to not update the # of items

### LINKED ISSUE #1669: jBPM Quarkus DevUI seems to not update the # of items
In the right top corner tab, the number of instances and tasks are not updated as you create new instances.

Steps to reproduce the issue:
1.Build incubator-kie-tools
2.Run the DevUI using 'pnpm start' and make sure DevUI is up in the link http://localhost:8080/kie/q/dev-ui/org.jbpm.jbpm-quarkus-devui/process-instances
Here, we see no process instances are found and 'Process Instances' on top right panel is also shows no instances.
3.Go to 'Process Definitions'. Start a process (Here, I am starting 'hiring' process clicking the play button near it.
4.Add the necessary input fields and click 'Start' to start the process.
5.We can see process instance created, but the count of process instance is now 1, which is not getting reflected in 'Process Instances' above

bncriju:
The wrong behavior is rectified by making the Process Instances, Tasks and Jobs to get dynamically updated, as and when user defines a process instance.

Closes https://github.com/apache/incubator-kie-issues/issues/1669

Please note that earlier pull request got auto closed while renamed the branch name to match the correct kie-issue and the comments in the previous pull request are already incorporated in this one.


packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/JBPMDevUIEventPublisher.java
```diff
@@ -0,0 +1,82 @@
...
+            case "ProcessInstanceStateDataEvent":
+                maybeRun(onProcessEvent);
+                break;
+            case "UserTaskInstanceStateDataEvent":
+                maybeRun(onTaskEvent);
+                break;
+            case "JobEvent":
+                maybeRun(onJobEvent);
+                break;
+            default:
```
pefernan: maybe we can get rid of the default

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/DataIndexCounter.java
```diff
@@ -0,0 +1,102 @@
...
+            vertxTimer = vertx.setPeriodic(500, id -> {
+                this.emit();
+            });
+            this.emit();
+        });
+        refreshCount();
+    }
+
+    public void refresh() {
+        vertx.setTimer(500, id -> {
```
pefernan: This timer looks like to short, in slow laptops (like main) this may not work, but I think we can go with it for now.

pefernan (changes requested):
@bncriju great work! I cannot get the counts initialized to `0` the first time I open Dev UI  for first time (they stay as `-`). Would you mind taking a look?

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/JBPMDevUIEventPublisher.java
bncriju: Done

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/DataIndexCounter.java
bncriju: Done and adjusted to 1000 instead of 500

bncriju:
Thanks for the comments @pefernan. Still cannot get the counts initialized to 0 the first time I open Dev UI for first time (they stay as -), but this works fine as expected when we open process-compact-architecture.

bncriju:
Done all the changes, and could you have a review once again

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/DataIndexCounter.java
```diff
@@ -0,0 +1,104 @@
...
+    public void stop() {
+        vertx.cancelTimer(vertxTimer);
+    }
+
+    private void emit() {
+        emitter.emit(count);
+    }
+
+    private void refreshCount() {
+        LOGGER.info("Refreshing data for query: {}", query);
```
pefernan: @bncriju I'd change this to `debug` to have the logs a bit more clean

pefernan (changes requested):
@bncriju great work, I'd just clean up a bit the logs to avoid printing info on every count refresh and that would be all.

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/DataIndexCounter.java
```diff
@@ -0,0 +1,104 @@
...
+    private String path;
+
+    public DataIndexCounter(String query, String graphField, WebClient dataIndexWebClient, String path) {
+        if (dataIndexWebClient == null) {
+            throw new IllegalArgumentException("dataIndexWebClient is null");
+        }
+        this.query = query;
+        this.field = graphField;
+        this.dataIndexWebClient = dataIndexWebClient;
+        this.path=path;
```
pefernan: I think this path is not needed, since it's always an empty string.

danielzhe:
Good job, @bncriju ! 

I just found some minor issues in the jbpm-dev-ui:
1. The Process Instances, Tasks and Jobs does not automatically update in the first run. I mean, if we have zero process and start a new one, the count just start if I refresh the page (pressing F5). After that, they update automatically without the need of refresh.

2. It looks like the counters doesn't decrement. I started 3 process, waited them to be finished and the counters still at 3. For the process instances it may be correct, since they are completed, but I'm not sure about the tasks. Even applying the filters, I can't see any task:
image

I don't know if this is an issue, but I think it doesn't looks correct. It can be an issue related to filters in tasks either. I'm not sure. What do you think @pefernan ?

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/DataIndexCounter.java
danielzhe: @pefernan In fact, it is not, specially in the `jbpm-dev-ui` where the default path is `http://0.0.0.0/kie`. Without this code, the `DataIndexCounter` fail [here](https://github.com/apache/incubator-kie-tools/pull/2814/files#diff-40c015a84a4818f22053e2fe1620d27260e24625fa029982b1b289f738e4a8e9R89), because it calls `http://0.0.0.0/graphql` instead of `http://0.0.0.0/kie/graphql`.

User can set a URL like `https://localhost:8080/some/long/path` for the setting `kogito.data-index.url`, so we need that path that is not part of the domain, because as you can see [here](https://github.com/apache/incubator-kie-tools/blob/be87d4b6bd7844cf175f74dd25ffc3bca4d19c59/packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/JBPMDevuiJsonRPCService.java#L81), the `WebClientOptions` only contains the **host**, port and if it is ssl or not, not the full path (subdomain).

bncriju: Agreeing with Daniel. This was necessary which is why it is given.

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/DataIndexCounter.java
bncriju: Done

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/src/main/java/org/jbpm/quarkus/devui/runtime/rpc/DataIndexCounter.java
pefernan: got it @danielzhe, I was missing that the the application may have a different context path configured...

pefernan:
@danielzhe yes, issue 1 is something I already found. Looks like the new subscriptions added to the counter Multi aren't updated with the new values.. I provided @bncriju a PR with a fix (https://github.com/bncriju/incubator-kie-tools/pull/2)

Regarding 2.. yes, that's a filter thing, we are currently showing the count of all elements (process instances, tasks or jobs) indexed in data-index. When we first created we wanted to show a snapshot of what was present on the system, but we can definitelly change it. In the end I think that it's just matter of deciding what we want to display there, I don't have a strong opinion.
IMO It would be cool to have different counters for each label and show different data, but at this point I think Quarkus doesn't allow that, maybe in future versions.

pefernan (approved):
great job @bncriju! I think we have it :D

bncriju:
thanks Pere and Daniel for all the inputs and suggestions on this. now, in good position to merge seeing the ci build status.

pefernan (approved):

danielzhe (approved):
Great job!

yesamer (approved):

yesamer:
Merging, the failing 'CI :: License headers' is unrelated and already fixed.

PR #2846, opened 2025-01-10
kie-issues#1636: DMN Editor throw an error opening Included Models tab

### LINKED ISSUE #1636: DMN Editor throw an error opening Included Models tab
image
Once user open a DMN model that includes another DMN model and click on the Included models tab (ie. there is a Card representing the included model), the attached error appears in the **browser console log**.

The **browser console log** is the only place where the issue can be spotted. User can still interact with the webapp, even this error is present in the **browser console log**.

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-issues/issues/1636 DMN Editor throw an error opening Included Models tab

### DIFF
--- a/packages/dmn-editor/src/includedModels/IncludedModels.css
+++ b/packages/dmn-editor/src/includedModels/IncludedModels.css
@@ -20,3 +20,9 @@
 .kie-dmn-editor--selected-model-to-include-error {
   color: var(--pf-global--danger-color--100);
 }
+.kebabtoggle-wrapper {
+  color: grey;
+}
+.kebabtoggle-wrapper:hover {
+  color: black;
+}
--- a/packages/dmn-editor/src/includedModels/IncludedModels.css
+++ b/packages/dmn-editor/src/includedModels/IncludedModels.css
@@ -20,9 +20,9 @@
 .kie-dmn-editor--selected-model-to-include-error {
   color: var(--pf-global--danger-color--100);
 }
-.kebabtoggle-wrapper {
+.kie-dmn-editor--model-card-kebabtoggle-wrapper {
   color: grey;
 }
-.kebabtoggle-wrapper:hover {
+.kie-dmn-editor--model-card-kebabtoggle-wrapper:hover {
   color: black;
 }
--- a/packages/dmn-editor/src/includedModels/IncludedModels.tsx
+++ b/packages/dmn-editor/src/includedModels/IncludedModels.tsx
@@ -612,7 +612,7 @@ function IncludedModelCard({
               shouldOpen={() => setRemovePopoverOpen(true)}
             >
               <div
-                className="kebabtoggle-wrapper"
+                className="kie-dmn-editor--model-card-kebabtoggle-wrapper"
                 onClick={(ev) => {
                   ev.stopPropagation();
                   ev.preventDefault();
--- a/packages/dmn-editor/src/includedModels/IncludedModels.tsx
+++ b/packages/dmn-editor/src/includedModels/IncludedModels.tsx
@@ -611,15 +611,14 @@ function IncludedModelCard({
               position={PopoverPosition.bottom}
               shouldOpen={() => setRemovePopoverOpen(true)}
             >
-              <Button
-                variant={"plain"}
+              <div
                 onClick={(ev) => {
                   ev.stopPropagation();
                   ev.preventDefault();
                 }}
               >
                 <KebabToggle />
-              </Button>
+              </div>
             </Popover>
           </CardActions>
         )}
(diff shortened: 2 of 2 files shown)


yesamer:
@Kusuma04-dev thank you for your PR!
Can you please share a screenshot of the change? 
I mean, what is the impact of changing that Button to a div?

Kusuma04-dev:
> @Kusuma04-dev thank you for your PR! Can you please share a screenshot of the change? I mean, what is the impact of changing that Button to a div?

Before fix: There is a console error when we open included models which is telling button inside button shouldn't be there .
Screenshot 2025-01-10 at 1 58 39 PM



After fix:
Screenshot 2025-01-10 at 2 27 43 PM

yesamer (commented):
Thank you @Kusuma04-dev .
I suspect that `Button` element was added just for styling purposes.
If you notice, in the UI the Kebab Element has a different style with your changes.

 Said so, your changes are correct from my point of view. if my above suspect is true, you can even totally remove that element, and not replace it with a div, unless we need that to stop the event propagation to the Kebab element. (to be checked)
 
 My only doubt is about the element's style. Should we preserve it?

jomarko:
I postpone my review until there is an answer for @yesamer comment.

Kusuma04-dev:
> Thank you @Kusuma04-dev . I suspect that `Button` element was added just for styling purposes. If you notice, in the UI the Kebab Element has a different style with your changes.
> 
> Said so, your changes are correct from my point of view. if my above suspect is true, you can even totally remove that element, and not replace it with a div, unless we need that to stop the event propagation to the Kebab element. (to be checked)
> 
> My only doubt is about the element's style. Should we preserve it?

Hi Yeser, Yeah i have kept div tag to stop event propogation ,also i don't see difference in the style ,as i have attached one ss with full screen and another with half screen , it might be looking different .

yesamer (approved):
@Kusuma04-dev Thank you for the clarification. The little difference I see in the style is on the Kebab component color: in the first screenshot it looks darker than the second one. However, it's just a minor difference, LGTM.

ljmotta:
@Kusuma04-dev As @yesamer pointed it out, previously, the buttom was gray (not black) and changed color on hover. I don't think it's a problem to remove the hover if we don't have another solution for it.

https://github.com/user-attachments/assets/2f334b38-87fb-4e1f-89d9-6c3a79786d3d

Kusuma04-dev:
> @Kusuma04-dev As @yesamer pointed it out, previously, the buttom was gray (not black) and changed color on hover. I don't think it's a problem to remove the hover if we don't have another solution for it.
> 
>  previous-kebab.mp4

Hi Luiz, i have added css styles to make it look like before.

packages/dmn-editor/src/includedModels/IncludedModels.css
```diff
@@ -20,3 +20,9 @@
 .kie-dmn-editor--selected-model-to-include-error {
   color: var(--pf-global--danger-color--100);
 }
+.kebabtoggle-wrapper {
+  color: grey;
+}
+.kebabtoggle-wrapper:hover {
+  color: black;
+}
```
jomarko:
```suggestion
.kie-dmn-editor--model-card-kebabtoggle-wrapper {
  color: grey;
}
.kie-dmn-editor--model-card-kebabtoggle-wrapper:hover {
  color: black;
}
```

packages/dmn-editor/src/includedModels/IncludedModels.tsx
```diff
@@ -611,15 +611,15 @@ function IncludedModelCard({
               position={PopoverPosition.bottom}
               shouldOpen={() => setRemovePopoverOpen(true)}
             >
-              <Button
-                variant={"plain"}
+              <div
+                className="kebabtoggle-wrapper"
```
jomarko:
```suggestion
                className="kie-dmn-editor--model-card-kebabtoggle-wrapper"
```

jomarko (commented):
Thank you for the PR. I can confirm there is no error in the browser console log 🥳 .

The only suggestion I have is to keep kie-tools css classes naming. We prefix custom classes in a way `kie-[something like package]--` and then the new class is `kie-[something like package]--[actual new name]`. Maybe other colleagues will have better suggestion for new css names.

Good job!

packages/dmn-editor/src/includedModels/IncludedModels.tsx
Kusuma04-dev: Done @jomarko ,Thank you.

jomarko (approved):
Thank you!

klbowers (approved):
Thanks Kusuma, LGTM!

PR #2849, opened 2025-01-10
kie-issues#1741: Sandbox: Add “Open Boxed Expression Editor” button in the DMN Runner table output columns

### LINKED ISSUE #1741: Sandbox: Add “Open Boxed Expression Editor” button in the DMN Runner table output columns
Create an "open in Boxed Expression Editor" button in the DMN Runner form output card that will open the expression of the respective evaluated Decision node. Figure 1 shows an example of the final component. For a nested output column (an output with a complex ItemDefinition), the button should be placed in the cell with the Decision node name as shown in Figure 2. Notice that the green border is to emphasize the button location and is not part of the final design. The button should use the same icon as the "Open Data Type" present in the Data Type component. (Figure 3)
image-2024-10-09-19-05-09-465
image-2024-10-09-19-10-48-113
image-2024-10-09-19-11-48-054

jomarko:
Closes apache/incubator-kie-issues#1741

The dmn runner outputs table introduces a button in the header cell top right corner for navigating into the bound expression. It means the boxed expression editor is open for the given dmn runner result.

Screenshot 2025-01-27 122459


jomarko:
fixing column resize feature 
image

jomarko:
@tiagobento @kbowers-ibm @ljmotta or anyone else, I will write down my analysis of the problem https://github.com/apache/incubator-kie-tools/pull/2849#issuecomment-2618631310, it is the thing that runner output columns are not resizable, there is missing the resize handler. It was introduced by changes regarding `readOnly` mode https://github.com/apache/incubator-kie-tools/pull/2538/files#diff-ab6fc87d76da643cd7a7f3cd903f077619837939303defc50b2d0e208a2d5ca5. 

The missing resize handler in dmn runner outputs columns is a problem because of this two things
#### Too long header text
Makes header unreadable and in combination with new arrow up button quite 'messy'.
image

#### Too long data text
Makes data unreadable
image


So here are some options I see. Please comment them or bring other so we can decide what option is the best.

### 01 'Weak' ReadOnly mode
We could change the `isReadOnly` of `StandaloneBeeTable` to be slightly 'weak'. I mean, we could allow to display resize handler even if `isReadOnly` is set to `true`

**Pros**
- it would fix the issue of unreadable header or data if the text is too long

**Cons**
- it decreases the level 'readonly' feeling. for example, readonly in google spreadsheets does not allow to resize column, if I am not wrong
- it affects more places, not just dmn runner outputs. for example opened included decision node

### 02 'allowResizing' property
We could introduce additional property `allowResizing` next to `isReadOnly` in `StandaloneBeeTable`

**Pros**
- we would be able to fix just the dmn runner outputs scenario without an effect on other `StandaloneBeeTable` usages

**Cons**
- it increases `StandaloneBeeTable` props complexity

### 03 'width' computation
We could reuse the column width computation for given text as we have here: https://github.com/apache/incubator-kie-tools/blob/main/packages/dmn-editor/src/boxedExpressions/getDefaultColumnWidth.tsx directly in `DmnRunnerOutputsTable` when we define `beeTableColumns` variable.

**Pros**
- We do not need to touch `StandaloneBeeTable` code
- We keep strong readonly 'feeling' on all `StandaloneBeeTable` usages

**Cons**
- we fix only the scenario of a long text in the header, if a long text is in data cell, it still remains unreadable

klbowers:
Thanks for the detailed analysis @jomarko - you've clearly put a lot of thought and effort into this.
My vote would be for Option 3 as I think it would probably be part of any refactoring efforts on the DMN Runner table anyway. This then allows you to continue with your task, and while the con is an issue, it's not like it's introduced by/related to your PR. I think best would be to fix the header as you describe and raise the issue of unreadable long text in data cell as tech debt if it's not already an existing issue.

packages/online-editor/src/dmnRunner/DmnRunnerTable.tsx
```diff
@@ -137,6 +144,18 @@ export function DmnRunnerTable() {
                           i18n={i18n.dmnRunner.table}
                           jsonSchemaBridge={jsonSchemaBridge}
                           results={results}
+                          openBoxedExpressionEditor={
+                            !isLegacyDmnEditor
+                              ? (nodeId: string) => {
+                                  const newDmnEditorEnvelopeApi = props.editor?.getEnvelopeServer()
+                                    .envelopeApi as unknown as MessageBusClientApi<NewDmnEditorEnvelopeApi>;
```
ljmotta: Why the `as unknown` is required?

packages/online-editor/src/dmnRunner/DmnRunnerTable.tsx
```diff
@@ -32,8 +32,12 @@ import setObjectValueByPath from "lodash/set";
 import cloneDeep from "lodash/cloneDeep";
 import { DmnRunnerProviderActionType } from "./DmnRunnerTypes";
 import { DmnRunnerExtendedServicesError } from "./DmnRunnerContextProvider";
+import { MessageBusClientApi } from "@kie-tools-core/envelope-bus/dist/api";
+import { NewDmnEditorEnvelopeApi } from "@kie-tools/dmn-editor-envelope/dist/NewDmnEditorEnvelopeApi";
+import { EmbeddedEditorRef } from "@kie-tools-core/editor/dist/embedded";
+import { useSettings } from "../settings/SettingsContext";
 
-export function DmnRunnerTable() {
+export function DmnRunnerTable(props: { editor: EmbeddedEditorRef | undefined }) {
```
ljmotta: I'm don't think we want to have an `undefined` `editor` as props of the `DmnRunnerTable`. Also, the `getEnvelopeServer()` is the only property of `editor` that is used, which could be accordlying typed to avoid casting.

packages/online-editor/src/editor/EditorPage.tsx
```diff
@@ -468,6 +468,7 @@ Error details: ${err}`);
                 workspaceFile={file.workspaceFile}
                 workspaces={workspaces}
                 dmnLanguageService={dmnLanguageService}
+                editor={editor}
```
ljmotta: As mentioned, we could pass the `editorEnvelopeServer`.

packages/online-editor/src/editor/EditorPageDockContextProvider.tsx
```diff
@@ -186,7 +189,7 @@ export function EditorPageDockContextProvider({
       case PanelId.DMN_RUNNER_TABLE:
         return (
           <DmnRunnerErrorBoundary>
-            <DmnRunnerTable />
+            <DmnRunnerTable editor={editor} />
```
ljmotta: Same.

packages/boxed-expression-component/src/BoxedExpressionEditorContext.css
```diff
@@ -99,3 +99,9 @@
   font-size: smaller;
   color: #1b515f;
 }
+
+.kie-tools--bee--header-cell-element-extension {
```
ljmotta: I don't want to be picky here, but for this particular case, we don't use the `kie-tools--bee` prefix. For the `boxed-expression-component` we are using multiple classes.

packages/unitables-dmn/src/DmnRunnerOutputsTable.tsx
```diff
@@ -228,8 +246,33 @@ function OutputsBeeTable({ id, i18n, outputsPropertiesMap, results, scrollablePa
...
+  const onOpenBoxedExpressionHeaderButtonClick = useCallback(
+    (clickedDecisionId: string) => {
+      (results?.[0] ?? []).flatMap(({ decisionId: resultDecisionId }) => {
+        if (clickedDecisionId === resultDecisionId) {
+          openBoxedExpressionEditor?.(resultDecisionId);
+        }
+      });
+    },
+    [openBoxedExpressionEditor, results]
+  );
```
ljmotta: Well, everytime we click on the button we will search the array to check if `decisionId` is part of the `results`. Why not save the `result` `decisionIds` in a `set`?

```
const resultsDecisionIds = useMemo(() => results?.[0]?.reduce((set, result) => {
  set.add(result.decisionId);
  return set;
}, new Set()), [results]);

... 
onClick={() => resultsDecisionIds.has(decisionId) && openBoxedExpressionEditor?.(decisionId)}
```

(I didn't test the code above)

ljmotta (changes requested):
@jomarko Thanks for the PR, I've made some comments inline. Please let me know if you have any questions or need further explanation.

ljmotta:
@jomarko I agree with @kbowers-ibm . Both option 1 and 2 would require to persist the column size, or closing and opening the DMN Runner Table would reset to the original size.

packages/unitables-dmn/src/DmnRunnerOutputsTable.tsx
jomarko: cool, I like this incorporated

packages/boxed-expression-component/src/BoxedExpressionEditorContext.css
jomarko: I am sorry I used wrong prefix, as you mentioned and I noticed, we use multiple prefixes, and I am sorry, I didn't figure out from the codebase which prefix should I use in my PR. 

Is `header-cell-element-extension` fine for you? I see we have `header-cell` or `header-cell-info` already in use.

packages/online-editor/src/editor/EditorPageDockContextProvider.tsx
jomarko: done

packages/online-editor/src/editor/EditorPage.tsx
jomarko: done

packages/online-editor/src/dmnRunner/DmnRunnerTable.tsx
jomarko: done

packages/online-editor/src/dmnRunner/DmnRunnerTable.tsx
jomarko: done

jomarko:
### Column width checks
- [x] Single context entry 
- [x] Single context entry header longer 
- [x] Multiple context entries
- [x] Multiple context entries header longer 
- [x] Single Structure entry
- [x] Single Structure entry header longer 
- [x] Multiple Structure entries
- [x] Multiple Structure entries header longer 
- [x] Single List entry
- [x] Multiple List entries
- [x] built in type short name
- [x] built in type long name

packages/unitables-dmn/src/DmnRunnerOutputsTable.tsx
```diff
@@ -228,8 +250,38 @@ function OutputsBeeTable({ id, i18n, outputsPropertiesMap, results, scrollablePa
...
+      );
+    },
+    [openBoxedExpressionEditor, resultsDecisionIds]
+  );
+
+  /**
+   * Wrapping 'getDefaultColumnWidth' to add additional space for the '<ArrowUp />' icon in the runner outputs columns.
+   */
+  const getDefaultDmnRunnerOutputColumnWidth = useCallback((label, dataType) => {
+    return getDefaultColumnWidth({ name: label, typeRef: dataType }) + 100;
```
tiagobento: Please extract `100` to a nice constant and give it a long and descriptive name 😄

jomarko: @tiagobento good feedback, thank you. done

klbowers (approved):
This looks great - thanks for the changes to the widths Jozef! This is a much better solution than adding back resizing, as the bug with column sizes not persisting means it would still be messy. Whereas this is simple, effective and looks very clean 🙇

packages/unitables-dmn/src/DmnRunnerOutputsTable.tsx
```diff
@@ -173,8 +193,10 @@ function OutputsBeeTable({ id, i18n, outputsPropertiesMap, results, scrollablePa
           dataType,
           isRowIndexColumn: false,
           groupType: "dmn-runner-output",
-          width: DMN_RUNNER_OUTPUT_COLUMN_MIN_WIDTH,
-          minWidth: DMN_RUNNER_OUTPUT_COLUMN_MIN_WIDTH,
+          minWidth: Math.max(
+            (parentColumnMinWidth ?? columnMinWidth) / Object.entries(myObject).length,
+            columnMinWidth
+          ),
```
tiagobento, line 208: Missing hooks dependencies.

jomarko: @tiagobento thank you for spotting, changes pushed

thiagoelg (approved):
LGTM

ljmotta (approved):

PR #2851, opened 2025-01-13
kie-issues#1635: DMN Editor can not reopen List expression

### LINKED ISSUE #1635: DMN Editor can not reopen List expression
image
Screenshot 2024-11-26 094525

To reproduce the issue please follow the steps:
1. open new DMN model
2. add a decision node
3. define the decision node as List expression
4. close the DMN model
5. reopen the node
6. try to open the decision node that is internally defined as List expression

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-issues/issues/1635 DMN Editor can not reopen List expression
when we have created list expression and go back to home screen and opened dmn again then we are getting below error(not able to reopen the diagram) :
Before fix:
Screenshot 2025-01-13 at 11 35 57 AM

After fix:
Screenshot 2025-01-13 at 11 36 11 AM


yesamer (changes requested):
Thank you @Kusuma04-dev for your PR!

IINW, with this change the logic will automatically nest a List Expression inside the created List Expression, as shown in the below image

Screenshot 2025-01-13 at 08 43 12

This is not the expected behavior, we should keep the expression undefined so the user can manually assign the expression to set there.

Based on that, I think you should revert the change and continue your investigation, the bug is in another place, in my opinion.

Kusuma04-dev:
> Thank you @Kusuma04-dev for your PR!
> 
> IINW, with this change the logic will automatically nest a List Expression inside the created List Expression, as shown in the below image
> 
> Screenshot 2025-01-13 at 08 43 12
> This is not the expected behavior, we should keep the expression undefined so the user can manually assign the expression to set there.
> 
> Based on that, I think you should revert the change and continue your investigation, the bug is in another place, in my opinion.

Hi Yeser,Got it will investigate further.

jomarko (commented):
I think we need to update also this file https://github.com/apache/incubator-kie-tools/blob/49f1cd8dd201eb574ef41588e509e5e73a2d45b0/packages/boxed-expression-component/stories/boxedExpressions/List/List.stories.tsx#L45

Kusuma04-dev:
> I think we need to update also this file
> 
> https://github.com/apache/incubator-kie-tools/blob/49f1cd8dd201eb574ef41588e509e5e73a2d45b0/packages/boxed-expression-component/stories/boxedExpressions/List/List.stories.tsx#L45

Yes Jozef , Updated.Thank you.

jomarko:
@Kusuma04-dev I can confirm the issue is fixed. I can start modeling, define decision as List, close and reopen and continue modeling. However, the test failure seems to be related. Please let us know if you need a help with fixing it.

Kusuma04-dev:
> @Kusuma04-dev I can confirm the issue is fixed. I can start modeling, define decision as List, close and reopen and continue modeling. However, the test failure seems to be related. Please let us know if you need a help with fixing it.

Hi @jomarko ,Thanks for the confirmation,I have fixed the tests, can you please check.

yesamer (approved):
Thank you @Kusuma04-dev

jomarko (commented):
Thank you @Kusuma04-dev . I think I do not understand properly the fix of the test.

if this is the state after opening boxed list (notice number of rows, not name and data type):
Screenshot 2025-01-31 080828

why do we need to click add row 5 times?
`await bee.expression.asList().addEntryAtTop();`


that would result in list with 6 rows, or am I wrong? while `boxed-list-age-group.png` has 5 rows
boxed-list-age-group

tiagobento:
As discussed, looks like the `List` expression is the only expression type affected by this problem, with all others being able to handle `undefined` expressions normally. My recommendation is that you take a look at what other expression types do and replicate it for `List`. cc @danielzhe

jomarko (approved):
I haven't found an issue during manual re-check. Thank you for the PR.

tiagobento:
Love the simplicity of the fix!! Thanks for the patience @Kusuma04-dev @danielzhe!

tiagobento (approved):

PR #2860, opened 2025-01-16
kie-issues#1552: Test Scenario Editor: Update `@kie-tools/dmn-vscode-extension` to add the new Test Scenario Editor

### LINKED ISSUE #1552: Test Scenario Editor: Update `@kie-tools/dmn-vscode-extension` to add the new Test Scenario Editor
The new test scenario editor should be integrated into the dmn-editor extension.

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/1552

In this PR we finally introduce the first "alpha" version of the new Test Scenario Editor in our dmn-vscode extension.
The editor is not still completed, but I believe it's important to have this get merged to simply the future testing phase, and the development of the missing features (eg. https://github.com/apache/incubator-kie-issues/issues/1514)

https://github.com/user-attachments/assets/43b8d834-35db-4187-addb-568320203569

Don't expect any other feature change or update in this PR.
It's worth to mention I applied some changes on the CSS side to make the editor usable in VSCode.


packages/scesim-editor/src/TestScenarioEditor.tsx
```diff
@@ -395,7 +396,7 @@ export const TestScenarioEditorInternal = ({
   console.trace("[TestScenarioEditorInternal] File Status: " + TestScenarioFileStatus[scesimFileStatus]);
 
   return (
-    <div ref={testScenarioEditorRootElementRef}>
+    <div ref={testScenarioEditorRootElementRef} data-testid="kie-tools--scesim-editor--container">
```
jomarko: This is my mistake. I am sorry. Looking on the rest of the codebase, I think the value we should use should be probably something like: `kie-scesim-editor--container` or `kie-scesim-editor-container` or `kie-scesim-editor`

packages/scesim-editor/tests-e2e/__fixtures__/editor.ts
```diff
@@ -61,6 +61,6 @@ export class Editor {
   }
 
   public get() {
-    return this.page.getByTestId("test-scenario-editor");
+    return this.page.getByTestId("kie-tools--scesim-editor--container");
```
jomarko: This is my mistake. I am sorry. Looking on the rest of the codebase, I think the value we should use should be probably something like: `kie-scesim-editor--container` or `kie-scesim-editor-container` or `kie-scesim-editor`

jomarko (commented):
Thank you for the PR. I did only very basic sanity test. I will do more proper one once https://github.com/apache/incubator-kie-issues/issues/970 is implemented.

One observation is that extension homepage should be updated probably:
image


Second is the DMN Namespace mocked as `https:\kiegroup`, we should use proper http or https format and not mention kiegroup anymore

packages/scesim-editor/tests-e2e/__fixtures__/editor.ts
yesamer: Fixed

packages/scesim-editor/src/TestScenarioEditor.tsx
yesamer: Fixed

yesamer:
Thank you @jomarko, I updated the extension homepage to better reflect the current status. The gifs are both targeting the old editors, I guess we can replace them the moment we set the new editors as the default.
Removed mocked reference to kiegroup.

jomarko (approved):
Thank you @yesamer

tiagobento (approved):

PR #2863, opened 2025-01-16
[kie-issues#1755] - Display the correct git URL when creating a snippet, gist, or repo in the online editor

### LINKED ISSUE #1755: Incorrect git url in the alert when adding a gist/snippet/repo
Creating a gist and clicking on the link resulted in a URL ending with .git and gave me this error, removing the .git at the end gets you to the correct URL:

Image

From @ljmotta:
"The Gist on the default GitHub is generated with the .git as well, but GitHub knows how to resolve this URL. It looks like GitHub Enterprise doesn't. Might be a good idea to remove the .git from the GitHub provider."

LightGuard:
Closes: https://github.com/apache/incubator-kie-issues/issues/1755

* Send the correct URL to the alert
* Call the correct alert
* Use the provided URL if available

The code has been using `successfullyUpdatedGistOrSnippetAlert` instead of `successfullyCreatedGistOrSnippetAlert`. The alert was also not setup to properly make use of the provided URL when the alert was shown.

### DIFF
--- a/packages/online-editor/src/editor/Toolbar/GitIntegration/GitIntegrationAlerts.tsx
+++ b/packages/online-editor/src/editor/Toolbar/GitIntegration/GitIntegrationAlerts.tsx
@@ -225,12 +225,12 @@ export function useGitIntegrationAlerts(workspace: ActiveWorkspace) {
 
   const successfullyCreatedGistOrSnippetAlert = useGlobalAlert(
     useCallback(
-      ({ close }) => {
+      ({ close }, staticArgs?: { url: string }) => {
         if (!isGistLikeWorkspaceKind(workspace.descriptor.origin.kind)) {
           return <></>;
         }
 
-        const gistOrSnippetUrl = workspace.descriptor.origin.url;
+        const gistOrSnippetUrl = staticArgs?.url || workspace.descriptor.origin.url;
         return (
           <Alert
             variant="success"
--- a/packages/online-editor/src/editor/Toolbar/GitIntegration/CreateGistOrSnippetModal.tsx
+++ b/packages/online-editor/src/editor/Toolbar/GitIntegration/CreateGistOrSnippetModal.tsx
@@ -75,7 +75,7 @@ export const CreateGistOrSnippetModal = (props: {
   const [isGistOrSnippetLoading, setGistOrSnippetLoading] = useState(false);
 
   const {
-    alerts: { successfullyUpdatedGistOrSnippetAlert, errorAlert },
+    alerts: { successfullyCreatedGistOrSnippetAlert, errorAlert },
   } = useGitIntegration();
 
   const createGitHubGist: () => Promise<CreateGistOrSnippetResponse> = useCallback(async () => {
@@ -132,7 +132,7 @@ If you are, it means that creating this Snippet failed and it can safely be dele
       return (e.name = "https" && e.href.startsWith("https"));
     })[0].href;
 
-    return { cloneUrl, htmlUrl: json.links.html };
+    return { cloneUrl, htmlUrl: json.links.html.href };
   }, [bitbucketClient, env.KIE_SANDBOX_APP_NAME, isPrivate, props.workspace.name, selectedOrganization]);
 
   const createGistOrSnippet = useCallback(async () => {
@@ -214,8 +214,8 @@ If you are, it means that creating this Snippet failed and it can safely be dele
       });
 
       props.onClose();
-      successfullyUpdatedGistOrSnippetAlert.show({ url: gistOrSnippet.cloneUrl });
-      props.onSuccess?.({ url: gistOrSnippet.cloneUrl });
+      props.onSuccess?.({ url: gistOrSnippet.htmlUrl });
+      successfullyCreatedGistOrSnippetAlert.show({ url: gistOrSnippet.htmlUrl });
       return;
     } catch (err) {
       setError(err);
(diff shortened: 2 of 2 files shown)


jomarko (commented):
Thank you for the PR. I can confirm that creating of:
- github gist
- github repository
- bitbucket snippet
- bitbucket repository

does not include `.git` substring in the alert message shown after the operation is done.

However, the 'update' and 'push' operation for mentioned 4 cases still include `.git` substring. Should we handle also 'update' and 'push' operations?

packages/online-editor/src/editor/Toolbar/GitIntegration/GitIntegrationAlerts.tsx
```diff
@@ -225,12 +225,12 @@ export function useGitIntegrationAlerts(workspace: ActiveWorkspace) {
 
   const successfullyCreatedGistOrSnippetAlert = useGlobalAlert(
     useCallback(
-      ({ close }) => {
+      ({ close }, staticArgs?: { url: string }) => {
```
thiagoelg, line 228: Could you add this to all gist related alerts here?

LightGuard:
So.... turns out after much digging and debugging, simply adding the staticArgs doesn't solve the problem, and we'd actually need to change the WorkspaceOrigin definition to store the htmlUrl the first time we have it. This in turn would actually cause problems for all those instances currently in use and would upgrade later, potentially causing problems the next time they edit a model (I think, honestly, that's a bit of conjecture on my part). 

Regardless, this PR does actually solve the issue in the ticket, and further enhancement, really should go into a separate ticket.

jomarko:
@LightGuard hi, I am not author of the codebase we speak about, so my comment may be wrong, however here is my opinion.

### `WorkspaceOrigin` 
I understand this part of code as holder of data that we want to display later. Maybe this code can be improved somehow, not sure. I think to decide it, we need to get an answer for the paragraph below.

NOTE: I didn't check the relation of `WorkspaceOrigin` and `workspace.descriptor.origin.url` I just trust you there is such.

### `UI`
I see we use `workspace.descriptor.origin.url` in four places. Right?
- `VsCodeDropdownMenu`
- `GitIntegrationAlerts`
- `GitIntegrationContextProvider`
- `EditorToolbar`

If each of these four places has **different** logic, how to display `workspace.descriptor.origin.url` (e.g. somewhere we want `.git` suffix, somewhere we don't, somewhere we want protocol prefix, somewhere we don't ....) then I do not think there is an easy way to update `WorkspaceOrigin` to satisfy all usages of it.

If each of these four places has **same** logic, or should have same logic, how to display `workspace.descriptor.origin.url`, then probably it is worth to update `WorkspaceOrigin`. From the existing code it seems on a lot of `UI` places we remove the `.git` suffix, however, probably @thiagoelg could help with the answer, if that should be like that across all `UI` codebase.

thiagoelg:
> @LightGuard hi, I am not author of the codebase we speak about, so my comment may be wrong, however here is my opinion.
> ### `WorkspaceOrigin`
> 
> I understand this part of code as holder of data that we want to display later. Maybe this code can be improved somehow, not sure. I think to decide it, we need to get an answer for the paragraph below.
> 
> NOTE: I didn't check the relation of `WorkspaceOrigin` and `workspace.descriptor.origin.url` I just trust you there is such.
> ### `UI`
> 
> I see we use `workspace.descriptor.origin.url` in four places. Right?
> 
>     * `VsCodeDropdownMenu`
> 
>     * `GitIntegrationAlerts`
> 
>     * `GitIntegrationContextProvider`
> 
>     * `EditorToolbar`
> 
> 
> If each of these four places has **different** logic, how to display `workspace.descriptor.origin.url` (e.g. somewhere we want `.git` suffix, somewhere we don't, somewhere we want protocol prefix, somewhere we don't ....) then I do not think there is an easy way to update `WorkspaceOrigin` to satisfy all usages of it.
> 
> If each of these four places has **same** logic, or should have same logic, how to display `workspace.descriptor.origin.url`, then probably it is worth to update `WorkspaceOrigin`. From the existing code it seems on a lot of `UI` places we remove the `.git` suffix, however, probably @thiagoelg could help with the answer, if that should be like that across all `UI` codebase.

When pushing to a git repository or gist, we want the `.git` suffix, but when linking to a user, we don't. Let's handle these cases individually.
As I mentioned to @LightGuard earlier, changing how we use and implement the `WorkspaceOrigin` model could break existing workspaces for our users without a proper data migration to the IndexedDB stored workspaces.

LightGuard:
Instead of changing the logic, I thin it would be better to add another property, if possible: htmlUrl. That way we keep the reference to the .git folder (as I think it is used in other places for things like push, pull, branch, etc.) However, I think showing the HTML URL for the browser would be helpful so the user can easily access the repo in the browser.

tiagobento:
@LightGuard Adding another property would have to account for everyone that already has a Gist workspace cloned when they refresh their browser tab to a new version of KIE Sandbox.

LightGuard:
> @LightGuard Adding another property would have to account for everyone that already has a Gist workspace cloned when they refresh their browser tab to a new version of KIE Sandbox.

Right, which may not be that easy to do. We'd have to make a request to the git service and find the link, or similar? Unfortunately, some of the alerts and modals don't send the repo link when the alert/modal is shown and we assume that info is already in the workspace state.

LightGuard:
How do we want to proceed with this?

thiagoelg:
If this PR solves the initial bug described in the issue I think it's OK for now

thiagoelg (approved):

LightGuard:
Are we good to merge in that case?

jomarko (approved):
Since my review https://github.com/apache/incubator-kie-tools/pull/2863#pullrequestreview-2558206267 there was no code update pushed.

The discussion outcome is not clear to me. I understand that handling all git url cases and how we display the url in sandbox is much broader task. As the PR fixes at least some scenario, I am approving the PR.

PR #2866, opened 2025-01-21
kie-issues#1114: Containerize Playwright end-to-end tests to resolve screenshot comparison issues caused by OS differences

### LINKED ISSUE #1114: Containerize Playwright end-to-end tests to resolve screenshot comparison issues caused by OS differences
Currently, some Playwright tests use screenshot comparisons, which are prone to failure due to operating system differences, such as variations in fonts and UI rendering [1][2][3]. These inconsistencies lead to unreliable test results across different environments.

To address this issue, we need to standardize how tests are executed within the kie-tools monorepo. By leveraging containers, we can create a stable, consistent environment for running tests, independent of the host OS. This approach ensures reproducibility and resolves screenshot comparison failures.

[1] https://github.com/microsoft/playwright/issues/8161
[2] https://github.com/microsoft/playwright/issues/16104
[3] https://github.com/microsoft/playwright/issues/6907

### LINKED ISSUE #2839: Help installing Playwright with Webkit on Fedora 41
I needed to run Playwright tests using Webkit and update a snapshot test, but Fedora Linux is not officially supported by Playwright (https://playwright.dev/docs/intro#system-requirements). Only Ubuntu and Debian are officially supported.

### Steps to Reproduce
Followed the official instructions for installing Playwright dependencies (https://github.com/apache/incubator-kie-tools?tab=readme-ov-file#step-1-bootstrap) with the following command:
   ```bash
   PLAYWRIGHT_BASE__installDeps=true pnpm bootstrap
   ```
output:
   ```
   packages/playwright-base install$ run-script-if --bool "$(build-env playwrightBase.installDeps)" --then "pnpm playwright install --with-deps"
   │ [run-script-if] Boolean condition supplied.
   │ [run-script-if] Conditions are checked using the operator 'and'.
   │ [run-script-if] Condition is 'true'. Running _then_ command(s).
   │ [run-script-if] Running 1 command(s): ['pnpm playwright install --with-deps']
   │ [run-script-if] Running 'pnpm playwright install --with-deps'
   │ BEWARE: your OS is not officially supported by Playwright; installing dependencies for ubuntu20.04-x64 as a fallback.
   │ Installing dependencies...
   │ Switching to root user to install dependencies...
   │ sh: line 1: apt-get: command not found
   │ Failed to install browsers
   │ Error: Installation process exited with code: 127
   │ [run-script-if] Error executing 'pnpm playwright install --with-deps'.
   ```

### How did I partially fix
4. Tried manual installation with:
   ```bash
    cd packages/playwright-base; pnpx playwright install
    BEWARE: your OS is not officially supported by Playwright; downloading fallback build for ubuntu20.04-x64.
    Downloading Chromium 131.0.6778.33 (playwright build v1148) from https://playwright.azureedge.net/builds/chromium/1148/chromium-linux.zip
    161.3 MiB [====================] 100% 0.0s
    Chromium 131.0.6778.33 (playwright build v1148) downloaded to /home/fantonan/.cache/ms-playwright/chromium-1148
    BEWARE: your OS is not officially supported by Playwright; downloading fallback build for ubuntu20.04-x64.
    Downloading Chromium Headless Shell 131.0.6778.33 (playwright build v1148) from https://playwright.azureedge.net/builds/chromium/1148/chromium-headless-shell-linux.zip
    100.9 MiB [====================] 100% 0.0s
    Chromium Headless Shell 131.0.6778.33 (playwright build v1148) downloaded to /home/fantonan/.cache/ms-playwright/chromium_headless_shell-1148
    BEWARE: your OS is not officially supported by Playwright; downloading fallback build for ubuntu20.04-x64.
    Downloading Firefox 132.0 (playwright build v1466) from https://playwright.azureedge.net/builds/firefox/1466/firefox-ubuntu-20.04.zip
    87.6 MiB [====================] 100% 0.0s
    Firefox 132.0 (playwright build v1466) downloaded to /home/fantonan/.cache/ms-playwright/firefox-1466
    BEWARE: your OS is not officially supported by Playwright; downloading fallback build for ubuntu20.04-x64.
    You are using a frozen webkit browser which does not receive updates anymore on ubuntu20.04-x64. Please update to the latest version of your operating system to test up-to-date browsers.
    BEWARE: your OS is not officially supported by Playwright; downloading fallback build for ubuntu20.04-x64.
    Downloading FFMPEG playwright build v1010 from https://playwright.azureedge.net/builds/ffmpeg/1010/ffmpeg-linux.zip
    2.3 MiB [====================] 100% 0.0s
    FFMPEG playwright build v1010 downloaded to /home/fantonan/.cache/ms-playwright/ffmpeg-1010   
   ```
   This successfully installed Chromium and Firefox, but Webkit does not work.

How can I install Playwright with Webkit on Fedora Linux? Are there workarounds or additional steps I should take?
Thanks in advance

ljmotta:
Closes https://github.com/apache/incubator-kie-issues/issues/1114
Closes https://github.com/apache/incubator-kie-tools/issues/2839

## Description
This PR introduces containerized end-to-end (E2E) Playwright tests to ensure consistent and reliable test execution across environments. By default, after this PR, running the E2E tests locally and in the CI will use containers instead of running natively in the OS.

Tests were executed in the following environments:
 - macOS
   - arm64
   - Sonoma 14.6.1
   - Podman Desktop v1.15
   - Podman Engine 5.3.1
   - Docker CLI 27.2.1
 - macOS
   - amd64
   - Sonoma 14.5
   - Podman Desktop v1.16
   - Podman Engine 5.3.2
   - Docker CLI 274.0
 - Windows WSL
   - amd64
   - Ubuntu 22.04.3 LTS
   - Docker CE CLI 25.0.2
   - Docker CE Engine 25.0.2

This PR updates multiple screenshots, as the changes introduced allowed lowering the snapshot threshold from 10% to 0.1%!

## Known Issues
 - Running the online-editor containerized test-e2e script on macOS requires starting the services in a separate terminal.
 - Performance: Tests may run slower on macOS arm64 due to architecture differences.
 - Flakiness: Retries are enabled because tests can occasionally fail without valid reasons when running in the restricted containerized environment.
 - Debugging tests is not currently supported but could be added in the future.
 - The Docker CLI is required to run the containerized tests.

## Trying it out
- Install the dependencies.
- Build the `@kie-tools/playwright-base` image with the following command:
```sh
KIE_TOOLS_BUILD__buildContainerImages=true pnpm -F @kie-tools/playwright-base image:docker:build
```
- In a package that contains Playwright e2e tests, run the following:
```sh
KIE_TOOLS_BUILD__runEndToEndTests=true pnpm test-e2e
```
> This command will start the required services and run the tests inside the container.

## Updating screenshots
To update screenshots, use the `pnpm test-e2e:container:shell` script to open a shell inside the container:
```sh
pnpm test-e2e:container:shell
# In the container shell, manually call the `test-e2e:run` passing the `-u` or `--update-snapshots` flag
pnpm test-e2e:run -u
# or
pnpm test-e2e:run --update-snapshots
```

To improve performance, you can filter which tests will run by adding the `-g` flag:
```sh
pnpm test-e2e:run -u -g "test name"
```


jomarko:
Thank you @ljmotta . Please, let us know this is ready for a verification. I will try on non-MAC environment.

ljmotta:
@jomarko This PR can be tested for macOS and Linux. I still need to make some adjustments for native Windows (non WSL).

packages/playwright-base/package.json
```diff
@@ -18,6 +18,7 @@
     "index.js"
   ],
   "scripts": {
+    "image:docker:build": "run-script-if --bool \"$(build-env containerImages.build)\" --then \"docker build --platform=linux/amd64 -t kie-tools-playwright-container-image:$(build-env root.version) . -f Containerfile --output type=docker\"",
```
thiagoelg:
```suggestion
    "build:dev": "run-script-if --bool \"$(build-env containerImages.build)\" --then \"pnpm image:docker:build\"",
    "build:prod": "pnpm build:dev",
    "image:docker:build": "docker build --platform=linux/amd64 -t kie-tools-playwright-container-image:$(build-env root.version) . -f Containerfile --output type=docker",
```

This way, we isolate the env var check to the `build:*` scripts and can call the `image:docker:build` directly if necessary.

packages/online-editor/package.json
```diff
@@ -24,7 +24,16 @@
...
     "start:kie-sandbox-accelerator-quarkus": "npm --prefix ./node_modules/@kie-tools/kie-sandbox-accelerator-quarkus run start",
     "test": "run-script-if --ignore-errors \"$(build-env tests.ignoreFailures)\" --bool \"$(build-env tests.run)\" --then \"jest --silent --verbose\"",
-    "test-e2e": "run-script-if --ignore-errors \"$(build-env endToEndTests.ignoreFailures)\" --bool \"$(build-env endToEndTests.run)\" --then  \"pnpm rimraf ./dist-tests-e2e\" \"pnpm test-e2e:run\"",
+    "test-e2e": "run-script-if --ignore-errors \"$(build-env endToEndTests.ignoreFailures)\" --bool \"$(build-env endToEndTests.run)\" --then \"pnpm rimraf ./dist-tests-e2e\" \"pnpm test-e2e:run\" --else \"pnpm test-e2e:container\"",
+    "test-e2e:container": "run-script-if --bool \"$(build-env containerizedEndToEndTests.run)\" --then  \"pnpm rimraf ./dist-tests-e2e\" \"pnpm test-e2e:container:check\"",
+    "test-e2e:container:bash": "docker exec -it kie-tools-playwright-containerization-online-editor /bin/bash -c \"cd kie-tools/packages/online-editor && exec /bin/bash\"",
+    "test-e2e:container:build": "cross-env KIE_TOOLS_BUILD__buildContainerImages=true pnpm -F playwright-base image:docker:build",
```
thiagoelg: I think we should handle the playwright-base image as a dependency, and as long as the dependencies are built, this image should exist.

It seems wrong to call the build script of a different package here, wdyt?

thiagoelg (commented):
Awesome PR @ljmotta! This will make our e2e tests more stable and concise!

I've made some comments below about the package.json scripts; let me know what you think.

packages/online-editor/package.json
ljmotta: This would simplify a lot!

ljmotta:
Status update regarding Windows. My current setup with Docker CLI + Docker WSL didn't work, as WSL doesn't handle Windows symlinks [1][2], which is heavily used by `pnpm`, so mounting the Windows volume in the container doesn't work. I'll try a different approach with Docker CLI + Podman Desktop to check. Podman works a little different, but it also uses WSL behind the scenes so I'm not too confident, but I'll give it a shot.

[1] https://github.com/microsoft/WSL/issues/5118
[2] https://github.com/microsoft/WSL/issues/12250

tiagobento:
Maybe try hard links @ljmotta? I think we had a similar issue with `maven-m2-repo-via-http-inage`.

ljmotta:
@tiagobento The symlinks are located in the `node_modules` directory, so switching to hard links might not be straightforward. After experimenting, I made some progress with volumes using Docker Desktop and HyperV. However, I encountered two additional issues:

1. It appears that playwright Windows browsers cannot be used inside the container. This means the browsers need to be reinstalled. Furthermore, the playwright binary is unavailable, requiring a full reinstallation of playwright, including erasing node_modules [1].
2. To address the first issue, all untracked files (e.g., node_modules) need to be removed. However, since pnpm creates recursive structures with symlinks, the `git clean -fdx` command cannot be used [2][3].

Even if the second issue were resolved, having to fix the first issue every time a container is started would be impractical. Considering these challenges, I am leaning toward not supporting Windows development without WSL for this pull request.

[1] https://github.com/microsoft/playwright/issues/23098#issuecomment-1558343627
[2] https://github.com/git-for-windows/git/issues/3358
[3] https://github.com/pnpm/pnpm/issues/4401

tiagobento:
@ljmotta I'm ok with not supporting Windows without WSL here, just please add a new statement to the README making that explicit. And thank you so much for the effort!

packages/playwright-base/Containerfile
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Using bookworm (debian 12) due to compatibility with GHA ubuntu-latest (Ubuntu 22.04)
+FROM node:20-bookworm
```
tiagobento: Please update to Node.js 22 after
- #2869

packages/playwright-base/README.md
```diff
@@ -17,6 +17,48 @@
 
 ## @kie-tools/playwright-base
 
+## Overview
+
+This package collects common configurations to run end-to-end Playwright tests.
+
+## Using containers to generate screen shots
```
tiagobento:
```suggestion
## Using containers to generate screenshots
```

packages/root-env/env/index.js
```diff
@@ -93,6 +97,9 @@ module.exports = composeEnv([], {
         run: str2bool(getOrDefault(this.vars.KIE_TOOLS_BUILD__runEndToEndTests)),
         ignoreFailures: str2bool(getOrDefault(this.vars.KIE_TOOLS_BUILD__ignoreEndToEndTestFailures)),
       },
+      containerizedEndToEndTests: {
+        run: str2bool(getOrDefault(this.vars.KIE_TOOLS_BUILD__runContainerizedEndToEndTests)),
+      },
```
tiagobento: So I think the `KIE_TOOLS_BUILD__runEndToEndTests` env var should still be the one use to toggle the execution of the E2E tests, and this new flag should only toggle the MODE with which the tests run.

```bash
KIE_TOOLS_BUILD__runEndToEndTests=false
# E2E tests don't run and `KIE_TOOLS_BUILD__runContainerizedEndToEndTests` is ignored.
```

```bash
KIE_TOOLS_BUILD__runEndToEndTests=true
KIE_TOOLS_BUILD__runContainerizedEndToEndTests=true
# E2E tests run inside containers
```


```bash
KIE_TOOLS_BUILD__runEndToEndTests=true
KIE_TOOLS_BUILD__runContainerizedEndToEndTests=false (default)
# E2E tests run on host machine (without containers involved)
```

WDYT?

packages/dmn-editor/tests-e2e/__containerization__/playwright-docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    environment:
+      - KIE_TOOLS_BUILD__runEndToEndTests=true
+      - KIE_TOOLS_BUILD__runContainerizedEndToEndTests=true
+    volumes:
+      - type: bind
+        source: ./../../../../
+        target: /home/node/kie-tools-playwright/incubator-kie-tools
+
+  # The nginx reverse-proxy is used to forward `host.docker.internal` requests to `localhost.
+  nginx:
```
tiagobento: Why not call it `reverse-proxy`?

packages/dmn-editor/tests-e2e/__containerization__/playwright-docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    image: nginx:latest
```
tiagobento: I'm not the biggest fan of `latest` tags as you know... `nginx` is pretty stable, but still, I guess we should be the ones deciding when to change versions of a dependency.

packages/dmn-editor/tests-e2e/__containerization__/playwright-docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    container_name: kie-tools-playwright-containerization-dmn-editor-nginx
```
tiagobento: Same... `nginx` is just a tool, I think rather this name could convey intent

ljmotta:
@tiagobento I've renamed `KIE_TOOLS_BUILD__runContainerizedEndToEndTests` to `KIE_TOOLS_BUILD__containerizedEndToEndTests`, and instead of using a new variable, I've made it a property of `buildEnv.endToEndTests`. I've also updated the scripts to follow the new behavior you described:

```
KIE_TOOLS_BUILD__runEndToEndTests=true ?
  KIE_TOOLS_BUILD__containerizedEndToEndTests=true ? 
    run_on_container : 
    run_on_host : 
  not_run
```
Currently, KIE Sandbox tests aren't working on macOS containers. I'm not sure why. I've tried debugging it, but I haven't found any obvious cause. The DMN and BPMN Editors are stuck in an infinite loading state and aren't fetching their respective envelopes. The shared worker is stuck in the `git.init` method, which calls the `fs.exists` method. From what I can see, the `fs.exists` method doesn't exist 😅.

tiagobento:
Did you try serving `online-editor` with something like [`http-server`](https://www.npmjs.com/package/http-server), for instance, instead of the Webpack Dev Server (`pnpm start`)?

ljmotta:
@tiagobento `http-server` doesn't work too. I've tested with macOS in a `amd64` machine, and the `online-editor` Google Chrome and Chormium tests passed with Docker and Podman.

@thiagoelg gave me the idea of removing the `amd64` compatibility that Docker and Podman provide and running the container natively on the `arm` macOS. After removing the `platform` specification from the image build script and `docker-compose` file, the `online-editor` Chromium tests passed🎉  (Unfortunelly doesn't exist Google Chrome for Linux `arm64`). Everything points to a bug in the platform compatibility.

EDIT:
I see three ways we can proceed with this PR:
1. Remove the `platform` flag, and disable **all** Google Chrome tests for Linux `arm64`.
2. Keep the `platform` (since it works for the other packages) and disable Chromium and Google Chrome tests in the `online-editor` package.
3. Make the `arm64` macOS user generate an additional image for `arm64` that will be used for the `online-editor` and without Google Chrome

tiagobento:
Thank you @ljmotta. From what I understand I believe the option benefiting the larger surface area would be option 2. 

Please confirm if the following statement is correct:

---

If we go with option 2:
- Developers on macOS arm64 trying to update `online-editor` screenshots won't be able to do so, and will need to find an alternative setup while we figure out what's the problem there.
- Windows users will need to rely on WSL to update screenshots and we won't be pursuing finding the solution for non-WSL.

---

If this is accurate, then you have my blessing to go with it 😛

ljmotta:
Yeah, that's correct. And just to be more clear. With option 2, the macOS arm64 users will not be able to run _**any**_ `online-editor` Chromium and Google Chrome test inside the container.

thiagoelg (approved):
Awesome work @ljmotta! LGTM

ljmotta:
I just double checked all screenshots. It's looking good. @tiagobento could you review the PR one more time?

packages/online-editor/playwright.config.ts
```diff
@@ -61,6 +62,35 @@ const customConfig = defineConfig({
       timeout: 240000,
     },
   ],
+  // Override
+  projects: process.env.KIE_TOOLS_ONLINE_EDITOR__SKIP_PLAYWRIGHT_FOR_ARM64
```
tiagobento: Not using `build-env` here and directly accessing an env var. This is not aligned with the repository's convention...

packages/scesim-editor/tests-e2e/__containerization__/playwright-docker-compose.yml
```diff
@@ -0,0 +1,64 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+services:
+  playwright:
+    platform: linux/amd64
+    container_name: kie-tools-playwright-containerization-scesim-editor
+    image: dev.local/${USER}/kie-tools-playwright-container-image:${PROJECT_VERSION}
```
tiagobento: This name is not parameterized, and the ${USER} component there seems unnecessary. WDYT about

`dev.local/apache-kie-tools/playwright-e2e-tests-image`?

packages/online-editor/playwright.config.ts
tiagobento: Not using build-env to access an environment variable...

packages/online-editor/tests-e2e/__containerization__/playwright-docker-compose.yml
```diff
@@ -0,0 +1,103 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+services:
+  playwright:
+    platform: linux/amd64
+    container_name: kie-tools-playwright-containerization-online-editor
+    image: dev.local/${USER}/kie-tools-playwright-container-image:${PROJECT_VERSION}
```
tiagobento: Images should have its tag version aligned with `streamName`, not with `version`.

Also the ${USER} component seems unnecessary..

WDYT about `dev.local/apache-kie-tools/playwright-e2e-tests-image`?

packages/online-editor/playwright.config.ts
ljmotta: This env var is only used for a workaround. I didn't add it to the `build-env` as it runs a `bash` script:

```
"test-e2e:container:up:darwin": KIE_TOOLS_ONLINE_EDITOR__SKIP_PLAYWRIGHT_FOR_ARM64=$( [[ $(uname -p) == \"arm\" ]] ...
```

I'm not sure how we could add this to the `env.js` file...

ljmotta: Well, I could export the variable in the script, to use it in the `playwright.config.ts`... 😅

tiagobento (approved):

PR #2868, opened 2025-01-22
kie-issues#1773: DMN Editor: Render badges with evaluation status on Decision nodes

### LINKED ISSUE #1773: DMN Editor: Render badges with evaluation status on Decision nodes
This ticket belongs to series of tickets related to dmn evaluation highlights rendering. Previous tickets are:
- https://github.com/apache/incubator-kie-issues/issues/1527
- https://github.com/apache/incubator-kie-issues/issues/1547
- https://github.com/apache/incubator-kie-issues/issues/1595
- https://github.com/apache/incubator-kie-issues/issues/1741

In this ticket we will render badges with the evaluation status on Decision nodes. Figure 1 shows an error badge in the "Loan Pre-Qualification" Decision node and a "success" badge in the "Risk Score" Decision node. Badges must disappear when "Evaluation Highlights" is turned off.

 
Image

jomarko:
Closes: https://github.com/apache/incubator-kie-issues/issues/1773

### TODO
- [x] Create story that would demonstrate badges, it should be very similar story as we have for badges in the boxed expression editor
- [x] use existing `evaluationResults` prop instead of `evaluationStatus` that was added in the draft

#### Success badge - current proposal
Screenshot 2025-01-22 120830

#### Failure badge - current proposal
image


packages/dmn-editor/src/diagram/nodes/Nodes.tsx
```diff
@@ -402,6 +403,13 @@ export const DecisionNode = React.memo(
       );
     });
 
+    const isEvaluationHighlightsEnabled = useDmnEditorStore((s) => s.diagram.overlays.enableEvaluationHighlights);
+    const { evaluationResults } = useDmnEditor();
+    const evaluationResultsClassName =
```
thiagoelg: This should be a `useMemo` to avoid unnecessary re-renders

thiagoelg (commented):
This looks good @jomarko, thanks for the PR!
Added a single comment regarding re-renders.

tiagobento:
@jomarko Would it be possible to make the badges not hide the border of the nodes? Also, how do they look like when they're hovered/selected? Another feedback I have it regarding the alignment. I think the `x` and `✓` are too left-aligned. Recording a video always helps with reviews :D

jomarko:
Hi @tiagobento , 

### 01 hover
There is no behavior for hovering. We didn't define it in the feature definition phase. What is your suggestion for the 'hover' situation?

### 02 alignment
Should be fixed. I didn't record a video as it is enough to start playbook as `pnpm -F @kie-tools/dmn-editor start`

### 03 Border
this is the most problematic as the current border of nodes has width `1.5px`. From reading some [internet threads](https://stackoverflow.com/questions/9080633/can-a-css-pixel-be-a-fraction), it seems as not ideal to use fractions and pixels together. 

I tried to margin my pseudo element with:
1. `1px`
2. `1.5px`
3. `2px`

however there is some rounding and it is never correct result. 

#### 1px
image

#### 1.5px, 2px
for both the same result, 1.5 is rounded to 2
image


What I could do is to change `NodeStyle.ts` where we define default stroke width `export const DEFAULT_NODE_STROKE_WIDTH = 1.5;`, however any change we would do there (1px or 2px) would affect all tests.

tiagobento:
@jomarko Right... Hmmm... I think the 2px compromise is better? I'd go with the 2px :D

packages/dmn-editor/src/DmnEditor.tsx
```diff
@@ -71,7 +71,10 @@ export type DmnEditorRef = {
   getCommands: () => Commands;
 };
 
-export type EvaluationResults = Record<string, any>;
+/**
+ * TODO we need to implement mapping  between (extended-services-api/src/dmnResult.ts) and ("success" | "failure")
```
ljmotta: Could you please create an issue for this TODO?

ljmotta (commented):
@jomarko Thanks for this PR! Nice addition to the DMN Runner evaluation highlights! I've just made a comment inline.

packages/dmn-editor/src/DmnEditor.tsx
jomarko: @ljmotta done

packages/dmn-editor/src/diagram/nodes/Nodes.tsx
jomarko: @thiagoelg done

tiagobento:
Great job making the PR pass the checks! Is this ready for review? Thank you @jomarko.

jomarko:
@tiagobento yes, it should be ready for a re-review

ljmotta (approved):

tiagobento (approved):

PR #2800, opened 2024-12-12
kie-issues#1584: [kn-plugin-workflow] added --wait to `kn workflow deploy` 

treblereel:
Closes https://github.com/apache/incubator-kie-issues/issues/1584

PR #2874, opened 2025-01-26
kie-issues#1586: [kn-plugin-workflow] Evaluate if we should install operator via cli

### LINKED ISSUE #1586: [kn-plugin-workflow] Evaluate if we should install operator via cli
something like:

kn workflow install <-- Installs the Operator
kn workflow status <-- Tells you what version of the Operator you have installed
kn workflow uninstall <-- Deletes the Operator
Take a look at the Camel K CLI for inspiration as they have something very similar.

treblereel:
closes: https://github.com/apache/incubator-kie-issues/issues/1586

tests will be added after https://github.com/apache/incubator-kie-tools/pull/2800 gets merged. (We need `kind` for e2e tests)


packages/kn-plugin-workflow/pkg/command/operator/status.go
```diff
@@ -0,0 +1,67 @@
...
+
+package operator
+
+import (
+	"github.com/apache/incubator-kie-tools/packages/kn-plugin-workflow/pkg/common"
+	"github.com/ory/viper"
+	"github.com/spf13/cobra"
+)
+
+func NewStatusOperatorCommand() *cobra.Command {
```
ricardozanini, line 28: I think we can add the default `sonataflow-operator-system` namespace here:

https://github.com/apache/incubator-kie-tools/blob/main/packages/sonataflow-operator/config/default/kustomization.yaml#L19

packages/kn-plugin-workflow/pkg/common/k8sclient/goapi.go
```diff
@@ -23,8 +24,53 @@ import (
...
-	return GetNamespace()
+var selfSubjectAccessGVR = schema.GroupVersionResource{
+	Group:    "authorization.k8s.io",
+	Version:  "v1",
+	Resource: "selfsubjectaccessreviews",
+}
+
+var kubeconfigInfoDisplayed = false
+
+func (m GoAPI) CanICreate(resourcePath string, namespace string) (bool, error) {
```
ricardozanini:
```suggestion
func (m GoAPI) IsCreateAllowed(resourcePath string, namespace string) (bool, error) {
```

packages/kn-plugin-workflow/pkg/common/k8sclient/goapi.go
```diff
@@ -23,8 +24,53 @@ import (
...
+			}
+
+			allowed, _, _ := unstructured.NestedBool(result.Object, "status", "allowed")
+			return allowed, nil
+		}
+	}
+	return false, nil
+}
+
+func (m GoAPI) CanIDelete(name string, namespace string) error {
```
ricardozanini:
```suggestion
func (m GoAPI) IsDeleteAllowed(name string, namespace string) error {
```

ricardozanini (commented):
Very nice! So now, on our e2e we can install KIND, OLM, and run `kn workflow operator install`? Then we perform a simple deployment.

packages/kn-plugin-workflow/pkg/command/operator/status.go
treblereel: good question, to install the operator from operatorhub into some namespace, we should have a namespace and OperatorGroup.
by default (minikube/kind), after we installed olm, where two Operator groups
```kubectl get operatorgroup --all-namespaces

NAMESPACE   NAME               AGE
olm         olm-operators      7m46s
operators   global-operators   7m46s
```

so to use `sonataflow-operator-system` as a namespace for sonataflow operator, we should create a name and Operator group before installing the operator inseft.

I am going to implement it in the upcomming issue.
@ricardozanini tdyt ?

treblereel:
@ricardozanini @domhanak @wmedvede 

After the rebase, I tested kn workflow deploy --wait and discovered an sad issue: port forwarding only works if the operator is installed from `sonataflow-operator/operator.yaml`. When installed from OperatorHub (`https://operatorhub.io/install/sonataflow-operator.yaml`), port forwarding doesn't work.

Do you have any ideas what might be causing this?

ricardozanini:
@treblereel 

> Do you have any ideas what might be causing this?

No clue. Do you have the kubectl `--wait` command? So we can debug. Maybe we are missing a small config in the Hub installation in the CSV. 🤔

treblereel:
> @treblereel
> 
> > Do you have any ideas what might be causing this?
> 
> No clue. Do you have the kubectl `--wait` command? So we can debug. Maybe we are missing a small config in the Hub installation in the CSV. 🤔

That is what I do: check diffs. What is good, I cached it right now.
No clue as well, but it's just a matter of time

treblereel:
@ricardozanini nope, port forwarding works well, the reason is that operator, installed from OperatorHub use `quay.io/kiegroup/kogito-swf-devmode:1.44` which doesn't propagate  `/q/devui`. Is it expected bevavior?

packages/kn-plugin-workflow/pkg/common/operator.go
```diff
@@ -91,3 +166,273 @@ func FindServiceFiles(directory string) ([]string, error) {
...
+
+func (m OperatorManager) CheckOLMInstalled() error {
+	resources, err := m.dynamicClient.Resource(catalogSourcesGVR).List(context.Background(), v1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("❌ ERROR: OLM (Operator Lifecycle Manager) isn't installed, install OLM first : %v\n", err)
+	}
+	var operatorSource string
+
+	if m.isOpenshift {
+		operatorSource = "community-operators"
```
ricardozanini: The downstream namespace name is different. @jakubschwan do you know which NS we use downstream to fetch the catalog?

So here we might have to try both `community-operators` and `(?) - TBD`.

domhanak: iirc downstream would be `openshift-operators` . I guess we should parametrize this?

treblereel: I tested openshift local, and as fat as I remember, there were such buildin namespace.
It's possible to pass the requred namespace like this: `operator install --namespace <your_operator_namespace>`

ricardozanini: I think we can do a try on every possible solutions if `--namespace` is empty.

treblereel: @ricardozanini you mean iter over a set of namespaces until we find one?

ricardozanini (approved):

domhanak (approved):
Thanks

PR #2888, opened 2025-01-31
kie-issues#1581: New DMN Editor is getting crashed when working with Decision Table Annotation

### LINKED ISSUE #1581: New DMN Editor is getting crashed when working with Decision Table Annotation
This issue is to report that the new DMN editor is getting crashed and unable to open the DMN file when working with Decision Table Annotation. 

For your reference, we have attached steps to reproduce that can be used to replicate the issue locally in the new DMN editor (main branch)

1. Open a Decision and add few annotations
2. Open the DMN in Text Editor
3. Switch the Editor to DMN editor
4. You will see a error message unable to open the diagram
5. If we notice a empty <kie:width/> tag is created in xml which causing the error. 
6. Delete the empty <kie:width/> tag and reopen the model in editor

https://github.com/user-attachments/assets/81bc149b-85a2-4f82-8361-83e53feeea5a

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-issues/issues/1581 New DMN Editor is getting crashed when working with Decision Table Annotation

### DIFF
--- a/packages/boxed-expression-component/src/expressions/DecisionTableExpression/DecisionTableExpression.tsx
+++ b/packages/boxed-expression-component/src/expressions/DecisionTableExpression/DecisionTableExpression.tsx
@@ -851,7 +851,7 @@ export function DecisionTableExpression({
               : DECISION_TABLE_ANNOTATION_DEFAULT_WIDTH;
 
         const nextValues = [...prev];
-        const minValuesLength = args.beforeIndex + 1 + args.columnsCount;
+        const minValuesLength = args.beforeIndex + args.columnsCount;
         nextValues.push(...Array(Math.max(0, minValuesLength - nextValues.length)));
         for (let i = 0; i < args.columnsCount; i++) {
           const widthIndex = args.beforeIndex + i + 1; // + 1 to account for the rowIndex column.


packages/dmn-editor/src/mutations/updateExpressionWidths.ts
```diff
@@ -34,6 +34,6 @@ export function updateExpressionWidths({
 
   widthsExtension["kie:ComponentWidths"] = [...widthsById.entries()].map(([k, v]) => ({
     "@_dmnElementRef": k,
-    "kie:width": v.map((vv) => ({ __$$text: vv })),
+    "kie:width": v.filter((width) => width !== undefined).map((vv) => ({ __$$text: vv })),
```
jomarko: @Kusuma04-dev I think this is not proper place to do a fix. I agree the issue is not reproducible with this changes, however I think similar approach just hide a more serious bug in our code.

From the attached video, There we can see, that table has 6 columns (row index, input, output, 3 annotations), while the xml has 7 `<kie:width>` tags. This `+-1` is suspicious. I think the fix of the issue should be to prevent generating 7 tags for 6 columns.

I would like to ask you to investigate more `DecisionTableExpression` callbacks. This seems to do a lot with `+1` logic, maybe it is the root cause?
```
setWidthsById(({ newMap }) => {
        const prev = newMap.get(id) ?? [];
        const defaultWidth =
          args.groupType === DecisionTableColumnType.InputClause
            ? DECISION_TABLE_INPUT_DEFAULT_WIDTH
            : args.groupType === DecisionTableColumnType.OutputClause
              ? DECISION_TABLE_OUTPUT_DEFAULT_WIDTH
              : DECISION_TABLE_ANNOTATION_DEFAULT_WIDTH;

        const nextValues = [...prev];
        const minValuesLength = args.beforeIndex + 1 + args.columnsCount;
        nextValues.push(...Array(Math.max(0, minValuesLength - nextValues.length)));
        for (let i = 0; i < args.columnsCount; i++) {
          const widthIndex = args.beforeIndex + i + 1; // + 1 to account for the rowIndex column.
          nextValues.splice(widthIndex, 0, defaultWidth);
        }
        newMap.set(id, nextValues);
      });
 ```
 
 Other callbacks that sounds as related are `onColumnAdded`, `onColumnUpdates`, `setAnnotationColumnWidth` - all in `DecisionTableExpression.tsx` . It is also interesting this issue is present only for adding annotations columns.

jomarko (commented):
Thank you @Kusuma04-dev . I would like to ask you to investigate the issue more.See detailed message inline.

packages/dmn-editor/src/mutations/updateExpressionWidths.ts
Kusuma04-dev: Hi @jomarko ,Yes you are correct.Issue is adding extra count in the above code and i have changed now.Thanks for the review.

jomarko (approved):
Thank you, I haven't found an issue during manual check.

yesamer (approved):

PR #2895, opened 2025-02-04
kie-issues#1769 [kn-workflow-plugin] Specify image on 'workflow run'

### LINKED ISSUE #1769: [kn-plugin-workflow] Specify image on 'workflow run'
The plugin user should be able to specify an image when running the workflow.

treblereel:
Closes: https://github.com/apache/incubator-kie-issues/issues/1769

### DIFF
--- a/packages/kn-plugin-workflow/pkg/command/run.go
+++ b/packages/kn-plugin-workflow/pkg/command/run.go
@@ -36,6 +36,7 @@ type RunCmdConfig struct {
 	PortMapping string
 	OpenDevUI   bool
 	StopContainerOnUserInput bool
+	Image string
 }
 
 const StopContainerMsg = "Press any key to stop the container"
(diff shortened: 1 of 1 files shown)


packages/kn-plugin-workflow/pkg/command/run.go
```diff
@@ -65,9 +66,12 @@ func NewRunCommand() *cobra.Command {
 	# Stop the container when the user presses any key
 	{{.Name}} run --stop-container-on-user-input=false
 
+	# Specify a custom container image to use for the deployment.
+	{{.Name}} run --image=<your_image>
```
domhanak, line 71: I suggest we add something like `By default, the apache-kie-sonataflow-devmode image is used`  Just to be more clear what image are users substituting.

domhanak (changes requested):
@treblereel the `image` as flag is little bit confusing imho. Not sure how to improve it thought, container_image seems confusing too to me. 

Also I am missing unit test for this change. From Integration Test perspective I think it is fine for now.

ricardozanini:
> @treblereel the image as flag is little bit confusing imho. Not sure how to improve it thought, container_image seems confusing too to me.


@domhanak wdym? The `--image` flag is to replace the runner image with a custom one for debug/patches or any other reasons a user would need to change the base image to run with CLI.

If not `--image`, which name would you propose?

domhanak:
I think we could improve the help text with what you said and it should be ok. Most probably better solvable by providing a good guide for the feature.

ricardozanini (approved):

domhanak (approved):

PR #2901, opened 2025-02-10
kie-issues#1566: Remove `pkg` dependency from `image-env-to-json` on kie-tools, as it's deprecated

### LINKED ISSUE #1566: Remove `pkg` dependency from `image-env-to-json` on kie-tools, as it's deprecated
We need to find another solution to build native binaries from linux/macOS/windows from a JS file.

Possible alternatives:
- https://nodejs.org/api/single-executable-applications.html (not available on an LTS version yet)
- https://github.com/nexe/nexe

### LINKED ISSUE #2578: [image-env-to-json] Make binary compatible with any architecture
Currently, `image-env-to-json` only supports the x64 architecture. When running on new Arm machines, the underlying node `pkg` library relies on glibc for x64. That will always fail on Darwin arm CPUs.

Ideally, this package should also offer a way to copy the correct binary despite the OS and Architecture so we avoid clients having to do this:

```json
{
    "copy:image-env-to-json": "run-script-os",
    "copy:image-env-to-json:linux:darwin": "cp ./node_modules/@kie-tools/image-env-to-json/dist/linux/image-env-to-json-standalone ./dist-dev/",
    "copy:image-env-to-json:win32": "pnpm powershell \"Copy-Item ./node_modules/@kie-tools/image-env-to-json/dist/linux/image-env-to-json-standalone ./dist-dev/\""
}
```

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/1566
Closes: https://github.com/apache/incubator-kie-tools/issues/2578

## Description
Fully rewrite using Golang to simplify generate native binaries.

Obs: Some usages weren't updated to use the new Darwin and Windows binaries, since they are placed inside a Linux container.


packages/image-env-to-json/Makefile
```diff
@@ -0,0 +1,84 @@
...
+	GOARCH = amd64
+	BIN_BUILD_DARWIN=$(BIN_DARWIN_AMD)
+	BIN_BUILD_LINUX=$(BIN_LINUX_AMD)
+	BIN_BUILD_WINDOWS=$(BIN_WINDOWS_AMD)
+endif
+
+build-all: build-darwin-amd64 build-darwin-arm64 build-linux-amd64 build-linux-arm64 build-win-amd64 build-win-arm64
+
+build-darwin:
+	CGO_ENABLED=0 GOOS=darwin GOARCH=$(GOARCH) go build -ldflags $(LDFLAGS) -o $(BIN_PATH)/$(BIN_BUILD_DARWIN) $(MAIN_PATH)
```
thiagoelg: I know it's not ideal, but in other packages, we use `cross-env` here to avoid OS-specific errors when cross-building.

packages/image-env-to-json/package.json
```diff
@@ -14,32 +14,31 @@
...
+    "build:linux": "pnpm setup:env make build-linux",
+    "build:linux:amd": "pnpm setup:env make build-linux-amd64",
+    "build:linux:arm": "pnpm setup:env make build-linux-arm64",
+    "build:prod": "rimraf -rf dist && run-script-os && pnpm test",
+    "build:prod:darwin": "rimraf dist && pnpm setup:env make build-all",
+    "build:prod:linux": "rimraf dist && pnpm setup:env make build-all",
+    "build:prod:win32": "rimraf dist && pnpm setup:env:win32 make build-all",
+    "build:win32": "pnpm setup:env:win32 make build-win",
+    "build:win32:amd": "pnpm setup:env:win32 make build-win-amd64",
+    "build:win32:arm": "pnpm setup:env:win32 make build-win-arm64",
```
thiagoelg: I think we should build all binaries every time (on build:dev and build:prod). It should be quick, and we can avoid issues where, for example, we test the binary in a macOS ARM environment but then run it inside a Linux x64 container.

It would also simplify this list of scripts in the package.json.

packages/image-env-to-json/package.json
```diff
@@ -14,32 +14,31 @@
...
+    "install": "go mod tidy",
+    "powershell": "@powershell -NoProfile -ExecutionPolicy Unrestricted -Command",
+    "setup:env": "cross-env PLUGIN_VERSION=$(build-env imageEnvToJson.version)",
+    "setup:env:win32": "pnpm powershell \"cross-env PLUGIN_VERSION=$(build-env imageEnvToJson.version)\"",
```
thiagoelg: Is powershell needed even with `cross-env`?

ljmotta: Yes, we need to set the environment variables before the `make`.

tiagobento:
Please add tests for when env vars contain JSON strings too. Thank you!

tiagobento (approved):

thiagoelg (approved):

PR #2902, opened 2025-02-11
kie-issues#1459: Decision and BKM nodes properties cannot be edited from inside BEE due to focus being lost after a character is typed

### LINKED ISSUE #1459: Decision and BKM nodes properties cannot be edited from inside BEE due to focus being lost after a character is typed
There is an issue in the DMN Editor. Once user tries to set properties of the header cell, he/she is not able to write down more than one character. Once the one character is typed, the focus is lost.

### Steps to reproduce
- Open a DMN Diagram
- Add a decision node
- Open the decision node
- Set its type as 'context'
- open the properties panel
- select the context header cell
- try to fill in 'Allowed answers' using keyboard

Screenshot 2024-09-02 125802

See the video:
https://github.com/user-attachments/assets/c8808a31-7a25-4197-aeda-1f3d4c29035f

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-issues/issues/1459 DMN Header cell properties loses focus after one character is typed


jomarko:
Looking

jomarko (commented):
I can confirm the properties panel for Decision nodes works fine now:
- Question
- Description
- Allowed values

However, we need to fix also BKM node properties panel:
- Description

Thank you for the PR @Kusuma04-dev

Kusuma04-dev:
> I can confirm the properties panel for Decision nodes works fine now:
> 
> * Question
> * Description
> * Allowed values
> 
> However, we need to fix also BKM node properties panel:
> 
> * Description
> 
> Thank you for the PR @Kusuma04-dev

Thanks for the review Jozef, Done.

jomarko (approved):
I think now the feature works properly.

tiagobento:
I'm not sure if this is a proper fix, to be honest. What happens when `description` is undefined?

`(state.dmn.model.definitions.drgElement![index] as Normalized<DMN15__tDecision>).description!.__$$text = ...` will fail, right? Can you please check this case?

tiagobento:
Now when `drgElement.allowedAnswers` is undefined, no property is added. How can you go from `drgElement.allowedAnswers` being undefined to it being populated? 🤔

Kusuma04-dev:
> Now when `drgElement.allowedAnswers` is undefined, no property is added. How can you go from `drgElement.allowedAnswers` being undefined to it being populated? 🤔

we have __$$text in every property , if property itself is undefined , do we need to access text and set it ?? I assume as text is defined as string ,the default value should be an empty string and is it necessary to populate it ??

tiagobento (changes requested):
This PR does not solve the problem originally reported in https://github.com/apache/incubator-kie-issues/issues/1459. What this PR does is making the Decision- and BkmProperties components no-op, so that values typed in to the Description, Accepted answers, and Question are not persisted in the DMN model anymore.

Let's try another way please @Kusuma04-dev. Please write an E2E test, linked to the originally reported issue, containing the verification that:
1) Focus is not lost when trying to edit one of those properties inside BEE.
2) The DMN model is correctly updated containing the characters typed in one of the properties.

Thanks!

packages/dmn-editor/src/propertiesPanel/BkmProperties.tsx
```diff
@@ -29,10 +29,11 @@ import { useDmnEditorStore, useDmnEditorStoreApi } from "../store/StoreContext";
 import { InlineFeelNameInput } from "../feel/InlineFeelNameInput";
 import { useDmnEditor } from "../DmnEditorContext";
 import { useResolvedTypeRef } from "../dataTypes/useResolvedTypeRef";
-import { useCallback, useMemo } from "react";
+import { useCallback, useMemo, useState } from "react";
```
ljmotta: Leftover

packages/dmn-editor/src/propertiesPanel/DecisionProperties.tsx
```diff
@@ -29,10 +29,11 @@ import { useDmnEditorStore, useDmnEditorStoreApi } from "../store/StoreContext";
 import { InlineFeelNameInput } from "../feel/InlineFeelNameInput";
 import { useDmnEditor } from "../DmnEditorContext";
 import { useResolvedTypeRef } from "../dataTypes/useResolvedTypeRef";
-import { useCallback, useMemo } from "react";
+import { useCallback, useMemo, useState } from "react";
```
ljmotta: Same

packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/propertiesPanelBase.ts
```diff
@@ -30,6 +30,10 @@ export abstract class PropertiesPanelBase {
     return this.page.getByTestId("kie-tools--dmn-editor--properties-panel-container");
   }
 
+  public beepanel() {
+    return this.page.getByTestId("kie-tools--dmn-editor--bee-properties-panel-container");
+  }
+
```
ljmotta: You don't need to add this method. You can use the `BeePropertiesPanelBase` `panel` method

packages/dmn-editor/tests-e2e/drgElements/checkDecisionNodeProperties.spec.ts
ljmotta, line 1: Please, add the test in the `changeDecisionProperties` file

ljmotta (commented):
@Kusuma04-dev Thanks for the changes. Please, add a new test in the `changeBkmProperties` as well. Also, for the Decision test, follow the test name pattern to keep it consistent with the codebase.

danielzhe (approved):
Great job! That was a very annoying issue.
I'm approving, but just address the details that @ljmotta pointed.

tiagobento:
@Kusuma04-dev Can you please explain how your changes fix the focus problem? Thank you!

Kusuma04-dev:
> @Kusuma04-dev Can you please explain how your changes fix the focus problem? Thank you!

Hi @tiagobento , While investigating the issue of the entire properties panel (bee+properties) being re-rendered, I found that the focus was being lost due to the state updates. Initially, I implemented a solution where I created an intermediate local state (localQuestion) and updated it in the onChange handler. Then, in the onBlur event, I synced the local state with the global state. This helped prevent unnecessary re-renders, allowing the focus to stay intact while typing. Here's the initial solution I used :
 ```
const [localQuestion, setLocalQuestion] = React.useState(decision.question?.__$$text || "");
<FormGroup label="Question">
  <TextArea
    aria-label={"Question"}
    type={"text"}
    isDisabled={isReadOnly}
    value={localQuestion}   {/* Local state for the input */}
    onChange={(newQuestion) => {
      setLocalQuestion(newQuestion);  {/* Update local state */}
    }}
    onBlur={() => {
      setState((state) => {
        (state.dmn.model.definitions.drgElement![index] as Normalized<DMN15__tDecision>).question = {
          __$$text: localQuestion,  {/* Sync local state with global state on blur */}
        };
      });
    }}
    placeholder={"Enter a question..."}
    style={{ resize: "vertical", minHeight: "40px" }}
    rows={6}
  />
</FormGroup>
```
After discussing this approach with Luiz, we discovered that TextField already manages the intermediate state and handles the onBlur correctly when the field unmounts. So we have switched from TextArea to TextField.

File: packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/decisionPropertiesPanel.ts [hunk 4]
```diff
@@ -82,18 +90,35 @@ export class DecisionPropertiesPanel extends PropertiesPanelBase {
...
+    await this.beepanel().getByPlaceholder("Enter allowed answers...").fill(args.newAllowedAnswers);
+    await this.beepanel().getByPlaceholder("Enter allowed answers...").press("Tab");
+  }
+
   public async getAllowedAnswers() {
     return await this.panel().getByPlaceholder("Enter allowed answers...").inputValue();
   }
+  public async getDecisionAllowedAnswers() {
+    return await this.beepanel().getByPlaceholder("Enter allowed answers...").inputValue();
+  }
```
jomarko, line 100: I think I do not understand the naming convestion used here,
- `setDecisionXXX`
- `getDecisionXXX`

why not only:
- `setXXX`
- `getXXX`

jomarko (commented):
As we touch tests, I have one question related to tests API we have.

Thank you for the PR.

packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/decisionPropertiesPanel.ts
jomarko: this is for whole `decisionPropertiesPanel.ts`

Kusuma04-dev: Hi @jomarko , we need to write properties for bee panel when we see in decisionProperties with panel method using test-id _"kie-tools--dmn-editor--properties-panel-container"_ but we have beeproperties panel with test-id _"kie-tools--dmn-editor--bee-properties-panel-container"_. So i have written like that before but i can see it's already present in beePropertiesPanelBase which i have used right now.

packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/propertiesPanelBase.ts
Kusuma04-dev: Thanks @ljmotta , changed it accordingly.

jomarko (commented):
Hi, thank you, I think the current state of the PR is not correct, currently we duplicate the same methods both in:
- https://github.com/apache/incubator-kie-tools/blob/982ec20185fb3f0887a972b704bca4efbeb92135/packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/decisionPropertiesPanel.ts
- https://github.com/apache/incubator-kie-tools/blob/982ec20185fb3f0887a972b704bca4efbeb92135/packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/beePropertiesPanel.ts


my understanding is we should keep only methods generic for all nodes/expression in `beePropertiesPanel` it should not contain methods specific for particular nodes/expressions. I think for example `allowedAnswers` is specific property not available for all nodes.

yesamer:
@jomarko @tiagobento @ljmotta Can you please review this PR again? Thank you

jomarko (approved):
Thank you, my only feedback is now we combine in source code two approaches:
```
<FormGroup>
     <TextArea/>
</FormGroup>
```     

vs.

```
<FormField/>
```

Not saying it is bug, just saying it loudly, so others can share opinion about it, if they want. I can imagine can slow down refactoring in future if for example we will upgrade patternfly or similar task.

However my manual check passed, approving the PR. Thank you!

klbowers (approved):

packages/dmn-editor/tests-e2e/__fixtures__/propertiesPanel/bee/bkmPropertiesPanel.ts
```diff
@@ -0,0 +1,48 @@
...
+    await this.descriptionProperties.setDescription({ ...args });
+  }
+
+  public async getDescription() {
+    return await this.descriptionProperties.getDescription();
+  }
+
+  get expressionHeaderCell() {
+    return new NameAndDataTypeCell(this.page.getByRole("columnheader", { name: "New BKM (<Undefined>)" }));
+  }
```
ljmotta:
```suggestion
  public expressionHeaderCell() {
    return new NameAndDataTypeCell(this.page.getByRole("columnheader", { name: "New BKM (<Undefined>)" }));
  }
```

packages/dmn-editor/tests-e2e/drgElements/changeDecisionProperties.spec.ts
```diff
@@ -158,4 +158,24 @@ test.describe("Change Properties - Decision", () => {
 
     await expect(diagram.get()).toHaveScreenshot("change-decision-position.png");
   });
+
+  test("should change the Decision node description after selecting Decision node", async ({
```
ljmotta:
```suggestion
  test("should change the Decision node properties using the BEE properties panel", async ({
```

packages/dmn-editor/tests-e2e/drgElements/changeDecisionProperties.spec.ts
```diff
@@ -158,4 +158,24 @@ test.describe("Change Properties - Decision", () => {
...
+    nodes,
+    bee,
+    beePropertiesPanel,
+  }) => {
+    await nodes.edit({ name: DefaultNodeName.DECISION });
```
ljmotta, line 171:
```suggestion
    test.info().annotations.push({
        type: TestAnnotations.REGRESSION,
        description: "https://github.com/apache/incubator-kie-issues/issues/1459",
      });
    await nodes.edit({ name: DefaultNodeName.DECISION });
```

packages/dmn-editor/tests-e2e/drgElements/changeBkmProperties.spec.ts
```diff
@@ -137,4 +137,11 @@ test.describe("Change Properties - BKM", () => {
 
     await expect(diagram.get()).toHaveScreenshot("change-bkm-position.png");
   });
+  test("should change the BKM node description after selecting BKM node", async ({ nodes, beePropertiesPanel }) => {
+    await nodes.edit({ name: DefaultNodeName.BKM });
```
ljmotta, line 148:
```suggestion
    test.info().annotations.push({
        type: TestAnnotations.REGRESSION,
        description: "https://github.com/apache/incubator-kie-issues/issues/1459",
      });
    await nodes.edit({ name: DefaultNodeName.BKM });
```

packages/dmn-editor/tests-e2e/drgElements/changeBkmProperties.spec.ts
```diff
@@ -137,4 +137,11 @@ test.describe("Change Properties - BKM", () => {
 
     await expect(diagram.get()).toHaveScreenshot("change-bkm-position.png");
   });
+  test("should change the BKM node description after selecting BKM node", async ({ nodes, beePropertiesPanel }) => {
```
ljmotta:
```suggestion
  test("should change the BKM node properties using the BEE properties panel", async ({ nodes, beePropertiesPanel }) => {
```

ljmotta (changes requested):
@Kusuma04-dev I've added some change requests. The test annotations are to mark the reason behind the tests.

packages/dmn-editor/tests-e2e/drgElements/changeBkmProperties.spec.ts
Kusuma04-dev: Thanks for the review @ljmotta . As the issue arises only when selecting decision/BKM node , i thought to keep it like this to understand better . And yeah your suggestion also looks good and modified it.

ljmotta (approved):
@Kusuma04-dev Thanks for your patience! LGTM

tiagobento (commented):
Thanks all!

tiagobento (approved):

PR #2905, opened 2025-02-12
NO-ISSUE: Enable Maven's local repo tail mechanism for safer dependency relationship between Maven-based packages

tiagobento:
After Quarkus was upgraded to 3.15.3, we now have access to

- https://github.com/quarkusio/quarkus/pull/43352

so we can leverage it to enable a profile that will prevent packages from installing themselves on ~/.m2/repository, making builds safer in regards to the dependency relationship between Maven-based packages being now explicitly required on package.json files. See https://github.com/apache/incubator-kie-tools/blob/main/packages/maven-base/pom.xml#L424


tiagobento:
Build is green. Can you please review when you have the time @ricardozanini @ljmotta @pefernan @thiagoelg?

thiagoelg (approved):
Nice!

ricardozanini (approved):

PR #2909, opened 2025-02-14
NO-ISSUE: update go from 1.22.9 to 1.22.12 to address recently CVEs

rgdoliveira:


ljmotta:
@rgdoliveira Hey there! thanks for addressing this. I have a question regarding these go upgrades. Shouldn't we upgrade the `go.mod` files of the packages too? Some packages are with the go version `1.21`, and some use the `toolchain` of `1.23.x`.

rgdoliveira:
@ljmotta I see, like in following files (also others) there is go 1.21 defined:
packages/dev-deployment-upload-service/go.mod
packages/extended-services/go.mod

 I'm not an expert in kie-tools repo and I don't know where those packages are used. but thats a good point if all package should be aligned to the same go version.

ljmotta:
Well, actually, if the CI is using the 1.22.12, all artifacts are generated with this new Golang version, so I don't think we need to update all `go.mod` files. What concerns me is the difference in the `toolchain` version of the `kn-plugin-workflow`. From the Golang docs [1]. From what I understood, declaring a version bigger than the one installed, will download this new version. So this update will not affect the `kn-plugin-workflow`.

[1] https://go.dev/doc/toolchain

domhanak:
@treblereel fyi

rgdoliveira:
Do you think we can merge this one and discuss about the `kn-plugin-workflow` some other place?

ljmotta:
@rgdoliveira I think we can. After some research, the `toolchain go1.23.4` will download the `1.23.4` toolchain but will not set it as default.

ljmotta (approved):

PR #2918, opened 2025-02-21
[kie-issues#1826] Sandbox: Render selected indicator in the DMN Runner form output card

jomarko:
Closes https://github.com/apache/incubator-kie-issues/issues/1826

### TODO

- [x] verify the old GWT editor is not affected by the changes in the EnvelopeAPI
- [x] verify no form output card is highlighted if no expression editor is opened
- [x] form output card highlight can be changed by click to the Node Edit button
- [x] form output card highlight can be changed by form output card arrow up button
- [x] check deletion of the node that has currently highlighted form output card

Screenshot 2025-02-21 150405

PR #2940, opened 2025-02-26
NO-ISSUE: Expose own Shared values to Micro-frontends Multiplying Architecture Envelopes themselves

tiagobento:
During @jomarko's work on 

- https://github.com/apache/incubator-kie-tools/pull/2918

[I mentioned](https://github.com/apache/incubator-kie-tools/pull/2918#issuecomment-2675804896) that this work was a great opportunity for using [Shared values](https://github.com/apache/incubator-kie-tools/pull/691). However, a Shared value owned by an Envelope was currently not being exposed to it via EnvelopeClient, and therefore couldn't be used by the components inside an Envelope.

Since for @jomarko's work, the state of which Boxed Expression is currently selected is _owned_ by the DMN Editor, it only made sense for the `dmn-editor-envelope` to own this Shared value. Investigating with him, we weren't able to find a way to consume it, then after insisting, I noticed that [EnvelopeServer](https://github.com/apache/incubator-kie-tools/blob/main/packages/envelope-bus/src/channel/EnvelopeServer.ts) was exposing its Shared values via the `shared` getter, but [EnvelopeClient](https://github.com/apache/incubator-kie-tools/blob/main/packages/envelope-bus/src/envelope/EnvelopeClient.ts) wasn't. This PR changes that.

More than making EnvelopeClient expose its Shared values, I'm also adding a new `shared` property to `KogitoEditorEnvelopeContextType` so that all Editors running inside Envelopes can access Shared values they might come to declare.

This PR turned out somewhat large, as lots of places needed the EnvelopeApi as a new type parameter, but real changes are in:
- `KogitoEditorEnvelopeContext.ts`
  - Adding a new `shared` property
- `EnvelopeClient.ts`; and
  - Adding a new `shared` getter
- `examples/micro-frontends-multiplying-architecture-todo-list-view`
  - Adding a new Shared value (`todoList__itemsCount`) to the Envelope, and consuming it in the Channel.

---
A simple video showcasing the new `itemsCount` Shared value owned by `TodoListEnvelopeApi`.

https://github.com/user-attachments/assets/da3f40fd-c0e9-4a6e-89cd-4bf6b82c72ef

---

Also in this PR:
- Fixing the top-level `on-changed-deps-only` script to always exclude changed packages, even if they're dependencies of other changed packages.

### DIFF
--- a/examples/micro-frontends-multiplying-architecture-base64png-editor/src/Base64PngEditor.tsx
+++ b/examples/micro-frontends-multiplying-architecture-base64png-editor/src/Base64PngEditor.tsx
@@ -41,7 +41,7 @@ const INITIAL_INVERT = "0";
  * envelopeContext All the features and information provided by the Apache KIE Tools Envelope.
  */
 interface Props {
-  envelopeContext: KogitoEditorEnvelopeContextType<KogitoEditorChannelApi, KogitoEditorEnvelopeApi>;
+  envelopeContext: KogitoEditorEnvelopeContextType<KogitoEditorEnvelopeApi, KogitoEditorChannelApi>;
 }
 
 /**
--- a/examples/micro-frontends-multiplying-architecture-base64png-editor/src/Base64PngEditor.tsx
+++ b/examples/micro-frontends-multiplying-architecture-base64png-editor/src/Base64PngEditor.tsx
@@ -19,7 +19,7 @@
 
 import * as React from "react";
 import { useCallback, useEffect, useImperativeHandle, useMemo, useRef, useState } from "react";
-import { EditorApi, KogitoEditorEnvelopeContextType } from "@kie-tools-core/editor/dist/api";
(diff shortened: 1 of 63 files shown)


tiagobento:
@jomarko After this PR is merged, you'll be able to:

- Change `DmnEditorFactory` to be typed with `NewDmnEditorEnvelopeApi` instead of `KogitoEditorEnvelopeApi`
- Add a new Shared value to `NewDmnEditorEnvelopeApi`
- Expose a new callback `onSelectedBoxedExpressionChanged` or similar, on `DmnEditorRoot`, then
- Use `envelopeContext.shared` inside `DmnEditorFactory` to call the `set` method of the new Shared value.

💪

jomarko:
Thank you @tiagobento , I will review this ASAP

jomarko:
I am sorry, I will probably get to this tomorrow.

tiagobento:
Added a video to the description.

thiagoelg (approved):
LGTM

ljmotta (approved):

PR #2903, opened 2025-02-11
NO-ISSUE: Bump jQuery and Bootstrap version

danielzhe:
I updated Bootstrap to latest 4.x version and the latest jQuery version compatible with the bootstrap 4.x

PR #2949, opened 2025-02-28
NO-ISSUE: Fix TaskForm `transitions` API call when running the jBPM Dev UI in a different host/origin

thiagoelg:
- Replaces the baseUrl of the endpoint for the `transitions` request with the current origin baseUrl, just like all other API calls;
- Also fixes the dev app to run on both WSL and Windows.


tiagobento (approved):

thiagoelg:
@danielzhe Can you test this on Windows without WSL?
This reverts the changes to the `application.properties` from this [PR](https://github.com/apache/incubator-kie-tools/pull/2903), but adds a `win32` specific script for starting the dev app.

ljmotta (approved):

PR #2973, opened 2025-03-07
NO-ISSUE: Fix `pnpm bootstrap` in Windows.

danielzhe:
More info: https://stackoverflow.com/questions/78519711/toolchain-not-available-error-prevents-me-from-using-any-go-commands/78643571#78643571

### DIFF
--- a/packages/image-env-to-json/go.mod
+++ b/packages/image-env-to-json/go.mod
@@ -1,6 +1,6 @@
 module github.com/apache/incubator-kie-tools/packages/image-env-to-json
 
-go 1.22
+go 1.22.0
 
 require (
 	github.com/santhosh-tekuri/jsonschema/v6 v6.0.1
--- a/packages/image-env-to-json/go.mod
+++ b/packages/image-env-to-json/go.mod
@@ -1,6 +1,6 @@
 module github.com/apache/incubator-kie-tools/packages/image-env-to-json
 
-go 1.22.0
+go 1.22.12
 
 require (
 	github.com/santhosh-tekuri/jsonschema/v6 v6.0.1


thiagoelg:
What was the error you were getting when running bootstrap?

I remember it working fine on my end when I tested it 🤔 

I'm not exactly sure how go builds work, but will this force a version when building the package instead of using the locally installed toolchain (that could be an updated version, like `1.22.12`)? If so, maybe we should use `go 1.22.12` in the go.mod file.

danielzhe:
> What was the error you were getting when running bootstrap?
> 
> I remember it working fine on my end when I tested it 🤔
> 
> I'm not exactly sure how go builds work, but will this force a version when building the package instead of using the locally installed toolchain (that could be an updated version, like `1.22.12`)? If so, maybe we should use `go 1.22.12` in the go.mod file.

```
│ go: downloading go1.22 (windows/amd64)
│ go: download go1.22 for windows/amd64: toolchain not available

```
"Toolchain not available"

I'm running in Windows 11 native, not WSL.

thiagoelg:
> > What was the error you were getting when running bootstrap?
> > I remember it working fine on my end when I tested it 🤔
> > I'm not exactly sure how go builds work, but will this force a version when building the package instead of using the locally installed toolchain (that could be an updated version, like `1.22.12`)? If so, maybe we should use `go 1.22.12` in the go.mod file.
> 
> ```
> │ go: downloading go1.22 (windows/amd64)
> │ go: download go1.22 for windows/amd64: toolchain not available
> ```
> 
> "Toolchain not available"
> 
> I'm running in Windows 11 native, not WSL.

Yes, I was testing on WIndows 11 native as well, but I guess this is some recent changes with Go 1.22 (see https://github.com/golang/go/issues/66175#issuecomment-2035182462).

According to this issue (and commits like this one: https://github.com/golang/oscar/commit/e571ccd1de7b51c34fc9d88582e2878f65d3a5f3), we should probably set it to `go 1.22.12` in the go.mod files to match the go version we are using for the repo.

danielzhe:
@thiagoelg Cool! I'm changing it to 1.22.12 then. Thanks!

thiagoelg (approved):

tiagobento (approved):

PR #2975, opened 2025-03-07
kie-issues#1865: On the DMN Editor, the identifiers rename is not being applied to Inputs in Decision Tables.

### LINKED ISSUE #1865: On the DMN Editor, the identifiers rename is not being applied to Inputs in Decision Tables
Since the Inputs in the Decision Tables are now expressions, if an identifier inside of those expressions is renamed in its source it also needs to be renamed inside the expressions in the Inputs of the Decision Tables.

**Steps to Reproduce:**

1. Create an Input node "My Input"
2. Create a Decision node from the Input Node
3. Edit the Decision node
4. Add a Decision Table
5. Go back to diagram and rename the "My Input" to whatever name

It should show the message confirming the change and if the user wants to apply it where the identifier is referenced, but what is happening is that the rename is applied and the input in Decision Table gets broken.

https://github.com/user-attachments/assets/86afce48-e601-48c9-aac4-27ce870a58fe

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1865

https://github.com/user-attachments/assets/31c91df4-94cc-4ca1-b6f8-8f4a722f1f60

### DIFF
--- a/packages/dmn-feel-antlr4-parser/src/parser/IdentifiersRepository.ts
+++ b/packages/dmn-feel-antlr4-parser/src/parser/IdentifiersRepository.ts
@@ -33,6 +33,7 @@ import {
   DMN15__tFor,
   DMN15__tFunctionDefinition,
   DMN15__tInformationRequirement,
+  DMN15__tInputClause,
   DMN15__tInputData,
   DMN15__tInvocation,
   DMN15__tItemDefinition,
@@ -209,9 +210,9 @@ export class IdentifiersRepository {
   }
 
   private loadIdentifiersFromInputData(drg: DmnInputData) {
-    this.addIdentifier({
+    this.addIdentifierContext({
       uuid: drg["@_id"] ?? "",
-      name: drg["@_name"],
+      identifierDefinedByTheContext: drg["@_name"],
       kind: FeelSyntacticSymbolNature.GlobalVariable,
       parentContext: undefined,
       typeRef: drg.variable?.["@_typeRef"],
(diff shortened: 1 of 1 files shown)


packages/dmn-feel-antlr4-parser/src/parser/IdentifiersRepository.ts
```diff
@@ -209,9 +210,9 @@ export class IdentifiersRepository {
   }
 
   private loadIdentifiersFromInputData(drg: DmnInputData) {
-    this.addIdentifier({
+    this.addIdentifierContext({
```
danielzhe, line 213: This change is to make more clear what this method does. The previous name gives the false sense that we was just "adding identifiers" but the true is that we are adding identifiers context.

packages/dmn-feel-antlr4-parser/src/parser/IdentifiersRepository.ts
```diff
@@ -209,9 +210,9 @@ export class IdentifiersRepository {
...
       uuid: drg["@_id"] ?? "",
-      name: drg["@_name"],
+      identifierDefinedByTheContext: drg["@_name"],
```
danielzhe, line 215: Some contexts define identifiers, others are just contexts for identifiers. This needs some refactoring in the future to be optional.

packages/dmn-feel-antlr4-parser/src/parser/IdentifiersRepository.ts
```diff
@@ -759,38 +760,57 @@ export class IdentifiersRepository {
...
-      name: "",
+      identifierDefinedByTheContext: "",
       kind: FeelSyntacticSymbolNature.LocalVariable,
       parentContext: parent,
     });
     parent.children.set(ruleInputElementNode.uuid, ruleInputElementNode);
     this.addExpression(parent, entryNode);
   }
 
+  private addDecisionTableInputEntryNode(parent: IdentifierContext, inputEntryNode: DMN15__tInputClause) {
```
danielzhe, line 773: This is the real bug fix.

What we are doing here is telling to the repository "hey, this object that is in the Decision Table is a context that recognize variables and have this expression inside of it. Do your thing."

tiagobento (approved):

tiagobento:
Thank you @danielzhe. I love how simple it was to fix it. Great job on the renaming refactor feature!!!

thiagoelg (approved):

PR #2985, opened 2025-03-11
kie-issues#1870: New Test Scenario Editor: Add error page for invalid scesim file

### LINKED ISSUE #1870: New Test Scenario Editor: Add error page for invalid scesim file
Using BAMOE Developer tools to open an invalid file with the .scesim extension correctly shows an error message in the bottom right, but the main editor page is blank. The DMN and BPMN behavior for a similar scenario is to show some sort of error page as well as the error message in the bottom right.

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/1870

Screenshot 2025-03-11 at 14 39 43

In case of invalid scesim, now an error message is shown in the editor.
Removed `TestScenarioMarshallerFallbackError` component, which is a duplicate of a component already present in the `TestScenarioEditor` root component.

### DIFF
--- a/packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
+++ b/packages/scesim-editor-envelope/src/TestScenarioEditorRoot.tsx
@@ -52,7 +52,6 @@ export type TestScenarioEditorRootProps = {
 };
 
 export type TestScenarioEditorRootState = {
-  error: Error | undefined;
   externalModelsByNamespace: TestScenarioEditor.ExternalDmnsIndex;
   externalModelsManagerDoneBootstraping: boolean;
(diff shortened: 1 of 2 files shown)


jomarko (commented):
I think we should keep all editors aligned. So if this PR introduce new design for the opening error page, I think we should adopt it also for DMN and BPMN, and PMML? what do you think?

Currently DMN is different:
image

yesamer:
@jomarko Good point, I agree.

Done!

Screenshot 2025-03-12 at 11 09 43

ljmotta (approved):

thiagoelg (approved):

PR #2986, opened 2025-03-11
NO-ISSUE: Add light gray color to overflow scroll bar for new Test Scenario and DMN Editors

ljmotta:
## Description
This PR sets a light gray color to the overflow scroll bar, preventing channels with dark theme to set it to black.

### DMN Editor Before
image

### DMN Editor After
image

### Scesim Editor Before
image

### Scesim Editor After
image

### DIFF
--- a/packages/dmn-editor/src/DmnEditor.css
+++ b/packages/dmn-editor/src/DmnEditor.css
@@ -20,7 +20,7 @@
 /* (begin) global */
 .kie-dmn-editor--root {
   height: 100%;
-  scrollbar-color: var(--vscode-scrollbarSlider-background) #f1f1f1;
+  scrollbar-color: rgba(121, 121, 121, 0.4) #f1f1f1;
 }
 
 .kie-tools--dmn-editor--diagram-container {
--- a/packages/dmn-editor/src/DmnEditor.css
+++ b/packages/dmn-editor/src/DmnEditor.css
@@ -20,6 +20,7 @@
 /* (begin) global */
 .kie-dmn-editor--root {
   height: 100%;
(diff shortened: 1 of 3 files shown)


tiagobento (approved):

packages/dmn-editor/src/DmnEditor.css
```diff
@@ -20,6 +20,7 @@
 /* (begin) global */
 .kie-dmn-editor--root {
   height: 100%;
+  scrollbar-color: var(--vscode-scrollbarSlider-background) #f1f1f1;
```
jomarko: Don't we want to use `var` also instead of the constant `#f1f1f1` ? Patternfly seems to have a lot of 'grey' colors already defined, like `--pf-global--BorderColor--200`. Not sure, just asking. My idea was to minimize the amount of 'grey' colors we use in tooling. However, I understand scrollbars are more like host machine component that part of tooling.

jomarko (commented):
Thank you, I understand this is useful usability change. Just one question to the code.

yesamer (approved):

packages/dmn-editor/src/DmnEditor.css
ljmotta: @jomarko Firstly, I made this fix for `vscode-extension` package, and that's why I used the `vscode` variable. This fix targets any channel that may set the scrollbar-color, so I've update it to use the `rgba` equivalent. Regarding the PatetrnFly variables, they cover a few colors, and we still would need to manually define some of then (e.g. this `rgba` color), so I'm not sure if we should proceed to use those variables.

PR #2992, opened 2025-03-12
[NO-ISSUE] Always switch to Editor tab when programmatically opening a Boxed Expression in the DMN Editor

tiagobento:
If the DMN Editor is in the Data Types or Included Models tab when the "openBoxedExpressionEditor" method is called, it will look like it was a no-op, so settings the correct tab is important to make sure this action always puts the Boxed Expression Editor in the foreground.


danielzhe (approved):

ljmotta (approved):

thiagoelg (approved):

jomarko (commented):
Thank you team for this FIX! @danielzhe @tiagobento @ljmotta @thiagoelg

PR #2994, opened 2025-03-12
kie-issues#1871: New Test Scenario Editor doesn't show the file path in then settings and creation panels

### LINKED ISSUE #1871: New Test Scenario Editor doesn't show the file path in then settings and creation panels
The New Test Scenario Editor doesn't show the file path in the settings and creation panels tab, only the file name. This behavior makes it impossible to know which file the dropdown references when opening the `scesim` file in a workspace with multiple projects. The image below illustrates this behavior.

Image

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1871

image

image

### DIFF
--- a/packages/scesim-editor/src/creation/TestScenarioCreationPanel.tsx
+++ b/packages/scesim-editor/src/creation/TestScenarioCreationPanel.tsx
@@ -206,6 +206,8 @@ function TestScenarioCreationPanel() {
                 name="dmn-select"
                 onToggle={setModelSelectOpen}
                 isOpen={isModelSelectOpen}
+                typeAheadAriaLabel={"Select a model..."}
+                placeholderText={"Select a model..."}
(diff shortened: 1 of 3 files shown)


tiagobento:
What's the difference between `./Another model.dmn` and `Another model.dmn`? Why are they two separate entries?

danielzhe:
> What's the difference between `./Another model.dmn` and `Another model.dmn`? Why are they two separate entries?

Different folders. See the folder structure at the left? (Lol, I wrote "right")

tiagobento:
Ok, maybe I should go see my eye doctor. I saw `./` instead of `../`. Thanks @danielzhe, sorry!

tiagobento (approved):

ljmotta (approved):

thiagoelg (approved):

PR #3010, opened 2025-03-18
NO-ISSUE: Enable Test Scenario Editor Classic compatibility with the new Test Scenario Editor

yesamer:
File generated / edited with the new Test Scenario Editor are not compatible with the Classic Editor.

The reason relies on the namespace introduced in the new Editor `xmlns="https://kie.org/scesim/1.8"`: that is mandatory for the new Editor. Unfortunately, the classic editor's unmarshaller is not able to parse a scesim file with that namespace. 
I made a tentative to add the namespace in the unmarshalling process, but with that change, the namespace would be mandatory for the old editor too, breaking the compatibility with the old scesim files.

https://github.com/user-attachments/assets/16fdfb7c-df33-4b90-9a20-f17b8fa84f50

As a workaround, the old editor now removes the namespace from the scesim file string representation before getting unmarshalled. Not a nice solution, but I guess it's reasonable based on the future deprecation and removal plans of the old editors.


danielzhe (approved):

tiagobento:
We did bump the version of the SceSim XSD, right? So the old saves as 1.7, and the new one saves as 1.8? Are there any differences between them? From your proposed solution from interoperability I'm assuming no, but asking just in case. Thanks!

yesamer:
@tiagobento No, we didn't bump the scesim version. Both the old editor and Business Central are compatible with 1.8. 
If my memory serves me correctly, we bumped scesim from 1.7 to 1.8 in the 7.36.0 release. 
Please consider 

`<ScenarioSimulationModel version="1.8">`

The version attribute is used in all editors to determine the scesim file version. The namespace is not currently required.

tiagobento (approved):

PR #3013, opened 2025-03-19
kie-issues#1583: [sonataflow-operator] Add external built image integrity validation

### LINKED ISSUE #405: Add external built image integrity validation
### Description

As we allow external images set in the SonataFlow resource introduced by [KOGITO-9265](https://issues.redhat.com/browse/KOGITO-9265), we must guarantee that the informed image is valid.

This means that the .spec.flow definition must match with the workflow definition in the image. Without it, theoretically, a user could define a "hello world" workflow in the .spec.flow and a highly complex one in the image itself.

Having a conciliation between the flow and the one served by the image is important for the operator to configure correctly the deployment in the topology.

The operator won't deploy a SonataFlow instance if the given image integrity doesn't match the definition. Preferably, the operator should do static analysis in the image.

**Workarounds**
If use our tooling to generate CRs, this won't be a problem since we can control every aspect of the deployment, but won't prohibit one from changing the flow as they please.

### Implementation ideas

_No response_

treblereel:
Closes: https://github.com/apache/incubator-kie-kogito-serverless-operator/issues/405


packages/sonataflow-operator/container-builder/examples/dockerfiles/SonataFlow.dockerfile
```diff
@@ -40,6 +40,7 @@ COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/
 COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/quarkus-app/*.jar /deployments/
 COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/quarkus-app/app/ /deployments/app/
 COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/quarkus-app/quarkus/ /deployments/quarkus/
+COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/classes/workflow.sw.json /deployments/app/workflow.sw.json
```
ricardozanini: So, should the builder image add the workflow file to the `/deployments/app`? Is there no other way to do it? This file can also be a YAML named anything (`bananas.sw.yml`).

So, if we imagine a custom builder or an image built by users, they must know where to put the file in their environment.

Docs and migration from a previous version will be severely impacted. Can you also add a configuration to the manager to skip this validation? It will be needed in case of a migration from a previous version.

Currently, users may build their workflow images as they wish. To migrate to this new version, they might need to turn the validation off and turn it on once everything is migrated.

treblereel:
blocked by https://github.com/apache/incubator-kie-tools/pull/3028

packages/sonataflow-operator/internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,128 @@
+// Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede, line 1: wrong licence format, copyright not needed

packages/sonataflow-operator/internal/controller/sonataflow_controller.go
```diff
@@ -118,6 +122,11 @@ func (r *SonataFlowReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
+	if err := r.Validate(ctx, workflow, req); err != nil {
+		klog.V(log.E).ErrorS(err, "Failed to validate SonataFlow")
```
wmedvede, line 126: I think the error message should include the name / namespace of the failing workflow. That information will be very helpful when we need to see the logs in case of failues.

packages/sonataflow-operator/internal/controller/validation/Validator.go
```diff
@@ -0,0 +1,41 @@
...
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	operatorapi "github.com/apache/incubator-kie-tools/packages/sonataflow-operator/api/v1alpha08"
+)
+
+var validators = []Validator{ImageValidator{}}
+
+type Validator interface {
```
wmedvede, line 28: Nice approach!

packages/sonataflow-operator/internal/controller/validation/Validator.go
```diff
@@ -0,0 +1,41 @@
...
+	Validate(ctx context.Context, client client.Client, sonataflow *operatorapi.SonataFlow, req ctrl.Request) error
+}
+
+// validate the SonataFlow object, right now it's only check if workflow is in deployment has image declared as that image
```
wmedvede, line 32: nitpick, validate here, goes with uppercase Validate instead.

packages/sonataflow-operator/internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,128 @@
...
+type ImageValidator struct{}
+
+var pattern = `^.*\.sw\.(yaml|yml|json)$`
+
+var readablePattern = `*.sw.(yaml|yml|json)`
+var compiled = regexp.MustCompile(pattern)
+
+func (v ImageValidator) Validate(ctx context.Context, client client.Client, sonataflow *operatorapi.SonataFlow, req ctrl.Request) error {
+	if sonataflow.HasContainerSpecImage() {
+		err := client.Get(ctx, req.NamespacedName, sonataflow)
```
wmedvede, line 44: If we are validating a `workflow'  that was just read the instant before calling the Validators sequence, I don't believe we need to re-read it again.

But if you do, I don't believe we should read it and storing the value in the parameter sonataflow *operatorapi.SonataFlow paramter, I think this could make the validator has side effects on that variable value. When multiple validators are executed I think this code wont be clear.

Another nitpick, when we do reading of the resource as part of the reconciliation, we normally guard the code with something like this:

```
	err := r.Client.Get(ctx, req.NamespacedName, workflow)
	if err != nil {
		if errors.IsNotFound(err) {
			not good, return the error or whatever, but we know exactly that the given resource don't exist at this time.
		}
		not good, return the error or whatever, something went wrong, we don't know what.
	}
```

packages/sonataflow-operator/internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,128 @@
...
+		equals, err := validateImage(sonataflow, sonataflow.Spec.PodTemplate.Container.Image)
```
wmedvede, line 45: if we don't do the consideration above, in case of error the sonataflow value is uncertain.

packages/sonataflow-operator/internal/controller/sonataflow_controller.go
wmedvede: This invocation to the validator should go after the code below.
That `if` indicates that the WF being deleted right now, so, we don't need to validate it.

	if workflow.DeletionTimestamp != nil {
		return r.applyFinalizers(ctx, workflow)
	}

packages/sonataflow-operator/internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,128 @@
...
+		if err != nil {
+			return err
+		}
+		if !equals {
+			return fmt.Errorf("Workflow, defined in the image %s doesn't match deployment workflow", sonataflow.Spec.PodTemplate.Container.Image)
```
wmedvede, line 50: workflow name/namepace would be nice to include in the message.

packages/sonataflow-operator/internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+
+	return nil, nil
+}
+
+func jsonFromDockerImage(reader io.Reader) (operatorapi.Flow, error) {
+	data, err := io.ReadAll(reader)
+	workflow := &operatorapi.Flow{}
+	if err = yaml.Unmarshal(data, workflow); err != nil {
```
wmedvede, line 124: Excuse me if I'm missing something, but do this Unmarshalling work both when the workflow inside the image is in json and yaml format? If so, maybe this function can be called workflowFromDockerImage.

packages/sonataflow-operator/internal/controller/validation/Validator.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede, line 1: same here, copyright we don't need

wmedvede (changes requested):

wmedvede:
Hi @treblereel I had tried the feature in OpenShift 4.15.17 with the following scenario

1. I have a workflow with a pre-built image
2. In SonataFlow I modify the `flow` intentionally to force validation to fail
3. When I deploy the workflow I can see this message:

`E0401 16:52:11.167885 1 sonataflow_controller.go:126] "Failed to validate SonataFlow" err="Get \"https://quay.io/v2/\": tls: failed to verify certificate: x509: certificate signed by unknown authority"`

Screenshot from 2025-04-01 18-55-10

``

It looks like it was not possible for the validator to read the image,

wmedvede:
@treblereel I'd recommend rebasing with main, since there has been some recent updates in sontaflow_controller.go

ricardozanini:
@treblereel the docker api needs to authenticate with OpenShift internal registry

domhanak:
@treblereel This PR is stale, could you please update it or close if not relevant anymore? Thanks.

tiagobento:
Closing as `sonataflow-operator` doesn't exist anymore in the 10.3.x+ stream.

PR #3020, opened 2025-03-20
[sonataflow-operator] NO-ISSUE: Removed hard-coded control-plane labels from operator manifests

ricardozanini:
Due to a problem with Community Operators CI blocking the 10.0.0 release on OpenShift Hub, we have to remove the `control-plane` label from the operator's manager. See https://github.com/k8s-operatorhub/community-operators/pull/5428.

Instead, we are using the `app.kubernetes.io/name` label from the common labels pattern. It will allow us to pass the CI and align our manifests with Kube patterns.

@wmedvede @jakubschwan, this is a critical update. Care to test it on your side, too, using OpenShift? I tested it locally and it's working accordingly.

@jakubschwan the main use case here is the upgrade from one version to another with these new labels.


wmedvede (approved):
LGTM!

jakubschwan (approved):
lgtm

PR #3031, opened 2025-03-26
[kie-issues#1878] Make it possible to import Data Types from Java classes on the new DMN Editor

### LINKED ISSUE #1878: Make it possible to import Data Types from Java classes on the new DMN Editor
Currently, the DMN 1.5 editor does not support importing Java classes. (It was available on the previous editor). See attached Image for how the previous DMN Editor did it.

For the new version, since the Data Types tab is slightly different, we can add a new action under the extra items in the header of the side panel (color, font, and icons are just for illustrating):

Image

A new empty state option needs to be added to, for when there are no Data Types created yet (color, font, and icons are just for illustrating):

Image

This option should not appear when DMN Editor is ran outside of VS Code Extensions.

When clicking on the "Import from Java class" buttons, a Modal should appear, similarly to what we had in the previous version.

Making this possible will require changes in the DMN Editor's component props, as well as integration with the Channel where it is running, for providing Java classes introspection capabilities.

Adarshvk98:
Closes [issue #1878](https://github.com/apache/incubator-kie-issues/issues/1878)  

Based on the initial development PR [#751](https://github.com/apache/incubator-kie-tools/pull/751).

**Showcase :** 

https://github.com/user-attachments/assets/912806da-cf80-4bec-8ab8-9bf12453e4d9

### Key Features

1. Integrates the Import Java Classes Component into the New DMN Editor
2. This feature is only available when the DMN Editor is run inside VS Code Extensions

### Important Considerations

When importing Java classes, the system checks for existing DMN data types that match the selected classes. If a match is found, the user is prompted to decide how to proceed. The user can choose to either:

- Replace the existing type
- Keep both versions as separate types (following the DMN Data Type naming convention: `<class-name>, or <class-name>-<count>` if multiple classes have the same name)

### Future Improvements

- Currently, attributes converted to DMN types do not support lists (collections). it will be good to add support in near future.


yesamer:
Well done @Adarshvk98! Can you please confirm the feature is still working in the old editor with your changes?

jomarko:
@Adarshvk98 the build is failing
```
 [run-script-if] Finished 'kie-tools--eslint ./src'
  Error: src/vscode/VsCodeNewDmnEditorFactory.tsx(56,51): error TS2314: Generic type 'EditorFactory<E, EnvelopeApi, ChannelApi>' requires 3 type argument(s).
  Error: src/vscode/VsCodeNewDmnEditorFactory.tsx(58,22): error TS2314: Generic type 'KogitoEditorEnvelopeContextType<EnvelopeApi, ChannelApi>' requires 2 type argument(s).
  ```

**Comment by Adarshvk98** [comment]
> Well done @Adarshvk98! Can you please confirm the feature is still working in the old editor with your changes?

@yesamer, yeah sure, I'll check and confirm.

**Comment by Adarshvk98** [comment]
> @Adarshvk98 the build is failing
> 
> ```
>  [run-script-if] Finished 'kie-tools--eslint ./src'
>   Error: src/vscode/VsCodeNewDmnEditorFactory.tsx(56,51): error TS2314: Generic type 'EditorFactory<E, EnvelopeApi, ChannelApi>' requires 3 type argument(s).
>   Error: src/vscode/VsCodeNewDmnEditorFactory.tsx(58,22): error TS2314: Generic type 'KogitoEditorEnvelopeContextType<EnvelopeApi, ChannelApi>' requires 2 type argument(s).
> ```

I'm checking...

**Comment by Adarshvk98** [comment]
> Well done @Adarshvk98! Can you please confirm the feature is still working in the old editor with your changes?

I’ve checked in the old editor, and it’s working fine.

**Review by jomarko — COMMENTED** [review]
Thank you for the PR. Here is my feedback in few points.

### 01 DESKTOP vs WEB
In code I see `settings.channelType === ChannelType.VSCODE_DESKTOP || settings.channelType === ChannelType.VSCODE_WEB`. I have tried just a desktop. Is here anyone who tried web?

### 02 search
imagine there is class `Person`. User is not able to search it using string like `son` or `person`. User is able to search it just using the proper lower/upper case match and the `startsWith` letters. For example `Pers` will find `Person`

maybe we should do similar updates as we do in https://github.com/apache/incubator-kie-tools/pull/3033 ?

### 03 select multiple classes
Once I selected two classes `Person` and `Person2`, the first one is always imported/loaded as empty structure. However, checking java code, `Person` it is not empty.
Screenshot 2025-04-01 114910
Screenshot 2025-04-01 114903

### 04 Activator hint
is currently: 
```
Not available, please check if 'Language Support for Java by Red Hat' extension is correctly installed and the Activator Java Class is present in your project.
```
from the ^ it is not so clear to me it should look like:https://github.com/yesamer/dmn-javaclasses-example/blob/main/src/main/java/Activator.java

Points 2 and 3 are probably also in old editor? Probably we can handle them in separate tickets potentially? What is your opinion aboit 1 and 4?

**Comment by Adarshvk98** [comment]
> Thank you for the PR. Here is my feedback in few points.
> 
> ### 01 DESKTOP vs WEB
> In code I see `settings.channelType === ChannelType.VSCODE_DESKTOP || settings.channelType === ChannelType.VSCODE_WEB`. I have tried just a desktop. Is here anyone who tried web?
> 

- I'm not sure about testing in VSCode Web, as I haven't explored this option before. I couldn't find a way to install a VSIX file in the Extensions tab. If someone could guide me or help me do so, I can test it.

> ### 02 search
> imagine there is class `Person`. User is not able to search it using string like `son` or `person`. User is able to search it just using the proper lower/upper case match and the `startsWith` letters. For example `Pers` will find `Person`
> 
> maybe we should do similar updates as we do in #3033 ?
> 
> ### 03 select multiple classes
> Once I selected two classes `Person` and `Person2`, the first one is always imported/loaded as empty structure. However, checking java code, `Person` it is not empty. Screenshot 2025-04-01 114910 Screenshot 2025-04-01 114903
> 

-  I agree that there is room for improvement in these areas. However, since I haven't touched the overall functionality of the `import-java-classes-component` package (apart from rearranging the component to make it reusable in the new DMN editor), I think these improvements can be considered for a later stage ?

> ### 04 Activator hint
> is currently:
> 
> ```
> Not available, please check if 'Language Support for Java by Red Hat' extension is correctly installed and the Activator Java Class is present in your project.
> ```
> 
> from the ^ it is not so clear to me it should look like:https://github.com/yesamer/dmn-javaclasses-example/blob/main/src/main/java/Activator.java
> 

- The error message is likely an existing message in the old DMN editor and can be considered an improvement.

> Points 2 and 3 are probably also in old editor? Probably we can handle them in separate tickets potentially? What is your opinion aboit 1 and 4?

- I think the current goal is to bring the same feature from the old DMN editor to the new one. While the mentioned points are improvements, I'm not sure if they should be addressed along with this or as a separate tickets. I'd like to hear your thoughts on this, @tiagobento  and @yesamer.

**Comment by jomarko** [comment]
### 01 DESKTOP vs WEB
try to search for "run:webmode" in package.json to run in WEB mode

**Comment by yesamer** [comment]
@Adarshvk98 I agree with you, the scope of this PR is to move the same existing functionality in the new editor, preserving all the existing features. If there's space for minor improvements, I guess we can consider it. 
However, about points 2 and 3 I would check if the same behavior reproduces in the old editor as well, in particular, point 3 it's very weird. Can you please check it?

**Comment by Adarshvk98** [comment]
About point 3, I checked the old editor, and it’s behaving a bit strangely. I tried with a similar example, using “Book” and “Book1” as classes. Initially, when importing two classes, it correctly displayed their attributes and expanded them. However, the second time I tried, the “Book” or “Book1” structure sometimes showed an empty structure.
image

Regarding point 2, I believe the function `getClasses` from `javaCodeCompletionService` handles the search, comparison, and retrieval of Java classes. As @jomarko  requested, we might need to modify the logic from the Java side. don’t have much control over the component side. Please correct me if I’m wrong about it. @yesamer

**Comment by yesamer** [comment]
@Adarshvk98 Thank you for checking, it seems to me that behavior is not a regression introduced in your changes, so I guess we can analyze them in a separate ticket WDYT @jomarko ?

@ljmotta Can you please review it as well?

**Review by jomarko — APPROVED** [review]
Approving the PR as my understanding is this:

### 1
 will be checked once the new VSIX extension is released. Currently for local testing we are blocked because `vscode-test-web` have trouble to install multiple extensions - we need java language support next to dmn vs code extension.
 
 
 ### 2
 Will have a separate ticket
 
 ### 3
 Will have a separate ticket
 
 ### 4
 There does not seem to be enough agreement how to improve. We will keep as it is.

File: packages/dmn-editor/src/dataTypes/DataTypes.tsx [hunk 1]
```diff
@@ -105,6 +111,18 @@ export function DataTypes() {
   );
   const dataTypesTree = useDmnEditorStore((s) => s.computed(s).getDataTypes(externalModelsByNamespace).dataTypesTree);
 
+  const [isOpen, setOpen] = useState(false);
```
yesamer: I would suggest using a more descriptive variable name. What is open?

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClasses.tsx
```diff
@@ -18,34 +18,91 @@
...
+    {...props}
+  />
+);
+
+const ImportJavaClasses = ({
+  javaCodeCompletionService,
+  gwtLayerService,
+  loadJavaClassesInDataTypeEditor,
+}: ImportJavaClassesProps) => {
+  const [isOpen, setOpen] = useState(false);
```
yesamer: @Adarshvk98 same

yesamer (approved):
@Adarshvk98 Thank you for implementing this important feature, you did a great job! Approving, I just have a couple of minor things.

packages/dmn-editor/src/dataTypes/DataTypes.tsx
Adarshvk98: Done - I’ve added a descriptive variable name for the variable.

packages/import-java-classes-component/src/components/ImportJavaClasses/ImportJavaClasses.tsx
Adarshvk98: I’ve added a descriptive variable name for the variable here as well.

packages/dmn-editor/src/dataTypes/ImportJavaClasses.tsx
```diff
@@ -0,0 +1,178 @@
...
+import * as React from "react";
+import { useCallback, useState } from "react";
+import { DropdownItem } from "@patternfly/react-core/dist/js/components/Dropdown";
+import { Spinner } from "@patternfly/react-core/dist/esm/components/Spinner";
+import PlusIcon from "@patternfly/react-icons/dist/js/icons/plus-icon";
+import { Tooltip } from "@patternfly/react-core/dist/js/components/Tooltip";
+import { Modal, ModalVariant } from "@patternfly/react-core/dist/js/components/Modal";
+import { Button } from "@patternfly/react-core/dist/js/components/Button";
+import { TextContent, Text, TextVariants } from "@patternfly/react-core/dist/js/components/Text";
+import { HelperText, HelperTextItem } from "@patternfly/react-core/dist/js/components/HelperText";
```
jomarko:
```suggestion
```

it is unused

packages/dmn-editor/src/dataTypes/useImportJavaClasses.ts
```diff
@@ -0,0 +1,282 @@
...
+        return { ...javaClass, name: newName } as JavaClass;
+      });
+
+      // Return the updated Java classes with renamed types
+      return updatePropertiesReferences(renamedJavaClasses, javaClassNameToDMNTypeNameMap);
+    },
+    [updatePropertiesReferences, buildName]
+  );
+
+  const generateUniqueDMNTypeNames = useCallback(
```
jomarko: interesting, comparing `generatingUniqueDMNTypeNames` with the line 60 in the same file I am not sure, do we always have the same logic for generating a unique name?

jomarko (commented):
Thank you for PR. Before an approval, I have one very specific corner case to discuss.

### A.dmn
Lets say this file has data type named `Person`

### B.dmn
Lets say this file includes `A.dmn` using default namesspace. So there is no prefix if `B.dmn` file want to access `Person` defined in `A.dmn`. Now imagine `A.dmn` want to import data type from  a java class also named `Person`. The feature correctly detect conflict [1], however if there is conflict with and external data type, we should probably not allow the option `Replace` what do you think?



### [1]
Screenshot 2025-04-04 154753

packages/dmn-editor/src/dataTypes/ImportJavaClasses.tsx
Adarshvk98: Thanks for pointing that out. I’ve removed the import statement

packages/dmn-editor/src/dataTypes/useImportJavaClasses.ts
Adarshvk98: There are two cases in which we need to rename the DMN type. The first case occurs when there are conflicting class names within the selected Java classes by the user. The second case happens when there is a conflict of class names from the existing DMN type.

The `generatingUniqueDMNTypeNames` function is used to solve the second case. Therefore, there are a few logical changes required, so I keep it separate. I’m not sure if this is what you were trying to point out here.

jomarko: ok

jomarko:
@Adarshvk98 please, what is your opinion about my point of conflicts with external data types https://github.com/apache/incubator-kie-tools/pull/3031#pullrequestreview-2742951305

Adarshvk98:
> @Adarshvk98 please, what is your opinion about my point of conflicts with external data types [#3031 (review)](https://github.com/apache/incubator-kie-tools/pull/3031#pullrequestreview-2742951305)

@jomarko you’re probably right. I think we shouldn’t let user replace the external modal types. but I'm not very sure about what action can be taken further , @tiagobento @yesamer 

A related scenario to consider is what happens when an external model is included that has a type that already exists in our DMN types?

yesamer:
> A related scenario to consider is what happens when an external model is included that has a type that already exists in our DMN types?

Good point @Adarshvk98. I would check that case. BTW, In my opinion, the case reported by @jomarko deserves attention, I would suggest opening a separate ticket for that.

packages/dmn-editor/src/dataTypes/useImportJavaClasses.ts
tiagobento:
```suggestion
  const generateUniqueDmnTypeNames = useCallback(
```

tiagobento (changes requested):
Please take my small comment about acronyms into consideration, and also somethings we discussed between ourselves:
- There's no need for the `window` object to be used as means to give the DmnEditor component access to methods related to importing Java classes. You can use `envelopeContext` directly in `dmn-editor-envelope` and remove the indirection.

Thanks!!

packages/dmn-editor/src/dataTypes/useImportJavaClasses.ts
Adarshvk98: I’ve made the changes. Please take a look

packages/dmn-editor/src/dataTypes/DataTypes.tsx
```diff
@@ -230,6 +243,16 @@ export function DataTypes() {
                         position={DropdownPosition.right}
                         isOpen={isAddDataTypeDropdownOpen}
                         dropdownItems={[
+                          settings?.isImportDataTypesFromJavaClassesSupported ? (
+                            <ImportJavaClassesI18nDictionariesProvider key={"import-java-classes"}>
+                              <ImportJavaClassesDropdownItem
+                                javaCodeCompletionService={settings?.javaCodeCompletionService!}
```
thiagoelg: This can't be `undefined` as per the `ImportJavaClassesDropdownItem` props type... Maybe add a check to the conditional above.

packages/dmn-editor/src/dataTypes/DataTypes.tsx
```diff
@@ -230,6 +243,16 @@ export function DataTypes() {
...
+                                onClick={handleImportJavaClassButtonClick}
+                              />
+                            </ImportJavaClassesI18nDictionariesProvider>
+                          ) : (
+                            <></>
```
thiagoelg: This will add an empty Dropdown Item if `settings?.isImportDataTypesFromJavaClassesSupported` is false. Won't that look weird?

packages/dmn-editor/src/dataTypes/DataTypes.tsx
```diff
@@ -311,6 +334,23 @@ export function DataTypes() {
                   editItemDefinition={editItemDefinition}
                 />
               )}
+              {isOpenImportJavaClassesWizard && (
+                <ImportJavaClassesI18nDictionariesProvider>
+                  <ImportJavaClassesWizard
+                    javaCodeCompletionService={settings?.javaCodeCompletionService!}
```
thiagoelg: Same issue here, `javaCodeCompletionService` can't be undefined, add it to the check above.

packages/dmn-editor/src/dataTypes/DataTypesEmptyState.tsx
```diff
@@ -54,9 +55,17 @@ export function DataTypesEmptyState({ onAdd, onPaste }: { onAdd: () => void; onP
               </Button>
             </EmptyStatePrimary>
             <br />
-            <br />
+            {settings?.isImportDataTypesFromJavaClassesSupported && (
+              <>
+                or
+                <br />
+                <br />
+                <ImportJavaClassesWrapper javaCodeCompletionService={settings?.javaCodeCompletionService!} />
```
thiagoelg: Last one 😄

packages/dmn-editor/src/dataTypes/ImportJavaClasses.tsx
```diff
@@ -0,0 +1,179 @@
...
+};
+
+const ImportJavaClassNameConflictsModal = ({
+  isOpen,
+  handleConfirm,
+  conflictsNames,
+}: {
+  isOpen: boolean;
+  handleConfirm: (options: JavaClassConflictOptions) => void;
+  conflictsNames: JavaClass[];
```
thiagoelg:
```suggestion
  confliectedJavaClasses: JavaClass[];
```

I think this better represents what the prop actually is, WDYT?

thiagoelg (commented):
This looks great, thanks @Adarshvk98!

Left some inline comments with some nitpicks regarding variable naming and undefined checks reported by the CI.

packages/dmn-editor/src/dataTypes/DataTypes.tsx
Adarshvk98: Thanks for taking the time to review this. I’ve made the changes. Please have a look.

packages/dmn-editor/src/dataTypes/DataTypes.tsx
Adarshvk98: Yes, I agree. I made those changes as well

packages/dmn-editor/src/dataTypes/DataTypes.tsx
Adarshvk98: done

packages/dmn-editor/src/dataTypes/DataTypesEmptyState.tsx
Adarshvk98: done

packages/dmn-editor/src/dataTypes/ImportJavaClasses.tsx
Adarshvk98: `confliectedJavaClasses` will be a good one. Thanks. I’ve renamed the variable.

thiagoelg (approved):

packages/dmn-editor/src/dataTypes/DataTypes.tsx
```diff
@@ -169,13 +182,37 @@ export function DataTypes() {
...
+        <ImportJavaClassesI18nDictionariesProvider key={"import-java-classes"}>
+          <ImportJavaClassesDropdownItem
+            javaCodeCompletionService={javaCodeCompletionService}
+            onClick={handleImportJavaClassButtonClick}
+          />
+        </ImportJavaClassesI18nDictionariesProvider>
+      );
+    }
+    return dropdownItems;
+  }, [isImportDataTypesFromJavaClassesSupported, javaCodeCompletionService]);
```
tiagobento: Missing dependency hooks here...

Adarshvk98: I was wondering if it was really necessary to pass the remaining `handleImportJavaClassButtonClick` and `pasteTopLevelItemDefinition` as dependencies?  because I think the only thing that would change would be the `isImportDataTypesFromJavaClassesSupported` boolean flag and the `javaCodeCompletionService`. I’m not sure if we should add them or not. WDYT?

Adarshvk98: Added the missing dependencies.

tiagobento: Adding dependencies pointed by ESLint helps us make sure our React apps are adhering to good reactivity practices. It's a defensive practice we do to not lose control when things change.

Although some people recommend not memoizing by default (I.e., useMemos and useCallbacks), or carefully crafting the list of dependencies for each hook, we found out that standardizing the use of memoization hooks combined with the exhaustive list of dependencies of that hook helped us keep the codebase resilient to changes.

That's more a convention on our side than anything else, really 😊

Adarshvk98: Okay, I agree. It makes sense now.

packages/dmn-editor-envelope/src/vscode/VsCodeNewDmnEditorFactory.tsx
```diff
@@ -0,0 +1,60 @@
...
+  JavaCodeCompletionChannelApi,
+} from "@kie-tools-core/vscode-java-code-completion/dist/api";
+
+import { DmnEditorInterface } from "../DmnEditorFactory";
+
+/**
+ * EXPOSED INTEROP API
+ *
+ * This API is exposed from the Envelope to be consumed on Java code.
+ */
```
tiagobento: Is this comment true here? Sorry I know there's been a lot of back and forth already but I guess this is not the case this usage of this type, right? It **was** true when we were still using the `window` object, but not anymore. In fact, I think this type might not even be being used right now. Can you please check @Adarshvk98?

tiagobento (changes requested):
Sorry, found something that needs double-checking.

packages/dmn-editor-envelope/src/vscode/VsCodeNewDmnEditorFactory.tsx
Adarshvk98: @tiagobento An interface `JavaCodeCompletionExposedInteropApi` is defined in line 41 and is being used. I believe this also refers to the same meaning as mentioned in the comments ?

jomarko:
@Adarshvk98 @tiagobento please let me know, hwat is the status of this PR. I would like to start my work on https://github.com/apache/incubator-kie-issues/issues/1912 that will touch same code area as #3031

packages/dmn-editor-envelope/src/vscode/VsCodeNewDmnEditorFactory.tsx
tiagobento: What I mean is that this interface no longer represents code that is called from Java code. It's simply a regular interface being used on `DmnEditorRoot`. My suggestion is that we remove the comment and move it to `DmnEditorRoot`.

Adarshvk98: Changes have been made

tiagobento (approved):

PR #3032, opened 2025-03-27
kie-issues#1889: Single-parameter number() function

### LINKED ISSUE #1889: Single-parameter number() function
Update the NumberFunction class - add new invoke() method with one parameter. It needs to be able to handle both String or Number parameter type. In case the parameter is string, it needs to behave based on this description approved in OMG DMN RTF group: 

Image

Basically it need to be able to convert a String representation of a number using just the string parameter. If the parameter is a number, it needs to behave based on this description approved in OMG DMN RTF group: 

Image

Columns of the table are: function definition, parameter type, return value, example. 

Note:
please create unit-test specifically targeting, directly, the Number function

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/1889

Added the new number(from) function in the Monaco editor autocompletion entries

### DIFF
--- a/packages/feel-input-component/src/FeelConfigs.ts
+++ b/packages/feel-input-component/src/FeelConfigs.ts
@@ -1073,8 +1073,8 @@ export const feelDefaultSuggestions = (): Monaco.languages.CompletionItem[] => {
         label: "number(from)",
         insertText: "number($1)",
         description: "Converts `from` to a number.",
-        parameters: [["from", "`string` representing a valid number"]],
-        examples: ['number( "1.1" ) = number( "1.1", "null", "null" ) = 1.1'],
+        parameters: [["from", "`string` or `number` representing a valid number"]],
+        examples: ['number( "1.1" ) = number( "1.1", "null", "null" ) = 1.1', "number( 5 ) = 5"],
       },
       {
         label: "number(from, grouping separator, decimal separator)",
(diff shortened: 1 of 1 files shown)


packages/feel-input-component/src/FeelConfigs.ts
```diff
@@ -1069,6 +1069,13 @@ export const feelDefaultSuggestions = (): Monaco.languages.CompletionItem[] => {
         parameters: [],
         examples: ["now()"],
       },
+      {
+        label: "number(from)",
+        insertText: "number($1)",
+        description: "Converts `from` to a number.",
+        parameters: [["from", "`string` representing a valid number"]],
```
jomarko: is it really a string?

looking on related backend pr https://github.com/apache/incubator-kie-drools/pull/6289/files

```
{"number(1947)", BigDecimal.valueOf(1947), null},
```

or
```
     * @param from : the input object to be converted into a BigDecimal.
```

jomarko (commented):
thank you yeser, I will do not manual review as the change is small, just one question before approval

packages/feel-input-component/src/FeelConfigs.ts
yesamer: @jomarko You're right, it can be a string OR a number

yesamer: Done.

jomarko (approved):

yesamer:
@ljmotta or @thiagoelg Can you please approve it? :)

thiagoelg (approved):

PR #3040, opened 2025-04-01
kie-issues#1512: DMN Editor: Context result row change causes expression data type to be reverted

### LINKED ISSUE #1512: DMN: Context result row change causes expression data type is reverted
There is an interesting issue in the context boxed expression.  In case user changes context root data type more times and then tries to change the result row value, the root data type is unexpectedly reverted.

Please follow the steps below to understand the issue.
- create new node - *A*
- set node *A* type as number
- open node *A*
- set it root expression as *context*
- set the context result row as literal expression and put a numeric value, for example *1*
- now in the context header cell, change the data type to *Any*
- now double click the result row value *1* to edit it, change it to new value, *2*
- once this change is committed, the context header is unexpectedly reverted from *Any* to *number*

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-issues/issues/1512
Before fix:
Data type/label is getting changed to the initially selected value.
After fix:
Data type/label is with the updated values irrespective of changing values.

### DIFF
--- a/packages/boxed-expression-component/src/expressions/ContextExpression/ContextResultExpressionCell.tsx
+++ b/packages/boxed-expression-component/src/expressions/ContextExpression/ContextResultExpressionCell.tsx
@@ -66,8 +66,6 @@ export function ContextResultExpressionCell(props: {
           const ret: Normalized<BoxedContext> = {
             ...prev,
             contextEntry: newContextEntries,
-            "@_label": newExpression?.["@_label"] ?? prev["@_label"],
-            "@_typeRef": newExpression?.["@_typeRef"] ?? prev["@_typeRef"],
           };
 
           return ret;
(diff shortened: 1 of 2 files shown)


yesamer (approved):
LGTM, thanks @Kusuma04-dev

jomarko (commented):
Thank you @Kusuma04-dev . I can confirm the fix works. However, I believe we should add tests to similar changes. I have prepared a PR with tests https://github.com/Kusuma04-dev/incubator-kie-tools/pull/1. Once merged and the build is green I can approve #3040.

jomarko (approved):

Kusuma04-dev:
3 tests which are failing here are not related to the changes in this PR but related to https://github.com/apache/incubator-kie-tools/pull/3045

tiagobento (approved):

PR #3049, opened 2025-04-06
[sonataflow-operator] Knative outdated revisions are not cleaned-up for workflows with the knative deploymentModel and gitops profile

### LINKED ISSUE #3048: [sonataflow-operator] Knative outdated revisions are not cleaned-up for workflows with the knative deploymentModel and gitops profile
When we deploy a workflow with the gitops profile that uses the knative deploymentModel, the outdated knative revisions are not being cleaned up.

 
When we deploy workflows that are linked with the knative eventing system, as part of the initialization procedure the K_SINK env variable is injected by the knative operator. That's part of the regular initialization. Additionally, when we use the knative deploymentModel, that injection produces a new knative serving revision, which is also part of the regular initialization.

However, when the injection is produced in revision N (without the K_SINK injected) to produce the revision N+1 (with the K_SINK injected), revision N remains in the system but is inoperable, since lacks of the K_SINK. This produces a negative impact since the initialization pod might remain trying to restart due to the K_SINK health check, etc.

 
So, considering that revision N is completely useless, we are basically removing it.

wmedvede:
Closes: #3048

### DIFF
--- a/packages/sonataflow-operator/internal/controller/profiles/preview/states_preview.go
+++ b/packages/sonataflow-operator/internal/controller/profiles/preview/states_preview.go
@@ -229,7 +229,7 @@ func (h *deployWithBuildWorkflowState) Do(ctx context.Context, workflow *operato
 
 func (h *deployWithBuildWorkflowState) PostReconcile(ctx context.Context, workflow *operatorapi.SonataFlow) error {
 	// Clean up the outdated Knative revisions, if any
-	if err := knative.CleanupOutdatedRevisions(ctx, h.C, h.Cfg, workflow); err != nil {
+	if err := knative.CleanupOutdatedRevisions(ctx, h.Cfg, workflow); err != nil {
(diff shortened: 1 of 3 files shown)


packages/sonataflow-operator/internal/controller/profiles/preview/states_preview.go
```diff
@@ -238,7 +229,10 @@ func (h *deployWithBuildWorkflowState) Do(ctx context.Context, workflow *operato
 
 func (h *deployWithBuildWorkflowState) PostReconcile(ctx context.Context, workflow *operatorapi.SonataFlow) error {
 	// Clean up the outdated Knative revisions, if any
-	return h.cleanupOutdatedRevisions(ctx, workflow)
+	if err := knative.CleanupOutdatedRevisions(ctx, h.C, h.Cfg, workflow); err != nil {
```
ricardozanini: You don't need to pass the client since it's globally accessible now.

wmedvede: good catch

ricardozanini (approved):

PR #3066, opened 2025-04-16
kie-issues#1912: DMN VS Code Extension: Evaluation highlights feature should not be available

### LINKED ISSUE #1912: DMN VS Code Extension: Evaluation highlights feature should not be available
The DMN VS Code extension has available the 'evaluation highlights' feature, or at least the toggle for its activation, however it should not be available in the VS Code. It should be available in the Sandbox only.

This is because the evaluation highlights works only together with the Dmn Runner component.

Image

jomarko:
Closes: https://github.com/apache/incubator-kie-issues/issues/1912

### Before
Screenshot 2025-04-16 113413

### After
Screenshot 2025-04-16 113253


packages/dmn-editor/src/DmnEditor.tsx
```diff
@@ -167,6 +167,14 @@ export type DmnEditorProps = {
    * This is shown on the ErrorBoundary fallback component, when an uncaught error happens.
    */
   issueTrackerHref?: string;
+  /**
+   * Dmn Editor Overlays panel contains 'Evaluation Highlights' toggle
+   * However this toggle is supported only for specific channels, (currently Sandbox only)
+   * The Dmn Editor should not have knowledge about the channel it is opened in.
+   * Use this boolean flag to allow or disallow the feature use.
+   * Default value is false.
```
klbowers: Maybe me being slightly nitpicky, but I think this comment belongs more in the PR description rather than here.
I feel like this could be simplified to follow the format of the others, e.g.
A flag to enable 'Evaluation Highlights' on the DMN Editor

klbowers (approved):
Thanks @jomarko, looks good to me! Left a minor comment, but maybe others have a different opinion.

yesamer (approved):
LGTM @jomarko

packages/dmn-editor/src/DmnEditor.tsx
thiagoelg: "A flag to enable 'Evaluation Highlights' on supported channels (only ONLINE for now)"

I think this would be enough to give context.

thiagoelg (commented):
Should the standalone DMN Editor support Evaluation Highlights? It uses the STANDALONE channel type.

jomarko:
@thiagoelg @kbowers-ibm thank you, feedback incorporated

_Should the standalone DMN Editor support Evaluation Highlights? It uses the STANDALONE channel type._ - I do not think so.

thiagoelg (approved):

PR #3069, opened 2025-04-16
NO-ISSUE: Upgrade go to 1.23.8

thiagoelg:
Some dependencies (with security updates) require a newer version of Go. 

ref https://github.com/apache/incubator-kie-tools/pull/2999


thiagoelg:
@ricardozanini  The sonataflow-operator and its dependencies were built successfully and passed all tests, so I think they're fine, but feel free to test them.

Out of topic: It would be nice if our pipelines produced the .tar.gz files for the built images and added them to the artifacts 🤔

examples/kie-sandbox-commit-message-validation-service/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/apache/incubator-kie-tools/examples/kie-sandbox-commit-message-validation-service
 
-go 1.21
+go 1.24.1
```
ricardozanini: Use 1.24, please, so we don't force a required patch version on users.

ricardozanini (changes requested):
Thanks, but I think it's unnecessary to jump to 1.24 from 1.22 directly since there's 1.23 in-between and the x/net dependency requires 1.23: https://github.com/golang/net/blob/master/go.mod#L3

Also, we need to upgrade the image too: https://github.com/apache/incubator-kie-tools/blob/main/packages/sonataflow-operator/images/manager.yaml#L20.

Maybe we can use this version from `env` instead of hardcoding it?

ricardozanini (approved):
Nice! Many thanks!

tiagobento (approved):

tiagobento:
Waiting on @jstastny-cz's eyes here too.

devbox.lock
```diff
@@ -21,51 +21,51 @@
         }
       }
     },
-    "go@1.22.12": {
-      "last_modified": "2025-02-05T05:41:33Z",
-      "resolved": "github:NixOS/nixpkgs/5b2753b0356d1c951d7a3ef1d086ba5a71fff43c#go_1_22",
+    "go@1.24.1": {
```
jstastny-cz: is this harmless?

jstastny-cz (approved):
LGTM.

devbox.lock
thiagoelg: Nice catch, I didn't push the updated lockfile.

thiagoelg:
As of now the repo.eclipse.org repository is down (https://www.eclipsestatus.io/), breaking our CI bootstrap.

tiagobento:
It happens periodically and it's usually back in a few hours @thiagoelg. Let's hope it is still the case. cc @yesamer

yesamer:
https://www.eclipsestatus.io

PR #3070, opened 2025-04-16
incubator-kie-tools-3068: Workflow metadata event must use the sonataflow.org/name annotation instead of SonataFlow.Name

### LINKED ISSUE #3068: Workflow metadata event must use the sonataflow.org/name annotation instead of SonataFlow.Name
Instead of using the SonataFlow.Name as value for the "name", event, we must use the SonataFlow annotation sonataflow.org/name if present.

wmedvede:
closes: #3068

### DIFF
--- a/packages/sonataflow-operator/api/v1alpha08/conversion.go
+++ b/packages/sonataflow-operator/api/v1alpha08/conversion.go
@@ -102,9 +102,18 @@ func ToCNCFWorkflow(workflowCR *SonataFlow, context context.Context) (*cncfmodel
 	}
 
 	cncfWorkflow.ID = workflowCR.ObjectMeta.Name
-	cncfWorkflow.Key = workflowCR.ObjectMeta.Annotations[metadata.Key]
-	cncfWorkflow.Description = workflowCR.ObjectMeta.Annotations[metadata.Description]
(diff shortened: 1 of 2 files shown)


packages/sonataflow-operator/api/v1alpha08/conversion.go
```diff
@@ -102,9 +102,18 @@ func ToCNCFWorkflow(workflowCR *SonataFlow, context context.Context) (*cncfmodel
 	}
 
 	cncfWorkflow.ID = workflowCR.ObjectMeta.Name
-	cncfWorkflow.Key = workflowCR.ObjectMeta.Annotations[metadata.Key]
-	cncfWorkflow.Description = workflowCR.ObjectMeta.Annotations[metadata.Description]
-	cncfWorkflow.Version = workflowCR.ObjectMeta.Annotations[metadata.Version]
+	if key, ok := workflowCR.ObjectMeta.Annotations[metadata.Key]; ok {
```
wmedvede, line 105: btw @ricardozanini if the annotation is not present, we shouldn't force anything.  If we do e.g. workflowCR.ObjectMeta.Annotations[metadata.Description] and the annotation is not present, we get the "".

ricardozanini (approved):

PR #3078, opened 2025-04-24
kie-issues#119: VS Code Extensions should keep their Editors in sync even when files change externally

### LINKED ISSUE #3734: VS Code Extensions should keep their Editors in sync even when files change externally

danielzhe:
Closes: https://github.com/apache/incubator-kie-tools/issues/3734

**This also works with .bpmn and .scesim files.**

Demo of the behavior, with .dmn files:
https://github.com/user-attachments/assets/78cbf013-38f5-44dc-bec9-420822231a4b


packages/vscode-extension/src/VsCodeKieEditorController.ts
```diff
@@ -257,6 +273,35 @@ export class VsCodeKieEditorController implements EditorApi {
...
+      this.yesKeepTheChanges,
+      this.noDiscardTheChanges
+    );
+    if (selection === this.noDiscardTheChanges) {
+      await (this.document.document as VsCodeKieEditorCustomDocument).revert(undefined!);
+    }
+    this.hasPendingExternalChanges = false;
+  }
+
+  public startListenToFileChanges() {
```
tiagobento:
```suggestion
  public startListeningToExternalFileChanges() {
```

packages/vscode-extension/src/VsCodeKieEditorController.ts
```diff
@@ -257,6 +273,35 @@ export class VsCodeKieEditorController implements EditorApi {
...
+    this.fileWatcher?.dispose();
+    this.fileWatcher = vscode.workspace.createFileSystemWatcher("**/*.{dmn,bpmn,scesim}");
```
tiagobento: I guess we need a parameter here, or infer those extensions from the other parameters we already have, since this package is generic and format-agnostic.

jomarko (commented):
Thank you for the PR @danielzhe . After I click on the "No, discard and reload from disk", the VS Code still mark the file as 'dirty'. Is that expected?

Screenshot 2025-04-25 074755

danielzhe:
@jomarko It was expected. We probably need to re-save the file inside VSCode to clear that status because it doesn't know the file is already saved. What do you think?

packages/vscode-extension/src/VsCodeKieEditorController.ts
danielzhe: Fixed. Ready for your re-review, @tiagobento

tiagobento:
@danielzhe Can we do it programmatically? I mean, when the user selects that it wants VS Code to reflect what's in the file system, there's nothing to be saved, right?

danielzhe:
@tiagobento Yes, I'm going to "trick" the VS Code by saving the file again programmatically to fix the issue reported by @jomarko .

danielzhe:
@jomarko @tiagobento All changes applied! Now we programmatically save the file if the user chooses to reload the changes from disk. It is ready for your review.

PR #3085, opened 2025-04-29
[kie-issues#1938] Remove userinfo call from Management Console to support Azure as IDP

### LINKED ISSUE #1938: KIE management console - userinfo call preventing compatibility with Azure as IDP
Using Azure as Identity Provider, there is currently a blocking issue with the way the authentication is implemented in the KIE management console webapp in the main branch.

Once the tokens are granted from the authorization code, the webapp performs a call to the oauth userinfo end point: https://github.com/apache/incubator-kie-tools/blob/main/packages/runtime-tools-management-console-webapp/src/authSessions/AuthSessionsService.ts#L304
With Azure, this end point requires a token issued specifically to access the Microsoft Graph API. If the authentication initially requested a specific scope for the runtime API, it will not have access to the MS Graph API without a token exchange & the userinfo will always fail with 401 - invalid audience (the audience of the JWT has to be the Graph API one - whereas requesting a resource service API scope will lead to the audience being the uuid of the requested resource API).

As of today, there is no way around this userinfo call, & as far as I've understood, its only purpose is to retrieve the username to display in the list of connected runtimes.

Removing the userinfo call or providing a way around it & computing the username from the claims extracted from the tokens would unblock this use case, ie:

`username: claims.preferred_username ?? claims.name ?? claims.email ?? claims.sub;`

jamedard:
- Remove userinfo end point call after authorization in the KIE management console to support Azure as Identity Provider as it cannot work with the same token as the runtime API if the latter requires a custom scope. The user info is only used for username computation which relies on claims extracted from the tokens with these code changes.
- Make the OIDC default scopes configurable with an environment variable, just like the client ID

Closes https://github.com/apache/incubator-kie-issues/issues/1938


thiagoelg:
Hi @jamedard, thanks for the PR, this is a great first contribution!

Could you please update the `packages/kogito-management-console/README.md` with the new env var? That's where we document the environment variables for the container image.

jamedard:
Hi @thiagoelg ,
Thank you.
I have updated the readme.md.

thiagoelg (approved):
LGTM

Thanks @jamedard!

tiagobento (approved):

PR #3086, opened 2025-04-29
[patternfly-v5-upgrade] online-editor UI updates 

klbowers:
Addresses various problems with spacing, icons and other UI changes that have occurred due to the Patternfly update.

There's still various spacing differences, so it depends on how close we want to get to the original. The DMN Editor screenshots with the file status icons stacked on top of each other is a result of the toolbar taking up slightly more space so it's getting compressed. When expanded slightly they will show side by side as expected


tiagobento:
This is green and I think the differences are negligible. Merging this hoping to expedite progress. cc @fantonangeli.

tiagobento (approved):

fantonangeli:
@tiagobento  if you say these differences are acceptable, I'm very happy to expedite progress.
Thanks a lot for checking this PR.
cc @kumaradityaraj

PR #3089, opened 2025-04-29
kie-tools#2641: Upgrade PatternFly to v5

fantonangeli:
Closes https://github.com/apache/incubator-kie-tools/issues/1877
Closes https://github.com/apache/incubator-kie-tools/issues/2641

**Special thanks:** @kumaradityaraj and @kbowers-ibm for the work in this PR.

**Notes:** I'm opening this PR on behalf of @kumaradityaraj which have no permissions for this.
This PR reopen https://github.com/apache/incubator-kie-tools/pull/2853 as we moved the branch under `apache/incubator-kie-tools` repository.

Upgrade Patternfly from v4 to v5 to the full repository following PatternFly's migration.

guide: https://www.patternfly.org/get-started/upgrade/#patternfly-5

**Updated packages**
@kie-tools-core/editor
@kie-tools-examples/base64png-editor
@kie-tools-examples/uniforms-patternfly
@kie-tools-examples/webapp
@kie-tools/boxed-expression-component
@kie-tools/dashbuilder-component-table
@kie-tools/dashbuilder-component-uniforms
@kie-tools/dashbuilder-component-victory-charts
@kie-tools/dashbuilder-editor
@kie-tools/dashbuilder-viewer
@kie-tools/dashbuilder-viewer-deployment-webapp
@kie-tools/dev-deployment-dmn-form-webapp
@kie-tools/dmn-editor
@kie-tools/dmn-editor-envelope
@kie-tools/dmn-editor-standalone
@kie-tools/dmn-runner
@kie-tools/form
@kie-tools/form-dmn
@kie-tools/import-java-classes-component
@kie-tools/kie-bc-editors
@kie-tools/online-editor
@kie-tools/pmml-editor
@kie-tools/runtime-tools-components
@kie-tools/runtime-tools-management-console-webapp
@kie-tools/runtime-tools-process-dev-ui-webapp
@kie-tools/runtime-tools-process-enveloped-components
@kie-tools/runtime-tools-shared-enveloped-components
@kie-tools/runtime-tools-shared-webapp-components
@kie-tools/runtime-tools-swf-enveloped-components
@kie-tools/runtime-tools-swf-webapp-components
@kie-tools/runtime-tools-task-console-webapp
@kie-tools/scesim-editor
@kie-tools/serverless-logic-web-tools
@kie-tools/serverless-workflow-combined-editor
@kie-tools/serverless-workflow-dev-ui-webapp
@kie-tools/serverless-workflow-text-editor
[...]

PR #3095, opened 2025-04-30
[NO-ISSUE] `onChange` events fixes after PatternFly v5 upgrade

fantonangeli:
This PR fixes some onChange events that are not passing the value to the callback.
I suggest merging on `main` after https://github.com/apache/incubator-kie-tools/pull/3089 not to block the main PR.

cc @kumaradityaraj


packages/jbpm-quarkus-devui/dev/src/main/resources/custom-forms-dev/hiring.tsx
```diff
@@ -69,7 +69,7 @@ const Form__hiring: React.FC<any> = (props: any) => {
           placeholder={""}
           type={"text"}
           value={candidate}
-          onChange={set__candidate}
```
thiagoelg, line 72: @ljmotta I think we need to review the form generation code to avoid cases like this.

thiagoelg (approved):

packages/jbpm-quarkus-devui/dev/src/main/resources/custom-forms-dev/hiring.tsx
fantonangeli: @ljmotta @thiagoelg  Maybe we can create an issue in this epic for this?
https://github.com/apache/incubator-kie-issues/issues/468

ljmotta: I don't think is a good idea to break the feature to fix it latter. I'll try to provide a quick fix for it, and open a PR to the patternfly branch. I'll just need to double check all components used in the `form-code-generator-patternfly-theme` package. Like Select, Checkbox, Radio, DatePicker, TimePicker, etc.

thiagoelg: The idea is not to fix it later but to fix it now in a different PR pointing to main to unblock the original pf5 upgrade PR, just like what this PR is doing.

fantonangeli: +1 @thiagoelg

tiagobento (approved):

PR #3107, opened 2025-05-05
[NO-ISSUE] onChange events fixes after PatternFly v5 upgrade

kumaradityaraj:
This PR fixes some onChange events that are not passing the value to the callback.
@fantonangeli


kumaradityaraj:
@fantonangeli seems like you already updated those events and iwas working on the outdated branch. Sorry for this PR.

PR #3132, opened 2025-05-13
kie-issues#11: GitLab support on KIE Sandbox

### LINKED ISSUE #11: GitLab support on KIE Sandbox
- Allow users to import individual files from GitLab like we do with GitHub
- Assuming that an authentication mechanism is already provided
- Allow users to create repositories on GitLab like we do with GitHub
- Allow users to import private repositories from GitLab, like we do with GitHub
- Allow KIE Sandbox to configure with private instances of GitLab, not only gilab.com
   - This required creating a new AuthProvider type "gitlab" and configuring its UI on the Accounts modal

### LINKED ISSUE #1980: KIE Sandbox: Enhance GitLab group and nested group display in the dropdown list
Related to PR [#3132](https://github.com/apache/incubator-kie-tools/pull/3132)
- Currently, all groups and nested groups will be displayed in a dropdown list, which will not provide a hierarchical view of the parent-child relationships
- As a suggestion, integrating a tree view structure into the dropdown list could enhance the UX.

Image

Adarshvk98:
Closes [issue #11](https://github.com/apache/incubator-kie-issues/issues/11)  ,
Closes [kie-issues#1980](https://github.com/apache/incubator-kie-issues/issues/1980)

## Overview  
This PR introduces support for GitLab's structure and features within our KIE Sandbox.
 - Aligning with GitLab’s concepts of Groups, Projects, and Snippets, this implementation maps similar functionality using Groups and Projects, although GitLab doesn’t use the term “organization” like GitHub or Bitbucket.
 - The REST API endpoints themselves are consistent across GitLab’s three offerings: **GitLab.com**, **GitLab Self-Managed**, and **GitLab Dedicated**. Therefore, there is an assumption that it will work across private instances.
- There is no officially maintained GitLab client, so a simple fetch client was implemented as part of this [Reference material for the REST API](https://docs.gitlab.com/api/rest/)

---

### 1. Authentication  
- Added support for authentication using **[Personal Access Tokens (PAT)](https://docs.gitlab.com/user/profile/personal_access_tokens/#create-a-personal-access-token)** to securely interact with GitLab APIs.
- While creating a Personal Access Token, need to add the following scopes: `api`, `read_user`, `read_repository`, and `write_repository`.
### 2. Clone & Import Repository from URL  
- **Clone Project Repository:**  
  Support cloning entire projects via URL.  
 eg: `https://<domain>/<group>/<project>.git`  

- **Clone Specific Branch/Ref:**  
  Ability to clone a specific branch or reference from a project.  
  eg: `https://<domain>/<group>/<project>/-/tree/<branch>`  

- **Import File from Branch:**  
  Import individual files directly from a specified branch in the repository.  
  eg: `https://<domain>/<group>/<project>/-/blob/<branch>/<file-path>`  

- **User-Specific Project Support:**  
  Cloning and importing also supported for user-specific projects.
  eg: `https://<domain>/<user-id>/<project>.git`

### 3. Clone & Import Snippet from URL 
- **Clone Standalone User-Specific Snippet:**   
  Clone snippets that belong to an individual user outside any project context.  
  eg: `https://<domain>/snippets/<snippet-id>.git`

- **Import Standalone User-Specific Snippet File:**   
  Import snippet files directly by URL.  
  eg: `https://<domain>/snippets/<snippet-id>/raw/main/<file-name>`

- **Clone Project Snippet:**   
  Clone snippets associated with a specific project group.  
  eg: `https://<domain>/<group>/<project>/-/snippets/<snippet-id>.git`

- **Import Project Snippet File:**   
   Import snippet files belonging to a project snippet by direct URL access.    
   eg: `https://<domain>/<group>/<project>/-/snippets/<snippet-id>/raw/main/<file-name>`

### 4. Create Repository & Snippets
- **Create Repository From Workspace:**   
   - Lists groups under the dropdown for selection.    
   - Supports creating repositories under user scope, group scope, or subgroup scope as selectable options.

- **Create Snippet From Workspace:**    
   - Lists projects under the dropdown for selection.    
   - Supports creating snippets either under user scope or project scope.

### 5. Update GitLab Snippets
- Enabled updating existing snippets only if the current authenticated user is the owner of the snippet (verified ownership before enabling update).

---

### Issue Observed
To connect to gitlab.com, it is observed that need to disable TLS certificate verification to get a successful response from Git operations like ref-check, clone, push, and pull. However, it’s unclear how to resolve this issue since disabling TLS certificate verification would introduce security concerns. One possible solution is to send the `accept-encoding: “identity”` header from KIE Sandbox itself for GitLab authentication sessions. However it's not clear whether this is the right way to approach it. 

### Solution

- Introduced an new flag similar to `insecurelyDisableTlsCertificateValidation` called `disableEncoding`. This option will act as a flag to specific for disabling gzip encoding (forcing `accept-encoding: identity`).
- This can be loaded through an optional environmental variable, `KIE_SANDBOX_AUTH_PROVIDERS`, for Git providers. If enabled, it forces the proxy or server to skip compression. 

```
{
      id: "gitlab_dot_com",
      domain: "gitlab.com",
      supportedGitRemoteDomains: ["gitlab.com"],
      type: "gitlab",
      name: "GitLab",
      enabled: true,
      group: "git",
      disableEncoding: true,
}
```


jomarko (commented):
Thank you for the PR.

### 01
Maybe we should update this text to mention gitlab?

Screenshot 2025-05-19 133223

### 02
This red text seems as unexpected. maybe related to patternfly 5 update
Screenshot 2025-05-19 133135


### 03
Once I deleted repository on gitlab, that was previously used in sandbox, sandbox was still saying he is pushing successfully to a non existing location
Screenshot 2025-05-19 133804


### 04
Flash message with information about snippet update print url that can not be opened

Screenshot 2025-05-19 134155
Screenshot 2025-05-19 134142

Adarshvk98:
Thank you, @jomarko, for reviewing. I’ve outlined my points below. Please take a look.

>01 

II’ve included “GitLab Snippet” in the above sentences.

> 02

Yes, those changes are related to the Patternfly 5 update. Perhaps I can take a look and try to fix it.

> 03

I have observed that once a repository is deleted, it will be in a pending deletion state for the next 7 days. After that date, it will be permanently deleted from GitLab.com. During this time, we can commit or make any changes to the deleted repository until it is permanently deleted

> 04

I tried to reproduce this issue, but I couldn't. Perhaps you could provide more information, such as a reproducible set of steps. When I tried, I was able to open the snippet URL displayed in the flash message. I noticed that your snippet URL ends with the .git extension, which might be the reason. If the snippet URL ends with the .git extension, it will display a 404 page. Could you please double-check the snippet URLs that end with the .git extension?

jomarko:
### 04 Steps to reproduce
- Open Sandbox
- Try Sample DMN
- Share -> Create Gitlab SNippet, chose private visibility
- Shown flash message is OK (no '.git' suffix) ✅ 
- Move some decision node
- Sync Snippet
- The flash message contains '.git' suffix 🔴

Adarshvk98:
Thanks for sharing those reproducible steps, it helps.

I have fixed this [issue](https://github.com/apache/incubator-kie-tools/pull/3132/commits/49edf197ee1e8bd13859cee9cd901501ff52f39d)

@jomarko Please take a look now ?

jomarko:
@Adarshvk98 thank you, could you please attach screenshot or video, that `gist` works after your last commit?

jomarko (commented):
### 01 Pull Scenario
There is this pull scenario, that is quite confusing, not surte if it is just me
- In Sanbox, I have cloned a gitlab repository, this repository has a DMN model, that has three nodes ✅ 
- Clone was successful ✅ 
- I have opened the DMN model ✅ 
- I deleted one node, so the DMN model has decreased nodes cont from 3 to 2 ✅ 
- Teen I decided it was step I didn't want, so I clicked Sync->Pull and here is the situation strange. Please notice in the attached screenshot I am informed pull was successfull, but actually it was not, the DMN was not restored to have 3 nodes again 🔴 

image

jomarko:
### 02 Snippet from two places
I can use this URL `https://gitlab.com/jomarko1-group/jomarko1-project/snippets/4844651.git` from home screen
Screenshot 2025-05-22 083932


But I can not use the same URL from asset opened
image

jomarko:
### 03 Subgroups
It is expected Sandbox list subgroup sibling of group?
Screenshot 2025-05-22 091351


while gitlab has subgroup like nested in group?
image

Adarshvk98:
Thank you, @jomarko , for taking the time to test this. I’ve some observations that I’d like to share below.
>01 Pull Scenario

I’ve tested the same scenario with other Git providers (Github) and noticed the issue you’ve observed here as well. Which means that it’s a common issue with pull ?. All the Git operations were performed using the npm package called [isomorphic-git](https://isomorphic-git.org/).

>02 Snippet from two places

From the home screen, you can clone the snippet because it supports both cloning the repository and importing a single file. However, from within the workspace, under the “New File” dropdown, you can only import a single file using the URL. I noticed that the URL you provided here is a snippet clone URL with a `.git` extension, which can contain multiple files. Therefore, it won’t support. Try entering a single raw snippet file URL in the input field. Below is a screenshot to help you get the raw URL of the snippet.

image

image

> 03 Subgroups

Currently, all groups and nested groups will be displayed in a dropdown list. I would like to hear your feedback on how this should be displayed.
Note:
- Repositories can be created under user space, group space, or nested group space.
- Snippets can be created under user space or project space (where there’s no nested structure).

cc: @tiagobento

tiagobento (approved):

packages/online-editor/src/gitlab/GitlabClient.ts
```diff
@@ -0,0 +1,222 @@
...
+  createProjectPath = ({ group, project }: ProjectPath): string =>
+    encodeURIComponent([group, project].filter(Boolean).join("/"));
+
+  createProjectBasePathWithId = ({ group, project }: ProjectPath, basePath: GitlabBasePath = "/projects") => {
+    const id = this.createProjectPath({ group, project });
+    return id ? `${basePath}/${id}` : basePath;
+  };
+
+  checkSnippetLevel = ({ group, project }: ProjectPath): string => {
+    const isProjectSnippet = !lodash.isEmpty(group) || !lodash.isEmpty(project);
```
thiagoelg: Aren't `group` and `project` strings? This could be `!group || !project`, right?
If we really need lodash here, update the import to bring only the required function.

thiagoelg (approved):
I've noticed that the default scope it selects when creating a Snippet or Repository is a group instead of the user itself, and this differs from the GitHub behavior. It's not a blocker, but we should probably make it consistent.

Left another comment with a tiny nitpick, but everything else looks good. Thanks @Adarshvk98

packages/online-editor/src/gitlab/GitlabClient.ts
Adarshvk98: Updated the import of lodash

packages/online-editor/src/importFromUrl/ImportableUrlHooks.tsx
```diff
@@ -466,6 +549,12 @@ export function useImportableUrl(urlString?: string, allowedUrlTypes?: UrlType[]
       });
     }
 
+    // Gitlab
+    if (isGitRemoteDomainAllowed(url, "gitlab.com", AuthProviderType.gitlab)) {
```
porcelli, line 553: Shouldn't use the configurable domain here instead gitlab.com?

porcelli (commented):
overall looks good, I'll pull now and test local with some complex inner groups to confirm if it works.

just one question left regarding the hardcode of the gitlab.com domain that seemed overlooked

packages/online-editor/src/importFromUrl/ImportableUrlHooks.tsx
tiagobento: No. Just like we do for `github.com`.

packages/online-editor/src/editor/Toolbar/GitIntegration/LoadOrganizationsSelect.tsx
```diff
@@ -102,10 +121,30 @@ export const LoadOrganizationsSelect = (props: Props) => {
...
+        organizations: (projectsResponseJson ?? []).map(({ name, id }: { name: string; id: string }) => ({
+          name,
+          value: id,
+        })),
+      };
+    }
+    const groupsResponse = await gitlabClient.listGroups();
+    const groupsResponseJson = await groupsResponse.json();
+    return {
+      organizations: (groupsResponseJson ?? []).map(({ name, id }: { name: string; id: string }) => ({
```
porcelli: maybe use `full_path` instead of `name` so you show the full name and avoid ambiguities with different sub groups with same name.

use full_path could also be applied to project list that I can list different projects (under different sub-groups) but with same name.

Screenshot 2025-05-28 at 6 39 08 PM

porcelli (changes requested):
not critical, but would be helpful to show full path info instead of names

packages/online-editor/src/editor/Toolbar/GitIntegration/LoadOrganizationsSelect.tsx
Adarshvk98: @porcelli 
**Creating a repository**
I would like to know which one would be more suitable here, is either `full_path` or `full_name` ?
- `full_path` : Group paths (used in URLs). It’s all lowercase, uses dashes/underscores instead of spaces, and shows the full nesting path.
- `full_name` : Display-friendly name of the group, but in a human-readable format (with proper capitalization, spaces, slashes, etc.)

Attaching a screenshot in case for a reference
image

**Creating a snippets**

- The projects API response doesn’t have a `full_path`, but instead, it has `name` and `name_with_namespace`.

image

cc: @tiagobento

tiagobento: I'd say we go with `full_name`, as it's display-friendly.

tiagobento: For snippets I think `name_with_namespace` follows the same logic.

Adarshvk98: I would also suggest, as it displays a more readable format

- To display Groups --> use `full_name`
- To display Projects  --> use `name_with_namespace`

porcelli (approved):
thx for the updates @Adarshvk98

Adarshvk98:
Closes [issues - 1980](https://github.com/apache/incubator-kie-issues/issues/1980)

Now, groups and nested groups are displayed with a separator of forward slash, which can avoid having different subgroups with the same name and same applies for projects as well
image
image

jomarko:
@Adarshvk98 thank you for updates. 

### Ordering
I think we should apply some ordering to items. Please notice here, it may be difficult to find proper item without ordering.
Screenshot 2025-05-29 132800

### Pull
As my understanding is the pull scenario described in https://github.com/apache/incubator-kie-tools/pull/3132#pullrequestreview-2859954333 is a separate issue, could you please report ticket for it before we proceed with a merge?

### Snippet -> Repository
Question, is it expected, by design, not sure what word to use, that if Sandbox user created firstly Gitlab snippet, that Sandbox still allows to create Repository from the snippet? I mean this button is allowed even if I work with snippet already.
Screenshot 2025-05-29 133437

Adarshvk98:
Thanks, @jomarko , for suggesting this improvement. I’ve made those changes.
>Pull

For the pull scenario, I will create a Kie issue
>Snippet -> Repository

I think this is expected. We can convert a snippet into a repository, and this is implemented in the same way as GitHub today.

PR #3133, opened 2025-05-14
kie-issues#1963: Edges are not being auto-generated in special case of using namespace in href

### LINKED ISSUE #1963: DMN Editor: Edges are not being auto-generated in special case of using namespace in href
Image

Image

https://raw.githubusercontent.com/gitgabrio/drools/9b2dc8792328c9f45397beee7b08ae40ba4f5015/kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/LocalHrefs.dmn

The model above is special in the href format for the <*Requirement> elements. They are int the format "[NAMESPACE]#[ID]"

When such model is opened in tooling KIE Sandbox, it offers you to autogenerate diagram [1], what is expected, however, the autogeneration result [2] doesn't seem to be expected. There are missing edges between nodes.

aswathitv:
Closes [kie#1963](https://github.com/apache/incubator-kie-issues/issues/1963)

https://raw.githubusercontent.com/gitgabrio/drools/9b2dc8792328c9f45397beee7b08ae40ba4f5015/kie-dmn/kie-dmn-test-resources/src/test/resources/valid_models/DMNv1_5/LocalHrefs.dmn

When above model is opened in tooling KIE Sandbox, it offers you to autogenerate diagram however, the autogeneration results in missing edges between nodes.

After fix: edges are generated for autogeneration


jomarko (commented):
Hi @aswathitv . Thank you for the PR. I can confirm the fix works. However before merge, I believe we should explain somewhere in a code, as a comment probably, what this find/search mechanism was introduced. I mean, we should mention somewhere possible format `[NAMESPACE]#[ID]`. It is at least my opinion we should have a short comment about it in the source code. Thank you for the fix one more time.

jomarko (approved):

tiagobento (changes requested):
Ok, so although this works, it's not the correct approach to manually look for IDs like that, as the ID of a node is dependent on its namespace too, not only its ID part.

The consistent approach here is to make use of `externalModelTypesByNamespace.dmns`, making sure hrefs are compared as an atom, not as a "string containing values".

Iterating over a Map's values like that, to find out something that contains an ID is not a reliable approach, as you can have a node with ID "a", and another one with ID "aa", for example, making the `includes` check pass for both.

I understand the subject of IDs and HREFs is complicated, so let us know if you have additional questions @aswathitv.

Thanks!

packages/dmn-editor/src/normalization/autoGenerateDrd.ts
```diff
@@ -33,6 +33,8 @@ import { addShape } from "../mutations/addShape";
 import { addEdge } from "../mutations/addEdge";
 import { EdgeType, NodeType } from "../diagram/connections/graphStructure";
 import { PositionalNodeHandleId } from "../diagram/connections/PositionalNodeHandles";
+import { DmnDiagramNodeData } from "../diagram/nodes/Nodes";
+import * as RF from "reactflow";
```
tiagobento: Leftover?

packages/dmn-editor/src/normalization/autoGenerateDrd.ts
```diff
@@ -173,8 +175,16 @@ export async function autoGenerateDrd(args: {
...
+    const { namespace: sourceIdNamespace, id: source } = parseXmlHref(drgEdge.sourceId);
+    const { namespace: targetIdNamespace, id: target } = parseXmlHref(drgEdge.targetId);
+
+    const sourceNode = updatedNodesByIdWithExternalNodes.get(
+      sourceIdNamespace === args.model.definitions["@_namespace"] ? `#${source}` : drgEdge.sourceId
+    );
+
+    const targetNode = updatedNodesByIdWithExternalNodes.get(
+      targetIdNamespace === args.model.definitions["@_namespace"] ? `#${target}` : drgEdge.targetId
+    );
```
tiagobento: Please consider using `buildXmlHref` instead of manually prepending `#`. There are other parts of the codebase dealing with this kind of problem, you should be able to use a similar strategy to understand the "locality" relationship of a HREF depending on `thisDmn`.

tiagobento (changes requested):

packages/dmn-editor/src/normalization/autoGenerateDrd.ts
aswathitv: Added the required changes

tiagobento (approved):

PR #3134, opened 2025-05-14
kie-issues#1961: DMN Editor can not change node Data Type from Boxed Expression Editor

### LINKED ISSUE #1961: DMN Editor can not change node data type from boxed expression editor
The DMN Editor user has two options to change the node data type

### Diagram Editor
When user select a node, small UI element to change the node data type is shown over the node shape and user can change the node data type. This approach works pretty well ✅ 

### Expression Editor
The second option is to change the node data type once the boxed expression editor of given node is opened. Then Click on the root header cell shows a small popup where node name and node data type can be changed. Here unfortunately the data type change has no effect 🔴 

### Environment
The issue is reproducible as of 13th of May 2025 using https://sandbox.kie.org/dev/#/ chrome 136.0.7103.93 browser.

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/1961

## Description
With the PatternFly 5 upgrade, the onHide callback is no longer triggered when clicking outside the Popover component. Instead, onHidden is now the appropriate callback for this behavior.
This PR updates the component to use onHidden accordingly, ensuring it responds correctly when the popover is dismissed by an outside click.

Additionally, the tests have been reorganized into two main categories: "Pressing Enter" and "Clicking Outside". Within each category, the tests are further divided into the following scenarios:
- Change Name
- Change Data Type
- Change Name and Data Type

This restructuring improves clarity and makes it easier to identify new possible bugs.


thiagoelg (approved):

packages/boxed-expression-component/src/expressionVariable/ExpressionVariableMenu.tsx
```diff
@@ -146,7 +146,7 @@ export function ExpressionVariableMenu({
   // With this ref we ensure the "saveExpression" inside onHide will not be called
   const cancelEdit = useRef<boolean>(false);
 
-  const onHide = useCallback(() => {
```
jomarko, line 149: I think we should update also comment:
```
  // onCancel doesn't prevent the onHide call
  // With this ref we ensure the "saveExpression" inside onHide will not be called
  const cancelEdit = useRef<boolean>(false);
  ```
  to not mention onHide anymore. WDYT?

jomarko (commented):
Thank you @ljmotta for the prompt fix and tests. I can confirm the reported issue is fixed now.

One question to tests is: as we did `close` optional, do we have a test where we call `setName` without the `close` argument?

And one more observation is trying to use properties panel, there change of data type is not reflected into boxed expression editor cell. However it seems as working on https://sandbox.kie.org/dev . Could we somehow check if that is related to this PR? I am fine also to solve in separate PR, just let me know if you are able to reproduce it.

Screenshot 2025-05-14 165312

packages/boxed-expression-component/src/expressionVariable/ExpressionVariableMenu.tsx
ljmotta: Good catch, I'll update accordingly

ljmotta:
@jomarko I made it optional to make it compatible with the old API, as we have the option to not close the Popover. This is used in the "Change Name and Data Type" tests, as we need to change the Name without closing it. 

I'll check the Properties Panel behavior.

ljmotta:
@jomarko I could confirm that this is in sandbox/dev. You need to create a BKM node, and after adding the Boxed Context expression is not possible to change the Data Type using the Properties Panel.

ljmotta:
Based on the CI, it looks like I broke a few tests. I'll investige.

jomarko:
Thank you @ljmotta , please let me know once I can help with a re-review.

ljmotta:
@jomarko The PR is ready for a re-review

jomarko:
@ljmotta I am sorry, will not get to this sooner than Monday  26th of May 2025

klbowers (approved):

PR #3143, opened 2025-05-16
NO-ISSUE: Automatically refresh custom forms list on every fetch in the jbpm-quarkus-devui

thiagoelg:
Fetching the list of custom forms (`/forms`) now will initialize the Storage again, listing all files in the `custom-dev-forms` directory.

This solves the issue when new custom forms are generated using the extension while the jBPM Quarkus DevUI is running, but the forms list is not updated with the new forms.

This solution is not pretty. We should be able to hook into the "hot-reload" API from Quarkus, but I don't know how to do that and couldn't find any documentation online.
This fix should be good enough for development use cases, so I'm not too worried about it.


thiagoelg:
@pefernan, your review is appreciated here, as I think you were the last person to touch this functionality. :smile:

pefernan (approved):

tiagobento (approved):

Issue #3112, opened 2025-05-06
[kn-workflow] Provide an option to generate build resources for gitops profile


masayag:
The `kn-workflow` already supports generating k8s manifests for deployment on cluster with `gitops` profile.
With `gitops` profile it is expected for the user to specify the image to use and for the user to be responsible for building the image.
However, there is no direct reference to how to build that image, e.g. which builder image to use for the specific version compatible with kn-workflow and also which Dockerfile to use.

The kn-workflow can provide that information by offering the same Dockerfile and steps on how to use it, e.g.:
- By adding a flag to enable persistence, the user will get the required QUARKUS_EXTENSION to use which contains the persistence support compatible with that version.

masayag:
cc @ricardozanini

ricardozanini:
@treblereel can you evaluate this one and add to our Q2 epic?

treblereel:
After discussing with masayag, we’ve decided to add to `quarkus create` the ability to generate additional GitOps functionality via an extra argument.

The idea is to introduce a profile flag:

```
--profile=gitops  or  --gitops-profile
```

I prefer `--gitops-profile` because it won’t give end users the misleading impression that other profiles include special functionality.

And additionally:

```
--with-persistence
```

@masayag @ricardozanini

ricardozanini:
> I prefer --gitops-profile because it won’t give end users the misleading impression that other profiles include special functionality.


But in the future we might have :)

treblereel:
> > I prefer --gitops-profile because it won’t give end users the misleading impression that other profiles include special functionality.
> 
> But in the future we might have :)

or might not :)
I’m not one to optimize before it’s necessary :)

treblereel:
@masayag When we execute kn workflow quarkus create, the Dockerfile is generated automatically. So we only need to enable persistence for the `gitops` profile. Am I understanding correctly? Or maybe we want to add something else?

masayag:
> [@masayag](https://github.com/masayag) When we execute kn workflow quarkus create, the Dockerfile is generated automatically. So we only need to enable persistence for the `gitops` profile. Am I understanding correctly? Or maybe we want to add something else?

In our case, we need to provide the following quarkus extensions and maven args:
https://github.com/rhdhorchestrator/orchestrator-demo/blob/main/scripts/build.sh#L186-L197
```
    local base_quarkus_extensions="\
[...]

PR #3144, opened 2025-05-19
kie-issues#3112 [kn-workflow] Provide an option to generate build resources for gitop…

treblereel:
…s profile

issue: https://github.com/apache/incubator-kie-tools/issues/3112


treblereel:
for test run `kn workflow quarkus create --profile=gitops --with-persistence`

packages/kn-plugin-workflow/pkg/command/quarkus/create.go
```diff
@@ -65,13 +78,36 @@ func NewCreateCommand() *cobra.Command {
...
-func runCreate() error {
-	cfg, err := runCreateCmdConfig()
+type MavenDependency struct {
+	GroupId    string
+	ArtifactId string
+}
+
+var ExtensionPerProfile = map[string][]MavenDependency{
+	apiMetadata.GitOpsProfile.String(): {
+		MavenDependency{"org.kie","kie-addons-quarkus-persistence-jdbc"},
```
ricardozanini: We should start thinking about adding a configuration locally. The groupId here might change downstream, for example.

packages/kn-plugin-workflow/pkg/common/checks.go
```diff
@@ -197,3 +198,23 @@ func CheckProjectName(name string) (err error) {
 	}
 	return
 }
+
+func IsValidProfile(profile string) error {
+	var allProfiles = []apiMetadata.ProfileType{
+		apiMetadata.DevProfile,
+		apiMetadata.ProdProfile,
```
ricardozanini: This one has been deprecated. Can you remove it?

packages/kn-plugin-workflow/pkg/command/quarkus/create.go
```diff
@@ -65,13 +78,36 @@ func NewCreateCommand() *cobra.Command {
...
+		MavenDependency{"io.quarkus","quarkus-agroal"},
+		MavenDependency{"io.quarkus","quarkus-jdbc-postgresql"},
```
masayag: Should the version be included as well?

ricardozanini: IIRC once we build the CLI, we have the versions set in the config.

treblereel: @masayag we use bom's so it's ok not have versions here

```
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>${quarkus.platform.group-id}</groupId>
                <artifactId>${quarkus.platform.artifact-id}</artifactId>
                <version>${quarkus.platform.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
            <dependency>
                <groupId>org.kie.kogito</groupId>
                <artifactId>kogito-bom</artifactId>
                <version>${kie.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>
```

packages/kn-plugin-workflow/pkg/command/quarkus/create.go
treblereel: in this case, it's better to create an issue

packages/kn-plugin-workflow/pkg/common/checks.go
treblereel: hmm, apiMetadata.ProdProfile comes from sonataflow-operator, I would create a separate issue to remove it form sonataflow-operator and kn-workflow-plugin as well

packages/kn-plugin-workflow/pkg/command/quarkus/create.go
ricardozanini: @treblereel I agree, can you create it?

packages/kn-plugin-workflow/pkg/command/quarkus/create.go
```diff
@@ -50,13 +56,20 @@ func NewCreateCommand() *cobra.Command {
 	# Create a project with additional extensions
 	# You can add multiple extensions by separating them with a comma
 	{{.Name}} create --extensions kogito-addons-quarkus-persistence-postgresql,quarkus-core
+
+	# Specify a profile to use for the new Quarkus project (default: dev)
```
masayag: Do you think the valid values for the profile should be listed here? (e.g. dev/preview/gitops) otherwise, the profiles will be listed only in case of a wrong value.

masayag (commented):
Thank you Dmitrii!
Looks good to me, I'll give it a try next week. For now, all I have is a small comment about the usage.

ricardozanini (approved):

packages/kn-plugin-workflow/pkg/command/quarkus/create.go
```diff
@@ -50,13 +56,20 @@ func NewCreateCommand() *cobra.Command {
 	# Create a project with additional extensions
 	# You can add multiple extensions by separating them with a comma
 	{{.Name}} create --extensions kogito-addons-quarkus-persistence-postgresql,quarkus-core
+
+	# Specify a profile to use for the new Quarkus project (default: dev). Available options: dev, preview, gitops.
+	{{.Name}} create --profile=<profile_name>
+
+	# Add persistence support to the project (default: false)
+	{{.Name}} create --with-persistence
```
masayag: shouldn't setting the flag `--with-persistence=true` enable also the flag
 `--extensions kogito-addons-quarkus-persistence-postgresql`?

or is it needed to specify each individually? or is the addition of the maven dependency makes the quarkus persistence extension redundant?

treblereel: @masayag 

`--with-persistence=true` adds the following extensions:

- kie-addons-quarkus-persistence-jdbc
- quarkus-agroal
- quarkus-jdbc-postgresql

Do you mean that I also have to include `kogito-addons-quarkus-persistence-postgresql` ?

masayag: No, my question was about the generated Dockerfile:
In our case, we specify the extensions in the Dockerfile, e.g.
https://github.com/rhdhorchestrator/serverless-workflows/blob/main/pipeline/workflow-builder.Dockerfile#L24

```
ARG QUARKUS_EXTENSIONS=io.quarkiverse.openapi.generator:quarkus-openapi-generator:2.9.1-lts,org.kie:kie-addons-quarkus-monitoring-sonataflow,org.kie:kogito-addons-quarkus-jobs-knative-eventing,org.kie:kie-addons-quarkus-persistence-jdbc,io.quarkus:quarkus-jdbc-postgresql:3.15.4.redhat-00001,io.quarkus:quarkus-agroal:3.15.4.redhat-00001
```

However, in the generated Dockerfile files there is no mention for that QUARKUS_EXTENSIONS argument.
This makes me wonder how the builder image is being used to build the images.

I thought that for the gitops profile we'll provide a Dockerfile that relies on a builder image.
According to the `gitops` profile documentation this Dockerfile should be used:
https://sonataflow.org/serverlessworkflow/main/cloud/operator/gitops-profile.html#_building_workflow_container_images

treblereel: @masayag I add these three deps directly into the pom.xml

Ok, I got your point, let me double check

masayag: Just to pinpoint on the differences:
Having the dependencies in the pom.xml are good for local development.
But when building with the builder image using the Dockerfile, the QUARKUS_EXTENSION argument is used to specify the dependencies and the pom.xml is the one internal to the builder image.

treblereel:
After discussing with @masayag, I removed
```
quarkus create --profile=gitops --with-persistance
```

To generate a Dockerfile with the current builder, it’s enough to run
```
create  --with-persistence

```

masayag (approved):

treblereel:
@ricardozanini ^^

ricardozanini (approved):

PR #3155, opened 2025-05-26
kie-issues#1977: Import types from Java class - replace external data type should not be possible

### LINKED ISSUE #1977: Import types from Java class - replace external data type should not be possible
A.dmn
Lets say this file has data type named Person

B.dmn
Lets say this file includes A.dmn using default namespace. So there is no prefix if B.dmn file want to access Person defined in A.dmn. Now imagine B.dmn want to import data type from a java class also named Person. The feature correctly detect conflict [1], however if there is conflict with and external data type, we should probably not allow the option Replace.

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-issues/issues/1977
Before fix:
User can able to see replace option for external data type conflicts also.
After fix:
Replace option is disabled in case of external data type conflicts.

Screenshot 2025-05-26 at 10 11 54 PM

### DIFF
--- a/packages/dmn-editor/src/dataTypes/ImportJavaClasses.tsx
+++ b/packages/dmn-editor/src/dataTypes/ImportJavaClasses.tsx
@@ -34,7 +34,7 @@ import {
   useLanguageServerAvailable,
 } from "@kie-tools/import-java-classes-component";
 import { addTopLevelItemDefinition as _addTopLevelItemDefinition } from "../mutations/addTopLevelItemDefinition";
-import { JavaClassConflictOptions, useImportJavaClasses } from "./useImportJavaClasses";
+import { JavaClassConflictOptions, JavaClassWithConflictInfo, useImportJavaClasses } from "./useImportJavaClasses";
 import { Radio } from "@patternfly/react-core/dist/esm/components/Radio";
 
 const ImportJavaClassesWrapper = ({
@@ -54,7 +54,7 @@ const ImportJavaClassesWrapper = ({
         <ImportJavaClassNameConflictsModal
           isOpen={isConflictsOccured}
           handleConfirm={handleConflictAction}
-          confliectedJavaClasses={conflictsClasses}
+          conflictedJavaClasses={conflictsClasses}
         />
       )}
     </>
@@ -111,21 +111,26 @@ const ImportJavaClassesDropdownItem = ({
 const ImportJavaClassNameConflictsModal = ({
   isOpen,
   handleConfirm,
-  confliectedJavaClasses,
+  conflictedJavaClasses,
 }: {
   isOpen: boolean;
   handleConfirm: (options: JavaClassConflictOptions) => void;
-  confliectedJavaClasses: JavaClass[];
+  conflictedJavaClasses: JavaClassWithConflictInfo[];
 }) => {
-  const [action, setAction] = useState<JavaClassConflictOptions>(JavaClassConflictOptions.REPLACE);
-  const handleActionButtonClick = useCallback(() => handleConfirm?.(action), [handleConfirm, action]);
   const handleRadioBtnClick = useCallback((event: React.FormEvent<HTMLInputElement>) => {
     const selectedValue = event.currentTarget.name;
     setAction(selectedValue as JavaClassConflictOptions);
   }, []);
-  const internalConflicts = confliectedJavaClasses.filter((c) => !c.isExternalConflict);
-  const externalConflicts = confliectedJavaClasses.filter((c) => c.isExternalConflict);
-  const classNames = confliectedJavaClasses?.map((javaClass) => javaClass?.name);
+  const internalConflicts = conflictedJavaClasses.filter((c) => !c.isExternalConflict);
+  const externalConflicts = conflictedJavaClasses.filter((c) => c.isExternalConflict);
+  const hasInternalConflicts = internalConflicts.length > 0;
+  const hasExternalConflicts = externalConflicts.length > 0;
+  const [action, setAction] = useState<JavaClassConflictOptions>(
+    hasExternalConflicts ? JavaClassConflictOptions.KEEP_BOTH : JavaClassConflictOptions.REPLACE
+  );
+  const handleActionButtonClick = useCallback(() => handleConfirm?.(action), [handleConfirm, action]);
+
+  const classNames = conflictedJavaClasses?.map((javaClass) => javaClass?.name);
   return (
     <Modal
       title="Duplicate DMN Data Type Detected"
@@ -138,7 +143,7 @@ const ImportJavaClassNameConflictsModal = ({
       actions={[
         <Button
           key="import-java-classes-conflict-btn"
-          isDisabled={externalConflicts.length > 0 && action === JavaClassConflictOptions.REPLACE}
+          isDisabled={hasExternalConflicts && action === JavaClassConflictOptions.REPLACE}
           variant="primary"
           onClick={handleActionButtonClick}
         >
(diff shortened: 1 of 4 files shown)


Adarshvk98:
Thanks for the PR, @Kusuma04-dev. I have some reviews on it.

1. Right now, For me personally it’s a bit confusing to understand which class names here have conflicts with internal and which ones have conflicts with external for the first time. Perhaps we could add a label as a prefix to the external conflicts class, similar to how we display the list of DMN datatypes in editors. Please refer to the screenshot below for your reference. Let me know your thoughts or if you have better representations.

image

2. Once we have conflicts with external data types, we disable the replace radio button. In such cases, it might be a good idea to default to selecting the “Keep Both” radio button. This way, users don’t have to manually do additional sections in the UI. They can simply click on the import button. From the screenshot, I can see that the replace option is currently selected by default. WDYT?

packages/dmn-editor/src/dataTypes/ImportJavaClasses.tsx
```diff
@@ -131,21 +136,48 @@ const ImportJavaClassNameConflictsModal = ({
...
         </Button>,
       ]}
     >
       <TextContent>
         {classNames?.length === 1 ? (
           <Text component={TextVariants.p}>
-            An existing DMN type named <b>{classNames?.join()}</b> has been detected. This type is currently in use
-            within the system. How would you like to proceed?
+            An existing DMN type named{" "}
+            {internalConflicts.length > 0 ? <b>{internalConflicts[0]?.name}</b> : externalConflicts[0]?.name} has been
```
Adarshvk98: I could see that the condition `internalConflicts.length > 0` and `externalConflicts.length > 0` has been used multiple times. Perhaps for better understanding, you could store this value in a variable and reuse it wherever needed?

packages/dmn-editor/src/dataTypes/useImportJavaClasses.ts
```diff
@@ -188,21 +188,35 @@ const useImportJavaClasses = () => {
...
 
       // Use a traditional for loop for better performance
       for (let i = 0, len = updatedJavaClasses?.length; i < len; i++) {
         const javaClass = updatedJavaClasses?.[i];
         const fullClassName = javaClass.name;
-        if (dataTypeNames.has(fullClassName)) conflicts.push(javaClass);
-        else nonConflicts.push(javaClass);
+        const isExternalConflict = dataTypesTree.some((dataType) => {
+          return (
+            dataType.namespace && externalModelsByNamespace[dataType.namespace] && dataType.feelName === fullClassName
```
Adarshvk98: this check is being executed in every iteration. We could potentially avoid it by moving this inside the if condition `if (dataTypeNames.has(fullClassName))`. This way, once the class name matches, then only need to check for conflicts with external types ? WDYT ?

packages/import-java-classes-component/src/components/ImportJavaClasses/model/JavaClass.ts
```diff
@@ -26,11 +26,14 @@ export class JavaClass {
   public fields: JavaField[];
   /** It indicates if the fields has been loaded, in order to support empty fields Java Classes */
   public fieldsLoaded: boolean;
+  /** It indicates if there is an external conflict */
+  public isExternalConflict: boolean;
```
Adarshvk98: I think we shouldn't add  `isExternalConflict` here. May be, we could create a new type that extends `JavaClass` ? This is because this variable is only used to check for conflicts. So, instead of modifying the existing type, let’s create a new type for conflicts and non-conflicts, which should extend from `JavaClass`. WDYT?

Kusuma04-dev: Thanks for the review @Adarshvk98 , Yeah I agree and we can do the same.

Kusuma04-dev:
> Thanks for the PR, @Kusuma04-dev. I have some reviews on it.
> 
> 1. Right now, For me personally it’s a bit confusing to understand which class names here have conflicts with internal and which ones have conflicts with external for the first time. Perhaps we could add a label as a prefix to the external conflicts class, similar to how we display the list of DMN datatypes in editors. Please refer to the screenshot below for your reference. Let me know your thoughts or if you have better representations.
> 
> image
> 
> 2. Once we have conflicts with external data types, we disable the replace radio button. In such cases, it might be a good idea to default to selecting the “Keep Both” radio button. This way, users don’t have to manually do additional sections in the UI. They can simply click on the import button. From the screenshot, I can see that the replace option is currently selected by default. WDYT?

@jomarko WDYT on showing label while searching? I feel as we are already mentioning external types in popup. So do we still need this label?

Kusuma04-dev:
> > Thanks for the PR, @Kusuma04-dev. I have some reviews on it.
> > 
> > 1. Right now, For me personally it’s a bit confusing to understand which class names here have conflicts with internal and which ones have conflicts with external for the first time. Perhaps we could add a label as a prefix to the external conflicts class, similar to how we display the list of DMN datatypes in editors. Please refer to the screenshot below for your reference. Let me know your thoughts or if you have better representations.
> > 
> > image
> > 
> > 2. Once we have conflicts with external data types, we disable the replace radio button. In such cases, it might be a good idea to default to selecting the “Keep Both” radio button. This way, users don’t have to manually do additional sections in the UI. They can simply click on the import button. From the screenshot, I can see that the replace option is currently selected by default. WDYT?
> 
> @jomarko. @tiagobento  WDYT on showing label while searching? I feel as we are already mentioning external types in popup label is not needed. So Can you please share your thoughts ?

jomarko:
@Kusuma04-dev I am not sure I understand what exactly do we discuss.

Do we discuss the screenshot from https://github.com/apache/incubator-kie-tools/pull/3155#issue-3089733428 and adding some labels there?

Personally I would inform user as it is in screenshot, using text not label, just personal preference. I am not saying I will block other solution.

however the screenshot https://github.com/apache/incubator-kie-tools/pull/3155#issue-3089733428 is problematic in other aspect I believe.

From the screenshot, I do not see, how user can replace `Test` data type, what should be valid action for the user.

Kusuma04-dev:
> @Kusuma04-dev I am not sure I understand what exactly do we discuss.
> 
> Do we discuss the screenshot from [#3155 (comment)](https://github.com/apache/incubator-kie-tools/pull/3155#issue-3089733428) and adding some labels there?
> 
> Personally I would inform user as it is in screenshot, using text not label, just personal preference. I am not saying I will block other solution.
> 
> however the screenshot [#3155 (comment)](https://github.com/apache/incubator-kie-tools/pull/3155#issue-3089733428) is problematic in other aspect I believe.
> 
> From the screenshot, I do not see, how user can replace `Test` data type, what should be valid action for the user.

Thanks @jomarko . My suggestion was along the same lines — we can show this in the popup text, but perhaps we could rephrase it in a way that's easier to understand.

As the User selected external type which is in conflict , he can't replace anything including internal(Test) in the above case. If you have any other suggestion for the popup text ,Please suggest.

Kusuma04-dev:
Popup text changed as per the discussion and PR is ready for review @jomarko.
Screenshot 2025-05-27 at 3 38 10 PM

jomarko:
### 01 Overlap
Once I click I want to import types from java classes, the UI element still remains visible and overlap the other popup

Screenshot 2025-05-29 085125

jomarko:
### 02 Cancel
There is missing option to cancel the import operation. I think there should be cancel button or cross icon in the popup.
Screenshot 2025-05-29 085247

jomarko:
### 03 Replace
On my machine, the 'Replace' operation has no effect. Once I selected I want to 'Replace' the confilct, the existing DMN data type stayed unchanged.

jomarko:
### 04 Internal vs External
If there are both Internal and External conflicts. User is unable to replace internal conflicts, what is form my point of view valid operation. Please notice the 'Replace' option is disabled. I still think this does not follow my suggestions from private slack conversation. I think we need separate group of options for 'Internal' and separate groups of option for 'External' conflicts.
Screenshot 2025-05-29 085836

jomarko (commented):
My review is in.

Kusuma04-dev:
> ### 04 Internal vs External
> If there are both Internal and External conflicts. User is unable to replace internal conflicts, what is form my point of view valid operation. Please notice the 'Replace' option is disabled. I still think this does not follow my suggestions from private slack conversation. I think we need separate group of options for 'Internal' and separate groups of option for 'External' conflicts. Screenshot 2025-05-29 085836

Hi @jomarko ,Thanks for the review.I have updated the code to seperate options for both internal and external conflicts.
Screenshot 2025-05-30 at 1 24 13 AM

Kusuma04-dev:
> ### 03 Replace
> On my machine, the 'Replace' operation has no effect. Once I selected I want to 'Replace' the confilct, the existing DMN data type stayed unchanged.

Can you please validate now?

Kusuma04-dev:
All review comments are incorporated and PR is ready for re-review

danielzhe (approved):
Good job!

tiagobento (approved):

PR #3163, opened 2025-05-30
kie-tools-issues #3171: Enabling Java security OIDC properties to all workflows

ishanjogi89:
In this MR, the changes have been made to controller and deployment creators to push/merge the workflowDefaults config map. By injecting the defaults there, they will propagate to all Deployments /Knative Services the operator ever creates.


pkliczewski:
cc @gabriel-farache @masayag

pkliczewski:
@ricardozanini @wmedvede please review

packages/sonataflow-operator/api/v1alpha08/sonataflowplatform_types.go
```diff
@@ -66,6 +67,9 @@ type SonataFlowPlatformSpec struct {
 	// Settings for Prometheus monitoring
 	// +optional
 	Monitoring *PlatformMonitoringOptionsSpec `json:"monitoring,omitempty"`
+
+	 // +optional
+	WorkflowDefaults *WorkflowDefaults `json:"workflowDefaults,omitempty"`
```
gabriel-farache, line 72: If I understand correctly: you add this new field to allow users to add anything in the podSpec template of all deployed workflow? It would act like a global configuration.

But the Java security OIDC properties have to be set in the properties file, which maps to the `-managed-props` configmap generated and managed by the operator (and which the user cannot override).
From my understanding we only need the following properties to be set
```
# Properties to check for identity, needed to use $WORKFLOW.identity within the workflow
quarkus.oidc.auth-server-url=${auth-server-url}
quarkus.oidc.client-id=${client-id}
quarkus.oidc.credentials.secret=${client-secret}
quarkus.oidc.token.header=X-Authorization-<provider>
quarkus.oidc.token.issuer=any # needed in case the auth server url is not the same as the one configured; e.g: localhost VS the k8S service
```
and they should not be placed in the podSpec template from the best of my knowledge.

I imagined more something similar to the `eventing` field which, apart from generating Knative resources, also configured the workflows properties to communicate using eventing with DataIndex and JobService;

ricardozanini:
@ijogi-redhat thanks for your PR.

However, I'm failing to understand the requirements and what they are for. We already have the `.spec.properties` in the `SonataFlowPlatform` that sets the default user properties to every workflow in the same context scope.

Additionally, I'm not inclined to add attributes to the public SonataFlow API without a compelling reason.

pkliczewski:
@ricardozanini is there anything we can do to make progress with this change?

ricardozanini:
@pkliczewski I'm failing to understand the purpose of this change.

pkliczewski:
> @pkliczewski I'm failing to understand the purpose of this change.

sometime ago we discussed how to filter workflow instances by user id (user who triggered it). You suggested the way to do it was enabling java security. Am I missing/misunderstood anything?

ricardozanini:
> > @pkliczewski I'm failing to understand the purpose of this change.
> 
> sometime ago we discussed how to filter workflow instances by user id (user who triggered it). You suggested the way to do it was enabling java security. Am I missing/misunderstood anything?

This PR doesn't do this, and the way they are trying to push is not the correct way. We could discuss this to iron out the details, but the PR as it stands today won't be merged.

ishanjogi89:
> > > @pkliczewski I'm failing to understand the purpose of this change.
> > 
> > 
> > sometime ago we discussed how to filter workflow instances by user id (user who triggered it). You suggested the way to do it was enabling java security. Am I missing/misunderstood anything?
> 
> This PR doesn't do this, and the way they are trying to push is not the correct way. We could discuss this to iron out the details, but the PR as it stands today won't be merged.

@ricardozanini Upon further investigation and as you explained in your earlier comments , SonataFlowPlatformSpec already exposes the properties field we need. In that case I don't think we need any code changes; we only need to add or patch the field in the running Platform CR. It will look something like this :

1. Create the Secret and ConfigMap
# client-secret value
oc create secret generic oidc-client-secret \
     -n sonataflow-infra \
     --from-literal=cred=swf-client-secret

# non-secret values
oc create configmap oidc-common-props \
     -n sonataflow-infra \
     --from-literal=authServer=https://keycloak/realms/master \
     --from-literal=clientId=swf-client \
     --from-literal=tokenHeader=X-Authorization-Github \
     --from-literal=issuer=any

2. Patch SonataFlowPlatform

apiVersion: sonataflow.org/v1alpha08
kind: SonataFlowPlatform
metadata:
  name: dev
  namespace: sonataflow-infra
spec:
  properties:
    # 1. literals (quick, but exposes secret – we keep them in a CM/secret instead)
    # values:
    #   quarkus.oidc.auth-server-url: https://keycloak/realms/master
    #   quarkus.oidc.client-id:       swf-client

    # 2. pull from ConfigMap & Secret
    valueFrom:
      - configMapKeyRef:
          name: oidc-common-props
          properties:
            quarkus.oidc.auth-server-url: authServer
            quarkus.oidc.client-id:       clientId
            quarkus.oidc.token.header:    tokenHeader
            quarkus.oidc.token.issuer:    issuer
      - secretKeyRef:
          name: oidc-client-secret
          properties:
            quarkus.oidc.credentials.secret: cred

After applying the changes of sonataflow-platform.yaml, we will roll existing workflow pods once.

Please let me know if this a correct understanding and right approach.

cc: @pkliczewski @gabriel-farache

masayag:
I agree with Zanini.

The functionality required by this PR is provided by [SonataflowPlatform.spec.properties](https://github.com/apache/incubator-kie-tools/blob/main/packages/sonataflow-operator/api/v1alpha08/sonataflowplatform_types.go#L65)
And its description is what is needed:

> 	// Properties defines the property set for a given actor in the current context.
> 	// For example, the workflow managed properties. One can define here a set of properties for SonataFlow deployments
> 	// that will be reused across every workflow deployment.
> 	//
> 	// These properties MAY NOT be propagated to a SonataFlowClusterPlatform since PropertyVarSource can only refer local context sources.
> 	// +optional
> 	Properties *PropertyPlatformSpec `json:"properties,omitempty"`

The only thing that should be taken into account is that they become managed properties, meaning they will be applied to all workflows and unmodifiable from the managed props, only through changes to the SonataflowPlatform CR.

There is always an option to override it per-workflow via its `-props` configmap.

ricardozanini:
I'm closing this PR since I understand the requirement is already supported.

gabriel-farache:
> I agree with Zanini.
> 
> The functionality required by this PR is provided by [SonataflowPlatform.spec.properties](https://github.com/apache/incubator-kie-tools/blob/main/packages/sonataflow-operator/api/v1alpha08/sonataflowplatform_types.go#L65) And its description is what is needed:
> 
> > // Properties defines the property set for a given actor in the current context.
> > // For example, the workflow managed properties. One can define here a set of properties for SonataFlow deployments
> > // that will be reused across every workflow deployment.
> > //
> > // These properties MAY NOT be propagated to a SonataFlowClusterPlatform since PropertyVarSource can only refer local context sources.
> > // +optional
> > Properties *PropertyPlatformSpec `json:"properties,omitempty"`
> 
> The only thing that should be taken into account is that they become managed properties, meaning they will be applied to all workflows and unmodifiable from the managed props, only through changes to the SonataflowPlatform CR.
> 
> There is always an option to override it per-workflow via its `-props` configmap.

oh, ok, we can already set any properties we want in the `SonataFlowClusterPlatform` and they will be applied in the `-managed-props` of all deployed workflow, cool!
@ishanjogi89 now we need to test that based on the per-workflow procedure we have here https://www.rhdhorchestrator.io/main/docs/serverless-workflows/configuration/token-propagation/

Issue #3138, opened 2025-05-15
[sonataflow-devmode-image] Add dashboard extension to connect to the embedded PostgreSQL server


masayag:
When running workflow using the devmode image, an embedded PostgreSQL server is started by `zonky`
However, there is no option to access that database in that mode.

For example, when starting the devmode with setting `quarkus.embedded.postgresql.port=5432` and exposing the port `-p 5432:5432` in the command below (to prevent a random port as explained [here](https://github.com/quarkiverse/quarkus-embedded-postgresql):
```
podman run --rm --name backstage-internal-sonataflow \
  -e QUARKUS_HTTP_PORT=8080 -p 8080:8080 \
  -p 5432:5432 \
  -e KOGITO_SERVICE_URL=http://localhost:8080/ \
  -e QUARKUS_EMBEDDED_POSTGRESQL_DATA_DIR=/home/kogito/persistence \
  -e MAVEN_ARGS_APPEND="-Dquarkus.embedded.postgresql.port=5432" \
 quay.io/kubesmarts/incubator-kie-sonataflow-devmode:9.103.x-prod
```
Even with that, an access to PostgreSQL will not be allowed since internally postgresql is configured by `zonky` with:
1. `#listen_addresses = 'localhost'		# what IP address(es) to listen on;` in postgresql.conf
2. In hba.conf with
    ```
    host    all             all             127.0.0.1/32            trust
    host    all             all             ::1/128                 trust
   ```
meaning, zonky does not support connection from outside.
In addition, the fact `pdql` cli isn't packaged in the image  makes it impossible to get to the postgresql DB.

By adding a dashboard extension to enable access to the DB, we can spare a more complex configuration in which we use an external PostgreSQL server for development.

Somehow related to https://github.com/apache/incubator-kie-tools/issues/3137

p.s. even without the extension, changes to the image will be welcomed to support an access to the embedded pgsql.

masayag:
cc @ricardozanini

ricardozanini:
We should create an issue here: https://github.com/quarkiverse/quarkus-embedded-postgresql

So that the extension itself exposes a dashboard for users to run SQL queries in the devmode application directly from the devui.

masayag:
> We should create an issue here: https://github.com/quarkiverse/quarkus-embedded-postgresql
> 
> So that the extension itself exposes a dashboard for users to run SQL queries in the devmode application directly from the devui.

Do you think this one is sufficient or should I report a new one:
[...]

PR #3176, opened 2025-06-11
[NO_ISSUE] Inherit embedded-postgresql version from incubator-kie-kogito-runtimes

treblereel:
as part of https://github.com/apache/incubator-kie-tools/issues/3138

### DIFF
--- a/packages/maven-base/pom.xml
+++ b/packages/maven-base/pom.xml
@@ -132,7 +132,6 @@
     <version.junit>4.13.2</version.junit>
     <version.org.apache.commons.commons-compress>1.26.1</version.org.apache.commons.commons-compress>
     <version.org.iq80.snappy>0.5</version.org.iq80.snappy>
-    <version.io.quarkiverse.embedded.postgresql>0.6.0</version.io.quarkiverse.embedded.postgresql>
     <version.org.apache.mime4j>0.8.11</version.org.apache.mime4j>
     <version.org.freemarker>2.3.32</version.org.freemarker>
     <version.org.assertj>3.25.3</version.org.assertj>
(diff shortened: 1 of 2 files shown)


packages/maven-base/pom.xml
```diff
@@ -132,7 +132,7 @@
     <version.junit>4.13.2</version.junit>
     <version.org.apache.commons.commons-compress>1.26.1</version.org.apache.commons.commons-compress>
     <version.org.iq80.snappy>0.5</version.org.iq80.snappy>
-    <version.io.quarkiverse.embedded.postgresql>0.3.0</version.io.quarkiverse.embedded.postgresql>
+    <version.io.quarkiverse.embedded.postgresql>0.6.0</version.io.quarkiverse.embedded.postgresql>
```
ricardozanini: I think we have to upgrade on kogito-runtimes too. :D

ricardozanini (approved):

packages/maven-base/pom.xml
tiagobento: Can we make it aligned with whatever version we have in the "version" of Kogito we're pointing to?

treblereel:
it looks like [CI :: Build / run (ubuntu-latest, 2) (pull_request)](https://github.com/apache/incubator-kie-tools/actions/runs/15590223756/job/43937738556?pr=3170) is broken and not only for me

packages/maven-base/pom.xml
ricardozanini: @tiagobento, what do you mean? Both Kogito and this repo use this extension, and they must be aligned. What's your suggestion then?

tiagobento: Right now we're using Kogito 999-20250511-SNAPSHOT. What is the version of this extension on this version? Just saying we should be aligned with it, instead of 'main'.

When we bump the Kogito version, if the version of this extension changed there, we update here too.

Hope it makes sense.

ricardozanini: Yeah, makes sense. But this is not a broken change, so won't be a problem. But I'm curious why this extension must be declared here too. @treblereel do you know?

ricardozanini (approved):

treblereel:
@tiagobento may I ask you to take one more look

PR #3181, opened 2025-06-12
[kogito-data-index-images] Fix kie-issues#1981 - Add Data Index UI to Images; Small install.js refactor

ricardozanini:
Part of https://github.com/apache/incubator-kie-issues/issues/1981

In this PR:

1. Remove `jobs-service-allinone-image` since it is deprecated and we are not using it anymore
2. Refactor the `install.js` script in every sonataflow and kogito image to a new package
3. Integrate the Data Index UI with the Data Index Images.

See the beautiful thing in action:

https://github.com/user-attachments/assets/b916429f-a588-48d4-a8e2-b6cdbd6b3d6c

Usually, I don't like doing refactors in feature PRs, but I was obligated since adding the UI to the image required fixing a Maven base problem in the install phase.


packages/kogito-images-install-helper/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+// Constants for dependent packages
+const PYTHON_VENV_PKG = "@kie-tools/python-venv";
+const IMAGE_COMMON_PKG = "@kie-tools/sonataflow-image-common";
+
+/**
+ * Shared install routine for bumping versions, configuring Maven and updating YAMLs/features.
+ * @param {{ imageEnv: { buildTag:string, registry:string, account:string, name:string }, resourceDir?: string, finalImageName: string, requiresMvn: boolean }} options
+ */
+function runSharedInstall({ imageEnv, resourceDir = "./resources", finalImageName, requiresMvn = true }) {
```
tiagobento:
```suggestion
function runKogitoImageInstall({ imageEnv, resourceDir = "./resources", finalImageName, requiresMvn = true }) {
```

packages/kogito-images-install-helper/index.js
```diff
@@ -0,0 +1,95 @@
...
+  if (!finalImageName) {
+    throw new Error("'finalImageName' parameter is required");
+  }
+  if (!imageEnv) {
+    throw new Error("'imageEnv' parameter is required");
+  }
```
tiagobento: I think it's not a good idea to pass the `env` object around like this. It hurts the grep test, creates indirection, and start a "meta framework" around envs. I guess I'd prefer if you'd transform the env in four separate arguments so that we don't create too many abstractions on top of envs.

packages/kogito-images-install-helper/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+  // 1) Maven revision and tail
+  if (requiresMvn) {
+    const { version } = require(path.resolve(__dirname, "package.json"));
+    setupMavenConfigFile(
+      `
+  -Drevision=${version}
+  -Dmaven.repo.local.tail=${buildTailFromPackageJsonDependencies()}
```
tiagobento: Gotta double-check if this will use the current package's `package.json` to build the tail, or the directory where it was ran.

packages/kogito-images-install-helper/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+const { execSync } = require("child_process");
+const fs = require("fs");
+const path = require("path");
+const replaceInFile = require("replace-in-file");
+const { setupMavenConfigFile, buildTailFromPackageJsonDependencies } = require("@kie-tools/maven-base");
+
+// Constants for dependent packages
+const PYTHON_VENV_PKG = "@kie-tools/python-venv";
+const IMAGE_COMMON_PKG = "@kie-tools/sonataflow-image-common";
```
tiagobento: Why extract to constants?

packages/kogito-images-install-helper/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Shared install routine for bumping versions, configuring Maven and updating YAMLs/features.
```
tiagobento:
```suggestion
 * Install routine for Kogito images. Helps with bumping versions, configuring Maven, and updating YAMLs/features.
```

packages/kogito-images-install-helper/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "private": true,
+  "name": "@kie/kogito-images-install-helper",
+  "version": "0.0.0",
+  "description": "Shared install helper for Kogito images",
```
tiagobento:
```suggestion
  "description": "Install helper for Kogito images",
```

packages/kogito-jit-runner-image/install.js
```diff
@@ -17,48 +17,15 @@
...
-  process.platform === "win32"
-    ? `${pythonVenvDir}\\venv\\Scripts\\Activate.bat`
-    : `. ${pythonVenvDir}/venv/bin/activate`;
-
-execSync(
-  `${activateCmd} && \
-  python3 ${sonataflowImageCommonDir}/resources/scripts/versions_manager.py --bump-to ${env.kogitoJitRunnerImage.buildTag} --source-folder ./resources`,
-  { stdio: "inherit" }
-);
+// import the function you exported
```
tiagobento: Not sure I understand this comment.

tiagobento (changes requested):
@ricardozanini Thanks for the PR! Great work with the new package for reducing duplication. I posted a few comments inline, please consider them! Thank you!

packages/kogito-images-install-helper/index.js
ricardozanini: It's just a data structure. I'll change the variable name to avoid confusion with the environment variables, but breaking it into another structure or pass single parameters to make the method signature bigger is even worse, IMO.

packages/kogito-images-install-helper/index.js
ricardozanini: It uses the current `package.json`, which is ok since all of them share the same version?

packages/kogito-jit-runner-image/install.js
ricardozanini: Neither do I, it's a left over.

packages/kogito-images-install-helper/index.js
tiagobento: Not really. It should use the `package.json` of the package it's being invoked by. `kogito-images-install-helper` won't have all Maven-based packages mapped for all packages that depend on it, as `kogito-images-install-helper` shouldn't have to know what Maven dependencies their consuming packages have.

packages/kogito-images-install-helper/index.js
tiagobento: It's a little more code, but the boundaries are clear. Passing values from `env` (build-env) to `kogito-image-install-helper` should be an explicit operation, instead of a "coincidence" that both share the same data structure.

ricardozanini: I'll extract the data from env, but keep the same method signature to avoid exposing other attributes from the environment.

ricardozanini:
@tiagobento mind taking a second look?

wmedvede (approved):
I have checked the images build by including these commits and works as expected:

image

packages/kogito-base-builder-image/install.js
```diff
@@ -17,48 +17,18 @@
...
-// Find and read the -image.yaml file
-const resourcesPath = path.resolve(__dirname, "./resources");
-const files = fs.readdirSync(resourcesPath);
-const imageYamlFiles = files.filter((fileName) => fileName.endsWith("-image.yaml"));
-if (imageYamlFiles.length !== 1) {
-  throw new Error("There should only be one -image.yaml file on ./resources!");
-}
-const originalYamlPath = path.join(resourcesPath, imageYamlFiles[0]);
-let imageYaml = fs.readFileSync(originalYamlPath, "utf8");
+// import the function you exported
```
tiagobento: Leftover comment?

packages/kogito-base-builder-image/install.js
```diff
@@ -17,48 +17,18 @@
...
-// Write file and then rename it to match the image name
-fs.writeFileSync(originalYamlPath, imageYaml);
-fs.renameSync(originalYamlPath, path.join(resourcesPath, `${env.kogitoBaseBuilderImage.name}-image.yaml`));
-
-// Replace image URL in .feature files
-replaceInFile.sync({
-  files: ["**/*.feature"],
-  from: /@docker.io\/apache\/.*/g,
-  to: `@${imageUrl}`,
+// grab your package's env object
```
tiagobento: Leftover comment? They are in other files too.

packages/kogito-base-builder-image/install.js
```diff
@@ -17,48 +17,18 @@
...
+const { env } = require("./env");
+const { buildTag, registry, account, name: imageName } = env.kogitoBaseBuilderImage;
```
tiagobento, line 23: 😍

tiagobento (commented):
Thanks! Looking good! Just a minor comments on some comments that seem to be leftover.

tiagobento (approved):

PR #3183, opened 2025-06-17
kie-tools-issues#3184- [sonataflow-management-console-webapp] Management console is unable to sort workflow instances by ID

### LINKED ISSUE #3184: [sonataflow-management-console-webapp] Management console is unable to sort workflow instances by ID
Management console is unable to sort workflow instance by ID on workflow instances page.
Only way to fix this is to refresh the page. Other option to sort become unavailable.
When using other options to sort ( status ) the user is able to select different option, while with ID the other options become disabled.

kumaradityaraj:
✨ **Closes [3184](https://github.com/apache/incubator-kie-tools/issues/3184) – Management console is unable to sort workflow instances by ID**

---
### Issue
The Management Console was unable to sort the workflow instances list by **ID** on the Workflow Instances page. When selecting the ID option for sorting:
- The sorting did not apply.
- The only way to restore functionality was to refresh the page.
- All other sorting options (e.g., by status) became disabled or unselectable after choosing ID.

This was impacting user experience by blocking expected UI behavior when managing workflow instances.

### What Was Fixed
- Corrected the sorting logic for the **ID** column.
- Ensured that selecting ID as a sort option does not disable or block other sort fields.
- Now, sorting by ID correctly updates the list as expected, without requiring a page refresh.

### ✅ Expected Behavior
- The list of workflow instances should not be empty.
- Sorting by **ID** correctly applies and shows results in the expected order.
- Other sort options remain available and functional after using the ID sort.

---

### 📹 Demo

A video showcasing the updated behavior is attached.

https://github.com/user-attachments/assets/007fccad-3a64-42f6-a02f-7e9699b33c7b

### DIFF
--- a/packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx
+++ b/packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx
@@ -90,6 +90,38 @@ const WorkflowListTable: React.FC<WorkflowListTableProps & OUIAProps> = ({
   const [titleType, setTitleType] = useState<string>("");
   const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
   const [selectedWorkflowInstance, setSelectedWorkflowInstance] = useState<WorkflowInstance | null>(null);
+  const [sortByState, setSortByState] = useState<{ index: number; direction: "asc" | "desc" }>({
+    index: 2,
+    direction: "asc",
+  });
+
+  const getComparableValue = (instance: WorkflowInstance, columnKey: string): any => {
+    switch (columnKey) {
+      case "Id":
+        return instance.id;
+      case "Status":
+        return instance.state;
+      case "Created":
+        return new Date(instance.start);
+      case "Last update":
+        return new Date(instance.lastUpdate);
+      default:
+        return "";
+    }
+  };
+
+  const onSortInternal = (_event: React.SyntheticEvent, index: number, direction: "asc" | "desc") => {
+    const columnKey = columns[index];
+    const sorted = [...workflowInstances].sort((a, b) => {
+      const aVal = getComparableValue(a, columnKey);
+      const bVal = getComparableValue(b, columnKey);
+      if (aVal < bVal) return direction === "asc" ? -1 : 1;
+      if (aVal > bVal) return direction === "asc" ? 1 : -1;
+      return 0;
+    });
+    setSortByState({ index, direction });
+    setWorkflowInstances(sorted);
+  };
 
   const handleModalToggle = (): void => {
     setIsModalOpen(!isModalOpen);
@@ -361,11 +393,14 @@ const WorkflowListTable: React.FC<WorkflowListTableProps & OUIAProps> = ({
           <Tr ouiaId="workflow-list-table-header">
             {columns.map((column, columnIndex) => {
               let sortParams = {};
-              if (!isLoading && rowPairs.length > 0) {
+              if (!isLoading && rowPairs.length > 0 && columnIndex > 1 && columnIndex !== columns.length - 1) {
                 sortParams = {
                   sort: {
-                    sortBy,
-                    onSort,
+                    sortBy: {
+                      index: sortByState.index,
+                      direction: sortByState.direction,
+                    },
+                    onSort: onSortInternal,
                     columnIndex,
                   },
                 };


kumaradityaraj:
@fantonangeli Please review this PR.

packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx
```diff
@@ -90,6 +90,38 @@ const WorkflowListTable: React.FC<WorkflowListTableProps & OUIAProps> = ({
   const [titleType, setTitleType] = useState<string>("");
   const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
   const [selectedWorkflowInstance, setSelectedWorkflowInstance] = useState<WorkflowInstance | null>(null);
+  const [sortByState, setSortByState] = useState<{ index: number; direction: "asc" | "desc" }>({
+    index: 2,
+    direction: "asc",
+  });
+
+  const getComparableValue = (instance: WorkflowInstance, columnKey: string): any => {
```
fantonangeli, line 98: it's always better to use `useCallback` to reduce re-rendering of components

packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx
```diff
@@ -90,6 +90,38 @@ const WorkflowListTable: React.FC<WorkflowListTableProps & OUIAProps> = ({
...
+      case "Created":
+        return new Date(instance.start);
+      case "Last update":
+        return new Date(instance.lastUpdate);
+      default:
+        return "";
+    }
+  };
+
+  const onSortInternal = (_event: React.SyntheticEvent, index: number, direction: "asc" | "desc") => {
```
fantonangeli, line 113: it's always better to use `useCallback` to reduce re-rendering of components

packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx
fantonangeli, line 1: IDKW but in this file at line 22 there is a change from `Table` to `TableComposable` which GH doesn't show:
upstream/main:
```
import { Table } from "@patternfly/react-table/deprecated";
```
kumaradityaraj/sortWorkflow:
```
import { TableComposable, Thead, Tbody, Tr, Th, Td } from "@patternfly/react-table/dist/js/components/TableComposable";
```
you can see it running:
```
git difftool -y upstream/main packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx
```

https://github.com/apache/incubator-kie-tools/blob/main/packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx#L23

https://github.com/kumaradityaraj/kie-tools/blob/bc8556ec487ffb47155a69dce34bbc43a8796a5c/packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx#L22

Was this intentional?

packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx
thiagoelg, line 113: This will sort the already-fetched items, but if the Data Index is returning a "paginated" list of items, this sort is not necessarily correct.

@fantonangeli , can the Data-Index sort by ID?

thiagoelg:
@kumaradityaraj, please update the title and description of this PR to reference Apache KIE community issues.

packages/runtime-tools-swf-enveloped-components/src/workflowList/envelope/components/WorkflowListTable/WorkflowListTable.tsx
fantonangeli: > This will sort the already-fetched items, but if the Data Index is returning a "paginated" list of items, this sort is not necessarily correct.
> 
> @fantonangeli , can the Data-Index sort by ID?

@wmedvede @ricardozanini @domhanak 
It seems that the Data Index does not support sorting by `id`, even though it's documented  [here](https://sonataflow.org/serverlessworkflow/latest/data-index/data-index-core-concepts.html).
I tried using the following query variables:
```
{
  "offset": 0,
  "limit": 10,
  "orderBy": {
    "id": "ASC"
  }
}
```
Response: 
```
The variables input contains a field name 'id' that is not defined for input object type 'ProcessInstanceOrderBy' 
```
IMO we have a couple of ways:
- Fix the Data Index to support sorting by `id`
- Sort by processName instead and update the column label in the UI from `id` to `Name`, since that column displays both the process name and ID.

Let me know what you think, or if I'm missing something in the query.

fantonangeli (changes requested):
@kumaradityaraj Thank you for your PR.
I left 2 comments about the use of `useCallback` which we don't need to fix ATM as we should check how to sort by `id` on server-side

wmedvede:
Hello @fantonangeli 

I don't think the sorting by **"id"**  (which corresponds to the process instanced id)  is implemented for the ProcessInstances query, but also I don't see that the documentation is saying that it is possible.

Looking at here: 
https://sonataflow.org/serverlessworkflow/main/data-index/data-index-core-concepts.html  I don't see any example.

Also, if we look in the DI GraphqlUI in the sort options for the ProcessInstances query, we are not showing the **"id"** field as available for sorting.

We can sort for example by **"processId"**, but not by **"id"**

```
{
  ProcessInstances (orderBy: {processId: DESC} ) {
  	processId,
    state,
    serviceUrl,
    state
  }
}
```


Screenshot from 2025-06-19 19-10-25


While it could be technically possible to implement the sorting by **"id"**, in practice I don't think it't be useful, since the **"id"** (the process instance id) has values like:  **"a0ac78ad-54e6-4168-882e-ac2849706e69"** .

I think that instead of sorting by **"id"** , this query can internally sort by **"processId"**, and thus, users can group all the instances that belongs to the same ProcessDefinition.

@fantonangeli so yes, internally, when the use clicks on the sort arrow for that column, we internally we can sort by **"processId"** or **"processName"**, and I'd say that we should ideally use the same field we are using to display in the UI.

fantonangeli:
@wmedvede thank you so much for your very useful information.
I agree to use the same field we are using to display in the UI, so I would "vote" for `processName` and rename the column to `Process Name`.
@domhanak  Because you were the first one to discover the bug, do you agree with this solution?

fantonangeli:
Hi @domhanak, would you agree with changing the sort by "Id" to "Process Name"?

kumaradityaraj:
Closing this PR since it has outdated patternfly due to outdated branch; It was my mistake since I started working on the outdated branch. New PR link - https://github.com/apache/incubator-kie-tools/pull/3189

PR #3188, opened 2025-06-25
kie-issues#890: DMN Editor: Adding a Decision to a Decision Service

aswathitv:
Closes [kie#890](https://github.com/apache/incubator-kie-issues/issues/890)

- Adding a node to a decision service in a DRD adds the node in the decision services if expanded in all DRDs.

- Moving a node to a decision service in a DRD moves the node to the decision service if expanded in all DRDs.

- Moving a node to a decision service in a DRD removes the node if decision service is collapsed in other DRDs.

PR #3162, opened 2025-05-29
kie-issues#898: DMN Editor: Moving the divider line of or resizing a Decision Service

aswathitv:
Closes [kie#898](https://github.com/apache/incubator-kie-issues/issues/898)

- Resizing a Decision Service in a DRD resizes decision service in all other DRDs that contain it.

- Moving the divider line of a decision service in a DRD replicates the change in all other DRD that contains the decision service.

- Resizing a decision service in a DRD results in repositioning of neighbouring nodes in other DRD so as to avoid overlapping.

PR #3198, opened 2025-07-07
kie-issues#2011: Support for DMN 1.6 

aswathitv:
Part of [kie#2011](https://github.com/apache/incubator-kie-issues/issues/2011)
Supports DMN1.6 and new files are created with DMN1.6 namespaces.


jomarko (commented):
@aswathitv hi, it seems still some screenshots comparision is failing. could you please doublecheck if all screenshots were updated ?

.rat-excludes
```diff
@@ -348,6 +348,14 @@ DI.xsd
...
 # packages/dmn-marshaller/src/schemas/dmn-1_5/DMNDI15.xsd
 DMNDI15.xsd
+# packages/dmn-marshaller/src/schemas/dmn-1_6/DC.xsd
+DC.xsd
+# packages/dmn-marshaller/src/schemas/dmn-1_6/DI.xsd
+DI.xsd
+# packages/dmn-marshaller/src/schemas/dmn-1_6/DMN16.xsd
+DMN16.xsd
+# packages/dmn-marshaller/src/schemas/dmn-1_6/DMNDI15.xsd
+DMNDI15.xsd
```
jomarko, line 358: This does not look correct

aswathitv: For dmn1.6 isn't schema imported from DMNDI15?

tiagobento:
Added "order:4" for this to be merged after:

- https://github.com/apache/incubator-kie-tools/pull/3188
- https://github.com/apache/incubator-kie-tools/pull/3162; and
- https://github.com/apache/incubator-kie-tools/pull/3200

tiagobento:
Conflicts... @aswathitv

tiagobento:
Thanks! This is looking good. I don't think we should be upgrading `drools-and-kogito` in this PR, though. Can you please send a separate PR doing that so this PR can remain focused on DMN 1.6 alone? Thanks!

baldimir:
Hi, I see there are XSD schemas in dmn-marshaller schemas folder. I think there was a past discussion in our mailing list mentioning this is not allowed in Apache when we publish sources as the release artifact. Is that correct, please @LightGuard (you may know more)? We created an XSD jar for that, that contains all required XSDs (1), so they are not part of the sources. I would recommend to use it instead of bundling the XSDs, so we share one single solution for this problem in KIE. 

(1) https://github.com/apache/incubator-kie-drools/tree/main/kie-dmn/kie-dmn-xsd-resources

LightGuard:
Correct. I’d have to find the thread we talked about it, but it’s in the
archives from a few months ago.

tiagobento:
IDK how much extra work it would be, and it's okay for us to do it in the future, but if possible and not too much effort, creating aliases for the DMN types generated by `xml-parser-ts-codegen` would be great to more easily do these updates in the future. E.g.,

In `dmn-marshaller` we could have:

```ts
export type DMN_LATEST__tDefinitions = DMN15__tDefinitions
export type DMN_LATEST__tDecision = DMN15__tDecision
// ...
//etc
```

This way, when DMN 1.7 arrives we can simple update this file and account for any discrepancies that happen in the dependent modules. WDYT @aswathitv?

tiagobento:
Btw @LightGuard and @baldimir this is the issue you're looking for I guess: https://github.com/apache/incubator-kie-issues/issues/1926

aswathitv:
@tiagobento Aliases for types are added.

packages/dmn-editor/src/propertiesPanel/BoxedExpressionPropertiesPanelComponents/LiteralExpressionContentCell.tsx
```diff
@@ -21,7 +21,7 @@ import * as React from "react";
 import { useMemo } from "react";
 import { DescriptionField, ExpressionLanguageField } from "../Fields";
 import { BoxedExpressionIndex } from "../../boxedExpressions/boxedExpressionIndex";
-import { DMN15__tLiteralExpression } from "@kie-tools/dmn-marshaller/dist/schemas/dmn-1_5/ts-gen/types";
+import { DMN_LATEST__tLiteralExpression } from "@kie-tools/dmn-marshaller/src/index";
```
tiagobento: Oops. Importing from './src' is not ok.

packages/dmn-editor/src/propertiesPanel/BoxedExpressionPropertiesPanelComponents/useBoxedExpressionUpdater.ts
```diff
@@ -19,10 +19,7 @@
 
 import { useCallback, useMemo } from "react";
 import { ExpressionPath, getDmnObjectByPath } from "../../boxedExpressions/boxedExpressionIndex";
-import {
-  DMN15__tBusinessKnowledgeModel,
-  DMN15__tDecision,
-} from "@kie-tools/dmn-marshaller/dist/schemas/dmn-1_5/ts-gen/types";
+import { DMN_LATEST__tBusinessKnowledgeModel, DMN_LATEST__tDecision } from "@kie-tools/dmn-marshaller/src/index";
```
tiagobento: Same here. Please double-check if there are other places.

aswathitv: Done

packages/dmn-editor/src/propertiesPanel/BoxedExpressionPropertiesPanelComponents/LiteralExpressionContentCell.tsx
aswathitv: Changed to from "@kie-tools/dmn-marshaller"

tiagobento (approved):

.rat-excludes
tiagobento: @jomarko This is correct, as not all spec files have their version updated. DI and DC are examples. Marking as resolved, and thanks for the review!

PR #3211, opened 2025-07-24
kie-issues#891: DMN Editor: Removing a Decision from a Decision Service

### LINKED ISSUE #891: Decision Services & multiple DRDs: Removing a Decision from a Decision Service.
On the DMN Editor, dragging a Decision out of a Decision Service needs to remove that Decision from all DRDs that previously included it.

aswathitv:
Closes [kie#891](https://github.com/apache/incubator-kie-issues/issues/891)

- Dragging a decision out from a decision service removes it from all DRDs that contains the decision service.


packages/dmn-editor/src/diagram/Diagram.tsx
```diff
@@ -918,6 +918,32 @@ export const Diagram = React.forwardRef<DiagramRef, { container: React.RefObject
                         dsShape["dc:Bounds"]["@_y"] =
                           (parentShape["dc:Bounds"]?.["@_y"] ?? 0) + relativePosinCurrentDS.y;
                       }
+                      // Remove decision from decision service
+                      let isInside = true;
```
tiagobento, line 922: I don't think you need to check whether the Decision node is visually inside the Decision Service, as when the Decision Service is not in a collapsed state you can assume the Decision being present in the DRD means it now needs to be removed.

tiagobento (changes requested):
Please see my comment inline.

packages/dmn-editor/src/diagram/Diagram.tsx
aswathitv: Checking if inside to make sure the movement of contained decision is not within the decision service and to make sure it is moving out like removing.

tiagobento: Right, what was strange to me is that you're checking whether the Decision `isInside` for every single DRD. But I guess because the depiction is consistent across all DRDs, this won't fail, just some extra checks. I'm ok with this. Merging.

tiagobento (approved):

PR #3218, opened 2025-08-05
NO-ISSUE: Bump drools and kogito version to 999-20250713-local

aswathitv:
Bump kojito and drools version


jstastny-cz:
Hello @aswathitv shouldn't we update to a more recent "version" given that the relevant code in kogito landed last week?

aswathitv:
@jstastny-cz 999-20250713 is the latest version

tiagobento:
You can use any date @aswathitv

PR #3219, opened 2025-08-06
NO-ISSUE: Bump `drools-and-kogito` to `999-20250813-local`

aswathitv:
Bump kogito and drools version

PR #3225, opened 2025-08-13
NO-ISSUE: Replace Cekit Behave 'serve pages' with 'curl'

ricardozanini:
In this PR:

- Removed all "check if page served" from Cekit Behave with simple `curl` within the container. It will avoid using Docker host network, simplifying the test and decreasing the time to run by ~10 times.

We might have to do a few tries before merging it, since now we can run these tests on Mac too.

Fix https://github.com/apache/incubator-kie-tools/pull/3219.


ricardozanini:
CI is failing:

> Exception: ('Phrase \'"data":{"ProcessInstances"\' was not found in the output of running the \'curl -X POST -sS -H Content-Type:application/json --data-binary \'{"query":"{ProcessInstances{id}}"}\' [http://127.0.0.1:8080/graphql\](http://127.0.0.1:8080/graphql/)' command', '<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">\n<html><head>\n<title>500 Proxy Error</title>\n</head><body>\n<h1>Proxy Error</h1>\nThe proxy server could not handle the request<p>Reason: <strong>DNS lookup failure for: host.docker.internal</strong></p><p />\n</body></html>\n', '')

This solution only applies to the Mac runners. I'll look for another one that works for both.

packages/sonataflow-management-console-image/test-resources/tests/features/sonataflow-management-console.feature
```diff
@@ -21,15 +15,6 @@ Feature: Serverless Workflow Management Console images
   Scenario: verify that the data-index is available from the container
     When container is started with env
       | variable                    | value                                                 |
-      | SONATAFLOW_MANAGEMENT_CONSOLE_DATA_INDEX_ENDPOINT | http://${DOCKER_HOST}:4000/graphql |
+      | SONATAFLOW_MANAGEMENT_CONSOLE_DATA_INDEX_ENDPOINT | http://host.docker.internal:4000/graphql |
```
fantonangeli: @ricardozanini in order to simplify the test and avoid the "host IP" issue, can be an option to avoid using a real graphql this way?
- `SONATAFLOW_MANAGEMENT_CONSOLE_DATA_INDEX_ENDPOINT=http://none.localhost:4000/graphql`
- `curl http://localhost:8080/graphql`
- then check if the container logs contains:
```
HTTP: failed to make connection to backend: none.localhost
```

ricardozanini:
Folks, please prio reviewing this PR.

wmedvede (approved):

packages/sonataflow-management-console-image/test-resources/tests/features/sonataflow-management-console.feature
```diff
@@ -23,13 +17,4 @@ Feature: Serverless Workflow Management Console images
...
-      | property             | value                                    |
-      | port                 | 8080                                     |
-      | path                 | /graphql                                 |
-      | request_method       | POST                                     |
-      | request_body         | { "query": "{ProcessInstances{ id } }" } |
-      | content_type         | application/json                         |
-      | wait                 | 480                                        |
-      | expected_status_code | 200                                      |
-      | expected_phrase      | "data":{"ProcessInstances                |
+    And run curl -X POST -sS -H Content-Type:application/json --data-binary '{"query":"{ProcessInstances{id}}"}' http://127.0.0.1:8080/graphql in container and check its output contains "data":{"ProcessInstances"
```
fantonangeli, line 20: As in my previous comment I would evaluate modifying the test this way to allow MacOS users to run tests and avoid DOCKER_HOST settings issues. WDYT?

```suggestion
      | SONATAFLOW_MANAGEMENT_CONSOLE_DATA_INDEX_ENDPOINT | http://none.localhost:4000/graphql |
    Then container log should contain httpd -D FOREGROUND
    Then container log should not contain HTTP: failed to make connection to backend: none.localhost
    And run curl -X POST -sS -H Content-Type:application/json --data-binary '{"query":"{ProcessInstances{id}}"}' http://127.0.0.1:8080/graphql in container and check its output contains 503 Service Unavailable
    Then container log should contain HTTP: failed to make connection to backend: none.localhost
```

tiagobento (approved):

packages/sonataflow-builder-image/test-resources/tests/features/sonataflow-builder.feature
```diff
@@ -22,31 +22,16 @@ Feature: SonataFlow Builder Image Sanity Checks
...
-      | request_method       | GET               |
-      | expected_status_code | 200               |
-    And container log should match regex Extension io\.quarkus:quarkus-elytron-security-jdbc.* has been installed
-    And container log should match regex Installed features:.*kogito-serverless-workflow
-    And container log should match regex Installed features:.*kie-addon-knative-eventing-extension
-    And container log should match regex Installed features:.*smallrye-health
-    And container log should match regex Installed features:.*security-jdbc
\ No newline at end of file
+    Then container log should match regex Extension io\.quarkus:quarkus-elytron-security-jdbc.* has been installed
+    And container log should match regex Installed features:.*security-jdbc
```
fantonangeli, line 37: Do we miss the curl command here to verify the page is served?

fantonangeli (changes requested):

packages/sonataflow-management-console-image/test-resources/tests/features/sonataflow-management-console.feature
ricardozanini: @fantonangeli that could be an option, indeed. I reverted it back to`DOCKER_HOST` env var. It's nice since we can test end-to-end.

packages/sonataflow-management-console-image/test-resources/tests/features/sonataflow-management-console.feature
ricardozanini: Feel free to open a PR when you get the chance :)

packages/sonataflow-builder-image/test-resources/tests/features/sonataflow-builder.feature
ricardozanini: No, I've removed it on purpose since we are already testing the remote port in other tests. My goal was to decrease the time to takes to run the tests.

Issue #3197, opened 2025-07-03
enable workflow viewer without requiring unsafe csp policies


pkliczewski:
The workflow editor sources served statically by the orchestrator backend can only run by configuring the following csp for backstage backend:

```
 script-src: ["'self'", "'unsafe-inline'", "'unsafe-eval'"] 
 script-src-elem: ["'self'", "'unsafe-inline'", "'unsafe-eval'"] 
 connect-src: ["'self'", 'http:', 'https:', 'data:']
```

See:
https://github.com/parodos-dev/orchestrator-helm-chart/blob/main/charts/orchestrator/templates/rhdh-operator.yaml#L86

batzionb:
Adding some more details:

The csp exceptions that need to be added to enable the editor are:

script-src:' 'unsafe-inline'
connect-src:  'data:'

These are not enabled by default in a standard csp header.

To reproduce, you need to serve the files from a webserver which contains the rhdh csp:

default-src 'self';
base-uri 'self';
font-src 'self' https: data:;frame-ancestors 'self';
img-src 'self' data:;
object-src 'none';
script-src 'self' 
'unsafe-eval' https://cdn.segment.com;
script-src-attr 'none';
style-src 'self' https: 'unsafe-inline';
connect-src 'self' http: https:

handreyrc:
@batzionb I investigated the issue and the root cause is how resources such as css and scripts are being added to the DOM in runtime by the diagram editor's core. 

GWT uses script / style injector that fetches the \<HEAD\> from the DOM and literally injects text / content into it at runtime. The catch is that same approach was ported to our current  J2CL implementation of the editor.  The issue is not only a matter of including those resources in the DOM,  we also need to make them available to java code otherwise we cannot consume jquery, booststrap, patternfly... 
Take a look how it works : [StunnerLienzoCore.java](https://github.com/apache/incubator-kie-tools/blob/main/packages/serverless-workflow-diagram-editor/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/StunnerLienzoCore.java#L51)

[...]

PR #3226, opened 2025-08-16
kie-issues#3197: enable workflow viewer without requiring unsafe csp …

treblereel:
…policies

issue: https://github.com/apache/incubator-kie-tools/issues/3197


handreyrc (approved):
Greate work @treblereel!
It works fine with rhdh CSP enabled locally.

ricardozanini (approved):

PR #3242, opened 2025-08-19
kie-issues#1742: DMN Editor: Option to choose between FEEL and B-FEEL as expression language

aswathitv:
Part of [kie#1742](https://github.com/apache/incubator-kie-issues/issues/1742)

Added dropdown to choose between FEEL and B-FEEL.

image
image
image


tiagobento:
Before this PR users had the option to enter the expression language that they wanted. Not sure if our Decision Engine allows custom expression languages, but the Editor did. I think we need to keep that behavior. WDYT? cc @gitgabrio

gitgabrio:
> Before this PR users had the option to enter the expression language that they wanted. Not sure if our Decision Engine allows custom expression languages, but the Editor did. I think we need to keep that behavior. WDYT? cc @gitgabrio

@tiagobento I need to investigate our engine. I see the point, of course... @baldimir @yesamer FYI: any thought ?

yesamer:
FEEL defines a subset of expression language, so I guess we would keep the previous behavior, as suggested by @tiagobento, even if for most real case scenarios final users will refer to one of those 2 options.
Moreover, this editor could be used with different engines.

gitgabrio:
@yesamer I agree with you, anyway my doubt is how the engine actually behave.
I made the B-FEEL implementation not long time ago, and now I'm wondering what would happen if the user defines a different expression language (mind you, I did not test it yet).
Beside, FEEL version is also part of the DMN version definition, for the moment being (see `KieDMNModelInstrumentedBase.java` classes) but, potentially, a user could use an "older" FEEL version with the new DMN version: is my doubt clear ? Does this make sense ?

yesamer:
@gitgabrio Yes, it does. 

Speaking of the frontend, what about this solution? https://v5-archive.patternfly.org/components/menus/select#typeahead-with-create-option. **Multiple typeahead with create option**
In that way, the user can easily switch from FEEL - BFEEL as static options, but is free to add another one.

packages/dmn-editor/src/store/Store.ts
```diff
@@ -91,6 +91,9 @@ export interface State {
   navigation: {
     tab: DmnEditorTab;
   };
+  expressionLanguages: {
+    customLanguages: string[];
+  };
```
tiagobento: Leftover

packages/dmn-editor/src/store/Store.ts
```diff
@@ -206,6 +209,9 @@ export const defaultStaticState = (): Omit<State, "dmn" | "dispatch" | "computed
     activeItemDefinitionId: undefined,
     expandedItemComponentIds: [],
   },
+  expressionLanguages: {
+    customLanguages: [],
+  },
```
tiagobento: Leftover

packages/dmn-editor/src/propertiesPanel/GlobalDiagramProperties.tsx
```diff
@@ -109,18 +122,38 @@ export function GlobalDiagramProperties() {
...
+                  isDisabled={settings.isReadOnly}
+                  selections={thisDmn.model.definitions["@_expressionLanguage"]}
+                  placeholderText={"Enter an expression language..."}
+                >
+                  {allLanguages?.map((EXPRESSION_LANGUAGES: string) => (
+                    <SelectOption key={EXPRESSION_LANGUAGES} value={EXPRESSION_LANGUAGES}>
+                      {EXPRESSION_LANGUAGES}
+                    </SelectOption>
+                  ))}
+                </Select>
```
tiagobento: Duplicated code. Why not extract it into a "ExpressionLanguageSelect" component?

tiagobento (changes requested):

tiagobento (approved):

PR #3252, opened 2025-08-21
kie-issues#1127: New swf-editor package ts / react / react flow

### LINKED ISSUE #1127: EPIC - SWF Diagram Editor in ts / react (Milestone 0)
*Summary*
-----

This Epic encompasses the Milestone 0 tasks for the new SWF Editor based on ts / react / reactflow.

It cannot be considered a MVP once it does not implement all the functionalities and UX of the current editor.
The idea is to use this functional end-to-end to gather feedback and also use it as foundation to move forward with an MVP by customizing nodes and edges and adding functionalities on top of it.

What is not supported in this Milestone:

- Evelope, Integration API to embed e distribute the editor among the multiple channels (VsCode / Standalone / Combined editor)
- UX customization for nodes and edges and associated features such as icons, tooltips, styles...

*Goals*
-----

The goal is to deliver a completely functional environment covering the basic end-to-end for the new SW diagram editor. 

*Description*
-----

In order to get a functional end-to-end the following requirements must be addressed: 
- Storybook integration for running the application, load and visualize diagrams and testing purposes 
- Marshalling / umarshalling in order to get a valid model instance from JSON / YAML raw content that later can be rendered as a diagram
- Full Model support linked to react flow nodes and edges types. Basic reanderable shapes and lines shall be created for now to represent the different types of nodes and edges. (Final UX look&Feel and associated features won't be imlemented in this interation).
- Cache and performance optimizations inherited from the original DMN editor implementation
- SWF spec 0.8  do not store node positions in JSON /YAML so an auto-layout engine is necessary in order to position nodes and edges properly in the canvas

*Tasks in this Epic*
-----

All tasks listed below shall be delivered together for this first milestone. These tasks interdependent, otherwise it won't be possible to deliver a functional end-to-end scenario that can be assessed properly.  

 - [x] https://github.com/apache/incubator-kie-issues/issues/1128
 - [x] https://github.com/apache/incubator-kie-issues/issues/1129
 - [x] https://github.com/apache/incubator-kie-issues/issues/1631
 - [x] https://github.com/apache/incubator-kie-issues/issues/1653
 - [x] https://github.com/apache/incubator-kie-issues/issues/1652
 - [x] https://github.com/apache/incubator-kie-issues/issues/1641
 - [x] https://github.com/apache/incubator-kie-issues/issues/1642
 - [x] https://github.com/apache/incubator-kie-issues/issues/1660
 - [x] https://github.com/apache/incubator-kie-issues/issues/1663
 - [x] https://github.com/apache/incubator-kie-issues/issues/1644
 - [x]  https://github.com/apache/incubator-kie-issues/issues/2058

### LINKED ISSUE #1631: SWF-Editor: Create SWF package for ts/react flow editor
*Summary*
-----

This task aims to create a starting point for the development of the new SWF editor based on ts / react / react flow.
A new package shall be created to host the new editor development and it must follow as much as possible DMN editor's code structucture and standards.  This way we ensure compatibility with the current tooling architecture and standards.

*Description*
-----

-  Create swf-editor package on kie-tools
-  It shall be possible to build and run the package (build scripts)
-  The code shall be clean and functional
-  Zustand shall be set as state manager  
-  It shall be possible to render a simple workflow of generic nodes (shape and label) and edges from a static model (Marshallers will be tackled later) 

*Testing*
-----
Assert that the new package
- New package swf-editor created in the packages directory 
- The package builds properly in dev / prod like the other packages in the kie-tools
- it should be possible to run the the component by calling the run script

### LINKED ISSUE #1641: SWF-Editor: Create react flow node types
*Summary*
-----

The goal of this task is to create is to create basic renderable nodes for each state type in the SWF spec 0.8.

Note: Check current SWF J2CL implementation of the editor for reference.

*Description*
-----

- Create the all node types SVG + Data to be rendered by react flow
- The look&feel of the nodes shall be based on SVG to avoid image distortions
- Link the nodes to state types from the model

Note: The final look and feel of the nodes will be done in other task.

*Testing*
-----

-  By running the component it should be possible to visualize and render any valid diagram with all the node types supported in the speck 0.8.

*Dependencies*
-----
- https://github.com/apache/incubator-kie-issues/issues/1652

### LINKED ISSUE #1642: SWF-Editor: Create react flow edge types
*Summary*
-----

The goal of this task is to create is to create all renderable edge types described in the SWF 0.8 spec.

Note: Check current SWF J2CL implementation of the editor for reference.

*Description*
-----

- Implement React Flow edge type settings so the the library can identify what type of edge shall be rendered and its properties from the provided model instance  
- Add support to waypoints (CPs) 

*Testing*
-----

-  By running the component it should be possible to visualize and render any valid diagram with all the edge types supported in the speck 0.8.

*Dependencies*
-----
- https://github.com/apache/incubator-kie-issues/issues/1652

### LINKED ISSUE #1644: SWF-Editor: Add auto-layout capabilities to the diagram
*Summary*
-----

The goal of this task is to add auto-layout capabilities to the diagram.
Note: ELKjs is being used by DMN so consider using the same auto-layout engine.

*Description*
-----

-  Include and integrate the auto-layout lib to react flow
- Apply auto-layout when the content is set to be rendered
- The workflow layout must be top-down with orthogonal lines
- Apply calculated node positions to all nodes
- Apply calculated vertexes (CPs) for each edge to avoid lines crossing and overlaping (ELKjs provides vertexes coordinates)

*Testing*
-----
- By running the component it should be possible to verify that the layout of the nodes is top-down with ortogonal lines connecting them
- It should be no overlaps or line cossings

*Dependencies*
-----
- https://github.com/apache/incubator-kie-issues/issues/1641
- https://github.com/apache/incubator-kie-issues/issues/1642

### LINKED ISSUE #1652: SWF-Editor: Full SWF Model suport
*Summary*
-----

The goal of this task is to enable the full SWF 0.8 spec support, the model instance itself will be later used to render nodes.
Note: The editor takes a valid model instance or an error object as as props. 

*Description*
-----

The editor is completely agnostic of the content type because it takes a model intance to render the workflow or and error as props.  Thus it does not matter to the editor if the content is YAML or JSON. 
Spec links:
- [ts definitions](https://github.com/serverlessworkflow/specification/blob/0.8.x/specification.md#)

In oder the get the full SWF support:
-  Ensure the complete SWF spec support
-  Implement the model as source of truth for the renderable workflow by implementing the proper bidings and hooks to keep the model up-to-date.
- The model shall be kept up to date in both ways model to diagram and diagram to workflow, although the edit mode won't be enabled in this first implementantion it shall be considered to avoid major refactorings in the near future.
- Update the sample model to have complete state and transition samples to assertaing the SWf spec is fully covered

*Testing*
-----
Run the component then select  load the sample workflows available in storybook. 
- The diagram should be capable of parse, unmarshall and render the workflow based on the selected content.
- Model changes handled by the Zustand library

*Dependencies*
-----
- https://github.com/apache/incubator-kie-issues/issues/1653

### LINKED ISSUE #1653: SWF-Editor: Marshallers JSON / YAML
*Summary*
-----

The goal of this taks it so implement/ integrate marshallers in order to enable the editor to handle JSON / YAML content.
The marshallers shall be capable of transforming content into a renderable model and vice versa.

*Description*
-----

In order to populate the model from JSON / YAML we need a marshaller implementation capable of parsing content into the a renderable model.
Instead of implementing mashallers from scratch it should be possible to move the current marshallers implementation (J2CL) to an external lib consumed by the new editor and then consume it.

- Support JSON / YAML content
- Support unmarshall and marshall operations (JSON / YAML to Model and Model to JSON / YAML)
- Catch and print parser / marshaller errors on the console log
- Add YAML and JSON sample payloads to storybook so that content can be loaded and rendered by the Diagram editor

*Testing*
-----

Run the component then select  load the sample workflows available in storybook. The diagram should be capable of parse, unmarshall and render the workflow based on the selected content.

*Dependencies*
-----
- https://github.com/apache/incubator-kie-issues/issues/1631

### LINKED ISSUE #1660: SWF-Editor: Enable read-only mode
*Summary*
-----

The goal of this task is to implement the read only mode.  It should be possible to pass an initialization paramenter to disable / enable editing capabilities. 

*Description*
-----

At this point of the implentation it should be possible to move nodes and edges. The idea is to disable all the editing capabilities in the canvas by setting a configurable initialization flag.
With the read-oly mode enabled it should not be possible to move nodes and edges or peforform any operation that implies in updating the model.
However, it should still be possible to select nodes and edges.

*Testing*
-----

Run the component and load a sample workflow from storybook.
- With the addon "Controls loaded" find the toggle swith "isReadOnly" and set it to "True"
- When in readonly mode it should not be possible to move nodes around and the palette should not be visible

*Dependencies*
-----

- https://github.com/apache/incubator-kie-issues/issues/1641
- https://github.com/apache/incubator-kie-issues/issues/1642

### LINKED ISSUE #1663: SWF-Editor: Enable Zoombar
*Summary*
-----

The goal of this task is to enable reactflow native zoombar. 

*Description*
-----

React flow has a buit in zoom bar so the idea is to use as much as possible reactflow native components.

- Enable react flow native  toolbar
- Ensure that zoom in +, zoom out - and fit to screen [ ]  buttons are availabe and functional 

*Testing*
-----

Run the component and load a sample workflow from storybook.
- Assertain that the zoom bar is visible and that all of  its buttons are functional.

*Dependencies*
-----

- https://github.com/apache/incubator-kie-issues/issues/1641
- https://github.com/apache/incubator-kie-issues/issues/1642

### LINKED ISSUE #2058: SWF-Editor: Migrate from Patterfly 4 to Patternfly 5
*Summary*
-----

The goal of this task is to upgrade the Patternfly version of the new editor PoC from Patterfly 4 to Patternfly 5.

Note: The migration from Patterfly 4 to 5 is already meged in the main and to move forward with the new editor it must follow the same Patternfly version used in the main. 

*Description*
-----

- Update dependecies to Patterfly 5
- fix all patternfly references in the code to Patterfly 5 (CSS and ts)

*Testing*
-----

-  Check Patterfly 5 verstion in package.json
- No style complaints in the console or broken CSS in the diagram

handreyrc:
This PR includes the Milestone 0 tasks for the new swf editor (ts / react /react flow).

Check the epic below to track the tasks delivered:
Issue: [#1127](https://github.com/apache/incubator-kie-issues/issues/1127)

Github pages Demo:

[new-swf-diagram-editor](https://handreyrc.github.io/SWF_Pages/?path=/story/misc-empty--empty)


ricardozanini (approved):

packages/swf-editor/env/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+const { varsWithName, composeEnv } = require("@kie-tools-scripts/build-env");
+
+module.exports = composeEnv([require("@kie-tools/root-env/env"), require("@kie-tools-core/webpack-base/env")], {
+  vars: varsWithName({}),
+  get env() {
+    return {
+      swfEditor: {
+        storybook: {
+          port: "9901",
```
tiagobento: Same port as other packages. Please find a port that's not being used yet.

packages/swf-editor/src/store/Store.ts
```diff
@@ -0,0 +1,304 @@
...
+    edgeIdBeingUpdated: string | undefined;
+    dropTargetNode: DropTargetNode;
+    ongoingConnection: RF.OnConnectStartParams | undefined;
+    overlaysPanel: {
+      isOpen: boolean;
+    };
+    overlays: {
+      enableNodeHierarchyHighlight: boolean;
+      enableCustomNodeStyles: boolean;
+      enableEvaluationHighlights: boolean;
```
tiagobento: This value does not make sense for the SWF Editor.

tiagobento: It's a DMN-specific thing for Decision Tables and Boxed Conditionals.

packages/swf-editor/src/store/Store.ts
```diff
@@ -0,0 +1,304 @@
...
+//Handle layout data out of the model
+export type Layout = {
+  setNodePosition: (nodeId: string, newPosition: RF.XYPosition | undefined) => void;
+  setEdgeWaypoints: (edgeId: string, newWaypoints: RF.XYPosition[] | undefined) => void;
+  updateWaypointPosition: (edgeId: string, index: number, newPosition: RF.XYPosition | undefined) => void;
+};
+
+export class NodesPosition {
+  // Handle node position out of the model
+  [immerable] = false;
```
tiagobento: What does this do?

packages/swf-editor/src/swfUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Build edge Id string based on source, target and edgeType
+ */
+export const buildEdgeId = (
+  source: string | undefined,
+  target: string | undefined,
+  edgeType: SwfEdgeTypes | undefined
+) => {
+  return source + "_" + target + "_" + edgeType?.toString();
+};
```
tiagobento: There's a folder `diagram/edges/`. Why not put this function there?

tiagobento: I'm not a big fan of global "utils" files, as they tend age badly, grouping unrelated code into the same file, often creating unnecessary coupling and making our lives harder in the future when a refactor/modularization effort is necessary.

packages/swf-editor/src/evaluationHighlights/EvaluationHighlightsBadge.tsx
tiagobento, line 1: This file and folder do not belong in the SWF Editor. Please see the other comment about the same feature.

packages/swf-editor/src/SwfEditor.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+import "@patternfly/react-core/dist/styles/base.css";
+import "reactflow/dist/style.css";
+
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+import * as RF from "reactflow";
+import { ErrorBoundary, ErrorBoundaryPropsWithFallback } from "react-error-boundary";
+import { useCallback, useEffect, useImperativeHandle, useMemo, useRef, useState } from "react";
+//import { original } from "immer";
```
tiagobento: Why?

packages/swf-editor/src/SwfEditor.tsx
```diff
@@ -0,0 +1,289 @@
...
+import { Diagram, DiagramRef } from "./diagram/Diagram";
+import { SwfEditorContextProvider, useSwfEditor } from "./SwfEditorContext";
+import { SwfEditorErrorFallback } from "./SwfEditorErrorFallback";
+import { ComputedStateCache } from "./store/ComputedStateCache";
+import { Computed, createSwfEditorStore, defaultStaticState } from "./store/Store";
+import { SwfEditorStoreApiContext, StoreApiType, useSwfEditorStore, useSwfEditorStoreApi } from "./store/StoreContext";
+import { SwfDiagramSvg } from "./svg/SwfDiagramSvg";
+import { useEffectAfterFirstRender } from "./useEffectAfterFirstRender";
+import { INITIAL_COMPUTED_CACHE } from "./store/computed/initial";
+import "./SwfEditor.css"; // Leave it for last, as this overrides some of the PF and RF styles.
```
tiagobento: Shouldn't it be the last import?

packages/swf-editor/src/SwfEditor.tsx
```diff
@@ -0,0 +1,289 @@
...
+   */
+  onModelChange?: OnSwfModelChange;
+  /**
+   * To show information about execution results directly on the SWF diagram, use this prop.
+   */
+  evaluationResults?: EvaluationResults;
+  /**
+   * To show information about validation messages directly on the SWF diagram, use this prop.
+   */
+  validationMessages?: ValidationMessages;
```
tiagobento: Both do not belong in the SWF Editor.

packages/swf-editor/src/SwfEditor.tsx
```diff
@@ -0,0 +1,289 @@
...
+  onRequestToJumpToPath?: OnRequestToJumpToPath;
+  /**
+   * All paths inside the SWF Editor are relative. To be able to resolve them and display them as absolute paths, this function is called.
+   * If undefined, the relative paths will be displayed.
+   */
+  onRequestToResolvePath?: OnRequestToResolvePath;
+  /**
+   * Notifies the caller when the SWF Editor performs a new edit after the debounce time.
+   */
+  onModelDebounceStateChanged?: (changed: boolean) => void;
```
tiagobento: Are those being used? I'd like to avoid having dead-code introduced from the start.

packages/swf-editor/src/SwfEditor.tsx
```diff
@@ -0,0 +1,289 @@
...
+      // Avoid unecessary state updates
+      // if (model === original(state.swf.model)) {
+      //   return;
+      // }
+      // TODO: Swf model is not immerable so a draft cannot be created
+      // Consider the possibility of adding immerable to all sdk mddel classes
+      // Comparing strings as workaround
+      if (JSON.stringify(model) === JSON.stringify(state.swf.model)) {
+        return;
+      }
```
tiagobento: What makes the model not "immerable"? Maybe it's not suited to be a state too?

packages/swf-editor/stories/dev/DevWebApp.stories.tsx
```diff
@@ -0,0 +1,253 @@
...
+import * as React from "react";
+import { useCallback, useRef, useState } from "react";
+import type { Meta, StoryObj } from "@storybook/react";
+import "@patternfly/react-core/dist/styles/base.css";
+import { Flex, FlexItem } from "@patternfly/react-core/dist/js/layouts/Flex";
+import { Page, PageSection } from "@patternfly/react-core/dist/js/components/Page";
+import { SwfEditorWrapper } from "../swfEditorStoriesWrapper";
+import { SwfEditorProps, OnSwfModelChange, OnRequestToJumpToPath } from "../../src/SwfEditor";
+import { Specification } from "@severlessworkflow/sdk-typescript";
+import { Marshaller, getMarshaller, serializeSwf } from "../../src/marshallers/swfMarshaller";
```
tiagobento: This import going a little too far back? :D Why go outside the `src` folder just to re-enter it?

packages/swf-editor/src/marshallers/swfMarshaller.ts
tiagobento, line 1: Both new DMN and the upcoming BPMN Editor abstract the marshalling process away, purposefully, to couple itself only with the model's interfaces. Please consider the same strategy here. We don't want the editor to do any kind of heavy lifting when it comes to **file** formats. Graphical editors for DMN, BPMN, SWF, and whatever other format we might want to support in the future, IMHO, should not know the concept of a "file", let alone a "file format". By defining this hard boundary we can confidently keep the editors being Controlled Components, with easy to understand mechanics and expectations from our users.

packages/swf-editor/src/mutations/renameNode.ts
```diff
@@ -0,0 +1,37 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Specification } from "@severlessworkflow/sdk-typescript";
+
+// TODO: we need to update all references to ID/Name if we want to change a name.
```
tiagobento: Maybe map this TODO in a new issue? I remember there's an Epic for the new SWF Editor, no?

tiagobento (changes requested):
Hi! Thanks for the PR @handreyrc. Please consider my comments inline. Thank you!

tiagobento:
I would also like to point out the Palette in the video of the description, which contains a lot of identical rectangles. Can you please either address that or explain why that is necessary for this new version of the SWF Editor? Thanks!

packages/swf-editor/env/index.js
handreyrc: @handreyrc Done! Just curious, why it has to be unique port number in kie-tools?

packages/swf-editor/src/store/Store.ts
handreyrc: @tiagobento Done! 
The node hierarchy highlight makes sense to keep. We will need it also to highlight an executed path.

packages/swf-editor/src/store/Store.ts
handreyrc: @tiagobento This is a workaround for and issue I was facing with states in Zustand in this specific scenario.
SWF do not store nodes and edges positional dada. In this case I have to calculate the autolayout and apply those coordinates / waypoints to nodes and lines in the "computeDiagramData" before the diagram is rendered.
Otherwise the diagram would be rendered twice and that's pretty unpleasant visually. Without setting  "[immerable] = false;" for the nodes position / waypoints  zustand will wouldn't allow to change / set  new calculated coordinates in the  "computeDiagramData" because in that context they would be immutable.  
That's the way I found to make it work. I'm open to suggestions if there is  better way to do that.

packages/swf-editor/src/swfUtils.ts
handreyrc: @tiagobento Done!

packages/swf-editor/src/evaluationHighlights/EvaluationHighlightsBadge.tsx
handreyrc: @tiagobento I agree, I forgot this one behind. Thanks for spoting it!

packages/swf-editor/src/SwfEditor.tsx
handreyrc: @tiagobento There is a big issue with immer and the 0.8 SWF SDK. 
The SDK model is class based, and to make it work with immer it would be necessary to add "[immerable] = true;" to all model classes in the SDK.  It is not possible at the moment. So for now we have to compare strings.
This topic is something that has to be raised to the 0.8 SWF SDK maintainers.

packages/swf-editor/src/SwfEditor.tsx
handreyrc: @tiagobento Done!

packages/swf-editor/src/SwfEditor.tsx
handreyrc: @tiagobento I agree, it is removed!

packages/swf-editor/src/SwfEditor.tsx
handreyrc: @tiagobento We have subflows that would point to other files, however, it is not supported at the moment.
I was considering to keep it, but I agree that it is not in use now and it makes sense to remove it.

packages/swf-editor/src/SwfEditor.tsx
handreyrc: @tiagobento The SDK model is class based so it won't work with immer as is. The best we can do for now is to compare strings. I'm open to suggestions here.

packages/swf-editor/stories/dev/DevWebApp.stories.tsx
handreyrc: @tiagobento It is in the storybook, outside the src. As you can see we already have a similar existing case before it.

packages/swf-editor/src/marshallers/swfMarshaller.ts
handreyrc: @tiagobento The 0.8 SWF SDK is class based so the same interface approach wouldn't work without creating something in the middle generate / convert  it to interfaces. I understand that DMN editor takes a model instance and that validation / error handling is done in the envelope. It is something that can be considered when we will create the envelope. On the other hand, we may need to use the diagram editor without the envelope and maybe we would keep it as is. 
For now we need it to make it work without the envelope.

packages/swf-editor/src/mutations/renameNode.ts
handreyrc: @tiagobento I removed that TODO. It was more a note to myself. Edit mode is still under discussion anyway.

handreyrc:
> I would also like to point out the Palette in the video of the description, which contains a lot of identical rectangles. Can you please either address that or explain why that is necessary for this new version of the SWF Editor? Thanks!

@tiagobento Edit mode is pretty desirable, however, it still under discussion how it should work. It is definitely something we want but it is for the future. I preserved as much as I could of the edit mode as base for this future implementation.
For now the new editor will run in read-only mode only.

packages/swf-editor/env/index.js
tiagobento: When you run tests based on the dev webapps in parallel, we don't want ports to clash.

tiagobento:
@handreyrc Thanks, but I think we shouldn't introduce a new component with code that we're unsure whether or not will be needed in the future. Can you please remove the parts that are no-op/dead code for now? Since this new editor is heavily inspired by DMN Editor, it should be pretty easy to bring it back if necessary.

packages/swf-editor/src/marshallers/swfMarshaller.ts
tiagobento: In that case, my suggestion will be for you to create a "StandaloneSwfEditor" component that contains the marshalling logic, and leave the `SwfEditor` component agnostic when it comes to files and formats/extensions. Introducing a new editor with this logic inside it will make it increasingly hard to de-couple in the future when it evolves.

packages/swf-editor/src/SwfEditor.tsx
tiagobento: Well, if what's available doesn't work for us, we should be fixing the problem ourselves. The SWF specification doesn't account for its model classes to be used as React states, let alone immer-compatible. I don't think it is something to be raised with the 0.8 SWF SDK maintainers, as that's not what they created those classes for.

This to me seems like a workaround based on a model structure that is not compatible with the use we're doing here.

My recommendation is that we either:
1. Have a plan to address this ASAP; or
2. Fix it now by either owning the module or patching the SWF SDK in build- or runtime.

packages/swf-editor/src/SwfEditor.tsx
tiagobento: That is a consequence of using model types that are not compatible with the use we're doing here. See my comment above please.

packages/swf-editor/src/store/Store.ts
tiagobento: Yeah, I'm afraid this is "polluting" the codebase due to using incompatible model types to represent state. See my comment above about this please.

tiagobento:
I thought I'd said that in my previous review, but please rename the package to `serverless-workflow-editor`, as we currently don't have any `swf-`-prefixed packages, only `serverless-workflow-`.

handreyrc:
> @handreyrc Thanks, but I think we shouldn't introduce a new component with code that we're unsure whether or not will be needed in the future. Can you please remove the parts that are no-op/dead code for now? Since this new editor is heavily inspired by DMN Editor, it should be pretty easy to bring it back if necessary.

@tiagobento We need a Palette for dragging and drop nodes and means to connect those nodes. Those functionalities will remain anyway. What is uncertain for now is how that UX will look like. On top of that we will also need that for the new spec 1.0.
In my opinion we are safe here.

packages/swf-editor/src/marshallers/swfMarshaller.ts
handreyrc: @tiagobento I agree, we can move the raw content handling to the envelope in the next iteration when we will have the envelope.  It can be done with little effort.  For now we need it to have a renderable workflow.
WDYT?

packages/swf-editor/src/SwfEditor.tsx
handreyrc: @tiagobento The remaining issues are all related to the class based model provided by the SDK.
We are going to fix the SDK and I've already created and issue for that.
https://github.com/serverlessworkflow/sdk-typescript/issues/194

handreyrc:
> I thought I'd said that in my previous review, but please rename the package to `serverless-workflow-editor`, as we currently don't have any `swf-`-prefixed packages, only `serverless-workflow-`.

@tiagobento sorry, I missed this one. It is fixed "severless-workflow-editor" it is.

handreyrc:
@ricardozanini @tiagobento,

This PR is ready for another round!

* 0.8.4 ts SKD is now interface based and compatible with immer so all the workaronds were removed.
* Non immerable classes to store nodes and edges position were removed. Actually the issue is that it does not work with map so using arrays is the way to go.
* Toolbars were removed from edit mode since it is not a complete feature, I also removed the code around these functionalities.
* We also have a Github pages with the editor functional, check the link on top of the PR. 

Please, could you review it again?

Thanks for the guidance!

tiagobento:
@handreyrc Awesome! Thank you very much!!

I didn't have time to do a thorough review yet, but I looked at the Storybook (thanks for making it so easy to review it!)

1. Dev webapp is allowing nodes to be moved. Is that intentional?
2. Error edges seem to be broken? 
  - image
3. The "Empty" case doesn't seem to be empty? 😅
  - image

handreyrc:
> @handreyrc Awesome! Thank you very much!!
> 
> I didn't have time to do a thorough review yet, but I looked at the Storybook (thanks for making it so easy to review it!)
> 
> 1. Dev webapp is allowing nodes to be moved. Is that intentional?
> 2. Error edges seem to be broken?
> 
> * image
> 
> 3. The "Empty" case doesn't seem to be empty? 😅
> 
> * image

Hey @tiagobento ,

1) it's intentional, it should be possible to move nodes in the dev session. 
2) Error edges are not broken, we can handle workflow exceptions by using compesation transitions / edges. Those examples come from our webtools [kubesmarts](https://start.kubesmarts.org/#/sample-catalog). Anyway I renamed that workflow and created a new one with actual error transitions to make it easy to grasp.
3) Good catch, it is fixed!

Thanks a lot for rechecking my work!

jomarko (commented):
Hi team. Please, is the link for the github.io pages up to date? In the conversation I see you mention palette as component for dragging new elements to the canvas. However on my machine I do not see it and I didn't figure out a way to add a new node to the canvas.

Am I missing something? Could someone describe how the editor should be used?  I thought on high level it will be alligned with BPMN and DMN. That is why I would expect the palette to be there.

Or is my environment unsupported? Using windows and chrome for browsing the github.io.

image

handreyrc:
> Hi team. Please, is the link for the github.io pages up to date? In the conversation I see you mention palette as component for dragging new elements to the canvas. However on my machine I do not see it and I didn't figure out a way to add a new node to the canvas.
> 
> Am I missing something? Could someone describe how the editor should be used? I thought on high level it will be alligned with BPMN and DMN. That is why I would expect the palette to be there.
> 
> Or is my environment unsupported? Using windows and chrome for browsing the github.io.
> 
> image

Hey @jomarko,

It is up-to-date. The edit mode is the next step. We need a readonly editor integrated with the text editor before moving forward with an edit mode. The plan is to replace the J2CL canvas we are currently using and then add editing capabilities on top of it.
Despite of having the edit mode disabled initially, I thoght it would be a good idea to preserve the palette and other overlays as a good starting point for the edit mode future developments. However, as you can see in the history of this PR, I was asked to remove it because it is not something we would be using out of the box and at this point it is not completely clear how it should work. We still have to mature the requirements and UX and also figure out how it would go along with the text editor. 
It shouldn't be hard to put it back following the DMN footprints when the time comes. Anyway, we need to move forward with this initial PR to get means to collaborate and get other things going such as an edit mode MVP.

tiagobento:
> - 0.8.4 ts SKD is now interface based and compatible with immer so all the workaronds were removed.
> - Non immerable classes to store nodes and edges position were removed. Actually the issue is that it does not work with map so using arrays is the way to go.
> - Toolbars were removed from edit mode since it is not a complete feature, I also removed the code around these functionalities.
> - We also have a Github pages with the editor functional, check the link on top of the PR.

Thanks for addressing those change requests! Great seeing the SWF Editor decoupled from marshalling. Let's please wait for @jomarko's 👀 , but consider my approval! Thank you.

tiagobento (approved):

packages/serverless-workflow-editor/package.json
```diff
@@ -0,0 +1,117 @@
...
+    "@kie-tools/i18n-common-dictionary": "workspace:*",
+    "@kie-tools/xml-parser-ts": "workspace:*",
+    "@patternfly/react-core": "^5.4.1",
+    "@patternfly/react-icons": "^5.4.1",
+    "@patternfly/react-styles": "^5.4.0",
+    "@serverlessworkflow/sdk-typescript": "^0.8.4",
+    "@visx/text": "^3.3.0",
+    "d3-drag": "^3.0.0",
+    "d3-selection": "^3.0.0",
+    "elkjs": "^0.9.1",
```
jomarko, line 55: Could we please double check this report?

⚠️ Not Apache‑compatible: elkjs

elkjs is licensed under the Eclipse Public License 2.0 (EPL-2.0). [github.com], [github.com]
EPL‑2.0 is not compatible with Apache‑2.0 for redistribution or combination purposes, because it imposes distinct obligations (especially regarding patent grants and distribution rules).

packages/serverless-workflow-editor/stories/useCases/workflowSamples/WorkflowCompensationExample.mdx
jomarko, line 1: The used yellow color is difficult to see on the white background

image

packages/serverless-workflow-editor/stories/misc/empty/Empty.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { SwfEditorWrapper, StorybookSwfEditorProps } from "../../swfEditorStoriesWrapper";
+import { SwfEditor, SwfEditorProps } from "../../../src/SwfEditor";
+import { Specification, WorkflowValidator } from "@serverlessworkflow/sdk-typescript";
```
jomarko: some imports are unused

packages/serverless-workflow-editor/tests-e2e/__fixtures__/jsonModel/swf.ts
```diff
@@ -0,0 +1,38 @@
...
+import { SwfElement } from "./swfElement";
+import { Specification } from "@serverlessworkflow/sdk-typescript";
+
+export class Swf {
+  constructor(
+    public page: Page,
+    public swfElement: SwfElement
+  ) {}
+
+  private async getSwf(): Promise<object | undefined> {
```
jomarko: reported as unused

packages/serverless-workflow-editor/tests-e2e/__fixtures__/stories.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { Page } from "@playwright/test";
+
+export class Stories {
+  constructor(
+    private readonly page: Page,
+    private readonly baseURL?: string
+  ) {}
+
+  private getIframeURL(iframeId: string) {
```
jomarko, line 28: reported as unused

packages/serverless-workflow-editor/tests-e2e/readOnly/readOnlyDiagram.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  await editor.openApplicantWorkflow();
+  await expect(nodes.get({ name: "HandleNewApplicant" })).toBeVisible();
+  await editor.setIsReadOnly(true);
+});
+
+test.describe("Read Only mode - Diagram", () => {
+  test("should not be able to drag node", async ({ page, nodes }) => {
+    await nodes.move({ name: "HandleNewApplicant", targetPosition: { x: 0, y: 0 } });
+    await expect(page).toHaveScreenshot("readOnly-node-not-moved.png");
+  });
```
jomarko, line 31: maybe it would be add also a test, that `dblclick` has no effect

packages/serverless-workflow-editor/src/clipboard/Clipboard.ts
```diff
@@ -0,0 +1,93 @@
...
+};
+
+export function buildClipboardFromDiagram(rfState: RF.ReactFlowState, swfEditorState: State) {
+  const copiedNodesById = new Map<string, RF.Node<SwfDiagramNodeData>>();
+
+  const selectedNodesById = rfState
+    .getNodes()
+    .reduce((acc, n) => (n.selected ? acc.set(n.id, n) : acc), new Map<string, RF.Node<SwfDiagramNodeData>>());
+
+  const clipboard = [...selectedNodesById.values()].reduce<SwfEditorDiagramClipboard>(
```
jomarko, line 42: is this warning/message related just to my machine?

image

packages/serverless-workflow-editor/src/diagram/nodes/OutgoingStuffNodePanel.tsx
```diff
@@ -0,0 +1,211 @@
...
+}) {
+  const settings = useSettings();
+  const style: React.CSSProperties = React.useMemo(
+    () => ({
+      visibility: !settings.isReadOnly && props.isVisible ? undefined : "hidden",
+    }),
+    [props.isVisible, settings.isReadOnly]
+  );
+
+  const getEdgeActionTitle = React.useCallback((edgeType: string): string => {
```
jomarko, line 76: we should introduce `i18n` into serverless-workflow-editor

see https://github.com/apache/incubator-kie-tools/pull/3351 and similar

packages/serverless-workflow-editor/src/diagram/Diagram.tsx
```diff
@@ -0,0 +1,793 @@
...
+const edgeTypes: Record<EdgeType, any> = {
+  [EDGE_TYPES.compensationTransition]: CompensationTransitionEdge,
+  [EDGE_TYPES.dataConditionTransition]: DataConditionTransitionEdge,
+  [EDGE_TYPES.defaultConditionTransition]: DefaultConditionTransitionEdge,
+  [EDGE_TYPES.errorTransition]: ErrorTransitionEdge,
+  [EDGE_TYPES.eventConditionTransition]: EventConditionTransitionEdge,
+  [EDGE_TYPES.transition]: TransitionEdge,
+};
+
+const MIME_TYPE_FOR_SWF_EDITOR_NODE = "kie-swf-editor--node";
```
jomarko, line 117: reported unused

packages/serverless-workflow-editor/src/diagram/Diagram.tsx
```diff
@@ -0,0 +1,793 @@
...
+    useEffect(() => {
+      const edgeUpdaterSource = document.querySelectorAll(
+        ".react-flow__edgeupdater-source, .react-flow__edgeupdater-target"
+      );
+      if (ongoingConnection) {
+        edgeUpdaterSource.forEach((e) => e.classList.add("hidden"));
+      } else {
+        edgeUpdaterSource.forEach((e) => e.classList.remove("hidden"));
+      }
+    }, [ongoingConnection]);
```
jomarko, line 198: just wanted to ask, if there was some discussion about common editor parts and potentially extract them into a separate module? just asking, to have better picture about the future, not asking any change in this PR but I see the same code in DMN Editor

packages/serverless-workflow-editor/src/SwfEditor.css
```diff
@@ -0,0 +1,563 @@
...
+.kie-swf-editor--autolayout-panel-toggle-button:active {
+  filter: brightness(90%);
+}
+/* (end) autolayout panel toggle */
+
+/* (begin) evaluation-highlights panel toggle */
+.kie-swf-editor--evaluation-highlights-panel-toggle {
+  min-width: 190px;
+}
+/* (end) evaluation-highlights panel toggle */
```
jomarko: if I am not wrong, evaluation highlights a feature specific only for DMN, if that is true please remove it

if I understand correctly the PR discussion, this PR should not contain any code, that has currently no usage, is that correct?

jomarko (commented):
Thank you for the PR. Huge amount of the work. I put some comments/points/question inline. 

Unfortunately, I believe 100% proper review will be possible just with edit mode.

packages/serverless-workflow-editor/package.json
handreyrc: @jomarko, 

That's an interesting point. We are not tempering with the Elkjs source codebase, however, DMN, BPMN and the SWF code in this PR are consuming this lib as a dependency.

We should bring it up into discussion and maybe escalate it because if it is really an issue it concerns the kie editors in general manner. 

I don't think it should block this work but we must take it into account for the kie-editors.

WDYT?

packages/serverless-workflow-editor/stories/useCases/workflowSamples/WorkflowCompensationExample.mdx
handreyrc: @jomarko,

I made it darker. Anyway, there is a round of UX customization coming once this PR is meged so it should not be a big deal for now.

Thanks.

packages/serverless-workflow-editor/stories/misc/empty/Empty.stories.tsx
handreyrc: Fixed!

packages/serverless-workflow-editor/tests-e2e/__fixtures__/jsonModel/swf.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/tests-e2e/__fixtures__/stories.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/tests-e2e/readOnly/readOnlyDiagram.spec.ts
handreyrc: Double click in edit mode should trigger rename node. I added a test to check that.

packages/serverless-workflow-editor/src/clipboard/Clipboard.ts
handreyrc: I noticed that VsCode sometimes trigger false positives and warnings.
No warnings in my env.
image

packages/serverless-workflow-editor/src/SwfEditor.css
handreyrc: You are correct, if it is DMN specific it must be removed.
However, It is not totally true for all the cases. For example the overlay panel, palette, properties panel... all of it will be put back for the editing mode soon. There are also situations where it does not worthwhile to refactor the code for now and then have to undo it later. This PR is an entry point for team collaboration. it is a Milestone 0, so an MVP is the next step.

packages/serverless-workflow-editor/src/diagram/Diagram.tsx
handreyrc: Looking at the highlighted code above, connection rules can become pretty specific. I agree that maybe we could abstract it and then take specific connection rules as an input.
Some time ago I asked about that. By that time the answer was no. The editors were being developed independently of each other to a certain extent.

packages/serverless-workflow-editor/src/diagram/nodes/OutgoingStuffNodePanel.tsx
handreyrc: I checked and the i18n inplementation is relatively new so by the time I forked the code it wasn't there. I ported the i18n into  serverless-workflow-editor.

handreyrc:
Hey @jomarko,

I fixed the issues you pointed. 
This PR is ready for another review round.

Thanks!

packages/serverless-workflow-editor/src/i18n/locales/en.ts
```diff
@@ -0,0 +1,84 @@
...
+  off: "off",
+  overlaysPanel: {
+    snapping: "Snapping",
+    horizontal: "Horizontal",
+    vertical: "Vertical",
+    highlightSelectedNode: "Highlight selected node(s) hierarchy",
+  },
+  swfEditor: {
+    editor: "Editor",
+    unexpectedErrorOccured: "An unexpected error happened",
```
jomarko: `occured` vs `happened`

packages/serverless-workflow-editor/src/i18n/locales/en.ts
```diff
@@ -0,0 +1,84 @@
...
+    addSleepStateNode: "Add Sleep State",
+    addSwitchStateNode: "Add Switch State",
+    emptyDiagram: "Empty Diagram",
+    swfDiagramEmpty: "This SWF Diagram is empty",
+    diagramHasNodesOrOpenAnotherFile: "Make sure the workflow has nodes or try opening another file",
+    startByDraggingNodes: "Start by dragging nodes from the Palette",
+    overlays: "Overlays",
+    nodesSelected: (selectedNodesCount: number) => `${selectedNodesCount} nodes selected`,
+    edgesSelected: (selectedEdgesCount: number) => `${selectedEdgesCount} edges selected`,
+    nodeSelected: (nodeCount: number) => `${nodeCount} node`,
```
jomarko: looking on the usage in code, this is actually:
```
oneNode: "1 node"
```

packages/serverless-workflow-editor/src/i18n/locales/en.ts
```diff
@@ -0,0 +1,84 @@
...
+    edgeSelected: (edgeCount: number) => `${edgeCount} edge`,
```
jomarko: looking on the usage in code, this is actually:
```
oneEdge: "1 edge"
```

packages/serverless-workflow-editor/src/i18n/locales/en.ts
```diff
@@ -0,0 +1,84 @@
...
+export const en: SwfEditorI18n = {
+  ...en_common,
+  autoLayout: "Autolayout (beta)",
+  close: "Close",
+  none: "None",
+  cancel: "Cancel",
+  nodes: {
+    view: "View",
+    edit: "Edit",
+    output: "OUTPUT",
```
jomarko: It is somehow interesting, this is the only string we have completely in upper case. is this somehow related to specification? Is this somehow aligned with the rest of the UI?

jomarko (commented):
Thank you, changes looks good to me. Just sharing feedback for some translations.

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+    }
+  }
+
+  //clean up unused edges waypoits
+  if (edgeWaypoints) {
+    for (const id of edgeIds) {
+      if (!edgesById.get(id)) {
+        const i = edgeIds.indexOf(id);
+        if (i !== -1) {
+          edgeWaypoints.splice(1, i);
```
fantonangeli: If I go to: 
-"Dev - Web app"
- upload a WF with the drag&drop ie: `jsongreet.sw.json`
- click on Autolayout
- upload another WF ie: `callback-state-timeouts.sw.json`

I get:
```
SwfEditorErrorFallback.tsx:47 TypeError: Cannot assign to read only property 'length' of object '[object Array]'
```

In the real use cases, the user should not reach this issue, and we can also fix it in a separate PR, if needed, I think

packages/serverless-workflow-editor/src/i18n/locales/en.ts
```diff
@@ -0,0 +1,84 @@
...
+    overlays: "Overlays",
+    nodesSelected: (selectedNodesCount: number) => `${selectedNodesCount} nodes selected`,
+    edgesSelected: (selectedEdgesCount: number) => `${selectedEdgesCount} edges selected`,
+    nodeSelected: (nodeCount: number) => `${nodeCount} node`,
+    edgeSelected: (edgeCount: number) => `${edgeCount} edge`,
+    nodes: (nodeCount: number) => `${nodeCount} nodes`,
+    edges: (edgeCount: number) => `${edgeCount} edges`,
+    selected: "selected",
+    addingNodesMakingChanges: "Adding nodes or making changes to the Diagram will automatically create a DRD for you.",
+    removeDrd: "Remove DRD",
```
fantonangeli: Please review my suggestion or remove `addingNodesMakingChanges` if it's not used:
```suggestion
    addingNodesMakingChanges: "Adding nodes or making changes to the Diagram will automatically update the Serverless Workflow for you.",
```

packages/serverless-workflow-editor/src/store/computed/computeIndexes.ts
```diff
@@ -0,0 +1,290 @@
...
+    swfNodesByHref,
+  };
+}
+
+function buildSwfEdgesForNode(node: Unpacked<Specification.States>, index: number): SwfEdge[] {
+  const edges: SwfEdge[] = [];
+
+  //compensation
+  if (node.compensatedBy) {
+    edges.push(getCompensationTrasition(index, node));
```
fantonangeli:
```suggestion
    edges.push(getCompensationTransition(index, node));
```

packages/serverless-workflow-editor/src/store/computed/computeIndexes.ts
```diff
@@ -0,0 +1,290 @@
...
+  switch (node.type) {
+    case "sleep": {
+      const state = node as Specification.ISleepstate;
+
+      if (state.transition) {
+        edges.push(getTransition(index, state, state.transition));
+      }
+
+      if (state.onErrors) {
+        getErrorTrasitions(index, state, state.onErrors).forEach((errorTransitions) => edges.push(errorTransitions));
```
fantonangeli: There are also other parts of this file with the typo `Trasition -> Transition`
```suggestion
        getErrorTransitions(index, state, state.onErrors).forEach((errorTransitions) => edges.push(errorTransitions));
```

packages/serverless-workflow-editor/src/store/computed/computeIndexes.ts
```diff
@@ -0,0 +1,290 @@
...
+  index: number,
+  node: Unpacked<Specification.States>,
+  transition: string | Specification.ITransition
+): SwfEdge {
+  let transitionStr: string | undefined = undefined;
+  if (typeof transition === "object") {
+    transitionStr = transition.nextState;
+  } else {
+    transitionStr = transition;
+  }
```
fantonangeli: I see very similar parts of this file at lines: 186, 204, 222, 245, 258.
I would evaluate putting this logic in a reusable function.

packages/serverless-workflow-editor/src/store/computed/computeIndexes.ts
```diff
@@ -0,0 +1,290 @@
...
+      if (state.transition) {
+        edges.push(getTransition(index, state, state.transition));
+      }
+
+      if (state.onErrors) {
+        getErrorTrasitions(index, state, state.onErrors).forEach((errorTransitions) => edges.push(errorTransitions));
+      }
+
+      break;
+    }
```
fantonangeli: I think it's worth trying to aggregate these `swith cases`?

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+
+    const data: SwfDiagramNodeData = {
+      swfObject,
+      index,
+      parentRfNode: undefined,
+    };
+
+    // if there is no position calculated for the node in the state go for x:0 y:0
+    const i = nodeIds.indexOf(nodeName);
+    const position = nodeIds.indexOf(nodeName) === -1 ? { x: 0, y: 0 } : nodesPosition[i];
```
fantonangeli:
```suggestion
    const position = i === -1 ? { x: 0, y: 0 } : nodesPosition[i];
```

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+    nodeIds = [...nodeIds, nodeName];
+    nodesPosition = [...nodesPosition, position];
```
fantonangeli: We should avoid modifying the main function params and instead declare 2 new constants

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+    if (!targetAdjancyList) {
+      swfAdjacencyList.set(target, { dependencies: new Set([source]) });
+    } else {
+      targetAdjancyList.dependencies.add(source);
+    }
+
+    return edge;
+  };
+
+  // trasitions
```
fantonangeli: There are many other `trasition` in this file
```suggestion
  // transitions
```

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+        const i = nodeIds.indexOf(id);
+        if (i !== -1) {
+          nodesPosition.splice(1, i);
+          nodeIds.splice(1, i);
+        }
+      }
+    }
+  }
+
+  //clean up unused edges waypoits
```
fantonangeli:
```suggestion
  //clean up unused edges waypoints
```

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+    edgesById.set(edge.id, edge);
+    if (edge.selected) {
+      selectedEdgesById.set(edge.id, edge);
+    }
+
+    edges.push(edge);
+
+    swfEdges.push({ id, sourceId: source, targetId: target, swfObject });
+
+    const targetAdjancyList = swfAdjacencyList.get(target);
```
fantonangeli:
```suggestion
    const targetAdjacencyList = swfAdjacencyList.get(target);
```

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+  index: number,
+  node: Unpacked<Specification.States>,
+  transition: string | Specification.ITransition
+): EdgeArgs {
+  let transitionStr: string | undefined = undefined;
+  if (typeof transition === "object") {
+    transitionStr = transition.nextState;
+  } else {
+    transitionStr = transition;
+  }
```
fantonangeli, line 361: These conditions are replicated at lines 388, 406, 424, 447, 462, in this file as well, and can be unified together with the point [here](https://github.com/apache/incubator-kie-tools/pull/3252/changes#r2619758015)

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+      if (state.transition) {
+        edges.push(getTransitionArgs(index, state, state.transition));
+      }
+
+      if (state.onErrors) {
+        getErrorTrasitionsArgs(index, state, state.onErrors).forEach((errorargs) => edges.push(errorargs));
+      }
+
+      break;
+    }
```
fantonangeli, line 282: I would aggregate these `cases`

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+
+function getConditionTrasitionsArgs(
+  index: number,
+  node: Unpacked<Specification.States>,
+  conditions: Specification.Datacondition[] | Specification.Eventcondition[]
+): EdgeArgs[] {
+  const conditionTransitions: EdgeArgs[] = [];
+
+  conditions.forEach((condition) => {
+    // Specification.Enddeventcondition won't be created for now
```
fantonangeli: I wonder if we should create a GH Issue not to forget this.
You can skip this suggestion if not necessary.

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
```diff
@@ -0,0 +1,524 @@
...
+        transitionStr = transitionEventCondition.transition.nextState;
+      } else {
+        transitionStr = transitionEventCondition.transition;
+      }
+
+      conditionTransitions.push(
+        buildEdgeArgs(index, node, transitionStr, "edge_eventCondition", "eventConditionTransition")
+      );
+    } else {
+      // Specification.Enddatacondition won't be created for now
```
fantonangeli, line 424: I wonder if we should create a GH Issue not to forget this.
You can skip this suggestion if not necessary.

packages/serverless-workflow-editor/src/i18n/locales/en.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/src/i18n/locales/en.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/src/i18n/locales/en.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/src/i18n/locales/en.ts
handreyrc: @jomarko,

Actually, it comes from the original DMN file. I ported it from DMN and I ended up inheriting those issues.
https://github.com/apache/incubator-kie-tools/blob/main/packages/dmn-editor/src/i18n/locales/en.ts

I fixed all the issues you highlighted!

packages/serverless-workflow-editor/src/SwfEditor.tsx
```diff
@@ -0,0 +1,272 @@
...
+    }, ON_MODEL_CHANGE_DEBOUNCE_TIME_IN_MS);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [isDiagramEditingInProgress, onModelChange, swf.model]);
+
+  return (
+    <div ref={swfEditorRootElementRef} className={"kie-swf-editor--root"}>
+      <>
```
fantonangeli:
```suggestion
        <div
          className={"kie-tools--swf-editor--diagram-container"}
          ref={diagramContainerRef}
          data-testid={"kie-tools--swf-editor--diagram-container"}
        >
          <Diagram ref={diagramRef} container={diagramContainerRef} />
        </div>
```

handreyrc:
@jomarko,

I fixed all translation issues.
This PR is good for review again.

Thanks a lot for looking into it!

fantonangeli (changes requested):
@handreyrc thanks a lot for this huge milestone.
I reported some suggestions, but while I was doing the review you probably already fixed some of them.
Not in this PR, but I think in the future it would be beneficial for `dmn-editor` and `serverless-workflow-editor` to share a common package to "extend" as there are many parts which are similar. But I think you already evaluated this point with someone on the DMN side.

fantonangeli (commented):
I change my review from "request changes" to "comment" as I don't want to block the merge.

fantonangeli (approved):
Approving not to block this PR. Sorry for the pollution.

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: @fantonangeli,

It is fixed, I found a better place to do a cleanup on nodes and edges references.

packages/serverless-workflow-editor/src/i18n/locales/en.ts
handreyrc: It's not being used so I removed it along the property on top.

packages/serverless-workflow-editor/src/store/computed/computeIndexes.ts
handreyrc: I fixed the typo! Thanks

packages/serverless-workflow-editor/src/store/computed/computeIndexes.ts
handreyrc: I fixed the typo. Thanks

packages/serverless-workflow-editor/src/store/computed/computeIndexes.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
tiagobento: This is a strong indicator of a problem in state management. Very serious for non-readonly use-cases, and probably a source of many headaches in the future if not addressed, as there's a risk things will be build around this.

I'm okay to merge it with this known issue provided it is tackled soon.

Thanks!

tiagobento:
I agree with your comment @fantonangeli... In the new BPMN Editor I'm already experimenting with code reusability between the DMN and BPMN Editors. I don't want to create a deeply nested and coupled set of packages, though, so I'm taking it slowly.

If this reusability works well for DMN and BPMN we might consider expanding it to SWF as well!

packages/serverless-workflow-editor/src/store/computed/computeIndexes.ts
handreyrc: I aggregated the commom switch cases. Fixed!

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: This code was refactored. Fixed!

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: I searched in the code and fixed all.

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: Fixed!

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: It was mentioned that creating a start and end node may not be necessary.
There will be a next round to apply the UX.

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: It was mentioned that creating a start and end node may not be necessary.
There will be a next round to apply the UX.

packages/serverless-workflow-editor/src/SwfEditor.tsx
handreyrc: Fixed!

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
handreyrc: @tiagobento @fantonangeli,

I fixed the issue and refactored the code.

handreyrc:
@jomarko @fantonangeli ,

I addressed all the comments and fixed all the points raised so far.

Could you check it again, please?

jomarko:
Hello everyone, I am totally confused about one thing. So its currently only readonly mode implemented [1] or also the edit mode [2]? Could someone clarify the current state? It is crucial for a reviewer to know what scenarios should work. Until now, my understanding was only readonly mode should be tested. If I need to check also scenarios where model is somehow edited, manipulated I need more time for a review and a guidance what are such a scenarios.

[1]
https://github.com/apache/incubator-kie-tools/pull/3252#issuecomment-3627531617

[2]
https://github.com/apache/incubator-kie-tools/pull/3252#discussion_r2619511177

packages/serverless-workflow-editor/package.json
domhanak: Imho, It could certainly block upcoming release if merged. Is there any option to remove it or no?

packages/serverless-workflow-editor/src/store/computed/computeDiagramData.ts
fantonangeli: Can you please double check this is fixed?

fantonangeli:
> Hello everyone, I am totally confused about one thing. So its currently only readonly mode implemented [1] or also the edit mode [2]? Could someone clarify the current state? It is crucial for a reviewer to know what scenarios should work. Until now, my understanding was only readonly mode should be tested. If I need to check also scenarios where model is somehow edited, manipulated I need more time for a review and a guidance what are such a scenarios.
> 
> [1] [#3252 (comment)](https://github.com/apache/incubator-kie-tools/pull/3252#issuecomment-3627531617)
> 
> [2] [#3252 (comment)](https://github.com/apache/incubator-kie-tools/pull/3252#discussion_r2619511177)

@jomarko I think that ATM will only support read-only mode, just like it does currently.
My comment was referring to a particular case that the user should not reach, but it's nice to fix.

packages/serverless-workflow-editor/package.json
fantonangeli: It seems we are not the only ones having this [license issue](https://github.com/kieler/elkjs/issues?q=is%3Aissue%20license), but they have no plan to replace it.

I see this library it's used only in 4 points:
```
packages/dmn-editor/src/autolayout/autoLayoutInfo.ts|23| import ELK, * as Elk from "elkjs/lib/elk.bundled.js";
packages/dmn-editor/src/mutations/applyAutoLayoutToDrd.ts|23| import * as Elk from "elkjs/lib/elk.bundled.js";
packages/serverless-workflow-editor/src/autolayout/autoLayoutInfo.ts|20| import ELK, * as Elk from "elkjs/lib/elk.bundled.js";
packages/serverless-workflow-editor/src/mutations/applyAutoLayoutToSwf.ts|20| import * as Elk from "elkjs/lib/elk.bundled.js";
```
and we use it just for interfaces. Maybe we can create some custom interfaces in a separate PR and remove this dependency.

jomarko (approved):

packages/serverless-workflow-editor/package.json
thiagoelg: There are alternatives with compatible licenses, but I'm not sure about the effort to migrate to them...

- https://github.com/dagrejs/dagre (https://reactflow.dev/examples/layout/dagre)
- https://github.com/d3/d3-hierarchy
- https://github.com/iVis-at-Bilkent/cytoscape.js-fcose (https://ivis-at-bilkent.github.io/cytoscape.js-fcose/demo/demo-compound.html)

handreyrc:
@tiagobento,

If there are no objections we are good to merge this PR.

 Thanks!

domhanak (approved):

thiagoelg (approved):
Thanks @handreyrc for the monumental work on this PR, and thank you, everyone else, for the in-depth reviews!

Issue #3257, opened 2025-08-27
🐞 Bug report: ContextManager must only be set once caused by duplicated .cache.js loading


fanmeijian:
Description

When using @kie-tools/kie-editors-standalone (BPMN editor), the application crashes with:
`RuntimeException: The ContextManager must only be set once.`

This happens because both baseJsResources and referencedJsResources include the same .cache.js files.
As a result, the GWT runtime is initialized twice, causing duplicated registerGWTEditorProvider calls.

Steps to reproduce

1. Build the BPMN standalone editor using @kie-tools/kie-editors-standalone.
2. Integrate it into an Angular (or other SPA) project.
3. Run the app and open a BPMN editor.
4. Observe the error in the browser console:

`RuntimeException: The ContextManager must only be set once.`

n EditorResources construction:
```

baseJsResources: bpmnLanguageData.resources
  .filter((r) => r.type === "js")
  .flatMap((r) => r.paths)
  .map((p) => this.createResource({ path: p }, ["\\", "`", "$"])),
```

baseJsResources is currently including both .nocache.js and .cache.js files.

referencedJsResources also loads .cache.js files again.

This results in duplicate registration of GWT modules.

Workaround

Filter out .cache.js from baseJsResources:
```

.filter((r) => r.type === "js")
.flatMap((r) => r.paths)
.filter((p) => !p.includes(".cache.js"))   // prevent duplication
.map((p) => this.createResource({ path: p }, ["\\", "`", "$"]))
```

Expected behavior

baseJsResources should only include .nocache.js entrypoints.

.cache.js files should be loaded only via referencedJsResources.

Environment

@kie-tools/kie-editors-standalone version: [your version here]

Browser: [Chrome/Firefox...]

Framework: Angular 16 (but reproducible elsewhere)

tiagobento:
Closing as BPMN Editor (classic) was removed in 10.2.

PR #3258, opened 2025-08-27
kie-issues#3257: kie-editors-standalone: ContextManager must only be set once caused by duplicated .cache.js loading

fanmeijian:
resolve this issue https://github.com/apache/incubator-kie-tools/issues/3257


tiagobento:
BPMN Editor (classic) was removed and won't be present in 10.2.

PR #3267, opened 2025-09-08
kie-issues#2089: xml-parser-ts: idRandomizer should correctly take into account text elements

### LINKED ISSUE #2089: TS XML Parser: idRandomizer should correctly take into account text elements
The `idRandomizer` can normalize the XML adding missing ids and can re-generate ids and references without breaking the model. 
`idRandomizer` was first created for DMN taking into account xsd:IDREF and xsd:QName attributes. XSD can use xsd:IDREF and xsd:QName elements, requiring to update the idRandomizer, to take into account the __$$text value.

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/2089

### DIFF
--- a/packages/xml-parser-ts/src/idRandomizer.ts
+++ b/packages/xml-parser-ts/src/idRandomizer.ts
@@ -125,27 +125,35 @@ export class XmlParserTsIdRandomizer<M extends Meta> {
               attr
             )}: ${json})`
           );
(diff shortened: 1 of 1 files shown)


tiagobento (approved):
Thanks @ljmotta! This will be important for the upcoming BPMN Editor.

PR #3268, opened 2025-09-09
kie-issues#2090: DMN Editor: Does not render edge when required decision uses namespace in href

### LINKED ISSUE #2090: DMN Editor: Does not render edge when required decision uses namespace in href
DMN 1.6 supports referencing required decision [1] using two styles:

- with namespace in href `<requiredDecision href="http://www.example.org/Definitions01.xml#prebureauriskDec01"/>`
- without namespace in href `<requiredDecision href="#prebureauriskDec01"/>`

The problem is, when user starts to use namespace in href to reference node form the same file, the editor stop to display an edge between nodes.

### Steps to reproduce
- Start KIE Sandbox locally (`packages/online-editor @ apache/incubator-kie-tools`)
- Import this Gist: https://gist.github.com/tiagobento/368a169a2b7e2114c1184956e3d904b9
- Open both attached files and compere the displayed diagram

### Expected Diagram
Image

### Not Expected Diagram
Image

tiagobento:
Closes https://github.com/apache/incubator-kie-issues/issues/2090


packages/dmn-editor/src/mutations/deleteEdge.ts
```diff
@@ -44,7 +44,7 @@ export function deleteEdge({
   mode: EdgeDeletionMode;
   externalModelsByNamespace: ExternalModelsIndex | undefined;
 }) {
-  if (edge.dmnObject.namespace === definitions["@_namespace"]) {
+  if (!edge.dmnObject.normalizedNamespace) {
```
tiagobento, line 47: That's the fix for edge deletion. A normalized namespace equals `""` or `undefined` if it is the same as `thisDmn`'s namespace. We can only edit requirements on `thisDmn`.

packages/dmn-editor/tests-e2e/__fixtures__/editor.ts
```diff
@@ -62,4 +62,10 @@ export class Editor {
   public async changeTab(args: { tab: TabName }) {
     await this.page.getByRole("tab", { name: args.tab }).click();
   }
+
+  public async openDecisionWithNamespaceInHref() {
```
jomarko, line 66: maybe also this method name could reflect `NamespaceInRequiredDecisionHref` name, but is minor comment.

jomarko (approved):
I can confirm the fix works. I have used vsix extension generated by github action to verify manually. Thank you!

PR #3269, opened 2025-09-09
kie-tools#3215: [sonataflow] Move ubi8 jdk images to ubi9

### LINKED ISSUE #3215: [sonataflow] Move ubi8 jdk images to ubi9

wmedvede:
closes: #3215

### DIFF
--- a/packages/sonataflow-image-common/resources/modules/kogito-maven/common/configure
+++ b/packages/sonataflow-image-common/resources/modules/kogito-maven/common/configure
@@ -33,11 +33,6 @@ then
     mv /usr/share/apache-maven-"${MAVEN_VERSION}" ${MAVEN_HOME}
 else
     echo "---> Maven already present in the current context"
-    # TODO: Remove this temporary solution in favor of the -DnoMavenVersionCheck flag once it's available
-    # https://github.com/quarkusio/quarkus/issues/46105
-    # Installs Maven for version 3.9.6 in order to be compatible with the quarkus-maven-plugin minimum requirements (on Quarkus 3.15.3.1).
-    echo "---> Installing Maven 3.9.6"
-    ln -sf "${ALTERNATIVE_MAVEN_HOME}/bin/mvn" "${MAVEN_HOME}/bin/mvn"  
 fi
 
 mkdir "${KOGITO_HOME}"/.m2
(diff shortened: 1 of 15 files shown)


packages/sonataflow-image-common/resources/modules/kogito-maven/common/configure
```diff
@@ -33,11 +33,6 @@ then
     mv /usr/share/apache-maven-"${MAVEN_VERSION}" ${MAVEN_HOME}
 else
     echo "---> Maven already present in the current context"
-    # TODO: Remove this temporary solution in favor of the -DnoMavenVersionCheck flag once it's available
-    # https://github.com/quarkusio/quarkus/issues/46105
-    # Installs Maven for version 3.9.6 in order to be compatible with the quarkus-maven-plugin minimum requirements (on Quarkus 3.15.3.1).
-    echo "---> Installing Maven 3.9.6"
-    ln -sf "${ALTERNATIVE_MAVEN_HOME}/bin/mvn" "${MAVEN_HOME}/bin/mvn"
```
ricardozanini, line 40: There's a code block above unziping a Maven installation package. Mind taking a look? I guess we want to remove that too.

fantonangeli: @ricardozanini I think that `tar ...` command is needed by the command at line 33:
```
mv /usr/share/apache-maven-"${MAVEN_VERSION}" ${MAVEN_HOME}
```
I guess we can move the `tar` to that block.

ricardozanini: Nevertheless, this line should be removed from there.

wmedvede: @ricardozanini, ok, just to double check.

So we rely 100% on the mvn that is already installed in the ubi9 image, and we do no check nor installing here, right?

wmedvede: @ricardozanini I have removed the maven installation to just rely on the maven already present in the base image.

ricardozanini: @wmedvede it's ok to check if mvn is not there, but the untar should be in the branch where the `mvn` command check failed.

ricardozanini: But it's fine to remove it completely thou. The PR LGTM.

ricardozanini (approved):

fantonangeli (approved):

PR #3277, opened 2025-09-11
kie-issues#2096: DMN Editor: Does not load properties correctly sometimes

### LINKED ISSUE #2096: DMN Editor: Does not load properties correctly sometimes
There are scenarios where the DMN Editor doesn’t display the correct value in the properties panel.

In a few places in the DMN Editor’s property panels, we miss checking the value of the text input field, and it should be handled properly.

Adarshvk98:
Closes [kie-issue #2096](https://github.com/apache/incubator-kie-issues/issues/2096)

- Fixed text input undefined issue by replacing with empty string


jomarko (approved):
Thank you for this fixes @Adarshvk98 !

yesamer (approved):

PR #3290, opened 2025-09-22
DMN Editor: Copy of a Decision Service causes trouble to edges where target or source node is in the Decision Service

### LINKED ISSUE #2107: DMN Editor: Copy of a Decision Service causes trouble to edges where target or source node is in the Decision Service
Create a model with decision service and connect edges to it. Select all nodes and perform copy paste operation. Select the pasted set of nodes and move it to a different position. Unexpected edges are created in move operation.

Image

aswathitv:
Closes [kie#2107](https://github.com/apache/incubator-kie-issues/issues/2107)

Duplicate edge issue caused when copy pasting decision services resolved.

### DIFF
--- a/packages/dmn-editor/src/clipboard/Clipboard.ts
+++ b/packages/dmn-editor/src/clipboard/Clipboard.ts
@@ -165,7 +165,9 @@ export function buildClipboardFromDiagram(rfState: RF.ReactFlowState, dmnEditorS
             continue;
           }
 
-          accNode(decisionNode);
+          if (!selectedNodesById.has(decisionNode.id)) {
+            accNode(decisionNode);
+          }
         }
       }
 


tiagobento (commented):
Thanks for the PR! May I please ask you to explain the fix? It's not so obvious to understand how this simple change fixes the issue, I guess. Thanks! 🙏

aswathitv:
There are additional dmndi:DMNShape being added to dmn file for decisions due to the check for decision service. accNode is called for the decisions inside decision service as well by outer accNode call.

jomarko (changes requested):
Thank you for a PR, however I can not confirm the fix works. See the attached video.

https://github.com/user-attachments/assets/7a9ef6cc-6f43-40e0-a303-51252adf7c15

tiagobento:
@jomarko I think that issue on the video is a separate one. You're dragging the selection from a Decision contained by a Decision Service... and if I'm not wrong this won't work even without any copy/paste involved.

jomarko (approved):
@tiagobento ok, thank you for an explanation. I rechecked the scenario as dragging using different from copied set of nodes. Then I can confirm the issue is fixed.

@aswathitv sorry for an inappropriate first round of a review.

thiagoelg (approved):

PR #3291, opened 2025-09-23
kie-issues#2097: Unable to save the form changes in source fields - s…

jomarko:
…implified solution

This is part 2 for https://github.com/apache/incubator-kie-issues/issues/2097 and https://github.com/apache/incubator-kie-tools/pull/3284/files

It basically revert original solution and add `model.dispose()` call to another place, that was designed for it.


tiagobento (approved):

yesamer (approved):

jomarko:
PR updated, all `dispose` calls are needed to:
- switch between forms
- switch form **Source** and **Connections** tabs

danielzhe (approved):

thiagoelg (approved):

PR #3305, opened 2025-10-09
kie-issues#3294: i18n generic type is allowing properties that are not part of the type

### LINKED ISSUE #3294: Internationalization: i18n generic type is allowing properties that are not part of the type
The i18n generic type is allowing properties that are not part of the type.

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-tools/issues/3294,

Before fix:
When we add any unknown property in en.ts file , it is not throwing any error

After fix:
Adding unknown key will throw an error
Screenshot 2025-10-06 at 9 40 05 PM


mstuy:
@tiagobento please review or delegate as necessary.  Thanks.

packages/i18n/src/core/immutableDeepMerge.ts
```diff
@@ -19,25 +19,30 @@
...
     }
     if (typeof sourceValue === "string" || typeof sourceValue === "function") {
-      target[key] = sourceValue as string | DictionaryInterpolation;
+      target[key] = sourceValue as any;
     } else {
-      target[key] = deepMerge(createObjectCopy(target[key] as ReferenceDictionary), sourceValue as ReferenceDictionary);
+      target[key] = deepMerge(
+        createObjectCopy(target[key] as ReferenceDictionary<D>),
+        sourceValue as TranslatedDictionary<ReferenceDictionary<D>>
+      ) as any;
```
ljmotta: Could you please remove the cast to `any` if possible?

ljmotta (commented):
@Kusuma04-dev Thanks for the PR! I've made a comment inline regarding two `as any` cast.

packages/i18n/src/core/immutableDeepMerge.ts
Kusuma04-dev: Thanks for the review @ljmotta. If we need to replace any then it can be with `as unknown as ReferenceDictionary<D>[typeof key];`. Is this okay ?

ljmotta: @Kusuma04-dev Why do we need to cast it to `unknown`?

Kusuma04-dev: My bad. unknown is not needed Luiz, Made the change. Please have a look.

ljmotta (approved):
@Kusuma04-dev Thanks! LGTM

thiagoelg (approved):

Issue #3318, opened 2025-10-23
Ability to create multi platform images


RishiRajAnand:
Presently, container images only have the ability to build one architecture at a time, It does so by passing --arch flag while invoking kie-tools--image-builder script. Container images should support building for multiple architecture at same time. This can be achieved by making --arch flag to accept an array of archs instead of single architecture. The solution should not interfere with existing container packages functions.

PR #3319, opened 2025-10-23
kie-issues#3318: Enable kie-tools-image-builder to support multi platform builds

RishiRajAnand:
kie-issue: https://github.com/apache/incubator-kie-tools/issues/3318

The PR enables the kie-tools-image-builder to build multiple platform image. 
- The existing --arch flag is modified to accept array instead of single argument
- Compatible with existing container packages
- It retains the existing behavior for native builds, when no --arch is passed or the value of --arch  is "native", it triggers the native builds.
- Values can be passed to --arch flag like `kie-tools--image-builder build --arch="amd64,arm64"` or  --arch="amd64"

### DIFF
--- a/packages/image-builder/README.md
+++ b/packages/image-builder/README.md
@@ -18,7 +18,7 @@ Options:
   -f, --containerfile           Path to the Containerfile/Dockerfile  [string] [default: "Containerfile"]
   -c, --context                 Path to the build context  [string] [default: "./"]
       --build-arg               Build args for the builder in the format '<arg>=<value>', where <value> is a string (Can be used multiple times)  [array] [default: []]
-      --arch                    The target build architecture. If not provided will default to the native architecture  [string] [choices: "amd64", "arm64", "native"] [default: "native"]
+      --arch                    The target build architectures. If not provided will default to the native architecture  [array]
   -h, --help                    Show help  [boolean]
 
 Examples:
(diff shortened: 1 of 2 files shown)


tiagobento:
Why not change `arch` to be in the format `linux/amd64` instead of programmatically adding `linux/` to the configured values? That would be more flexible and make the code less complicated I think.

packages/image-builder/README.md
```diff
@@ -18,7 +18,7 @@ Options:
   -f, --containerfile           Path to the Containerfile/Dockerfile  [string] [default: "Containerfile"]
   -c, --context                 Path to the build context  [string] [default: "./"]
       --build-arg               Build args for the builder in the format '<arg>=<value>', where <value> is a string (Can be used multiple times)  [array] [default: []]
-      --arch                    The target build architecture. If not provided will default to the native architecture  [string] [choices: "amd64", "arm64", "native"] [default: "native"]
+      --arch                    The target build architectures. If not provided will default to the native architecture  [array]
```
tiagobento: Please add example values. Also, since this is an optional parameter, what is the default value?

RishiRajAnand:
> Why not change `arch` to be in the format `linux/amd64` instead of programmatically adding `linux/` to the configured values? That would be more flexible and make the code less complicated I think.

I added it that way so that it doesn't affect current behavior. I have now changed it to accept OS/ARCH format for better flexibility.

packages/image-builder/README.md
RishiRajAnand: I have added it as part of description and also default value

tiagobento:
@RishiRajAnand Thanks. You'll need to update all the places that are using this parameter too. Like
`buildImage({ ...args, arch: ["amd64"] }, imageFullNames);` and others.

RishiRajAnand:
> @RishiRajAnand Thanks. You'll need to update all the places that are using this parameter too. Like `buildImage({ ...args, arch: ["amd64"] }, imageFullNames);` and others.

okay, i added it

tiagobento (approved):

PR #3325, opened 2025-10-27
NO-ISSUE: Upgrade Playwright base image to Ubuntu 24.04

thiagoelg:
## This PR:
- Upgrades Playwright base image to Ubuntu 24.04;
- Replaces `wget` call with `curl` (don't know why, but sometimes wget would cause HTTP 429 errors from GitHub);
- Replace Node.js `.tar.xz` artifact with `.tar.gz` version;

## Motivation:
Running e2e tests locally on macOS with Rancher Desktop was causing segfaults inside the Playwright container, generating a QEMU dump. For some (unknown to me) reason, the `ubuntu:24.04` base image solves this.

Reference: https://github.com/rancher-sandbox/rancher-desktop/issues/2109


thiagoelg:
Build failures are probably related to the Ubuntu upgrade itself. Browsers might be rendering a bit differently, and screenshots will need to be updated.

tiagobento (approved):

ljmotta (approved):

PR #3327, opened 2025-10-28
kie-tools#3308: [cors-proxy] CORS code-scanning issues

fantonangeli:
Closes https://github.com/apache/incubator-kie-tools/issues/3308

**Review request:** @mareknovotny , I cannot add you in the reviewers

**Description:**
We need to fix the code-scanning issues here:
- sonataflow-dev-app (low impact): https://github.com/apache/incubator-kie-tools/security/code-scanning/1043
- runtime-tools-process-dev-ui-webapp (low impact): https://github.com/apache/incubator-kie-tools/security/code-scanning/1042
- cors-proxy: https://github.com/apache/incubator-kie-tools/security/code-scanning/1041

PR #3353, opened 2025-11-18
kie-tools#3349: [cors-proxy] Implement domains filtering

### LINKED ISSUE #3349: [cors-proxy] Implement domains filtering
Implement domains filtering to allow only a list of domain to be contacted.

fantonangeli:
Closes https://github.com/apache/incubator-kie-tools/issues/3349

**Important:** This PR starts from and require the PR below to be merged first:
- https://github.com/apache/incubator-kie-tools/pull/3327 

**Description:**
Implement domains filtering to allow only a list of domain to be contacted.

**How to test:**
- download [test.sh](https://github.com/user-attachments/files/23614831/test.sh)
- run the following commands:
```
cd packages/sonataflow-dev-app
CORS_DOMAIN=https://localhost:9020 SONATAFLOW_DEV_APP_DELAY=0 pnpm start
--------------
cd packages/cors-proxy
CORS_PROXY__allowedOrigins="https://localhost:9020" CORS_PROXY__allowedHosts="localhost,*.github.com,*.openshiftapps.com" pnpm start
--------------
sh test.sh
````
- only the first request will be rejected

### CODE REVIEW

**Review by ricardozanini — APPROVED** [review]

**Comment by tiagobento** [comment]
Thanks for the PR! @fantonangeli Can we avoid the new "MODE" parameter? It's not really something I'm a big fan of, as it often leads to things working in development and not in production. I understand why you created it, but can't we solve it in a different way? A list of allowed/blocked domains, or a "feature flag" like "validate X", or "check Y"... 

Thanks!

**Comment by fantonangeli** [comment]
@tiagobento applied the multiple origins here (https://github.com/apache/incubator-kie-tools/pull/3327#issuecomment-3586423282) and removed the dev mode, which I agree can lead to bugs.
Can you please re-review both PRs?

**Comment by thiagoelg** [comment]
I'm running the `online-editor` and trying to connect to an OpenShift cluster, but I'm getting 403 responses from the CORS Proxy, even after adding `CORS_PROXY__allowHosts=*` to the start command.

```
[0] Starting CORS proxy...
[0] ====================================================
[0] Allowed Origins:            http://localhost:9001
[0] Port:                       8080
[0] Verbose:                    false
[0] Hosts to proxy with HTTP:   localhost:8787
[0] Allow hosts:                *
[0] ====================================================
[0] CORS proxy listening at port 8080
```

In the logs I can see:
```
Couldn't handle request correctly due to:  Invalid URL
```

image

The same issue is happening when using the CORS Proxy service for git requests.

File: packages/cors-proxy/src/proxy/ExpressCorsProxy.ts [hunk 1]
```diff
@@ -140,7 +141,18 @@ export class ExpressCorsProxy implements CorsProxy<Request, Response> {
   }
 
   private resolveRequestInfo(request: Request): ProxyRequestInfo {
+    const origin = request.header("origin");
     const targetUrl: string = (request.headers[CorsProxyHeaderKeys.TARGET_URL] as string) ?? request.url;
+    const parsedTargetUrl = new URL(targetUrl);
```
thiagoelg: The issue seems to be here. Things like `/api.rm2.thpm.p1.openshiftapps.com:6443/version` are not valid as a parameter for `new URL()` and throw errors due to the leading `/`.

packages/cors-proxy/src/proxy/ExpressCorsProxy.ts
```diff
@@ -140,7 +141,18 @@ export class ExpressCorsProxy implements CorsProxy<Request, Response> {
...
+
+    if (!this.args.allowHosts.some((pattern) => minimatch(parsedTargetUrl.hostname, pattern))) {
+      throw new Error(`The target URL in not allowed. Requested: ${targetUrl}`);
```
thiagoelg:
```suggestion
      throw new Error(`The target URL is not allowed. Requested: ${targetUrl}`);
```

packages/cors-proxy/src/proxy/ExpressCorsProxy.ts
thiagoelg: But even after doing `const parsedTargetUrl = new URL(targetUrl.slice(1))` this is the resulting URL:
```
[0] { targetUrl: '/api.rm2.thpm.p1.openshiftapps.com:6443/version' }
[0] {
[0]   parsedTargetUrl: URL {
[0]     href: 'api.rm2.thpm.p1.openshiftapps.com:6443/version',
[0]     origin: 'null',
[0]     protocol: 'api.rm2.thpm.p1.openshiftapps.com:',
[0]     username: '',
[0]     password: '',
[0]     host: '',
[0]     hostname: '',
[0]     port: '',
[0]     pathname: '6443/version',
[0]     search: '',
[0]     searchParams: URLSearchParams {},
[0]     hash: ''
[0]   }
[0] }
```

Notice that the hostname is empty (probably because the URL doesn't have the protocol), and it will make the `allowHosts` check fail.

fantonangeli:
I updated the command in the description to run the cors-proxy:
```
CORS_PROXY__allowedOrigins="https://localhost:9020" CORS_PROXY__allowedHosts="localhost,*.github.com,*.openshiftapps.com" pnpm start
```

fantonangeli:
CodeQL still show the SSRF warning [here](https://github.com/apache/incubator-kie-tools/security/code-scanning?query=pr%3A3353+tool%3ACodeQL+is%3Aopen+language%3Atypescript), but in the tests we have evidence that the ev var `CORS_PROXY__allowedHosts` is used successfully to validate the target URL.
It seems CodeQL is not able to understand the validation which is based on `minimatch` or because he doesn't know that the user who declares `CORS_PROXY__allowedHosts` in the deployment is not the same person which send the `target-url` in the request header.
As a solution, I think we can simply dismiss the alert after this PR is merged.
Wdyt?

packages/cors-proxy/src/index.ts
```diff
@@ -42,12 +42,28 @@ function getAllowedOrigins(): string[] {
...
+function getAllowedHosts(): string[] {
+  const hosts = process.env.CORS_PROXY_ALLOWED_HOSTS || "localhost,*.github.com";
+  const hostsList = hosts.split(",").map((o) => o.trim());
+
+  if (hostsList.some((o) => o === "")) {
+    throw new Error("Invalid host: empty hosts are not allowed in CORS_PROXY_ALLOWED_HOSTS.");
+  }
+
+  if (hostsList.some((o) => o === "*")) {
+    throw new Error('Invalid host: wildcard "*" is not allowed in CORS_PROXY_ALLOWED_HOSTS.');
```
thiagoelg: I believe this is the exact code that fixes the SSRF warning, right?

My only issue with it is that KIE Sandbox should be able to connect to any OpenShift cluster, for example, and there's no way to know beforehand what the cluster host is.

In `online-editor`, I've changed the `start:cors-proxy` command to this:
```
cross-env CORS_PROXY__origin=https://localhost:$(build-env onlineEditor.dev.port) CORS_PROXY__useHttpForHosts=\"localhost:$(build-env kieSandboxAcceleratorQuarkus.dev.port)\" CORS_PROXY__allowedHosts=\"*.*,localhost\" npm --prefix ./node_modules/@kie-tools/cors-proxy run start
```
Notice the `*.*`, which basically allows any domain with a `.` in it. It kind of defeats the purpose of this check, but at the same time, it's something we need for KIE Sandbox.

Anyway, I'm fine with it, and the `*.*,localhost` solution works for me. But WDYT?

thiagoelg: Also, would you mind updating the `start:cors-proxy` script in the `online-editor` package with the above? Thanks!

fantonangeli: I didn't think about the `*.*` case, thank you for bringing this up!
The issue with it is that a bad user can very easily use the cors-proxy to attack a server anonymously, as the user decides the `target-url`. So in this case, CodeQL is right.

> My only issue with it is that KIE Sandbox should be able to connect to any OpenShift cluster, for example, and there's no way to know beforehand what the cluster host is.

This is an issue in securing this proxy, especially if it is a public service. Can you ensure with others if there can be a way to define the `allowedHosts` whitelist?

fantonangeli:
Did a bit more investigation on the SSRF alert and tried the following code:
```
      const href=info.proxyUrl.href

      // still show js/request-forgery error
      if (!/^https?:\/\/example\.com\/.*$/.test(href)) {
        throw new Error(`The target URL is not allowed. Requested: ${info.proxyUrl.href}`);
      }

      // fixes js/request-forgery
      if (!/^http:\/\/example\.com\/query$/.test(href)) {
        throw new Error(`The target URL is not allowed. Requested: ${info.proxyUrl.href}`);
      }

      // still show js/request-forgery error
      if (!href.startsWith("https://example.com/")) {
        throw new Error(`The target URL is not allowed. Requested: ${info.proxyUrl.href}`);
      }

      const proxyResponse = await fetch(href, {
        method: req.method,
        headers: outHeaders,
        redirect: "manual",
        body: req.method !== "GET" && req.method !== "HEAD" ? req : undefined,
        agent: this.getProxyAgent(info),
      });
```

With this simplified code I can see that CodeQL will not be happy until there is a check on the **full** URL and not just the domain. But this is not possible for us.
So with the domain checks in this PR and dismissing the alert it's enough, IMO.

packages/cors-proxy/src/index.ts
fantonangeli: Because `*.*` is the same as `*` I would allow the common `*`, and like in my [comment](https://github.com/apache/incubator-kie-tools/pull/3353#issuecomment-3617652820), it seems we cannot make CodeQL 100% happy, so I just added a `console.warn` to discourage the * in production.

packages/online-editor/package.json
```diff
@@ -18,8 +18,8 @@
     "build:dev": "rimraf dist && webpack --config webpack.config.ts --env dev",
     "build:prod": "pnpm lint && pnpm test && rimraf dist && webpack && pnpm test-e2e",
     "lint": "run-script-if --bool \"$(build-env linters.run)\" --then \"kie-tools--eslint ./src\"",
-    "start": "concurrently 'pnpm start:cors-proxy' 'pnpm start:extended-services' 'pnpm start:kie-sandbox' 'pnpm start:kie-sandbox-accelerator-quarkus'",
-    "start:cors-proxy": "cross-env CORS_PROXY__origin=https://localhost:$(build-env onlineEditor.dev.port) CORS_PROXY__useHttpForHosts=\"localhost:$(build-env kieSandboxAcceleratorQuarkus.dev.port)\" npm --prefix ./node_modules/@kie-tools/cors-proxy run start",
+    "start": "concurrently --no-colors 'pnpm start:cors-proxy' 'pnpm start:extended-services' 'pnpm start:kie-sandbox' 'pnpm start:kie-sandbox-accelerator-quarkus'",
```
fantonangeli, line 21: `--no-colors ` is to fix the issue receiving this value in the code:
```
this.args.allowedOrigins=[ 'http://localhost:\x1B[33m9001\x1B[39m' ]
```

thiagoelg:
> With this simplified code I can see that CodeQL will not be happy until there is a check on the full URL and not just the domain. But this is not possible for us.
So with the domain checks in this PR and dismissing the alert it's enough, IMO.

Sounds good to me, thanks for looking into this @fantonangeli!

thiagoelg (approved):

thiagoelg:
I haven't merged this PR yet as it has the `DO NOT MERGE` label.
Let me know when it's ready, thanks!

fantonangeli:
> I haven't merged this PR yet as it has the `DO NOT MERGE` label. Let me know when it's ready, thanks!

Ooops, removed it now

PR #3367, opened 2025-12-02
kie-tools#3316: [sonataflow-deployment-webapp] generated inputs are sent only if are touched

### LINKED ISSUE #3316: [sonataflow-deployment-webapp] generated inputs are sent only if are touched
In `sonataflow-deployment-webapp` and the other SonataFlow apps, using the json schema below, we saw that, despite the default value and the field value, the custom `boolField` is sent to the WF endpoint only if the user modify the input:
```
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "service_input",
    "required": ["boolField"],
    "type": "object",
    "properties": {
      "boolField": {
        "type": "boolean",
        "defaultValue": false
      }
    }
}
```

if the user modify `boolField` the form send:
```
{"workflowdata":{"boolField":true}}
```
if the user leaves the default value without touching the form, the WF endpoint receives this data:
```
{"workflowdata":{}}
```

kumaradityaraj:
---
Closes https://github.com/apache/incubator-kie-tools/issues/3316
---

# 🚀  Generated inputs are sent only if are touched

### Issue Summary
In **sonataflow-deployment-webapp** and other SonataFlow applications, when using a JSON schema that includes a boolean property with a default value, the generated form does not send the field to the workflow endpoint unless the user explicitly modifies it.

---

### Fix: By default the boolean value is false and if the checkbox is checked the value is true.

---

## 🧪 How to Test
Follow these steps to build and run the webapp locally:
```bash
# From the project root
cd incubator-kie-tools/
pnpm -r -F sonataflow-deployment-webapp... build:dev
cd packages/sonataflow-deployment-webapp
pnpm start:dev-webapp
Navigate to - http://localhost:9023/#/workflows/service


#From [[sonataflow-deployment-webapp-consumer](https://github.com/fantonangeli/sonataflow-deployment-webapp-consumer)]
mvn clean package quarkus:dev
```

```bash
Change the schema as follows  or you can use your own schema:-
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "type": "object",
    "properties": {
      "boolField": {
        "type": "boolean",
        "defaultValue": true
      }
    },
    "required": [
        "boolField"
    ]
}
```
---
## Preview

https://github.com/user-attachments/assets/4890e2ad-9e43-43d5-af48-71211140d5a0

---


kumaradityaraj:
@fantonangeli  Please review this PR. Thank you.

packages/runtime-tools-components/src/components/FormRenderer/FormRenderer.tsx
```diff
@@ -51,14 +51,26 @@ export const FormRenderer = React.forwardRef<FormRendererApi, IOwnProps & OUIAPr
...
       const newModel = ModelConversionTool.convertDateToString(formModel, formSchema);
-      return validator.validate(newModel);
+      const booleanSafeModel = normalizeBooleans(newModel, formSchema);
+      return validator.validate(booleanSafeModel);
     });
 
-    // Converting Dates that are in string format into JS Dates so they can be correctly bound to the uniforms DateField
-    const formData = ModelConversionTool.convertStringToDate(model, formSchema);
+    // Converting Dates that are in string format into JS Dates so they can be correctly bound to the uniforms DateField and doing the same for boolField
+    const formData = normalizeBooleans(ModelConversionTool.convertDateToString(model, formSchema), formSchema);
```
fantonangeli: Changing from `convertStringToDate` to `convertDateToString` can create bugs.
I would suggest to extend the functionalities of `ModelConversionTool` (packages/runtime-tools-components/src/utils/ModelConversionTool.ts) to have everything in the same place.

packages/runtime-tools-components/src/components/FormRenderer/FormRenderer.tsx
fantonangeli, line 1: The correct test-schema for the field is:
```
    "boolField": {
      "type": "boolean",
      "defaultValue": true
    }
```
This schema should generate a checked boolean, but in this branch the `defaultValue` is ignored.

packages/uniforms-patternfly/tests/BoolField.test.tsx
```diff
@@ -77,6 +77,17 @@ test("<BoolField> - renders a input with correct value (default)", () => {
...
 
+test("<BoolField> - renders an input with correct value (default true)", () => {
+  render(
+    usingUniformsContext(<BoolField name="x" />, {
+      x: { type: Boolean, defaultValue: true },
+    })
+  );
+
+  expect(screen.getByTestId("bool-field")).toBeInTheDocument();
+  expect(screen.getByTestId("bool-field")).toHaveAttribute("checked");
```
fantonangeli: Can we test that the expected value is submitted when click on submit?

packages/uniforms-patternfly/tests/BoolField.test.tsx
fantonangeli, line 1: Because now we have the conversion logic, can you please add a test to verify it works well.
You can have a look at this code which tests the conversion string->date:
https://github.com/apache/incubator-kie-tools/blob/6a0ee07f97acd91f176b86bfe3cd56366bc69a67/packages/uniforms-patternfly/tests/DateField.test.tsx#L97

fantonangeli (changes requested):

packages/runtime-tools-components/src/utils/ModelConversionTool.ts
```diff
@@ -96,6 +119,13 @@ function convertModel(model: any, schema: Record<string, any>, ctx: ConversionCo
       return;
     }
 
+    if (value === undefined || value === null) {
```
fantonangeli: This code will not be executed as we have this before:
```
    if (value === null) {
      return;
    }
```

packages/runtime-tools-components/src/utils/ModelConversionTool.ts
```diff
@@ -96,6 +119,13 @@ function convertModel(model: any, schema: Record<string, any>, ctx: ConversionCo
...
+      if (propSchema.type === "boolean") {
+        obj[propertyName] = !!value;
```
fantonangeli: because of the:
```
if (value === undefined || value === null) {...
```
value will never be `true`. What did you mean to write?

packages/runtime-tools-components/src/utils/ModelConversionTool.ts
```diff
@@ -96,6 +119,13 @@ function convertModel(model: any, schema: Record<string, any>, ctx: ConversionCo
       return;
     }
 
+    if (value === undefined || value === null) {
+      if (propSchema.type === "boolean") {
```
fantonangeli: why the need of nested ifs?

packages/runtime-tools-components/src/components/FormRenderer/FormRenderer.tsx
```diff
@@ -58,7 +58,10 @@ export const FormRenderer = React.forwardRef<FormRendererApi, IOwnProps & OUIAPr
     });
 
     // Converting Dates that are in string format into JS Dates so they can be correctly bound to the uniforms DateField
-    const formData = ModelConversionTool.convertStringToDate(model, formSchema);
+    const formData = ModelConversionTool.normalizeBooleans(
```
fantonangeli: 3 nested calls? Let's avoid too much complexity

packages/runtime-tools-components/src/utils/ModelConversionTool.ts
```diff
@@ -25,6 +25,28 @@ export class ModelConversionTool {
   public static convertStringToDate = (model: any, schema: Record<string, any>): any => {
     return doConvert(model, schema, (value) => new Date(value));
   };
+
+  public static normalizeBooleans = (model: any, schema: Record<string, any>): any => {
+    return doConvert(model, schema, (value) => {
+      if (value === "boolean") {
```
fantonangeli: Do we ever reach this code?
What did you mean with `value === "boolean"`?

packages/runtime-tools-components/src/utils/ModelConversionTool.ts
```diff
@@ -25,6 +25,28 @@ export class ModelConversionTool {
...
+      return value;
+    });
+  };
+
+  public static applySchemaDefaults(model: any, schema: Record<string, any>): any {
+    const newModel = { ...model };
+    if (schema.properties) {
+      Object.keys(schema.properties).forEach((key) => {
+        const prop = schema.properties[key];
+        if (newModel[key] === undefined && prop.default !== undefined) {
```
fantonangeli: I think this function is in the right direction.
For booleans we should add a value even if there is no default value in the schema

packages/uniforms-patternfly/tests/BoolField.test.tsx
```diff
@@ -109,3 +120,29 @@ test("<BoolField> - renders a wrapper with unknown props", () => {
   expect(screen.getByTestId("wrapper-field").getAttribute("data-y")).toBe("y");
   expect(screen.getByTestId("wrapper-field").getAttribute("data-z")).toBe("z");
 });
+
+test("<BoolField> - respects default false from model", () => {
+  render(usingUniformsContext(<BoolField name="x" />, { x: { type: Boolean } }, { model: { x: false } }));
+
+  const input = screen.getByTestId("bool-field");
+  expect(input).not.toBeChecked();
+});
```
fantonangeli: It seems to be already checked on line 91.
We already have a test which checks if BoolField is unchecked without a model plus a test to check if BoolField is checked when the model has value=true
```suggestion
```

packages/uniforms-patternfly/tests/BoolField.test.tsx
```diff
@@ -109,3 +120,29 @@ test("<BoolField> - renders a wrapper with unknown props", () => {
...
+
+test("<BoolField> - respects default true from model", () => {
+  render(usingUniformsContext(<BoolField name="x" />, { x: { type: Boolean } }, { model: { x: true } }));
+
+  const input = screen.getByTestId("bool-field");
+  expect(input).toBeChecked();
+});
```
fantonangeli: It seems to be already checked on line 91
```suggestion
```

packages/uniforms-patternfly/tests/BoolField.test.tsx
```diff
@@ -109,3 +120,29 @@ test("<BoolField> - renders a wrapper with unknown props", () => {
...
+
+test("<BoolField> - respects schema default true", () => {
+  render(usingUniformsContext(<BoolField name="x" />, { x: { type: Boolean, defaultValue: true } }));
+  const input = screen.getByTestId("bool-field");
+  expect(input).toBeChecked();
+});
```
fantonangeli: It seems to be already checked on line 80
```suggestion
```

packages/uniforms-patternfly/tests/BoolField.test.tsx
```diff
@@ -109,3 +120,29 @@ test("<BoolField> - renders a wrapper with unknown props", () => {
...
+
+test("<BoolField> - respects schema default false", () => {
+  render(usingUniformsContext(<BoolField name="x" />, { x: { type: Boolean, defaultValue: false } }));
+  const input = screen.getByTestId("bool-field");
+  expect(input).not.toBeChecked();
+});
```
fantonangeli: Repetition of other tests
```suggestion
```

packages/uniforms-patternfly/tests/BoolField.test.tsx
```diff
@@ -77,6 +77,17 @@ test("<BoolField> - renders a input with correct value (default)", () => {
   expect(screen.getByTestId("bool-field")).not.toHaveAttribute("checked");
 });
 
+test("<BoolField> - renders an input with correct value (default true)", () => {
+  render(
+    usingUniformsContext(<BoolField name="x" />, {
+      x: { type: Boolean, defaultValue: true },
+    })
+  );
```
fantonangeli: In the real use-case `defaultValue` in the schema is translated to `default` by the backend.
So the frontend receive `default` in the `openapi.json` and what we need to test is:
```
  render(usingUniformsContext(<BoolField name="x" />, { x: { type: Boolean } }, { model: { x: true } }));
```
Which is already tested in line 92

fantonangeli (changes requested):
Testing this PR with a schema without the default (see below) value, still show the error from the WF endpoint:
```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "required": ["boolField"],
  "properties": {
    "boolField": {
      "type": "boolean"
    }
  }
}
```
Please test both cases with and without `defaulValue` plus checking and not the checkbox.

packages/runtime-tools-components/src/utils/ModelConversionTool.ts
```diff
@@ -84,6 +104,7 @@ function convertModel(model: any, schema: Record<string, any>, ctx: ConversionCo
 
   Object.keys(model).forEach((propertyName) => {
     const property = schema.properties[propertyName];
+    const propSchema = lookupSchemaPropertyProps(schema.properties[propertyName], ctx);
```
fantonangeli: This seems not to be used

packages/runtime-tools-components/src/utils/ModelConversionTool.ts
```diff
@@ -25,6 +25,26 @@ export class ModelConversionTool {
...
+        if (newModel[key] !== undefined) {
+          return;
+        }
+        if (prop.default !== undefined) {
+          newModel[key] = prop.default;
+          return;
+        }
+        if (prop.type === "boolean") {
+          newModel[key] = false;
+        }
```
fantonangeli:
```suggestion
        if (prop.type === "boolean" && newModel[key] === undefined) {
          newModel[key] = prop.default ?? false;
        }
```

fantonangeli (changes requested):
Some more suggestions

packages/uniforms-patternfly/tests/BoolField.test.tsx
fantonangeli, line 1: @kumaradityaraj  We only need to clean this file,  then this PR is ready

packages/uniforms-patternfly/tests/BoolField.test.tsx
fantonangeli: If this test is similar to line 92, can you please remove it

fantonangeli (approved):
Thanks a lot @kumaradityaraj

ricardozanini (approved):

fantonangeli:
Hi @kumaradityaraj, because the last commit here is from 9th December, can you please sync this branch, so the macos CIs will also run?

fantonangeli:
@kumaradityaraj can you please sync this branch?

domhanak:
@kumaradityaraj @fantonangeli Could we please add a test for this?

I am unable to build the environment and validate the error using the `How to test` steps, so will try again in fresh workspace tomorrow.

fantonangeli:
I don't know what changed @domhanak , I tried a lot with some examples from kogito-examples, but the properties below seem totally ignored:
```
quarkus.http.port=4000
kogito.data-index.url=http://localhost:4000
```
The SonataFlow Dev App requires the DI to be on port 4000. The real app will use the same port where it is served as it contacts the graphql with a relative link.
If you want to test with the simple mocked data, which I created, you can just run: `pnpm start` and nothing more.

fantonangeli:
@domhanak  I created a patch to run the Greeting Example with OpenAPI and GraphQL on 4000 port, but there are bugs with the UI, not related to this PR.
I guess there have been changes which I was not aware of. 

Anyway, I created a Quarkus app, containing the `BoolField` WF used to test this PR. Please note it is based on versions `10.1.0`:
https://github.com/fantonangeli/sonataflow-testing-deployment

fantonangeli:
@domhanak I just realized last night I forgot to attach the patch and now I recreated it
[swf-greeting-examples-on-4000.patch](https://github.com/user-attachments/files/25230777/swf-greeting-examples-on-4000.patch)

domhanak (approved):

PR #3382, opened 2025-12-19
kie-issues#2197: Allow customizing Playwright test runs in CI without code changes

### LINKED ISSUE #2197: Allow customizing Playwright test runs in CI without code changes
Currently, it’s not possible to customize the Playwright test run in CI without modifying the source code. This creates friction because:

Running tests on the host machine can require different configuration values compared to CI.
Any change to adapt CI settings (timeouts, retries, workers, etc.) requires a code update and commit, which is not ideal for dynamic environments.

## Proposed Solution
Introduce a mechanism to override Playwright configuration via environment variables (e.g., PLAYWRIGHT_BASE__workers, PLAYWRIGHT_BASE__retries)

This would allow CI pipelines to inject custom values dynamically. Local runs to remain unaffected.

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/2197

This PR adds the option to change the Playwright configuration through environment variables.

```
# Enable running tests on Apple Silicon Google Chrome. Overrides PLAYWRIGHT_BASE__enableGoogleChromeProject.
PLAYWRIGHT_BASE__enableGoogleChromeTestsForAppleSilicon=true

# Enable/disable Chromium tests
PLAYWRIGHT_BASE__enableChromiumProject=true

# Enable/disable Google Chrome tests
PLAYWRIGHT_BASE__enableGoogleChromeProject=true

# Enable/disable Webkit tests
PLAYWRIGHT_BASE__enableWebkitProject=true

# Set project timeout
PLAYWRIGHT_BASE__projectTimeout=90000

# Set "expect" statement timeout
PLAYWRIGHT_BASE__expectTimeout=30000

# Set the maxDiffPixelRatio for the test assertion.
PLAYWRIGHT_BASE__maxDiffPixelRatio=0.001

# Set the number of retries for each test.
PLAYWRIGHT_BASE__retries=2

# Set the number of workers used when executing the test suite. Increasing this number can increase the tests flakyness.
PLAYWRIGHT_BASE__workers=6
```

This PR also:
- Creates a new CLI (`pw-e2e-container`) to handle Playwright specific containerization configurations instead of bloating the `package.json` with more commands.
- Remove 0.1 maxDiffPixel on `boxed-expression-component` and `scesim-editor`.
- Uses `start-server-and-test` on all packages.
- Improvements on `online-editor` test logs.

Now, it is possible to enable/disable suite for a specific project without using the `shell`. Example for run only webkit tests with more workers:
```sh
KIE_TOOLS_BUILD__runEndToEndTests=true PLAYWRIGHT_BASE__enableWebkitProject=true PLAYWRIGHT_BASE__enableGoogleChromeProject=false PLAYWRIGHT_BASE__enableChromiumProject=false PLAYWRIGHT_BASE__workers=6 pnpm test-e2e
```


packages/playwright-base/README.md
```diff
@@ -19,35 +19,15 @@
...
 
 ## Using containers to generate screenshots
 
 Each operating system has slight variations in UI, even within the same browser. These differences can cause screenshot comparison tests to fail. To address this issue and ensure a stable environment with consistent test results locally and in CI, containers can be used. Running Playwright tests inside a container that is also used in the CI environment makes screenshot tests reproducible, regardless of the host OS.
 
 > **ℹ️ NOTE**
 >
-> Due to compatibility issues, this containerization solution cannot yet be used on native Windows and requires running it directly within WSL (Windows Subsystem for Linux). Also, Linux arm64 doesn't support Google Chrome, and due to this caveat, some tests will be disabled for this arch.
+> Due to compatibility issues, this containerization solution cannot yet be used on native Windows and requires running it directly within WSL (Windows Subsystem for Linux). Also, Linux arm64 doesn't support Google Chrome, and due to this caveat, Google Chrome tests in macOS arm are disabled in some packages.
```
thiagoelg: > Also, Linux arm64 doesn't support Google Chrome, and due to this caveat, Google Chrome tests in macOS arm are disabled in some packages.

This is a bit confusing... Can you replace it with something like:

> Also, Google Chrome is not available for Linux on ARM64, so any Google Chrome tests running under a Linux ARM64 container will be disabled. This is usually the case when running Linux containers on Apple Silicon machines, so Google Chrome tests are disabled in some packages under macOS ARM64.

packages/playwright-base/README.md
```diff
@@ -79,6 +59,76 @@ pnpm test-e2e:container:shell
...
+  containerized-playwright shell --container-name my_playwright_container      Start the container (if needed) and open an interactive bash shell in
+  --container-workdir incubator-kie-tools/packages/my-package                  the package workdir.
+
+  containerized-playwright shell --container-name e2e --container-workdir      Open an interactive shell with extra environment variables forwarded to
+  incubator-kie-tools/packages/foo --additional-env DEBUG=true                 docker-compose.
+
+  containerized-playwright shell --container-name e2e --container-workdir      Forward multiple env variables in a single --additional-env option
+  incubator-kie-tools/packages/foo --additional-env FOO=bar,BAZ=qux            using comma-separated pairs.
+
+  containerized-playwright clean                                               Stop and remove Playwright-related containers using the base
```
thiagoelg: `containerized-playwright` should be `pw-e2e-container`, no?

packages/playwright-base/README.md
```diff
@@ -79,6 +59,76 @@ pnpm test-e2e:container:shell
...
 
+## Playwright-base CLI
+
+A CLI tool to assist running containerized Playwright tests. It loads environment variables and passes them through the docker-compose file, and helps with starting the test suite, opening an interactive shell inside the container, and cleaning up.
+
+```sh
+pnpm pw-e2e-container --help
+pw-e2e-container <command>
+
+Commands:
```
thiagoelg, line 70: It would be nice to also have a list of possible options/flags here (check the image-builder README).

packages/playwright-base/README.md
```diff
@@ -79,6 +59,76 @@ pnpm test-e2e:container:shell
...
+  containerized-playwright shell --container-name e2e --container-workdir      Forward multiple env variables in a single --additional-env option
+  incubator-kie-tools/packages/foo --additional-env FOO=bar,BAZ=qux            using comma-separated pairs.
+
+  containerized-playwright clean                                               Stop and remove Playwright-related containers using the base
+                                                                               docker-compose file.
+```
+
+## Installing Playwright dependencies on host machine
+
+> **i Warning**
```
thiagoelg:
```suggestion
> **ℹ️ Warning**
```

packages/playwright-base/README.md
```diff
@@ -79,6 +59,76 @@ pnpm test-e2e:container:shell
...
+```
+
+or
+
+```sh
+# in the `kie-tools` root
+PLAYWRIGHT_BASE__installDeps=true pnpm bootstrap -F playwright-base
+```
+
+> **i NOTE**
```
thiagoelg:
```suggestion
> **ℹ️ NOTE**
```

packages/playwright-base/env/index.js
```diff
@@ -25,11 +25,58 @@ module.exports = composeEnv([require("@kie-tools/root-env/env")], {
...
+      default: "90000",
+      description: "The timeout of a test in the project (Chromium, Google Chrome and Webkit)",
+    },
+    PLAYWRIGHT_BASE__expectTimeout: {
+      default: "30000",
+      description: "The timeout of an expect assertion",
+    },
+    PLAYWRIGHT_BASE__maxDiffPixelRatio: {
+      default: "0.001",
+      description: "The threshold of the maximum difference on screenshot comparisons (in percentage)",
```
thiagoelg: Is `0.001` = 0.001% or 0.1%? If it's 0.1% I would use `(percentage in decimal form)` just to be a bit clearer.

```suggestion
      default: "0.001",
      description: "The threshold of the maximum pixel difference on screenshot comparisons (percentage in decimal form)",
```

packages/playwright-base/README.md
ljmotta: I changed the cli name, and I forgot to update the readme :face_with_peeking_eye:

packages/playwright-base/env/index.js
```diff
@@ -25,11 +25,58 @@ module.exports = composeEnv([require("@kie-tools/root-env/env")], {
       default: "false",
       description: "Toggles the installation of Playwright dependencies. Can be `true` or `false`.",
     },
+    PLAYWRIGHT_BASE__enableGoogleChromeTestsForAppleSilicon: {
+      default: "true",
+      description: "Enable Google Chrome tests for ARM OSs. Overrides PLAYWRIGHT_BASE__enableGoogleChromeProject.",
+    },
```
thiagoelg:
```suggestion
    PLAYWRIGHT_BASE__enableGoogleChromeTestsOnARM: {
      default: "true",
      description: "Enable or disable Google Chrome tests on ARM OSes. Overrides PLAYWRIGHT_BASE__enableGoogleChromeProject.",
    },
```

packages/playwright-base/package.json
```diff
@@ -12,20 +12,33 @@
...
+    "src"
   ],
   "scripts": {
     "build:dev": "run-script-if --bool \"$(build-env containerImages.build)\" --then \"pnpm image:docker:build\"",
     "build:prod": "pnpm build:dev",
     "image:docker:build": "docker build --platform=linux/amd64 -t dev.local/apache-kie-tools/playwright-e2e-tests-image:$(build-env root.streamName) . -f Containerfile --output type=docker",
-    "install": "run-script-if --bool \"$(build-env playwrightBase.installDeps)\" --then \"pnpm playwright install --with-deps\""
+    "install": "pnpm install:cli && pnpm install:image",
+    "install:cli": "rimraf dist && tsc",
+    "install:image": "run-script-if --bool \"$(build-env playwrightBase.installDeps)\" --then \"pnpm playwright install --with-deps\""
```
thiagoelg: Why `install:image`? This is installing playwright dependencies only.

packages/playwright-base/src/bin.ts
```diff
@@ -0,0 +1,319 @@
...
+    PLAYWRIGHT_BASE__enableChromiumProject: buildEnv.playwrightBase.enableChromiumProject,
+    PLAYWRIGHT_BASE__enableGoogleChromeProject: buildEnv.playwrightBase.enableGoogleChromeProject,
+    PLAYWRIGHT_BASE__enableWebkitProject: buildEnv.playwrightBase.enableWebkitProject,
+    PLAYWRIGHT_BASE__projectTimeout: buildEnv.playwrightBase.projectTimeout,
+    PLAYWRIGHT_BASE__expectTimeout: buildEnv.playwrightBase.expectTimeout,
+    PLAYWRIGHT_BASE__maxDiffPixelRatio: buildEnv.playwrightBase.maxDiffPixelRatio,
+    PLAYWRIGHT_BASE__retries: buildEnv.playwrightBase.retries,
+    PLAYWRIGHT_BASE__workers: buildEnv.playwrightBase.workers,
+  };
+}
```
thiagoelg, line 47: If you run `pnpm build-env --print-dotenv` or `--print-dotenv:self` you'll get basically this, but this is fine.

packages/playwright-base/env/index.js
thiagoelg: The issue with Google Chrome on ARM also occurs on Linux ARM64 machines, so I think this env var should be more generic in this sense.

packages/playwright-base/src/bin.ts
```diff
@@ -0,0 +1,319 @@
...
+  return merged;
+}
+
+function prettyPrintError(error: unknown) {
+  console.error(`\x1b[31m[${CLI_NAME}] %s\x1b[0m`, error instanceof Error ? error.message : String(error));
+}
+
+function isAppleSilicon(): boolean {
+  return process.platform === "darwin" && process.arch === "arm64";
+}
```
thiagoelg: Following up on the comments above, using `process.arch === "arm64"` should be enough, since the Google Chrome issue happens inside Linux ARM64 containers, no matter the host OS.

packages/playwright-base/src/bin.ts
```diff
@@ -0,0 +1,319 @@
...
+PLAYWRIGHT_BASE__enableGoogleChromeTestsForAppleSilicon=${buildEnv.playwrightBase.enableGoogleChromeTestsForAppleSilicon}
+PLAYWRIGHT_BASE__enableChromiumProject=${buildEnv.playwrightBase.enableChromiumProject}
+PLAYWRIGHT_BASE__enableGoogleChromeProject=${isGoogleChromeEnabled}
+PLAYWRIGHT_BASE__enableWebkitProject=${buildEnv.playwrightBase.enableWebkitProject}
+PLAYWRIGHT_BASE__projectTimeout=${buildEnv.playwrightBase.projectTimeout}
+PLAYWRIGHT_BASE__expectTimeout=${buildEnv.playwrightBase.expectTimeout}
+PLAYWRIGHT_BASE__maxDiffPixelRatio=${buildEnv.playwrightBase.maxDiffPixelRatio}
+PLAYWRIGHT_BASE__retries=${buildEnv.playwrightBase.retries}
+PLAYWRIGHT_BASE__workers=${buildEnv.playwrightBase.workers}
+extraEnv=${JSON.stringify(extraEnv)}
```
thiagoelg: `pnpm build-env --print-dotenv` would also help here.

packages/playwright-base/src/bin.ts
```diff
@@ -0,0 +1,319 @@
...
+CI=${!!argv.ci}
+PLAYWRIGHT_BASE__enableChromiumProject=${buildEnv.playwrightBase.enableChromiumProject}
+PLAYWRIGHT_BASE__enableGoogleChromeProject=${isGoogleChromeEnabled}
+PLAYWRIGHT_BASE__enableWebkitProject=${buildEnv.playwrightBase.enableWebkitProject}
+PLAYWRIGHT_BASE__projectTimeout=${buildEnv.playwrightBase.projectTimeout}
+PLAYWRIGHT_BASE__expectTimeout=${buildEnv.playwrightBase.expectTimeout}
+PLAYWRIGHT_BASE__maxDiffPixelRatio=${buildEnv.playwrightBase.maxDiffPixelRatio}
+PLAYWRIGHT_BASE__retries=${buildEnv.playwrightBase.retries}
+PLAYWRIGHT_BASE__workers=${buildEnv.playwrightBase.workers}
+extraEnv=${JSON.stringify(extraEnv)}
```
thiagoelg: Also in here

thiagoelg (changes requested):
Thanks for this PR @ljmotta! This is an awesome addition to our e2e testing architecture.

I've left some comments inline.

packages/playwright-base/env/index.js
ljmotta: @thiagoelg Currently, tests run on an amd64 container, so we don't have an incopatibility issue. Now, this PR keeps compatibility with what we currently have, and Google Chrome tests were being skipped on `online-editor` and `scesim-editor` even when running on a amd64 container.

ljmotta: I believe we could create a new issue to re-asses the need for skipping tests on Google Chrome for darwin arm64.

packages/playwright-base/src/bin.ts
ljmotta: I'm not sure how this would help creating the object. Like run the command with `execSync` and parse the result to create an object?

packages/playwright-base/src/bin.ts
ljmotta: I could extract it to a function as the same env variables are used on `run` and `shell`. The `PLAYWRIGHT_BASE__enableGoogleChromeProject` is calculated by the CLI, so using `pnpm build-env --print-dotenv` would print a wrong value

packages/playwright-base/README.md
ljmotta: Actually, I'll update to something like WDYT @thiagoelg :
> Due to compatibility issues, this containerization solution cannot yet be used on native Windows and requires running it directly within WSL (Windows Subsystem for Linux). Additionally, Google Chrome is not available for Linux on ARM64, so any Chrome-based tests in an ARM64 container must be disabled. At present, tests run on AMD64 containers to ensure all browsers and test scenarios execute safely.

We don't have any tests running on arm64, some tests are disabled on Apple Silicon due to historical reasons.

packages/playwright-base/env/index.js
thiagoelg: Sure, sounds good to me. Or we could just stop skipping those tests and see what the CI thinks 👀

packages/playwright-base/src/bin.ts
thiagoelg: Yes, something like that, but if you feel it doesn't help that much, it's OK to leave it like this.

ljmotta: I'll keep as it is, as `pnpm build-env --print-dotenv` will use the caller context. The result will vary depending on the package.

ljmotta:
@thiagoelg Thanks for the review. I've addressed your comments. Now, regarding the apple silicon specific environment variable, I checked it locally and the tests are do passing on a darwin arm64 machine, so I have removed all code related to this.

thiagoelg (approved):

tiagobento:
Thanks @ljmotta. Great work standardizing the containerized approach for E2E testing with Playwright.

A couple comments/questions:
- I noticed a slight difference in the font on our screenshots inside the FEEL expressions. Is that expected? What is the reason for this change?
- Since `pw-e2e-container` is inside the `playwright-base` package... WDYT about naming it "playwright-base-container` instead?

ljmotta:
@tiagobento the screenshots on `boxed-expression-component` were outdated as they were using the `0.1` maxDiffPixel instead of the `0.001` maxDiffPixel that was in place for the other tests. Some time ago the playwright image was updated, probably causing the font difference.

Sure, I'll update it the CLI name.

ljmotta:
@tiagobento Could you please re-review?

tiagobento (approved):
Approved! Thanks!!

PR #3388, opened 2026-01-06
kie-issues#3387: DMN Runner missing validation information about the number range 

### LINKED ISSUE #3387: DMN Runner missing validation information about the number range
**Limitation**  
In the DMN Runner panel, when users enter a number outside the allowed range defined in the data type constraints, no validation error or warning is displayed.

This missing feedback causes confusion and can lead to incorrect evaluations, especially in models where number ranges are critical to decision logic.

There is currently no workaround. Users must manually verify the allowed range in the Data Types Editor before entering values in the DMN Runner.

Previously, when a value did not match the declared constraint, the DMN Runner displayed validation information. The expected behavior is that the system should continue to show appropriate validation messages whenever input values fall outside the defined constraints.

Image

rajalakshmys-27:
Closes : [kie-issues#3387](https://github.com/apache/incubator-kie-tools/issues/3387)

Fix for DMN Runner missing validation information about the number range

image (4)

image (5)

### DIFF
--- a/packages/uniforms-patternfly/src/wrapField.tsx
+++ b/packages/uniforms-patternfly/src/wrapField.tsx
@@ -48,9 +48,18 @@ filterDOMProps.register(
   "menuAppendTo"
 );
 
+interface ValidationError {
+  instancePath: string;
+  message: string;
+  keyword: string;
+  params: Record<string, any>;
+  schemaPath?: string;
+  data?: any;
+}
+
 type WrapperProps = {
   id: string;
-  error?: boolean | object;
+  error?: boolean | ValidationError;
   errorMessage?: string;
   help?: string;
   showInlineError?: boolean;
@@ -99,7 +108,7 @@ export default function wrapField(
       {...filterDOMProps(props)}
     >
       {children}
-      {error === true || (error && typeof error === "object") ? (
+      {error ? (
         <FormHelperText>
           <HelperText>
             <HelperTextItem variant="error">{errorMessage}</HelperTextItem>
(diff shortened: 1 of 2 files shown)


packages/uniforms-patternfly/src/wrapField.tsx
```diff
@@ -50,7 +50,7 @@ filterDOMProps.register(
 
 type WrapperProps = {
   id: string;
-  error?: boolean;
+  error?: boolean | object;
```
jomarko: with similar change, I wonder, does backend return still `boolean` in some scenarios? Or from now on it is always object. is this object always somehow unique or it has some known structure?

jomarko (commented):
thank you for the pr. I believe your fix works, just trying to understand if both boolezn and object are needed in kie-tools code

packages/uniforms-patternfly/src/wrapField.tsx
rajalakshmys-27: Thank you for the review.

To clarify, the error type in Uniforms depends on the schema bridge in use:

- With uniforms-bridge-simple-schema-2, error can still be a plain boolean (true), with the actual message provided separately via errorMessage.

- With uniforms-bridge-json-schema (AJV), error is a structured object that follows a known format (instancePath, message, keyword, params, etc.). 

In this PR, the type was updated from boolean | object to boolean | ValidationError. 

This documents the expected AJV shape when present, while still allowing the legacy boolean case. Since the codebase includes both bridges, we need to support both scenarios.

jomarko (approved):
Thank you for the PR!

thiagoelg (approved):

PR #3395, opened 2026-01-14
kie-issues#2209: CORS Proxy sometimes break with gzip encoded responses

### LINKED ISSUE #2209: CORS Proxy sometimes break with gzip encoded responses
Some gzipped responses (like from IBM watsonx, Bitbucket, or Gitlab) break CORS Proxy handling, causing it to drop the request and return an error.

thiagoelg:
Closes https://github.com/apache/incubator-kie-issues/issues/2209

This PR introduces support for streaming responses in the CORS Proxy service, enabling it to handle gzipped responses.

Disabling encoding is still supported via the `DISABLE_ENCODING` header, but it shouldn't be necessary anymore.


ljmotta (approved):

packages/cors-proxy/src/proxy/ExpressCorsProxy.ts
```diff
@@ -60,26 +59,33 @@ export class ExpressCorsProxy implements CorsProxy<Request, Response> {
...
         }
       });
 
-      // TO DO: Figure out why this gzip encoding is broken with insecure tls certificates!
-      if (req.headers[CorsProxyHeaderKeys.INSECURELY_DISABLE_TLS_CERTIFICATE_VALIDATION] === "true") {
-        outHeaders["accept-encoding"] = "identity";
-      }
       // Force uncompressed response if encoding is disabled via header
       if (req.headers[CorsProxyHeaderKeys.DISABLE_ENCODING] === "true") {
         outHeaders["accept-encoding"] = "identity";
```
fantonangeli, line 81: Hi @thiagoelg , I personally had issues with Brotli (`content-encoding: br`) compression.
If you have the possibility,  can you try disabling `br` compression, like accepting only `gzip, deflate, zstd`?
If the issue is with `br` maybe is not a `cors-proxy` issue, but it's related to our network, and it's like temporary?

thiagoelg: I'm not sure how to test that...

I've tried making requests to https://httpbin.org/brotli using the CORS Proxy, and it works fine (with or without the `accept-encoding` header).
I've also tested different encodings available here: https://httpbin.org/#/Response_formats

fantonangeli: You can test using the header:
```
Accept-Encoding: gzip, deflate, br, zstd
```
and if you receive this, you are testing it with brotli:

```
content-encoding: br
```
I tried to query this brotli service without the CORS Proxy, with success. 
Last week it was impossible for me to receive any brotli compressed content, receiving `NS_ERROR_INVALID_CONTENT_ENCODING` in the browser. This is why I thought it was a temporary issue not related to CORS Proxy.

thiagoelg: Yep, it works fine.

image

thiagoelg: But if I comment out the `res.removeHeader("content-encoding");` line from CORS Proxy, then the request fails with `net::ERR_CONTENT_DECODING_FAILED 200 (OK)`.
image


But I think that makes sense, as `node-fetch` in CORS Proxy is already decompressing the response and returning raw data to the browser.

fantonangeli (approved):

PR #3397, opened 2026-01-19
kie-issue#1192: On the DMN Editor's Boxed Expression Editor, FEEL autocompletion does not suggest the type properties - Part 2 of 2

### LINKED ISSUE #1192: On the DMN Editor's Boxed Expression Editor, FEEL autocompletion does not suggest the type properties - Part 2 of 2
On a `date and time duration` it should appears the suggestion to access the `days`, `hours`, `minutes` and `seconds` as shown on the FEEL handbook [1]. The same apply for all built-in types.

example of duration:
Screenshot 2024-05-10 163518

example of date:
image

[1] https://kiegroup.github.io/dmn-feel-handbook/#feel-values

This issue is split into two parts. This ticket is related to Part 2.
The Part 1 is here: https://github.com/apache/incubator-kie-issues/issues/1416

danielzhe:
Closes: https://github.com/apache/incubator-kie-issues/issues/1192

https://github.com/user-attachments/assets/1a418756-6868-4dcb-9436-7e272e673953


packages/dmn-feel-antlr4-parser/src/parser/grammar/visitor/FeelVisitorImpl.ts
```diff
@@ -0,0 +1,296 @@
...
+    switch (nodeType) {
+      case 1:
+        return BuiltInTypes.Boolean;
+      case 22:
+      case 23:
+        return BuiltInTypes.Number;
+      case 24:
+        return BuiltInTypes.String;
+      default:
+        return BuiltInTypes.Any;
```
thiagoelg, line 300: What are these numbers? (1, 22, 23, 24)

danielzhe: Great question! I'm changing it to use the `public static` integer where it comes from.

danielzhe: @thiagoelg Updated to use the proper public static integer.

thiagoelg (approved):

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    ["context merge", BuiltInTypes.Context],
+    ["count", BuiltInTypes.Number],
+    ["date", BuiltInTypes.Date],
+    ["date and time", BuiltInTypes.DateAndTime],
+    ["day of week", BuiltInTypes.String],
+    ["day of year", BuiltInTypes.Number],
+    ["decimal", BuiltInTypes.Number],
+    ["decision table", BuiltInTypes.Any],
+    ["distinct values", BuiltInTypes.List],
+    ["duration", BuiltInTypes.DaysAndTimeDuration],
```
ljmotta, line 47: I believe this could return `YearsAndMonthsDuration` too, or not?

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    ["index of", BuiltInTypes.List],
+    ["insert before", BuiltInTypes.List],
+    ["invoke", BuiltInTypes.Any],
+    ["is", BuiltInTypes.Boolean],
+    ["list contains", BuiltInTypes.Boolean],
+    ["list replace", BuiltInTypes.List],
+    ["log", BuiltInTypes.Number],
+    ["lower case", BuiltInTypes.String],
+    ["matches", BuiltInTypes.Boolean],
+    ["max", BuiltInTypes.Number],
```
ljmotta: `max` returns `any` as it will return the same type of the list elements.

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    ["mean", BuiltInTypes.Number],
+    ["median", BuiltInTypes.Number],
+    ["meets", BuiltInTypes.Boolean],
+    ["met by", BuiltInTypes.Boolean],
+    ["min", BuiltInTypes.Number],
```
ljmotta: Same as `max`

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    ["nn max", BuiltInTypes.Number],
+    ["nn mean", BuiltInTypes.Number],
+    ["nn median", BuiltInTypes.Number],
+    ["nn min", BuiltInTypes.Number],
+    ["nn mode", BuiltInTypes.Number],
+    ["nn stddev", BuiltInTypes.Number],
+    ["nn sum", BuiltInTypes.Number],
+    ["not", BuiltInTypes.Boolean],
+    ["now", BuiltInTypes.DateAndTime],
+    ["Number", BuiltInTypes.Number],
```
ljmotta: `number`?

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    ["split", BuiltInTypes.List],
+    ["sqrt", BuiltInTypes.Number],
+    ["started by", BuiltInTypes.Boolean],
+    ["starts with", BuiltInTypes.Boolean],
+    ["starts", BuiltInTypes.Boolean],
+    ["stddev", BuiltInTypes.Number],
+    ["string", BuiltInTypes.String],
+    ["string join", BuiltInTypes.String],
+    ["string length", BuiltInTypes.Number],
+    ["subList", BuiltInTypes.List],
```
ljmotta: `sublist`

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+    ["subString after", BuiltInTypes.String],
+    ["subString before", BuiltInTypes.String],
+    ["subString", BuiltInTypes.String],
```
ljmotta: `substring`

packages/dmn-feel-antlr4-parser/src/parser/grammar/visitor/FeelVisitorImpl.ts
```diff
@@ -0,0 +1,299 @@
...
+      );
+    }
+
+    return new VisitorResult({ text: ctx.getText(), dataType: beforeDot.dataType.properties.get(afterDot[0].text) });
+  };
+
+  public override visitFnInvocation = (ctx: FnInvocationContext) => {
+    if (ctx.children?.length != 2) {
+      return undefined!;
+    }
```
ljmotta, line 258: Shouldn't this throw an error instead of returning `undefined!`? Could you please add a comment for this scenario?

ljmotta (commented):
@danielzhe Thanks for this PR, I've made some inline comments.

thiagoelg (changes requested):
I'm removing my approval for now, as I think it's worth having a look at @ljmotta comments

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
danielzhe: Yes, can be both. I can change it to `any`, but the autocomplete will not work for that specific case then.

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
danielzhe: humm, ok, changing it!

packages/dmn-feel-antlr4-parser/src/parser/grammar/visitor/FeelVisitorImpl.ts
danielzhe: No, because it can be a malformed expression. We don't want to crash it in that case, we just return undefined because there is nothing we can do.

danielzhe:
@ljmotta Changes applied!

packages/dmn-feel-antlr4-parser/src/parser/grammar/visitor/FeelVisitorImpl.ts
thiagoelg: I don't know if this is a limitation of the Feel parser (and its class declarations), but in this case, can't we return an instance of `VisitorResult` without the `dataType`?
`undefined!` feels hacky, and bypassing the typing checks could cause unexpected issues along the way.

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
ljmotta: I'm not sure how to proceed here. Ideally, we would keep the `duration` working as expected, for that we need to change `FeelFunctionReturningTypes` to support multiple return types (`DataType[]`) instead of changing it to `any`.

Any ideas @thiagoelg ?

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
```diff
@@ -110,16 +110,16 @@ export class FeelFunctionReturningTypes {
...
+    ["substring before", BuiltInTypes.String],
+    ["substring", BuiltInTypes.String],
     ["sum", BuiltInTypes.Number],
     ["time", BuiltInTypes.Time],
     ["today", BuiltInTypes.Date],
     ["union", BuiltInTypes.List],
     ["upper case", BuiltInTypes.String],
     ["week of year", BuiltInTypes.Number],
-    ["years and months duration", BuiltInTypes.YearsAndMonthsDuration],
+    ["years and months duration", BuiltInTypes.Any],
```
ljmotta: @danielzhe Why changing to `Any`? Shouldn't this be `YearsAndMonthsDuration`?

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
danielzhe: The point here is that the only way to be sure what is the return is to parsing the content of the expression. We can do it, but in another ticket because this change wasn't even really required to be here, I'm just expanded to cover more cases that is not in the ticket and prepared the code for future changes.

packages/dmn-feel-antlr4-parser/src/parser/grammar/visitor/FeelVisitorImpl.ts
danielzhe: Done! I changed it to support `undefined` as a valid result.

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
danielzhe: Probably I drink too much coffee. Fixed now! 😆

packages/dmn-feel-antlr4-parser/src/parser/grammar/visitor/FeelVisitorImpl.ts
thiagoelg: Awesome, thanks!

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
ljmotta: Oh, okay. Could you please create a new issue to track this work? Also it would be good to add a comment to this specific line with the issue link. For now, I think it is okay to keep it `DaysAndTimeDuration`.

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
```diff
@@ -0,0 +1,125 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { DataType } from "./DataType";
+import { BuiltInTypes } from "./BuiltInTypes";
+
+export class FeelFunctionReturningTypes {
+  public static readonly Index: ReadonlyMap<string, DataType> = new Map<string, DataType>([
```
jomarko, line 24: Are we sure we define return type for each DMN function? The list of dmn functions in this file

Comparing the count of functions with /home/jomarko/apache/kie-tools/packages/feel-input-component/src/i18n/locales/en.ts 

seems to have more entries. That is reason for my question.

jomarko (commented):
Thank you for the PR. I put one question inline. Also one question here. To me it is not clear the relationship between `date` and `date and time`.

I would expect that `date` is subset of the `date and time`. In other words. I would expect `date and time` contains all properties of `date` plus some `time` specifics.

If you look on the screenshots, I would expect `date and time` offers me also a `day`. Is my expectation wrong?

Not a DMN specification expert. So I may be wrong. Please let me know. Thank you.


Image
Image

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
danielzhe: I just get the types from [here](https://github.com/apache/incubator-kie-tools/blob/cae1c6fe2d21f8e37bb255ac1bc7d1bace4ea753/packages/dmn-feel-antlr4-parser/src/parser/ReservedWords.ts#L43) and moved to another class adding the returning type. If we have anything that was not being handled before, it is not the focus of this PR and we should create a proper ticket for that, @jomarko

danielzhe: About the issue with "date and time" and "date", your expectation is correct, but this is not an issue related to this PR. Can you create a new ticket for it? @jomarko

jomarko: I double checked:
- packages/feel-input-component/src/FeelConfigs.ts
- packages/feel-input-component/src/i18n/locales/en.ts
- packages/dmn-feel-antlr4-parser/src/parser/ReservedWords.ts
- packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts

it seems that first two listed files does not contain description for two functions:
- `decision table`
- `invoke`

### decision table
Is so complex with the amount of parameters, that I am actually not sure how the description should look like

### invoke
is marked as deprecated in https://kiegroup.github.io/dmn-feel-handbook/#invoke-namespace-modelname-decisionname-parameters

jomarko (approved):
New issue for `date` and `date and time` created and assigned to you.
apache/incubator-kie-issues#2247

New issue for functions description created and assigned to you
apache/incubator-kie-issues#2248

feel free to reassign or unassign, was not sure who should be assignee

Approving this PR form my side. Thank you.

packages/dmn-feel-antlr4-parser/src/parser/FeelFunctionReturningTypes.ts
danielzhe: https://github.com/apache/incubator-kie-issues/issues/2250

thiagoelg (approved):

PR #3403, opened 2026-01-22
kie-issue#2216: Dev Deployments: Support recursive/nested token interpolation and remove non-standard JSON Patch operation

### LINKED ISSUE #2216: Dev Deployments: Support recursive/nested token interpolation and remove non-standard JSON Patch operations
`k8s-yaml-to-apiserver-requests` has two issues that limit its flexibility:

### 1. Tokens can't reference other tokens
Right now, tokens like `${{ tokenName }}` only do simple string replacement. You can't nest tokens inside other tokens or query complex data structures.
For eg, this doesn't work:
`host: ${{ $.resources.Route['${{ $.uniqueName }}'].status.ingress[0].host }}`
This makes it hard to build dynamic configurations where one token value depends on another, or when you need to look up data from arrays or nested objects in the token map.

### 2. Using a custom JSON Patch operation
it uses a custom `CheckTypeOperation` that's not part of the **JSON Patch standard (RFC 6902)**. This makes the code less compatible with standard tools and harder to maintain.

Adarshvk98:
Closes [kie-issue#2216](https://github.com/apache/incubator-kie-issues/issues/2216)

This PR adds test coverage for token interpolation in Kubernetes YAML resources and enhances the deployment system to support JSON path syntax with recursive and nested token references.

 Changes

- Added [jsonpath-rfc9535](https://www.npmjs.com/package/jsonpath-rfc9535) dependency for RFC-compliant JSON path handling
- Added unit tests for token interpolation and JSON patch operations
- KubernetesService to merge parameter tokens during resource patching
- Updated deployment YAML generators to use JSON path syntax


packages/k8s-yaml-to-apiserver-requests/src/patchK8sResourceYaml.ts
```diff
@@ -17,23 +17,15 @@
  * under the License.
  */
 
-import { Operation, TestOperation, applyPatch, getValueByPointer } from "fast-json-patch";
+import { Operation, TestOperation, applyPatch } from "fast-json-patch";
```
thiagoelg, line 20: Does the `jsonpath-rfc9535` library offer similar types so that we can remove this dependency?

packages/k8s-yaml-to-apiserver-requests/src/patchK8sResourceYaml.ts
```diff
@@ -81,14 +47,6 @@ export function patchK8sResourceYaml(k8sResourceYaml: string, patches: ResourceP
           : applyPatch(updatedResource, patch.testFilters, false, false).every(({ test }) => Boolean(test));
```
thiagoelg, line 47: We should be using the `exec` method from the `jsonpath-rfc9535` library here, right?

thiagoelg (changes requested):
Thanks for the PR @Adarshvk98!
I really appreciate the addition of tests for the `k8s-yaml-to-apiserver-requests` library.

Left some comments inline, let me know what you think.

packages/k8s-yaml-to-apiserver-requests/src/patchK8sResourceYaml.ts
Adarshvk98: No, from my understanding, `jsonpath-rfc9535` and `fast-json-patch` are completely different libraries serving different purposes:

- `jsonpath-rfc9535`: Implements JSONPath (RFC 9535) for querying/selecting values from JSON documents
- `fast-json-patch`: Implements JSON Patch (RFC 6902) for modifying JSON documents with operations like add, remove, replace, test

packages/k8s-yaml-to-apiserver-requests/src/patchK8sResourceYaml.ts
Adarshvk98: From what I understand, No, we shouldn't use jsonpath-rfc9535 here. this library is for JSONPath queries (used for token interpolation), while here uses JSON Patch (RFC 6902) operations.

packages/k8s-yaml-to-apiserver-requests/src/patchK8sResourceYaml.ts
thiagoelg: You're completely right. I got confused with the RFC names... they should really have named them a bit differently hahaha

thiagoelg (approved):

packages/k8s-yaml-to-apiserver-requests/tests/jsonPatch/patchK8sResourceYaml.test.ts
ljmotta, line 1: Please consider inlining the test case values or creating a test table with clear given / expected fields. The current structure makes it difficult to visualize the different test scenarios.

Test table example (property names are just a suggestion):
```ts
[{
    name: "should add a label to existing labels",
    given: {
        yaml: `
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-app
  labels:
    app: my-app
spec:
  replicas: 1`,
        patch: [{
            op: "add",
            path: "/metadata/labels/environment",
            value: "production",
        }]
    },
    expected: "environment: production"
},
// ...
]
```

Alternatively, inlining the values directly in the test cases would also be fine, even if it means duplicating some content between tests:
```ts
    it("should add a label to existing labels", () => {
        const result = patchK8sResourceYaml(`
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-app
  labels:
    app: my-app
spec:
  replicas: 1`, [{
            jsonPatches: [{
                op: "add",
                path: "/metadata/labels/environment",
                value: "production",
            }]}]);

        expect(result).toContain("environment: production");
    });

```

ljmotta (commented):
@Adarshvk98 Thanks for this PR! I’ve tested the Dev Deployments using kind, and everything is working as expected. The code looks great overall.

I’ve added a comment regarding the tests, the current test structure is quite hard to follow and visualize as the yaml, patch, and expected value for one scenario are split into their own variables.

packages/k8s-yaml-to-apiserver-requests/tests/jsonPatch/patchK8sResourceYaml.test.ts
Adarshvk98: Thank you for highlighting it. I agree it is a bit difficult to understand the current test data structure, and I like the suggested structure and have refactored into it.

ljmotta (approved):
@Adarshvk98 Thanks for your patience! The PR looks good to me. The tests are very easy to read and understand.

PR #3405, opened 2026-01-22
kie-issues#3404: DMN Runner evaluation highlights not working for Decision Tables and Boxed Context inside BKMs or Boxed Functions

### LINKED ISSUE #3404: Dmn Editor: DMN Runner evaluation highlights not working for Decision Tables and Boxed Context inside BKMs or Boxed Functions
Decision Tables and Boxed Contexts inside BKMs or Boxed Functions are not highlighting when Evaluation highlights are turned on.

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-tools/issues/3404,

Decision Tables inside BKMs were not showing evaluation highlights because evaluation hit IDs are returned as per the Decision in /dmnResult API. So now we are taking evaluation hit IDs from all Decisions that invoke the BKM to correctly display row highlights when the BKM is opened in the Editor.

Example:
Screenshot 2026-01-22 at 7 22 14 PM


Kusuma04-dev:
Getting evaluationHitIds are merged recently and updated the latest snapshots here: https://github.com/apache/incubator-kie-tools/pull/3402 , With this latest snapshot changes only we can test this.

jomarko (commented):
Thank you for this PR. I see it as an improvement of the evaluation highlights feature.

However, I am not sure if we are consistent enough in introduced changes. I understand we are from now on showing evaluation hits count inside BKM nodes - that is perfect.

The PR says BKM nodes will show evaluation hits count for:
- Decision Table ✅ 
- Context ❓ this is something I do not understand

The original evaluation highlights feature was introduced to show evaluation hits count for:
- Decision Table
- Conditional IF expression

So I think also #3405 should show evaluation hits count in BKM only for:
- Decision Table
- Conditional IF expression

what do you think?

### Manual check
Here I attach file that show #3405 works perfect for Decision tables but not for Conditional IF. Also I do not think it works for context as PR tittle states.

[Untitled.dmn.txt](https://github.com/user-attachments/files/24817455/Untitled.dmn.txt)

Image
Image
Image
Image

### Code comment
I tend to say `bkmEvaluationResults` computation could be moved somehow into computation of `evaluationResultsByNodeId`, so it means moved into 
`transformExtendedServicesDmnResult` in `DmnRunnerContextProvider.tsx`. Such refactoring would mean:
- we do not modify `BoxedExpressionScreen.tsx`, we keep even `? evaluationResultsByNodeId?.get(activeDrgElementId ?? "")?.evaluationHitsCountByRuleOrRowId`
- we have all `rowHitsCount` computation concentrated in single place

I didn't tried to complete such refactoring, but it seems to me as doable. Maybe I am wrong. Sorry if that is the case and you have tried already.

Kusuma04-dev:
Thanks for the review @jomarko , Yes it should display for both Decision Table and Conditional IF expression. I am working on it.
Yes you are right. I will move it to the contextProvider. Thanks for pointing out.

Kusuma04-dev:
These changes needs to be checked along with :https://github.com/apache/incubator-kie-tools/pull/3415

jomarko (approved):
Thank you for the PR. I can confirm when using #3415 the changes are working fine.

thiagoelg (approved):

PR #3437, opened 2026-02-11
kie-issues-2243: KIE Sandbox: DMN Runner inputs should be renamed when InputData nodes are renamed in the DMN model

### LINKED ISSUE #2242: KIE Sandbox: DMN Runner inputs constrained as enums should "survive" enum renames
Reproducer:
- Create a DMN model and add a string Data Type with enum constraints with values "foo" and "bar".
- Open DMN Runner in Table mode and enter some rows, using the "foo" and "bar" values.
- Go back to the Data Type and rename "bar" to "baz".
- The DMN Runner Table will not replace "bar" to "baz" automatically, requiring users to manually go back to the table and "fix" the values.

### LINKED ISSUE #2243: KIE Sandbox: DMN Runner inputs should be renamed when InputData nodes are renamed in the DMN model
It's very frustrating to lose inputs just because an InputData node was renamed.

Reproducer:
- Create a DMN model with some Inputs.
- Enter some values in the DMN Runner table
- Rename your Inputs in the DMN model.
- Values are lost in the DMN Runner table.

ljmotta:
Closes https://github.com/apache/incubator-kie-issues/issues/2243
Closes https://github.com/apache/incubator-kie-issues/issues/2242

## On this PR
- Renaming the InputData will update the DMN Runner form/table without losing the inputs
- Renaming a property will update the DMN Runner form/table without losing the inputs
- Updating the enum will update the DMN Runner form/table
- Fix a bug on DMN Editor where constraints using "allowedValues" weren't updated accordingly. DMN 1.5 spec states that enums should be defined through "typeConstraint" attribute, so even updating an enum that uses the deprecated property "allowedValues" would create this new attribute. With an outdated "allowedValues" and an updated "typeConstraint" attribute, the Open API Schema of the DMN would have a merge between both attributes (old + new). This PR fixes this and updates "allowedValues".

---

### Renaming Input Data nodes

Uploading Screen Recording 2026-02-10 at 22.19.15.mov…

### Updating enums

https://github.com/user-attachments/assets/10f88dc2-8cb6-4004-8654-20cc55ad56e5

### Renaming Input Data nodes

https://github.com/user-attachments/assets/1fb24045-5f00-46bf-b744-cc3ecf91a1cb

### Updating enums

https://github.com/user-attachments/assets/0480b5df-92fe-4db4-8a0c-17687a00ed63

https://github.com/user-attachments/assets/4ae99995-1b6e-43c3-abf5-18a97615e429

---

Enum.dmn file:
[Enum.dmn.txt](https://github.com/user-attachments/files/25222977/Enum.dmn.txt)


thiagoelg (approved):

jomarko (commented):
Thank you for the PR. I found only one scenario, when I am not sure, if the behavior is expected, or better to say, if we are aware of it. It is about renaming Input Node, but clicking **Just Rename**

### Initial state
Image

### Rename
I just rename, I do not change all occurrences

Image

### Final state
- It is expected output is not showing result
- However, input card was not renamed. So is the requirement of DMN Runner Cards to be shown properly to have valid DMN Model?


Image

ljmotta:
@jomarko Thanks for your review. The DMN must be valid before the DMN Runner form is updated. By using the "Just rename" option, the Decision node will not be updated, making this an invalid DMN.

jomarko (approved):

packages/dmn-editor-standalone/tests-e2e/__fixtures__/prettier.ts
```diff
@@ -0,0 +1,20 @@
+import * as prettier from "prettier";
+import xmlPlugin from "@prettier/plugin-xml";
+
+// Prettier loads configuration files using dynamic `import()` calls inside `prettier.resolveConfig()`.
+// However, when Playwright detects the use of `__dirname`, it opts into its CommonJS-based test runner
+// (see https://github.com/microsoft/playwright/issues/37890).
```
thiagoelg, line 25: If that's the case, can't we just replace `__dirname` with `import.meta.dirname`?

See https://nodejs.org/api/esm.html#importmetadirname

ljmotta: To use `import.meta.dirname` we would need to include `"type": "module"` into the `package.json`, which would lead to other problems. For example we would require to add  the type `module` in several other packages that are used by `dmn-editor-standalone`. Also, it would require to add the file extension into the imports.

PR #3440, opened 2026-02-13
kie-issues#2244: KIE Sandbox: DMN Runner Outputs table stealing the focus from Inputs table

### LINKED ISSUE #2244: KIE Sandbox: DMN Runner Outputs table stealing the focus from Inputs table
It's impossible to keep focused and use keyboard navigation on DMN Runner Inputs table, in table mode. The Outputs table steals the focus and makes editing values very frustrating.

The same problem happens with other elements in the screen as well, not only the Inputs table.

ljmotta:
Closes: https://github.com/apache/incubator-kie-issues/issues/2244

This PR prevents the `focus()` call when the activeElement is not from the current component hierarchy.

### Before

https://github.com/user-attachments/assets/2bd24ce0-7286-43ab-8c11-35d9cfc963e5

### After

https://github.com/user-attachments/assets/cf1a8c07-841e-4658-ab16-82ecc5ef2b31

### DIFF
--- a/packages/boxed-expression-component/tests-e2e/features/keyboard/keyboard.spec.ts
+++ b/packages/boxed-expression-component/tests-e2e/features/keyboard/keyboard.spec.ts
@@ -17,25 +17,120 @@
  * under the License.
  */
 
-import { test } from "../../__fixtures__/base";
+import { expect, test } from "../../__fixtures__/base";
+
+const ACTIVE_CLASS_REGEXP = /(^|\s)active(\s|$)/;
 
 test.describe("Keyboard", () => {
-  test.skip(true, "https://github.com/apache/incubator-kie-issues/issues/542");
   test.describe("Navigation", () => {
-    test("should correctly navigate", async () => {
-      // enter, shift+enter, tab, shift+tab, escape
-    });
+    test("should correctly navigate", async ({ bee, page, useCases }) => {
+      await useCases.openLoanOriginations("bureau-strategy-decision-service", "bureau-call-type");
+      const decisionTable = bee.expression.asDecisionTable();
+      await decisionTable.cellAt({ row: 1, column: 1 }).select();
 
-    test.describe("Arrow keys", () => {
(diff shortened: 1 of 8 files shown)


thiagoelg (approved):

jomarko (commented):
I can confirm the fix works. however from my point of view, we should maybe not postpone anymore the task https://github.com/apache/incubator-kie-issues/issues/543

This PR touches selection logic and we miss tests for it:
- https://github.com/apache/incubator-kie-tools/blob/main/packages/boxed-expression-component/tests-e2e/features/keyboard/keyboard.spec.ts
- https://github.com/apache/incubator-kie-tools/blob/main/packages/boxed-expression-component/tests-e2e/features/keyboard/keyboard.spec.ts

ljmotta:
@jomarko Thanks for your review. Regarding the tests, I have added two scenarios of tests checking the keyboard navigation on boxed expression editor.

packages/boxed-expression-component/tests-e2e/features/keyboard/keyboard.spec.ts
```diff
@@ -17,25 +17,120 @@
...
 
-    test.describe("Arrow keys", () => {
-      test("should correctly navigate", async () => {
-        // arrow up/down/left/right
-      });
+      await page.keyboard.type(`"test2"`);
+      await expect(decisionTable.cellAt({ row: 2, column: 1 }).content).toHaveClass(ACTIVE_CLASS_REGEXP);
+
+      await page.keyboard.press("Escape");
+      await expect(decisionTable.cellAt({ row: 1, column: 1 }).content).not.toContainText(`"test2"`);
```
jomarko: I think here should be row: 2 instead of row: 1

jomarko (commented):
Amazing tests, thank you!

I found just one think that should be updated in both of them, let me know what do you think.

jomarko (approved):

PR #3486, opened 2026-03-05
NO-ISSUE: Quarkus Dev UI: Blank screen when trying to open any page after quarkus upgrade

nrknithin:
The Quarkus 3.20.3 → 3.27.2 upgrade included a Vaadin Router 1.7.5 → 2.0.0 change that shortened Dev UI page URLs:

- **Old:** `/q/dev-ui/org.jbpm.jbpm-quarkus-devui`
- **New:** `/q/dev-ui/jbpm-quarkus-devui`

Static resources were still being served at the full namespace path, while the iframe resolved relative URLs against the new short-name page URL — causing `envelope.js` to be fetched from the wrong path and Quarkus's Dev UI catch-all returning HTML instead of JS.

### Changes
1. **`DevConsoleProcessor.java` ( jBPM):** Updated `BASE_RELATIVE_URL` to use the short artifact name matching Quarkus's new page routing, and registered static resource routes with explicit priority so they are evaluated before Quarkus's Dev UI catch-all
   - jBPM: `dev-ui/jbpm-quarkus-devui`

### Test plan
On a Kogito application with the jbpm-quarkus-devui module:
- `mvn clean quarkus:dev` → Dev UI pages (Process Instances, Tasks, Jobs, Forms) load without blank screen


thiagoelg (approved):
LGTM!

danielzhe (approved):

PR #3538, opened 2026-04-14
[incubator-kie-issues#2029]  Dependency management simplification

### LINKED ISSUE #2029: Dependency management simplification
This issue is to track effort to properly separate the BoMs for Quarkus and Springboot in Drools.

Ideally it should be possible to have a BoM for drools/quarkus projects, a BoM for drools/Springboot projects and a BoM for pure Drools project.

As demontrated by PRs on quarkus upgrade, version of the same libraries are declared/duplicated in multiple boms/poms, making harder to maintain/keep in sync them all over the repositories.

Scope of this ticket is to:

- identify all of them
- group all of them in one single pom.xml inside the "drools" repository and have all of them inherited throughout the codebase, so that such version would be declared only in one single place
- one of the observed pattern is that the version is declared in some pom, and then the library itself is declared in some dependencyManagement tag in children pom - which some how break the concept of dependencyManagement in two different places (the pom where the version is declared as property and the one where it is used in the dependencyManagement) and lead to the duplication - ideally, we should need one single dependencyManagement declaration all over the stack
- enforce clearly separated pom hierarchies between quarkus and springboot frameworks

This comes from suggestions of @gitgabrio and discussions with @mariofusco

yesamer:
Fixes: https://github.com/apache/incubator-kie-issues/issues/2029

kie-tools

Depends on:
https://github.com/apache/incubator-kie-drools/pull/6633
https://github.com/apache/incubator-kie-kogito-runtimes/pull/4241
https://github.com/apache/incubator-kie-kogito-apps/pull/2317


packages/maven-quarkus-bom/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "private": true,
+  "name": "@kie-tools/maven-quarkus-bom",
```
gitgabrio: I would suggest to keep the name of the folder and the name of the artifacts consistent, to make things easier to maintain (it is also a sort of good-rule in the maven world, even if optional)

examples/process-accelerator/pom.xml
```diff
@@ -25,9 +25,9 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.kie</groupId>
-    <artifactId>kie-tools-maven-base</artifactId>
+    <artifactId>kie-tools-quarkus-bom</artifactId>
```
gitgabrio, line 28: Here the parent artifact is  `kie-tools-quarkus-bom` , but elsewhere it is `kie-tools-maven-quarkus-bom`; I think they should be the same; beside, if there is one single name to identify the same object (the artifactId and the folder) things would be easier to follow, IMO

examples/process-business-calendar/pom.xml
```diff
@@ -27,9 +27,9 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.kie</groupId>
-    <artifactId>kie-tools-maven-base</artifactId>
+    <artifactId>kie-tools-maven-quarkus-bom</artifactId>
```
gitgabrio: Here the parent artifact is `kie-tools-maven-quarkus-bom `, but elsewhere it is `kie-tools-quarkus-bom`; I think they should be the same (`kie-tools-quarkus-bom`)

packages/kogito-jobs-service-postgresql-image/resources/app/pom.xml
```diff
@@ -25,7 +25,7 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.kie</groupId>
-    <artifactId>kie-tools-maven-base</artifactId>
+    <artifactId>kie-tools-quarkus-bom</artifactId>
     <version>${revision}</version>
     <relativePath>../../node_modules/@kie-tools/maven-base/pom.xml</relativePath>
```
gitgabrio: RelativePath  to be updated

packages/maven-base/pom.xml
```diff
@@ -122,27 +122,15 @@
...
-    <version.org.junit.jupiter>5.13.4</version.org.junit.jupiter>
-    <version.org.mockito>5.18.0</version.org.mockito>
     <version.org.kie.j2cl.tools.yaml.mapper>0.4</version.org.kie.j2cl.tools.yaml.mapper>
-    <version.io.netty>4.1.131.Final</version.io.netty>
 
     <!-- These versions are overrides for transitive dependencies, to fix security vulnerabilities.
            They need to be checked with Quarkus and Spring Boot upgrades and eventually removed, if they are not needed anymore. -->
-    <version.tomcat.embed.core>10.1.54</version.tomcat.embed.core>
-    <version.apache.commons.lang3>3.18.0</version.apache.commons.lang3>
     <version.angus.mail>2.0.5</version.angus.mail>
```
gitgabrio: This is already defined in kie-parent, so it should be inherited

packages/maven-base/pom.xml
gitgabrio: This is already defined in kie-parent, so it should be inherited

packages/maven-quarkus-bom/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>kie-tools-quarkus-bom</artifactId>
+  <packaging>pom</packaging>
+
+  <name>KIE Tools :: Quarkus BOM</name>
+  <description>Bill of Materials for KIE Tools Quarkus-based projects</description>
+
+  <properties>
+    <version.io.quarkus>3.27.2</version.io.quarkus>
```
gitgabrio: 🤔  I do not really like it, TBH....
I've the impression it is better to set `kogito-apps-quarkus-bom` so that:
1. we do not need to keep the version in sync all over different repos
2. we automatically inherit the plugin configuration

Then, kie-tools-maven-base could be imported; importing it, would require to duplicate some properties or configuration, but maybe the burden to maintain such duplication in the same repo is easier.

packages/quarkus-bom/quarkus-bom.iml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module version="4">
```
gitgabrio, line 2: IINW, IDE specific files should not be versioned

repo/MANUAL.md
```diff
@@ -226,6 +226,28 @@ All Maven-based packages should declare it as a `dependency` on their `package.j
 </parent>
 ```
 
+In case of Maven-based modules based on Quarkus, the following parent should be used:
+
+```xml
+<parent>
+  <groupId>org.kie</groupId>
+  <artifactId>kie-tools-maven-quarkus-bom</artifactId>
```
gitgabrio: IINW, this should be kie-tools-quarkus-bom

repo/MANUAL.md
```diff
@@ -226,6 +226,28 @@ All Maven-based packages should declare it as a `dependency` on their `package.j
...
+  <relativePath>./node_modules/@kie-tools/maven-quarkus-bom/pom.xml</relativePath>
+</parent>
+```
+
+In case of Maven-based modules based on Spring Boot, the following parent should be used:
+
+```xml
+<parent>
+  <groupId>org.kie</groupId>
+  <artifactId>kie-tools-maven-spring-boot-bom</artifactId>
```
gitgabrio: IINW, this should be kie-tools-spring-boot-bom

yesamer: Fixed.

repo/MANUAL.md
yesamer: Fixed

packages/quarkus-bom/quarkus-bom.iml
yesamer: maven-base and other maven modules have it.

packages/maven-base/pom.xml
yesamer: Removed.

packages/maven-base/pom.xml
yesamer: Removed.

packages/kogito-jobs-service-postgresql-image/resources/app/pom.xml
yesamer: Fixed.

packages/maven-quarkus-bom/package.json
yesamer: @gitgabrio I used this naming just to be consistent with the existing maven-base approach. I'm fine with your suggestion, but I'd ask @thiagoelg what is his input.

examples/process-accelerator/pom.xml
yesamer: @gitgabrio Related to the previous comment //cc @thiagoelg

packages/maven-quarkus-bom/package.json
thiagoelg: Right now we have:
- Package directory: `maven-quarkus-bom`
- Package name: `@kie-tools/maven-quarkus-bom`
- Artifact name: `kie-tools-quarkus-bom`
Is that correct?

I agree that using different names might be confusing, but naming the package `@kie-tools/kie-tools-quarkus-bom` doesn't sound good either.

Maybe this is better?
- Package directory: `quarkus-bom`
- Package name: `@kie-tools/quarkus-bom`
- Artifact name: `kie-tools-quarkus-bom`

WDYT?

gitgabrio: @thiagoelg Well, my point was to have the same name on folder and artifact, it is a good rule in maven world for good reason. I'm not sure why the name ->must<- be different, if I infer correctly from your answer. But if this is the case, I've no strong opinion on the name of the folder, since it is extremely "internal"

packages/maven-quarkus-bom/pom.xml
yesamer: @gitgabrio The reason for this choice is that I’m aware the kie-tools monorepo provides a global parameter to manage the Quarkus version (QUARKUS_PLATFORM_version).
@thiagoelg, we need your input here. Do we need to explicitly declare the Quarkus version so that it is managed by the kie-tools monorepo via QUARKUS_PLATFORM_version, or can we fully delegate version management to kogito-apps-quarkus-bom, as @gitgabrio is suggesting?

thiagoelg: kie-tools should follow the same Quarkus version from the upstream repositories.

The `QUARKUS_PLATFORM_version` env var is used in other places as well, not just POMs (some container images and Sonataflow packages), so we can't just straight remove it. Ideally, the default value of `QUARKUS_PLATFORM_version` should come from the `kogito-apps-bom` built by the `drools-and-kogito` package automatically; this way, it's always in sync.

I'm not sure how to programmatically retrieve a property from a POM, but if it's possible, that would be nice. Otherwise, we can maintain the `QUARKUS_PLATFORM_version` value manually for now.

gitgabrio: @thiagoelg @yesamer 

`mvn help:evaluate -Dexpression=<property-name> -q -DforceStdout`

yesamer: @thiagoelg Can you please then clarify if that highlighted property is then required or we can remove that?

thiagoelg: We can remove it

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/pom.xml
```diff
@@ -141,7 +141,6 @@
       <plugin>
         <groupId>io.quarkus</groupId>
         <artifactId>quarkus-extension-maven-plugin</artifactId>
-        <version>${version.quarkus}</version>
```
tiagobento, line 144: This will make it default to the latest version, if I'm not mistaken

packages/jbpm-quarkus-devui/package.json
```diff
@@ -30,9 +30,10 @@
     "start:win32": "pnpm build:dev && mvn -f ./dev/pom.xml clean quarkus:dev -Dquarkus.http.host=127.0.0.1  -Dkogito.service.url=http://127.0.0.1:8080/kie -Dkogito.jobs-service.url=http://127.0.0.1:8080/kie -Dkogito.data-index.urll=http://127.0.0.1:8080/kie"
   },
   "dependencies": {
-    "@kie-tools/maven-base": "workspace:*"
+    "@kie-tools/maven-quarkus-bom": "workspace:*"
   },
   "devDependencies": {
+    "@kie-tools/maven-base": "workspace:*",
```
tiagobento: Please keep `maven-base` as a "dependency" instead of a "devDependency" the Maven repo tail mechanism depends on it.

packages/maven-base/pom.xml.template
```diff
@@ -122,94 +122,25 @@
...
-    <version.org.junit.jupiter>5.13.4</version.org.junit.jupiter>
-    <version.org.mockito>5.18.0</version.org.mockito>
     <version.org.kie.j2cl.tools.yaml.mapper>0.4</version.org.kie.j2cl.tools.yaml.mapper>
-    <version.io.netty>4.1.131.Final</version.io.netty>
-
-    <!-- These versions are overrides for transitive dependencies, to fix security vulnerabilities.
-           They need to be checked with Quarkus and Spring Boot upgrades and eventually removed, if they are not needed anymore. -->
-    <version.tomcat.embed.core>10.1.54</version.tomcat.embed.core>
-    <version.apache.commons.lang3>3.18.0</version.apache.commons.lang3>
-    <version.angus.mail>2.0.5</version.angus.mail>
```
tiagobento, line 148: Why do we need to remove all those dependencies from here? Are they tightly coupled with Quarkus or Spring Boot?

packages/maven-quarkus-bom/package.json
tiagobento: Let's follow whatever `maven-base` used to do.

tiagobento (commented):
Thanks for the draft PR! Left some minor comments inline.

packages/jbpm-quarkus-devui/jbpm-quarkus-devui-runtime/pom.xml
yesamer: @tiagobento I double-checked:

`kie-tools-quarkus-bom`
    |
`jbpm-quarkus-devui-parent`
    |                                 
`jbpm-quarkus-devui`

In `kie-tools-quarkus-bom`,  the `quarkus-extension-maven-plugin` is declared.

packages/maven-quarkus-bom/package.json
yesamer: @tiagobento That means, the current status is right, correct?

packages/maven-quarkus-bom/pom.xml
yesamer: After trying it out with @gitgabrio, we confirmed that it is not possible to use kogito-apps-quarkus-bom as a parent with the current setup. This is due to Maven’s version replacement mechanism, which ends up using the project.version defined in the leaf POM.

packages/jbpm-quarkus-devui/package.json
yesamer: @tiagobento Done

packages/maven-base/pom.xml.template
yesamer: @tiagobento The dependencies you mentioned are already pulled in transitively via the <artifactId>kogito-apps-bom</artifactId> imported at line 138, which was already present before this PR. Since these dependencies are duplicated, we can safely remove them and rely on the single declaration provided by kie-parent.
These are framework-agnostic dependencies, including some that are used only for testing, so no additional imports are required.

gitgabrio (approved):

examples/process-accelerator/pom.xml
thiagoelg: I don't think we need the `maven` prefix in the package name. Having it implies that there's also a non-Maven BOM.

To me, naming the artifacts `kie-tools-quarkus-bom` and `kie-tools-spring-boot-bom` works, and then naming the packages `quarkus-bom` and `spring-boot-bom`, since `kie-tools` is already implied.

I'm not against having the folder named the same as the artifactId, but that's not the standard in the repository right now, and it would look out of place. In the future, we can investigate whether it makes sense to do so and then update all packages that apply.

thiagoelg (approved):

PR #3541, opened 2026-04-15
NO-ISSUE: Add flags to allow for local overrides of the Quarkus and Kogito versions

thiagoelg:
Added two new env vars:
- `KIE_TOOLS_BUILD__kogitoVersionSetPropertySkip`
- `KIE_TOOLS_BUILD__quarkusVersionSetPropertySkip`

Setting them to `true` allows developers to override the `KOGITO_RUNTIME_version` and `QUARKUS_PLATFORM_version` env vars values without changing the `packages/maven-base/pom.xml` file.

This is important for quick (and ephemeral) tests of different Kogito or Quarkus versions, without generating unwanted changes.

### Usage example:

```sh
export KIE_TOOLS_BUILD__kogitoVersionSetPropertySkip=true
export KIE_TOOLS_BUILD__quarkusVersionSetPropertySkip=true
# Using a `999-<date>-local` version here will trigger a new build of `drools-and-kogito`.
export KOGITO_RUNTIME_version=10.2.0
export QUARKUS_PLATFORM_version=3.27.3

pnpm bootstrap
pnpm -F @kie-tools/jbpm-quarkus-devui... build:dev
```

In this scenario, `@kie-tools/jbpm-quarkus-devui` will be built against a new Quarkus version and the published `10.2.0` version of Apache KIE Kogito Runtime, without changing the `packages/maven-base/pom.xml` file.

---

The default behavior remains the same as before.

### DIFF
--- a/devbox.lock
+++ b/devbox.lock
@@ -1,6 +1,10 @@
 {
   "lockfile_version": "1",
   "packages": {
+    "github:NixOS/nixpkgs/nixpkgs-unstable": {
+      "last_modified": "2026-04-09T19:38:50Z",
+      "resolved": "github:NixOS/nixpkgs/b0188973b4b2a5b6bdba8b65381d6cd09a533da0?lastModified=1775763530&narHash=sha256-BuTK9z1QEwWPOIakQ1gCN4pa4VwVJpfptYCviy2uOGc%3D"
+    },
     "glibcLocales@2.40-66": {
       "last_modified": "2025-05-16T20:19:48Z",
       "resolved": "github:NixOS/nixpkgs/12a55407652e04dcf2309436eb06fef0d3713ef3#glibcLocales",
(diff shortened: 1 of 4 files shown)


ljmotta (approved):

tiagobento:
If I start with

```
export KOGITO_RUNTIME_version=10.2.0
export QUARKUS_PLATFORM_version=3.27.3
```

Then I change it to different values, will it continue working as expected? I'm concerned that writing those values to maven.config might make it impossible to change them without having to manually delete/update the maven.config file.

thiagoelg:
> If I start with
> 
> ```
> export KOGITO_RUNTIME_version=10.2.0
> export QUARKUS_PLATFORM_version=3.27.3
> ```
> 
> Then I change it to different values, will it continue working as expected? I'm concerned that writing those values to maven.config might make it impossible to change them without having to manually delete/update the maven.config file.

As long as you run `pnpm bootstrap`, yes, it should work. The maven.config files will be automatically updated.

Running `pnpm bootstrap` after changing the `KOGITO_RUNTIME_version` env var is already required, even without these changes.

tiagobento (approved):

PR #3555, opened 2026-04-27
[incubator-kie-issues#1933] The new React-based BPMN Editor - more bug fixes

klbowers:
This PR fixes the following bugs in the BPMN Editor epic: https://github.com/apache/incubator-kie-issues/issues/1933 

1. [bug] BPMN Editor: Message start event with a message is not generating an itemDefinition
2. [bug] BPMN Editor: Correlation Key and Subscription not cleaned up after deletion
3. ~~[bug] BPMN Editor: Message Start Event-itemDefinition Not Assigning Value to structureRef~~
        - Problem with the engine, an empty structureRef shouldn't cause an error

**Messages:**

- Add/get messages with automatic ItemDefinition linking (shared __messageItemDefinition)
- Message itemRef updated when it gets a Correlation with a Data Type
- Rename messages (updates name + linked ItemDefinition ID + all event references)
- Delete messages (removes from definitions + cleans event references + removes shared ItemDefinition if last message)

**ItemDefinitions:**

- Add/get ItemDefinitions by data type (reuses existing, creates new if needed)
- Delete ItemDefinitions (removes from definitions + cleans all references in properties, I/O specs, messages, correlation properties)
- Auto-cleanup unused default-type ItemDefinitions (String, Boolean, Float, Integer, Object)

**Correlations:**

- Delete orphaned correlation subscriptions (removes subscriptions with invalid correlation keys or property bindings)
- Add required dataPath elements in the CorrelationSubscription: `language="java" `and `evaluatesToTypeRef="[IDREF-to-an-itemDefinition]`"


thiagoelg:
Thanks @kbowers-ibm!

I tried to recreate [this BPMN](https://gist.github.com/thiagoelg/0df718158a375bace8292ca703bdca8c#file-travelers_old-bpmn), but it's failing with:
```
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:3.27.2:build (default) on project process-event-driven-travelers: Failed to build quarkus application: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
[ERROR] 	[error]: Build step org.kie.kogito.quarkus.common.deployment.KogitoAssetsProcessor#generateModel threw an exception: org.kie.memorycompiler.KieMemoryCompilerException: [org/acme/travel/TravelersMessageConsumer__403851DA_EFA3_4FF2_8EED_2AB05ABA2581.java (51:31) : incompatible types: java.lang.Object cannot be converted to org.acme.travel.Traveler]
[ERROR] 	at org.kie.memorycompiler.KieMemoryCompiler.compileNoLoad(KieMemoryCompiler.java:140)
[ERROR] 	at org.kie.memorycompiler.KieMemoryCompiler.compileNoLoad(KieMemoryCompiler.java:108)
[ERROR] 	at org.drools.quarkus.util.deployment.DroolsQuarkusResourceUtils.compileGeneratedSources(DroolsQuarkusResourceUtils.java:148)
[ERROR] 	at org.kie.kogito.quarkus.common.deployment.KogitoAssetsProcessor.createGeneratedBeanBuildItemsFromJavaSources(KogitoAssetsProcessor.java:309)
[ERROR] 	at org.kie.kogito.quarkus.common.deployment.KogitoAssetsProcessor.generateModel(KogitoAssetsProcessor.java:189)
[ERROR] 	at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:732)
[ERROR] 	at io.quarkus.deployment.ExtensionLoader$3.execute(ExtensionLoader.java:874)
[ERROR] 	at io.quarkus.builder.BuildContext.run(BuildContext.java:255)
[ERROR] 	at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2651)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2630)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1622)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1589)
[ERROR] 	at java.base/java.lang.Thread.run(Thread.java:840)
[ERROR] 	at org.jboss.threads.JBossThread.run(JBossThread.java:501)
```

Here's the BPMN created with the new editor: https://gist.github.com/thiagoelg/0df718158a375bace8292ca703bdca8c#file-travelers_new-bpmn

Edit: I was trying to recreate the travelers.bpmn file from the [process-event-driven example](https://github.com/apache/incubator-kie-tools/tree/main/examples/process-event-driven).

klbowers:
Hi @thiagoelg 
Thanks for the review! 

The problem was that:
BPMN messages were not linked to their ItemDefinitions, as all messages defaulted to the generic __messageItemDefinition with structureRef="Object" instead of referencing specific data types (e.g., Traveler, ProcessedTraveler), causing compilation failures.

I've implemented a solution that links messages to ItemDefinition at three levels:

1. Message Creation/Selection (addOrGetMessages.ts)

- Added optional dataType parameter
- Updates message's @_itemRef to match the correct ItemDefinition when creating or retrieving messages

2. UI Filtering (MessageSelector.tsx, EventDefinitionProperties.tsx)

- MessageSelector filters dropdown to show only type-compatible messages
- EventDefinitionProperties extracts data type from event's dataInput/dataOutput and passes it to message operations

3. Data Mapping Sync (_dataMapping.ts)

- Added updateMessageItemRefForMessageEvent() helper
- Automatically updates message's itemRef when data mappings are added/changed
- Handles scenario where message is assigned before data mappings exist

Result: Messages now automatically link to correct ItemDefinitions whether assigned during message selection or when data mappings are added later

thiagoelg (approved):
Thanks @kbowers-ibm 

I was able to write the travelers.bpmn from scratch and test it with the process-event-driven example. 🎉

jomarko (commented):
Thank you for the PR @kbowers-ibm 

### [bug] BPMN Editor: Message start event with a message is not generating an itemDefinition

I think this is really fixed 🟢 

### [bug] BPMN Editor: Correlation Key and Subscription not cleaned up after deletion
This is advanced feature of BPMN. I am not sure if there is any good documentation of it. I tried to open mentioned travelers.bpmn new editor to somehow understand it, but it seems new editor does not load that file in a way, that would display all data. For example properties panel does not show message for message events. 

Anyway, I have created this file using the new editor.

[a.bpmn.txt](https://github.com/user-attachments/files/28088810/a.bpmn.txt)

It is using correlations. When I delete all correlation stuff, there remains `collaboration` 🟠 

### [bug] BPMN Editor: Message Start Event-itemDefinition Not Assigning Value to structureRef

I can confirm `strucutreRef` is set in 
```
  <itemDefinition id="__messageItemDefinition" structureRef="Object" />
```

however again comparing with `travelers` there were assignment to `structureRef` different from `Object`, for example:
```
  <bpmn2:itemDefinition id="_messageItem" structureRef="org.acme.message.Message" />
```

is it expected **new** editor set all message structure ref to Object? 🟠

jomarko:
minor thing

image

klbowers:
Hi @jomarko Thank you very much for the review! 

[bug] BPMN Editor: Correlation Key and Subscription not cleaned up after deletion


> It seems new editor does not load that file in a way, that would display all data. For example properties panel does not show message for message events.

Yeah that's a problem we've had elsewhere, I'll look into the exact fields that aren't displaying and create a separate ticket for it, thank you! 

> It is using correlations. When I delete all correlation stuff, there remains collaboration 🟠

That's intentional, the old bpmn editor always has a collaboration+participant so I wanted to keep the functionality the same.

[bug] BPMN Editor: Message Start Event-itemDefinition Not Assigning Value to structureRef

> is it expected new editor set all message structure ref to Object? 🟠

Yes we decided to create a default value of Object for all itemDefinitions, unless specified otherwise.
And we've also created a generic itemDefinition that relates to all messages, which has the value:
`  <itemDefinition id="__messageItemDefinition" structureRef="Object" /> 
`
That way we don't clutter the xml with too many itemDefinitions (as before each message had a corresponding itemDefinition)

I'll create a ticket for the long Text issue as well, and look into where else that issue is present.

Thank you for being so thorough! :)

ljmotta (commented):
@kbowers-ibm Thanks for addressing those bugs. I had two problems while reproducing the `hotel.bpmn` from the `process-event-driven` example.

1. Create the correlation first, update the process id later. I've started replicating the same file with a different `processId`. After I've renamed the `processId` to match the orignal file I've realized the `processRef` from the `participant` element wasn't updated accordingly, resulting in a build error.
2. ~~Create the `message` first, use it on the correlation later. Creating the message on the Event node, and then using it to create the correlation will not properly set the message type reference resulting on a build error.~~
2. Creating a data output on the `message` node will update the `message` `itemRef`, overriding the `correlation` `itemRef`.

I don't think 1 is related to this PR, but 2 definetely is. The `message` `itemDefinition` is not being updated accordingly.

packages/bpmn-editor/src/mutations/_dataMapping.ts
```diff
@@ -225,15 +225,34 @@ import { addOrGetProcessAndDiagramElements } from "./addOrGetProcessAndDiagramEl
 import { visitFlowElementsAndArtifacts } from "./_elementVisitor";
 import { DEFAULT_DATA_TYPES } from "./addOrGetItemDefinitions";
 
+function updateMessageItemRefForMessageEvent(
+  definitions: Normalized<BPMN20__tDefinitions>,
+  event: any,
```
ljmotta: Can't we use a specific type here to avoid `any` being propagated?

packages/bpmn-editor/src/mutations/_dataMapping.ts
```diff
@@ -225,15 +225,34 @@ import { addOrGetProcessAndDiagramElements } from "./addOrGetProcessAndDiagramEl
...
+  itemSubjectRef: string
+) {
+  const messageEventDef = event.eventDefinition?.find((ed: any) => ed.__$$element === "messageEventDefinition");
+  if (!messageEventDef?.["@_messageRef"]) return;
+
+  const messages = definitions.rootElement?.filter((e) => e.__$$element === "message") ?? [];
+  const message = messages.find((m) => m["@_id"] === messageEventDef["@_messageRef"]);
```
ljmotta:
```suggestion
  const message = definitions.rootElement?.find((e) => e.__$$element === "message" && e["@_id" === messageEventDef["@_messageRef"]) ?? [];
```

packages/bpmn-editor/src/mutations/deleteOrphanedCorrelationSubscriptions.ts
```diff
@@ -0,0 +1,84 @@
+/*
+ * IBM Confidential
+ * PID 5900-AR4
+ * Copyright IBM Corp. 2026
+ */
+/*
```
ljmotta: leftover?

ljmotta (commented):
While reviewing the code, I identified a couple of issues:
- It is currently possible to create multiple output mappings on a Message event, but only the last one is actually applied to the itemRef.
- The default type `__messageItemDefinition` is not cleared after defining output mappings, which results in an inconsistent state.

I don’t think we should move forward with the `__messageItemDefinition` approach.
If we need to define a type for a message event, a better alternative would be to introduce a dedicated type field—similar to what we already do for variables. The current approach of auto-generating `__messageItemDefinition` seems to only ensure that the BPMN compiles, rather than correctly handling the `itemRef`.
Instead, we should provide a clear and explicit way for users to define this value. If such a definition is implicitly derived (e.g., from Output Mapping or Correlation), the build should fail rather than silently auto-filling it.

There are still some unresolved cases that we need to clarify:
 - What should the `itemRef` be when a message event contains multiple output mappings?
 - What should the `itemRef` be when both correlation and output mappings are defined?

Before proceeding, we should define clear rules for how `itemRef` is derived and validated in these scenarios. Otherwise, we risk introducing ambiguous or inconsistent behavior that will be difficult to maintain.

---
I need more time to review the code, I've added just a few comments inline

klbowers:
Hi @ljmotta 
Thanks for the feedback!

> Create the correlation first, update the process id later. I've started replicating the same file with a different processId. After I've renamed the processId to match the orignal file I've realized the processRef from the participant element wasn't updated accordingly, resulting in a build error.

I've fixed this and will create a ticket for it shortly.

> Creating a data output on the message node will update the message itemRef, overriding the correlation itemRef.

This was a mistake, the message itemRef has nothing to do with correlations and is solely controlled by the data mapping. I've removed the all code relating to changing the messages itemRef through correlations.

> It is currently possible to create multiple output mappings on a Message event, but only the last one is actually applied to the itemRef.

It's now only possible to create one data mapping on messages nodes, following the functionality of the old bpmn editor.

> What should the itemRef be when a message event contains multiple output mappings?

No longer possible 

> What should the itemRef be when both correlation and output mappings are defined?

Only output mappings determines the itemRef

klbowers:
Regarding __messageItemDefinition, its primary purpose is to reduce XML bloat.

In the old editor, every message gets its own itemDefinition using the format ${messageName}Type. That itemDefinition is then linked to a data type via the structureRef property whenever the message's data mapping is changed.

In the new editor, all messages are assigned a shared messageItemDefinition by default. When a message's data mapping is updated, the message is linked directly to the appropriate itemDefinition instead which has the appropriate type in it's structureRef already.

This approach is more consistent with the BPMN specification, since messages should reference the ID of the target itemDefinition, which is expected to be unique. In contrast, the ${messageName}Type pattern used by the old editor does not guarantee uniqueness and can result in unnecessary duplication of itemDefinition elements.

packages/bpmn-editor/src/mutations/addOrGetItemDefinitions.ts
```diff
@@ -34,16 +35,20 @@ export const DEFAULT_DATA_TYPES = {
 export function addOrGetItemDefinitions({
   definitions,
   dataType,
+  id,
```
ljmotta, line 38: I believe this is a leftover. I couldn't see where it is used.

packages/bpmn-editor/src/propertiesPanel/correlations/Correlations.tsx
ljmotta, line 1: This file contain many changes to correlation elements outside of a mutation. We should refactor some methods, as changes to the definitions should happen on a mutation.

packages/bpmn-editor/src/propertiesPanel/itemDefinitionRefSelector/ItemDefinitionRefSelector.tsx
```diff
@@ -132,15 +134,25 @@ export function ItemDefinitionRefSelector({
...
-    />
+    <div
+      style={{
+        ...(isDisabled && {
+          backgroundColor: "var(--pf-global--disabled-color--300)",
+          borderRadius: "3px",
+          opacity: 0.6,
+        }),
+      }}
+    >
```
ljmotta, line 145: This could be moved to the `css` file.

packages/bpmn-editor/src/propertiesPanel/GlobalProperties.tsx
```diff
@@ -259,7 +259,19 @@ export function GlobalProperties() {
...
+                          (e) => e.__$$element === "collaboration"
+                        );
+                        if (collaboration && collaboration.__$$element === "collaboration") {
+                          const participant = collaboration.participant?.find(
+                            (p) => p["@_processRef"] === oldProcessId
+                          );
+                          if (participant) {
+                            participant["@_processRef"] = newId;
+                          }
+                        }
```
ljmotta, line 274: This entire change should be on a mutation, including the `process["@_id"] = newId;`

packages/bpmn-editor/src/propertiesPanel/propertiesManager/PropertiesManager.tsx
```diff
@@ -303,6 +311,73 @@ export function PropertiesManager({ p }: { p: undefined | WithVariables }) {
                         }}
                       />
                     </GridItem>
+                    <GridItem span={5}>
+                      <ItemDefinitionRefSelector
+                        value={entry["@_itemRef"]}
+                        onChange={(newItemDefinitionRef) => {
+                          bpmnEditorStoreApi.setState((s) => {
```
ljmotta, line 318: As my previous comments, this should be on a mutation

ljmotta (approved):
@kbowers-ibm Thanks for addressing the previous feedback. I've manually tested and it is looking good.
While reviewing the PR, I noticed that some updates to the definitions are happening outside of a mutation, which breaks editor isolation. Additionally, the Correlation.tsx file is not following the mutation pattern and would benefit from refactoring.
I’m approving this PR for now, but please open a follow-up PR to address these points.

thiagoelg:
Merging as this PR fixes some important issues with the BPMN Editor. A new follow-up PR should be opened to address @ljmotta comments.

Thank you @kbowers-ibm!

PR #3556, opened 2026-04-28
kie-issues#2289: Automated E2E Test Suite for BPMN Editor

### LINKED ISSUE #2289: Automated E2E Test Suite for BPMN Editor
**Overview**

Kickoff an automated test suite for the BPMN Editor using Storybook and Playwright, ensuring consistent quality and faster feedback cycles. The DMN Editor serves as inspiration for structure, tooling, and best practices.

**Tests**

**Node Operations**
- Move, rename, resize, copy/paste, morph nodes
- Multi-select, undo/redo, keyboard shortcuts

**Boundary Events**
- Attach/detach boundary events to Tasks and Sub-Processes
- All event types (Timer, Error, Message, Signal, Escalation, Compensation, Cancel, Conditional)ts

rajalakshmys-27:
Closes : [kie-issues#2289](https://github.com/apache/incubator-kie-issues/issues/2289)

Introduces a comprehensive E2E test suite for the BPMN Editor using Playwright and Storybook,
inspired by the DMN Editor's testing architecture.

### Flow Elements
- Start Events — None, Message, Timer, Signal, Conditional, Error, Escalation, Compensation
- Intermediate Events — Catch/Throw variants with Message, Timer, Signal, Error, Escalation, Link, Compensation, Conditional
- End Events — None, Message, Error, Escalation, Signal, Compensation, Terminate
- Tasks — Generic, User, Service, Script, Business Rule with morphing
- Gateways — Exclusive, Parallel, Inclusive, Event-Based, Complex with morphing
- Sub-Processes — Regular, Event, Multi-Instance (Parallel/Sequential), Ad-Hoc
- Call Activities
- Custom Tasks — REST API call, gRPC API call
- Data Objects — Creation, naming, type configuration
- Structural Elements — Lanes, Groups, Text Annotations

### Properties Panel
- Name and documentation editing for all elements
- Event definitions (Timer, Message, Signal, Error, Escalation)
- Task-specific properties (Actors, Groups, Implementation, Scripts)
- Gateway default flow configuration
- Sequence flow conditions and priorities
- Multi-instance configuration (Parallel/Sequential)
- Data I/O mappings
- Sub-process Ad-Hoc settings

### Node Operations
- Drag & drop from palette to canvas
- Connected node creation with auto-connection
- Morphing — type transformation (e.g., Task → User Task, Gateway types)
- Rename — inline and via properties panel
- Delete — single and multiple
- Sequence flow creation
- Nested elements inside Sub-Processes
### Prerequisites
- Docker installed and running
- Repository cloned and dependencies installed 

### Running Tests Locally

**Option 1: Run tests on your local machine (fastest)**
```bash
cd packages/bpmn-editor
pnpm test-e2e:run
```

**Option 2: Run tests in Docker container (matches CI environment)**
```bash
cd packages/bpmn-editor
pnpm test-e2e:container:run
```

**Option 3: Interactive container shell (for debugging)**
```bash
cd packages/bpmn-editor
pnpm test-e2e:container:shell
# Inside container:
pnpm test-e2e:run
```

**To open last HTML report run:**
```bash
pnpm exec playwright show-report dist-tests-e2e/reports
```

**To clean up Docker containers:**
```bash
pnpm test-e2e:container:clean
```

- 2 tests are intentionally skipped (known bugs, tracked separately)


packages/bpmn-editor/tests-e2e/__screenshots__/chromium/flowElements/create-sequence-flow-intermediate-catch-event-to-end-event.png
jomarko, line 1: what is the intention of this file please? it is named as `create-sequence-flow` but actually in the screenshot there is no sequence flow

I think I have same question also for other screenshots in this shape, where filename suggest an existence of the connection between nodes but actually there is no visible connection between nodes.


Thank you for a clarification

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/flowElements/rest-api-call-task-in-process-flow.png
jomarko, line 1: is this diagram in an expected shape?

jomarko (commented):
Thank you this  important valuable PR @rajalakshmys-27 .

All my feedback is related to the screenshots testing.

1.  I put some questions inline
2. the build time is increased to 113 minutes comparing 20 minutes in this PR https://github.com/apache/incubator-kie-tools/pull/3555, it seems as too big jump in time consumption, what is the build time on your local machine? asking because current test suite runs non trivial amount of DMN tests already
3. The failing tests often mention screenshot comparison failed, I am not sure if we should use screenshot testing for all current scenarios, especially tests that has in screenshot only diagram without properties or other panel and single node should be testable just using some check of existence in html?

jomarko (commented):
Thank you this  important valuable PR @rajalakshmys-27 .

All my feedback is related to the screenshots testing.

1.  I put some questions inline
2. the build time is increased to 113 minutes comparing 20 minutes in this PR https://github.com/apache/incubator-kie-tools/pull/3555, it seems as too big jump in time consumption, what is the build time on your local machine? asking because current test suite runs non trivial amount of DMN tests already
3. The failing tests often mention screenshot comparison failed, I am not sure if we should use screenshot testing for all current scenarios, especially tests that has in screenshot only diagram without properties or other panel and single node should be testable just using some check of existence in html?

rajalakshmys-27:
@jomarko Thanks for the feedback! I’ve already started looking into the screenshot test failures and I’m working on fixing them. I’ll update the PR once those are resolved. Regarding the build time increase, I’m checking on it

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/flowElements/create-sequence-flow-intermediate-catch-event-to-end-event.png
rajalakshmys-27: Thank you for pointing this out! I have fixed this.

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/flowElements/rest-api-call-task-in-process-flow.png
rajalakshmys-27: Thank you for pointing this out! I have fixed this.

rajalakshmys-27:
Here's the current state:

- **Screenshots reduced by 50–60%.** Refactored tests doing simple existence checks to use direct DOM or JSON model verification. Screenshots are kept only where visual regression detection adds real value.
- **BPMN test suite (747 tests) runtime reduced by ~38%**, now taking ~20min of the 176min total CI pipeline.
- Of the **176min** total, ~**90min** is consumed by the pre-existing DMN Editor E2E suite alone. The remaining time is taken by the other test suites, build infrastructure setup, and CI environment overhead

packages/bpmn-editor/tests-e2e/__fixtures__/edges.ts
```diff
@@ -0,0 +1,120 @@
...
+    const waypoint = edge.locator(`[data-waypointindex="${args.waypointIndex}"]`);
+    await waypoint.dragTo(this.diagram.get(), { targetPosition: args.targetPosition });
+  }
+
+  public async deleteWaypoint(args: { from: string; to: string; waypointIndex: number }) {
+    const edge = await this.get(args);
+    const waypoint = edge.locator(`[data-waypointindex="${args.waypointIndex}"]`);
+    await waypoint.click({ button: "right" });
+    await this.page.getByText("Delete waypoint").click();
+  }
```
jomarko: Do we have tests where we test behavior of multiple waypoints on the edge? asking because if most tests use the `waypointIndex` equal to `0` maybe we could have a default value here? and then invocation in actual tests would be shorter and more readbale?

packages/bpmn-editor/tests-e2e/__fixtures__/editor.ts
```diff
@@ -0,0 +1,59 @@
...
+      await this.page.getByRole("button", { name: "Start Modeling" }).click();
+      await this.page.locator(".react-flow").waitFor({ state: "visible" });
+    }
+  }
+
+  public async openCustomTasks({ nodes }: { nodes: Nodes }) {
+    await this.page.goto(`${this.baseURL}/iframe.html?args=&id=features-customtasks--custom-tasks&viewMode=story`);
+    await nodes.delete({ name: "Rest API call Task" });
+    await nodes.delete({ name: "gRPC API call Task" });
+  }
```
jomarko, line 54: why do we delete some nodes as part of open custom tasks operation? it sounds as unexpected, maybe is just my not enough knowledge, could you please clarify?

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/end-event-message.png
jomarko, line 1: I think similar test should set `Message` in properties panel, or am I missing something, do we have such test elsewhere?

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/start-event-message.png
jomarko, line 1: similar question about `Message` in properties panel

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/start-event-signal.png
jomarko, line 1: Similar question about `Signal` in properties panel

packages/bpmn-editor/tests-e2e/propertiesPanel/changeSubProcessProperties.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    await expect(multiInstanceOption).toBeVisible({ timeout: 5000 });
+    await multiInstanceOption.click({ force: true });
+  });
+
+  test("should configure Sub-Process multi-instance parallel", async ({ subProcessPropertiesPanel, page }) => {
+    await subProcessPropertiesPanel.setMultiInstance({ type: "parallel" });
+    await subProcessPropertiesPanel.setCollectionExpression({ expression: "${orderItems}" });
+
+    await expect(page.locator(".kie-bpmn-editor--root")).toHaveScreenshot("subprocess-multi-instance-parallel.png");
+  });
```
jomarko, line 72: is the editor working for this scenario? the screenshot show a message `No matches` however I do not see in screenshot anything like `Create expression '${orderItems}'. Are expressions supported as collection input in BPMN editor?

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/subprocess-multi-instance-sequential.png
jomarko, line 1: similar question here as for multiinstance collection input, is the `no matches` expected?

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/user-task-full-configuration.png
jomarko, line 1: is the full configuration proper name here? I see in properties panel not all fields are set

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/gatewayPropertiesPanel.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const morphingToggle = selectedGateway.locator(".kie-bpmn-editor--node-morphing-panel-toggle > div");
+    await morphingToggle.waitFor({ state: "visible", timeout: 5000 });
+    await morphingToggle.click({ force: true });
+
+    const morphingPanel = this.page.locator(".kie-bpmn-editor--node-morphing-panel");
+    await morphingPanel.waitFor({ state: "visible", timeout: 5000 });
+
+    return morphingPanel;
+  }
```
jomarko: When I search for occurrences of
`page.locator(".kie-bpmn-editor--node-morphing-panel");`  
I get 14 matches. Based on my understanding after reading the code, we currently have 14 different places where node morphing is handled. However, most of these implementations are very similar—if not direct copies of each other.

### Concerns

The issues I’m concerned about are:

- **Code maintenance**  
  If the morphing implementation changes, maintaining and updating the codebase will be difficult.

- **Inconsistent locations**  
  In some cases, morphing is implemented directly in the test body, while in others it’s handled via a fixture.

- **Misleading fixture naming**  
  The `propertiesPanel` fixture name suggests actions related to the properties panel, but node morphing is actually an action performed on the diagram canvas.

### Why this matters

These points are especially important for future test contributors. Someone contributing to the codebase without prior knowledge may encounter 14 separate, non-unified implementations of the same morphing logic, which could be quite confusing.

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+      expect(
+        process.flowElement?.find((e: { __$$element: string }) => e.__$$element === "boundaryEvent")
+      ).toBeUndefined();
+      expect(process.flowElement?.find((e: { __$$element: string }) => e.__$$element === "task")).toBeDefined();
+
+      await expect(boundaryEventNode).not.toBeAttached();
+    });
+
+    test.skip("should delete task with boundary event", async ({ palette, nodes, jsonModel, diagram }) => {
+      // TODO: Enable when boundary event deletion logic is implemented
```
jomarko: I think we should refer to some kie-issue to track this properly? is there a kie-issue to implement such logic?

packages/bpmn-editor/package.json
```diff
@@ -16,9 +16,13 @@
...
+    "test-e2e:container:clean": "playwright-base-container clean",
+    "test-e2e:container:run": "start-server-and-test 'pnpm start' http://localhost:$(build-env bpmnEditor.storybook.port)/iframe.html \"playwright-base-container run --additional-env=KIE_TOOLS_PLAYWRIGHT_CONTAINER__PORT=$(build-env bpmnEditor.storybook.port) --container-workdir=incubator-kie-tools/packages/bpmn-editor --container-name=kie-tools-playwright-containerization-bpmn-editor -- $@\"",
+    "test-e2e:container:shell": "start-server-and-test 'pnpm start' http://localhost:$(build-env bpmnEditor.storybook.port)/iframe.html 'playwright-base-container shell --additional-env=KIE_TOOLS_PLAYWRIGHT_CONTAINER__PORT=$(build-env bpmnEditor.storybook.port) --container-workdir=incubator-kie-tools/packages/bpmn-editor --container-name=kie-tools-playwright-containerization-bpmn-editor'",
     "test-e2e:open": "pnpm exec playwright show-report dist-tests-e2e/reports",
-    "test-e2e:run": "pnpm exec playwright test --pass-with-no-tests"
+    "test-e2e:run": "pnpm exec playwright test --update-snapshots"
```
jomarko: I am not sure about this change, to me it sounds as it will always update screenshot so the test will never start to fail?

packages/bpmn-editor/tests-e2e/flowElements/boundaryEventCompensation.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+        "@_cancelActivity": false,
+        eventDefinition: [{ __$$element: "compensateEventDefinition" }],
+      });
+
+    const cancelActivity = await intermediateEventPropertiesPanel.getCancelActivity();
+    expect(cancelActivity).toBe(false);
+  });
+
+  test.skip("should not allow incoming sequence flows to compensation boundary event", async ({
+    // TODO: Enable when compensation boundary event validation is implemented
```
jomarko: do we have an existing kie-issue ticket?

jomarko (commented):
Thank you for the changes @rajalakshmys-27 . I really appreciate addition of the tests. I put some comments inline.

I am not sure about the exection time, especially:
> ~90min is consumed by the pre-existing DMN Editor E2E suite alone

Why is that this PR has 90 minutes of DMN test suite execution and other BPMN Editor PRs do not have such execution time, for example https://github.com/apache/incubator-kie-tools/pull/3555

packages/bpmn-editor/tests-e2e/__fixtures__/editor.ts
rajalakshmys-27: This is because by default the BPMN storybook adds Rest API call task, gRPC API call task.

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
rajalakshmys-27: I’ve already raised this issue and I’ll take action soon.

packages/bpmn-editor/tests-e2e/flowElements/boundaryEventCompensation.spec.ts
rajalakshmys-27: I’ve already raised this issue and I’ll take action soon.

packages/bpmn-editor/tests-e2e/__fixtures__/edges.ts
rajalakshmys-27: yes we do have tests for multiple waypoints! Checked all usages across the codebase, no test uses `waypointIndex: 0`, all 18 calls use `1` (14 calls) or `2` (4 calls). This is actually by design: the `Waypoints` component renders `slice(1, -1)` and starts indices at `1`, so index `0` doesn't exist in the DOM at all. Defaulting to `0` would silently target nothing, so keeping `waypointIndex` required is safer and clearer.

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/end-event-message.png
rajalakshmys-27: Thanks for pointing it out! You're right, I've updated the tests to cover that.

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/start-event-message.png
rajalakshmys-27: Thanks for pointing it out! You're right, I've updated the tests to cover that.

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/start-event-signal.png
rajalakshmys-27: Thanks for pointing it out! You're right, I've updated the tests to cover that.

packages/bpmn-editor/tests-e2e/propertiesPanel/changeSubProcessProperties.spec.ts
rajalakshmys-27: Thanks for pointing it out! Fixed.

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/subprocess-multi-instance-sequential.png
rajalakshmys-27: fixed

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/user-task-full-configuration.png
rajalakshmys-27: image
I can see all the three configurations set properly!  actors, groups, taskName

packages/bpmn-editor/package.json
rajalakshmys-27: thanks for pointing that out! I forgot to remove that

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/gatewayPropertiesPanel.ts
rajalakshmys-27: Thanks for the detailed feedback! You're right to flag this. After reviewing, node morphing is actually already centralised in the Nodes fixture (nodes.ts). Refactored the code

jomarko: I think I can not confirm this update, I see we have morphing still in `propertiesPanel`:
- EndEventPropertiesPanel 
- StartEventPropertiesPanel

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/task-multi-instance-parallel.png
jomarko, line 1: I have commented on this also in my previous review round. My point is the current screenshot show:
- user typed and expression in format `${variableName}`
- editor say there is no match in available variables

While I understand, what editor says is correct, variable with such name doesn't exist, as user, the UI flash box text **No matches** indicates to me some problem and some doubts, if my operation was saved or if it is valid.

Could we somehow clarify if the current editor behavior is by design and is intended to work in this way? I think on other places in the editor we display flash message to **Create expression** if a new value is spot and is not available in existing.

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/task-multi-instance-sequential.png
jomarko, line 1: same question as above

jomarko (commented):
Thank you for the updates

packages/bpmn-editor/tests-e2e/__screenshots__/chromium/propertiesPanel/task-multi-instance-parallel.png
rajalakshmys-27: Sorry for missing this in the previous round! This field is a variable selector, not an expression field (`allowExpressions={false}`), so `${orderItems}` is not the intended input here `orderItems` is. The "No matches" message is expected behavior. I've updated the test to define process variables upfront and use plain variable names for the collection input, which should make the screenshot clearer. The completion-condition field is separate and still supports free-form expressions.

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/gatewayPropertiesPanel.ts
rajalakshmys-27: thank you Jozef! I had missed these. Updated

jomarko (approved):
Thank you for all the updates. 

I am approving the PR, however still we still need another engineer to do a review.

I am still confused about the build time, but clearly it is somehow dependent on particular PR. When I checked other PRs where we do just an 'regular' update of code, not addition of hundreds of tests, even there build times are not same
- https://github.com/apache/incubator-kie-tools/pull/3557 **30 minutes**
- https://github.com/apache/incubator-kie-tools/pull/3560 **162 minutes**
- https://github.com/apache/incubator-kie-tools/pull/3555 **20 minutes**

rajalakshmys-27:
Thank you @jomarko for your review! 

Regarding build time variability, that's expected. CI uses partial builds, only building changed packages and their dependents, so scope varies per PR.

For this PR, 57m is explained by the CI log breakdown:
- Setup environment: 10m 48s
- Build upstream (PARTIAL): 15m 0s
- Build changed + downstream (PARTIAL): 26m 33s
- Bootstrap + artifacts + misc: ~5m

BMPN editor changes trigger 11 dependent packages (`@kie-tools/bpmn-editor` with 24 e2e test files, `bpmn-editor-envelope`, `bpmn-editor-standalone`, `bpmn-editor-on-webapp`, `online-editor`, `bpmn-vscode-extension`, `chrome-extension-pack-kogito-kie-editors`, `kie-editors-dev-vscode-extension`, `kie-sandbox-webapp-image`, `kie-sandbox-distribution`, `kie-sandbox-helm-chart`), each with full linting and TypeScript compilation.

thiagoelg:
> I am still confused about the build time, but clearly it is somehow dependent on particular PR. When I checked other PRs where we do just an 'regular' update of code, not addition of hundreds of tests, even there build times are not same

@jomarko That's normal. Our PR checks will scan the code changes to identify which packages are affected, then build their dependencies and dependents.

https://github.com/apache/incubator-kie-tools/pull/3557 30 minutes - No BPMN Editor tests yet, so the build was fast.
https://github.com/apache/incubator-kie-tools/pull/3560 162 minutes - Changes that affect the DMN Editor, so it forced the run of all e2e tests for the DMN Editor.
https://github.com/apache/incubator-kie-tools/pull/3555 20 minutes - Same as the first one, but with fewer affected dependents.

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/dataObjectPropertiesPanel.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public async getDocumentation(): Promise<string> {
+    return await this.documentationProperties.getDocumentation();
+  }
+
+  public async setItemSubjectRef(args: { itemSubjectRef: string }) {
+    const dataTypeInput = this.panel().locator('input[role="combobox"]').first();
+
+    await dataTypeInput.click();
```
ljmotta: Please, use the recommeded Playwright Locator API. For this case `getByRole("checkbox", { name: "<optional name>" })`

This could be possible inlined: `this.panel().getByRole("checkbox").first().click();`

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/dataObjectPropertiesPanel.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (await createOption.isVisible().catch(() => false)) {
+      await createOption.click();
+    } else {
+      await this.page.getByRole("option", { name: args.itemSubjectRef, exact: true }).click();
+    }
+  }
+
+  public async getItemSubjectRef(): Promise<string> {
+    const dataTypeInput = this.panel().locator('input[role="combobox"]').first();
+    return (await dataTypeInput.inputValue()).trim();
```
ljmotta: Same here.

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/endEventPropertiesPanel.ts
```diff
@@ -0,0 +1,194 @@
...
+  public async setTerminateDefinition(args: { endEventLocator: Locator }) {
+    await this.morphToEventType({ endEventLocator: args.endEventLocator, eventType: "Terminate" });
+  }
+
+  public async setMessageDefinition(args: { messageName: string; endEventLocator: Locator }) {
+    await this.morphToEventType({ endEventLocator: args.endEventLocator, eventType: "Message" });
+
+    const messageInput = this.panel().locator('input[role="combobox"]').first();
+    await messageInput.waitFor({ state: "visible", timeout: 10000 });
+    await messageInput.click();
```
ljmotta: Why are you using `waitFor` instead of relying on Playwright auto-wait for this case?

You could possibly inline this too.

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/intermediateEventPropertiesPanel.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  public async getDocumentation(): Promise<string> {
+    return await this.documentationProperties.getDocumentation();
+  }
+
+  public async selectEventDefinition(args: { eventType: string }) {
+    const selectedNode = this.page
+      .locator(".kie-bpmn-editor--intermediate-catch-event-node, .kie-bpmn-editor--intermediate-throw-event-node")
+      .first();
```
ljmotta: Do not rely on CSS classes for selecting elements. They are pretty volatile and changing then shouldn't break any tests. Playwright test should rely on UI. CSS selectors should be used only on last case scenario.
If any other locator is not available, we can introduce the `testid`.

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/intermediateEventPropertiesPanel.ts
```diff
@@ -0,0 +1,245 @@
...
+    await this.nodes.morphNode({
+      nodeLocator: selectedNode,
+      targetMorphType: args.eventType,
+    });
+  }
+
+  public async setTimerDefinition(args: { type: "date" | "duration" | "cycle"; value: string }) {
+    if (args.type === "duration") {
+      await this.panel().getByLabel("Fire once after duration").click();
+      const valueInput = this.panel().locator("#fire-once-input");
```
ljmotta: We shouldn't test using Element ID. Same as above. If any other locartor is not available, we can introduce `testid`.

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+
+test.describe("Add Boundary Event", () => {
+  test.describe("Basic attachment", () => {
+    test("should attach intermediate catch event to task", async ({ palette, jsonModel, page }) => {
+      await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 300 } });
+      await palette.dragNewNode({ type: NodeType.INTERMEDIATE_CATCH_EVENT, targetPosition: { x: 450, y: 300 } });
+
+      const boundaryEvent = await jsonModel.getFlowElement({ elementIndex: 1 });
+      expect(boundaryEvent.__$$element).toBe("boundaryEvent");
+      expect(boundaryEvent["@_attachedToRef"]).toBeDefined();
```
ljmotta, line 35: Please, move all `jsonModel` assertions to the end of the tests.

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+
+      const boundaryEventNode = page.locator(".kie-bpmn-editor--intermediate-catch-event-node").first();
```
ljmotta: Shouldn't we have a `node` fixture to retrieve the nodes? Also, you're using CSS classes, and this could be inlined, we don't need to create unecessary variables.

ljmotta (changes requested):
@rajalakshmys-27 Thanks for this PR. You have created many test cases! I still didn't review many, but I could see some things that must be addresed.

Please take a look into my comments and fix all occurences in the PR, I didn't comment in all of them. Also, please take a look into the Playwright Locators API documentation, it can be useful. CSS/DOM locators must be the last resort and `waitFor` are usually unecessary.

https://playwright.dev/docs/locators

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/dataObjectPropertiesPanel.ts
rajalakshmys-27: Thanks! I'll update it to use `getByRole("checkbox")`

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/endEventPropertiesPanel.ts
rajalakshmys-27: thanks

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
rajalakshmys-27: resolved

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/intermediateEventPropertiesPanel.ts
rajalakshmys-27: Refactored all CSS selectors to getByRole, getByTestId, and data-field attributes.

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/intermediateEventPropertiesPanel.ts
rajalakshmys-27: Fixed

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
rajalakshmys-27: fixed

rajalakshmys-27:
Added data-testid attributes to BPMN editor source components following the kie-tools--bpmn-editor--{component-name} convention. Previously only 1 of 63+ interactive components had data-testid. Tests now prioritise semantic locators (getByRole, getByLabel, etc.) where possible, getByTestId for the rest, and removed unnecessary waitFor calls.

packages/bpmn-editor/tests-e2e/__fixtures__/nodes.ts
```diff
@@ -0,0 +1,393 @@
...
+    const morphingToggle = args.nodeLocator.getByRole("button", { name: /morph/i });
+    await expect(morphingToggle).toBeVisible({ timeout: 5000 });
+    await morphingToggle.click({ force: true });
+
+    const morphingPanel = this.page.getByTestId("kie-tools--bpmn-editor--morphing-panel");
+    await morphingPanel.waitFor({ state: "visible", timeout: 5000 });
+
+    const morphingOption = morphingPanel.getByTitle(args.targetMorphType, { exact });
+    await expect(morphingOption).toBeVisible({ timeout: 5000 });
+    await morphingOption.click({ force: true });
```
thiagoelg: Will this fail if the current Node is already of type `targetMorphType`? (e.g., morphing a User Task node to User Task again).

packages/bpmn-editor/tests-e2e/__fixtures__/nodes.ts
```diff
@@ -0,0 +1,393 @@
...
+    const box = await args.nodeLocator.boundingBox();
+    if (!box) {
+      throw new Error("Node not visible - cannot retrieve bounding box for morphing");
+    }
+
+    await this.page.mouse.move(box.x + box.width / 2, box.y + box.height / 2);
+    await this.page.waitForTimeout(hoverDelay);
+
+    const morphingToggle = args.nodeLocator.getByRole("button", { name: /morph/i });
+    await expect(morphingToggle).toBeVisible({ timeout: 5000 });
```
thiagoelg: I'm noticing several places where the `timeout` option is enforced to `5000`.
This can be achieved by adding this to the playwright.config.ts file:
```
expect: {
    timeout: 30000,
},
```
After adding this config, you can remove all `{ timeout: 5000 }` from the code.

Learn more [here](https://playwright.dev/docs/test-timeouts).

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/gatewayPropertiesPanel.ts
```diff
@@ -0,0 +1,66 @@
...
+    return await this.nameProperties.getName();
+  }
+
+  public async setDocumentation(args: { newDocumentation: string }) {
+    await this.documentationProperties.setDocumentation({ ...args });
+  }
+
+  public async getDocumentation(): Promise<string> {
+    return await this.documentationProperties.getDocumentation();
+  }
```
thiagoelg: (this is a nitpick)

Instead of creating wrapper functions, you can make the `nameProperties` and `documentationProperties` public, then tests can call them directly.

thiagoelg: Even better, since all `{nodeName}PropertiesPanel.ts` fixtures include the `nameProperties` and `documentationProperties`, you could add them to the `PropertiesPanelBase` class and constructor.

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/intermediateEventPropertiesPanel.ts
```diff
@@ -0,0 +1,241 @@
...
+    const createOption = this.page.getByText(`Create Signal "${args.signalName}"`, { exact: true });
+    if (await createOption.isVisible().catch(() => false)) {
+      await createOption.click();
+    } else {
+      await this.page.getByRole("option", { name: args.signalName, exact: true }).click();
+    }
+
+    if (args.scope) {
+      const scopeSelect = this.panel().locator("select").first();
+      await scopeSelect.waitFor({ state: "visible", timeout: 10000 });
```
thiagoelg: Can we stick to the default timeout (the one set in the playwright.config.ts)? Or is it really necessary to have different timeouts in these cases?

thiagoelg (changes requested):
Thanks @rajalakshmys-27, great work adding tests to our BPMN Editor!

This is a partial review, as I haven't finished looking at all the changes yet.

packages/bpmn-editor/tests-e2e/__fixtures__/nodes.ts
rajalakshmys-27: No, it won’t fail. In this editor the morphing panel still shows the current type as a clickable option, so the flow completes without error. With the safety checks I added, if the toggle or option isn’t visible the method exits gracefully, and if the option is present it’s just a no‑op click.

packages/bpmn-editor/tests-e2e/__fixtures__/nodes.ts
rajalakshmys-27: I’ve removed the explicit timeout

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/gatewayPropertiesPanel.ts
rajalakshmys-27: Thanks for the suggestion, I’ve updated the code so `nameProperties` and `documentationProperties` are now public in `PropertiesPanelBase`, removing the redundant wrappers.

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/intermediateEventPropertiesPanel.ts
rajalakshmys-27: Resolved!

packages/bpmn-editor/src/diagram/nodes/Nodes.tsx
```diff
@@ -987,6 +992,7 @@ export const SubProcessNode = React.memo(
         </svg>
         <PositionalNodeHandles isTargeted={isTargeted && isValidConnectionTarget} nodeId={id} />
         <div
+          data-testid="kie-tools--bpmn-editor--node-subprocess"
```
ljmotta:
```suggestion
          data-testid="kie-tools--bpmn-editor--node-sub-process"
```

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/sequenceFlowPropertiesPanel.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/intermediateEventPropertiesPanel.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/gatewayPropertiesPanel.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/endEventPropertiesPanel.ts
ljmotta, line 1: Review this file.
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/startEventPropertiesPanel.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/subProcessPropertiesPanel.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/propertiesPanel/taskPropertiesPanel.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/customTasks.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/edges.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/editor.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/jsonModel.ts
ljmotta, line 1: Review this file.
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/nodes.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/__fixtures__/palette.ts
```diff
@@ -0,0 +1,146 @@
...
+    public page: Page,
+    public diagram: Diagram,
+    public nodes: Nodes
+  ) {}
+
+  public async dragNewNode(args: { type: NodeType; targetPosition: { x: number; y: number }; thenRenameTo?: string }) {
+    const { title, nodeName } = this.getNewNodeProperties(args.type);
+
+    await this.page
+      .getByTestId("rf__wrapper")
```
ljmotta: We shouldn't rely on properties from other libraries.

packages/bpmn-editor/tests-e2e/__fixtures__/edges.ts
```diff
@@ -0,0 +1,120 @@
...
+    public nodes: Nodes,
+    public diagram: Diagram
+  ) {}
+
+  public async get(args: { from: string; to: string }): Promise<Locator> {
+    const fromId = args.from.startsWith("_") ? args.from : await this.nodes.getId({ name: args.from });
+
+    const toId = args.to.startsWith("_") ? args.to : await this.nodes.getId({ name: args.to });
+
+    await this.page.locator(".react-flow__edge").first().waitFor({ state: "attached" });
```
ljmotta: We shouldn't rely on anything from a 3rd party library.

packages/bpmn-editor/tests-e2e/__fixtures__/jsonModel.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Page } from "@playwright/test";
+
+export class JsonModel {
+  constructor(
+    public page: Page,
+    public baseURL?: string
+  ) {}
+
+  public async getModel(): Promise<any> {
+    const modelElement = this.page.locator('[data-testid="storybook--bpmn-editor-model"]');
```
ljmotta: Should use the `getByTestId`

packages/bpmn-editor/tests-e2e/__fixtures__/palette.ts
ljmotta, line 1: Review this file.
 - CSS locator
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+    test("should move task with boundary event", async ({ palette, jsonModel, diagram, page }) => {
+      await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 300 } });
+      await palette.dragNewNode({ type: NodeType.INTERMEDIATE_CATCH_EVENT, targetPosition: { x: 450, y: 300 } });
+
+      const taskNode = page.getByTestId("kie-tools--bpmn-editor--node-task").first();
+      await expect(taskNode).toBeAttached();
+      await taskNode.scrollIntoViewIfNeeded();
+
+      const taskBox = await taskNode.boundingBox();
+      if (!taskBox) throw new Error("Task bounding box not found");
```
ljmotta: A test shouldn't have conditions or throw errors.

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+
+      await expect(diagram.get()).toHaveScreenshot("non-interrupting-boundary-event.png");
+    });
+  });
+
+  test.describe("Activity types", () => {
+    test("should attach to user task", async ({ palette, nodes, jsonModel, page, diagram }) => {
+      await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 300 } });
+
+      const task = page.locator(`[data-nodelabel="${DefaultNodeName.TASK}"]`);
```
ljmotta: CSS locator

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
ljmotta, line 1: Review this file:
 - CSS locator
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/addCustomTasks.spec.ts
ljmotta, line 1: Review this file:
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/addEndEvent.spec.ts
ljmotta, line 1: Review this file:
 - CSS locator
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/addGateway.spec.ts
ljmotta, line 1: Review this file:
 - CSS locator
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/addIntermediateCatchEvent.spec.ts
ljmotta, line 1: Review this file:
 - CSS locator
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/addIntermediateThrowEvent.spec.ts
ljmotta, line 1: Review this file:
 - CSS locator
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/addStartEvent.spec.ts
ljmotta, line 1: Review this file:
 - CSS locator
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/addSubProcess.spec.ts
ljmotta, line 1: Review this file:
 - CSS locator
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/addTask.spec.ts
ljmotta, line 1: Review this file:
 - CSS locator
 - If block
 - throw error
 - `waitFor` usage
 - using `let`

packages/bpmn-editor/tests-e2e/flowElements/addTask.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+      await nodes.morphNode({ nodeLocator: task, targetMorphType: "Script task" });
+
+      let taskElement = await jsonModel.getFlowElement({ elementIndex: 0 });
+      expect(taskElement.__$$element).toBe("scriptTask");
+
+      await page.mouse.move(0, 0);
+
+      await nodes.morphNode({ nodeLocator: task, targetMorphType: "Task", exact: true });
+
+      taskElement = await jsonModel.getFlowElement({ elementIndex: 0 });
```
ljmotta: Unnecessary?

packages/bpmn-editor/tests-e2e/flowElements/addTextAnnotation.spec.ts
ljmotta, line 1: Review this file:
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/flowElements/boundaryEventCompensation.spec.ts
ljmotta: Please link the issue into the test with `test.annotation`

packages/bpmn-editor/tests-e2e/flowElements/boundaryEventCompensation.spec.ts
ljmotta, line 1: Review this file:
 - If block
 - throw error

packages/bpmn-editor/tests-e2e/propertiesPanel/changeEndEventProperties.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      "This event ends the process successfully"
+    );
+  });
+
+  test("should configure Terminate event definition", async ({ endEventPropertiesPanel, page }) => {
+    const endEvent = page.getByTestId("kie-tools--bpmn-editor--node-end-event").first();
+
+    await endEventPropertiesPanel.setTerminateDefinition({
+      endEventLocator: endEvent,
+    });
```
ljmotta: Inline variables that are not re-used

packages/bpmn-editor/tests-e2e/propertiesPanel/changeEndEventProperties.spec.ts
ljmotta, line 1: Review this file:
 - remove variables that are not re-used

packages/bpmn-editor/tests-e2e/propertiesPanel/changeIntermediateEventProperties.spec.ts
ljmotta, line 1: Review this file:
 - remove variables that are not re-used

packages/bpmn-editor/tests-e2e/propertiesPanel/changeLaneProperties.spec.ts
ljmotta, line 1: Review this file:
 - if block

packages/bpmn-editor/tests-e2e/propertiesPanel/changeSequenceFlowProperties.spec.ts
ljmotta, line 1: Review this file:
 - remove variables that are not re-used
 - if block
 - throw error
 - `waitFor` usage

packages/bpmn-editor/tests-e2e/propertiesPanel/changeStartEventProperties.spec.ts
ljmotta, line 1: Review this file:
 - remove variables that are not re-used
 - if block
 - throw error

packages/bpmn-editor/tests-e2e/propertiesPanel/changeSubProcessProperties.spec.ts
ljmotta, line 1: Review this file:
 - remove variables that are not re-used

packages/bpmn-editor/tests-e2e/propertiesPanel/changeTaskProperties.spec.ts
ljmotta, line 1: Review this file:
 - remove variables that are not re-used
 - CSS locator

ljmotta (changes requested):

ljmotta:
@rajalakshmys-27 I've done a new round of review on the code. Please take a look into all files that I've added a comment into. I could see we still have many `waitFor` and CSS locators. Also we shouldn't have `if` blocks on a test or throw errors. Related to readability some tests create unnecessary variables and they could be inlined.

rajalakshmys-27:
@ljmotta Thanks for the detailed review! I've gone through all the commented files and addressed the `waitFor` calls, CSS locators, `if` blocks, error throws, and unnecessary variables. The changes are now updated.

packages/bpmn-editor/tests-e2e/__fixtures__/edges.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export class Edges {
+  constructor(
+    public page: Page,
+    public nodes: Nodes,
+    public diagram: Diagram
+  ) {}
+
+  public async get(args: { from: string; to: string }): Promise<Locator> {
```
ljmotta, line 36: Is it not possible to use the same approach we have on DMN Editor?

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
ljmotta: please, add a test annotation tag with the issue link so we don't lose track of this test

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+  await editor.open();
+});
+
+test.describe("Add Boundary Event", () => {
+  test.describe("Basic attachment", () => {
+    test("should attach intermediate catch event to task", async ({ palette, jsonModel, page }) => {
+      await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 300 } });
+      await palette.dragNewNode({ type: NodeType.INTERMEDIATE_CATCH_EVENT, targetPosition: { x: 450, y: 300 } });
+
+      const boundaryEventNode = page.getByTestId(/^kie-tools--bpmn-editor--node-intermediate-catch-event-/).first();
```
ljmotta: Why are you not using the nodes fixture to get the node?

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+      palette,
+      jsonModel,
+      diagram,
+      page,
+    }) => {
+      await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 300 } });
+      await palette.dragNewNode({ type: NodeType.INTERMEDIATE_CATCH_EVENT, targetPosition: { x: 450, y: 300 } });
+
+      const boundaryEventNode = page.getByTestId(/^kie-tools--bpmn-editor--node-intermediate-catch-event-/).first();
+      await expect(boundaryEventNode).toBeVisible();
```
ljmotta: What is the reason behind this assertion?

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+      jsonModel,
+      diagram,
+      page,
+      intermediateEventPropertiesPanel,
+    }) => {
+      await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 300 } });
+      await palette.dragNewNode({ type: NodeType.INTERMEDIATE_CATCH_EVENT, targetPosition: { x: 450, y: 300 } });
+
+      const boundaryEventNode = page.getByTestId(/^kie-tools--bpmn-editor--node-intermediate-catch-event-/).first();
+      await expect(boundaryEventNode).toBeVisible();
```
ljmotta: Same. Why do we have an assert for the "first" boundary event node?

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+    test("should delete boundary event", async ({ palette, jsonModel, page }) => {
+      await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 300 } });
+      await palette.dragNewNode({ type: NodeType.INTERMEDIATE_CATCH_EVENT, targetPosition: { x: 450, y: 300 } });
+
+      const boundaryEventNode = page.getByTestId(/^kie-tools--bpmn-editor--node-intermediate-catch-event-/).first();
+      await expect(boundaryEventNode).toBeVisible();
+      await boundaryEventNode.click();
+      await boundaryEventNode.press("Delete");
+
+      await expect(boundaryEventNode).not.toBeAttached();
```
ljmotta: A few things to make a more readble test:

If the intetation is to create an assertion to check if the node was indeed created, put it together with the block where the node was created.
Use the fixtures to get and manipulate the node.
Use the same type of assertion. If you want to use `toBeVisible()`, use its counterpart `not.toBeVisible()`. For this case, I believe `toBeAttached` would make more sense, as it does not take into account if the element is visible or not.


```suggestion
      await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 300 } });
      await palette.dragNewNode({ type: NodeType.INTERMEDIATE_CATCH_EVENT, targetPosition: { x: 450, y: 300 } });
      await expect(nodes.get(...)).toBeAttached();

      await nodes.delete(...)
      await expect(nodes.get(...)).not.toBeAttached();
```

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
```diff
@@ -0,0 +1,436 @@
...
+        targetPosition: { x: 100, y: 100 },
+      });
+
+      const callActivity = nodes.get({ name: "New Call Activity" });
+      await expect(callActivity).toBeAttached();
+
+      const box = await callActivity.boundingBox();
+      expect(box).not.toBeNull();
+
+      await page.mouse.move(box!.x + box!.width - 10, box!.y + box!.height / 2);
```
ljmotta: Shouldn't this be on a fixture?

ljmotta (commented):
@rajalakshmys-27 Thanks for addressing my previous changing request. We are almost there. I'm doing a new round of reviews, and I couldn't finish it yet but I've added a few comments, please take into account for the other tests as well. Some tests are not using all fixtures. Could you please double check?

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
rajalakshmys-27: I forgot to update this

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
rajalakshmys-27: The toBeVisible() here acts as a synchronization point before the dragTo() on the next line (since a DOM interaction (drag) follows immediately after.). However, after further review, .first() is unnecessary since there's always exactly one INTERMEDIATE_CATCH_EVENT node at this point in the test. I've removed .first() across all tests in the file.

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
rajalakshmys-27: removed

packages/bpmn-editor/tests-e2e/propertiesPanel/changeSequenceFlowProperties.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+test.describe("Change Properties - Conditional Sequence Flow from Gateway", () => {
+  test.beforeEach(async ({ palette, nodes, edges, sequenceFlowPropertiesPanel, diagram, page }) => {
+    await palette.dragNewNode({ type: NodeType.GATEWAY, targetPosition: { x: 100, y: 250 } });
+
+    await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 50 }, thenRenameTo: "High Amount" });
+
+    await palette.dragNewNode({ type: NodeType.TASK, targetPosition: { x: 300, y: 400 }, thenRenameTo: "Low Amount" });
+
+    await expect(nodes.getByType(NodeType.GATEWAY)).toBeVisible();
+    const gatewayId = (await nodes.getByType(NodeType.GATEWAY).getAttribute("data-nodehref")) ?? "";
```
ljmotta: shouldn't have a fallback.

packages/bpmn-editor/tests-e2e/flowElements/addBoundaryEvent.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+      await expect(nodes.getByType(NodeType.TASK)).toBeAttached();
+      await nodes.getByType(NodeType.TASK).scrollIntoViewIfNeeded();
+
+      const taskBox = await nodes.getNodeBounds({ id: await nodes.getIdByType(NodeType.TASK) });
+
+      await nodes.getByType(NodeType.TASK).dragTo(diagram.get(), {
+        sourcePosition: { x: 20, y: taskBox.height / 2 },
+        targetPosition: { x: 400, y: 350 },
+        force: true,
+      });
```
ljmotta: It is not clear why moving to `taskBox.height / 2`, and where is this on the editor. Fixed numbers are prefered, or this should be behind a fixture with a good name on it.

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+
+      const callActivity = nodes.get({ name: "New Call Activity" });
+      await expect(callActivity).toBeAttached();
+
+      await nodes.showNodeHandles({ name: "New Call Activity" });
+
+      const addTaskHandle = callActivity.getByTitle("Add Task");
+      await expect(addTaskHandle).toBeVisible();
+
+      await addTaskHandle.dragTo(diagram.get(), { targetPosition: { x: 300, y: 100 } });
```
ljmotta: Why are you not using `nodes.dragNewConnectedNode`?

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+
+      const callActivity = nodes.get({ name: "New Call Activity" });
+      await expect(callActivity).toBeAttached();
+
+      await nodes.showNodeHandles({ name: "New Call Activity" });
+
+      const addGatewayHandle = callActivity.getByTitle("Add Gateway");
+      await expect(addGatewayHandle).toBeVisible();
+
+      await addGatewayHandle.dragTo(diagram.get(), { targetPosition: { x: 300, y: 100 } });
```
ljmotta: Same here.

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+      });
+
+      await palette.dragNewNode({
+        type: NodeType.CALL_ACTIVITY,
+        targetPosition: { x: 350, y: 100 },
+      });
+
+      const gateway = nodes.getByType(NodeType.GATEWAY);
+      await expect(gateway).toBeVisible();
+      await expect(nodes.get({ name: "New Call Activity" })).toBeAttached();
```
ljmotta: Not using enum value.

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+      await expect(addSequenceFlowHandle).toBeVisible();
+
+      const callActivityBox = await nodes.getNodeBounds({ name: "New Call Activity" });
+
+      await addSequenceFlowHandle.dragTo(diagram.get(), {
+        targetPosition: {
+          x: callActivityBox.x + callActivityBox.width / 2,
+          y: callActivityBox.y + callActivityBox.height / 2,
+        },
+      });
```
ljmotta: Not using `dragNewConnectedNode`. Not clear to where the node is being drag to.

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+      await palette.dragNewNode({ type: NodeType.END_EVENT, targetPosition: { x: 1000, y: 150 } });
+
+      const startEvent = nodes.getByType(NodeType.START_EVENT);
+      await expect(startEvent).toBeAttached();
+
+      const prepareData = nodes.get({ name: "Prepare Data" });
+      const callActivity = nodes.get({ name: "Execute Subprocess" });
+      const processResults = nodes.get({ name: "Process Results" });
+      const endEvent = nodes.getByType(NodeType.END_EVENT);
```
ljmotta: unecessary variables.

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+
+      // Connect Start Event -> Prepare Data
+      await nodes.showNodeHandles({ id: await nodes.getIdByType(NodeType.START_EVENT) });
+      const handle1 = startEvent.getByTitle("Add Sequence Flow");
+      await expect(handle1).toBeVisible();
+      let targetBox = await nodes.getNodeBounds({ name: "Prepare Data" });
+      await handle1.dragTo(diagram.get(), {
+        targetPosition: { x: targetBox.x + targetBox.width / 2, y: targetBox.y + targetBox.height / 2 },
+      });
```
ljmotta: `nodes.dragNewConnectedNode`.

packages/bpmn-editor/tests-e2e/flowElements/addCallActivity.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+      // Connect Process Results -> End Event
+      await nodes.showNodeHandles({ name: "Process Results" });
+      const handle4 = processResults.getByTitle("Add Sequence Flow");
+      await expect(handle4).toBeVisible();
+      targetBox = await nodes.getNodeBounds({ id: await nodes.getIdByType(NodeType.END_EVENT) });
+      await handle4.dragTo(diagram.get(), {
+        targetPosition: { x: targetBox.x + targetBox.width / 2, y: targetBox.y + targetBox.height / 2 },
+      });
+
+      await diagram.resetFocus();
```
ljmotta: Why? If it is not clear, probably it should be behind a fixture.

ljmotta (approved):
I'll approve this PR, but here are comments that I've previously added/shared with you and they were not addressed. As before, I didn't cover the entire PR (only a few files), and I could see many things that require to be addressed. After each review, it is necessary to take a time to analyze the comments and the codebase, each comment can be applied for different places. (going back to go forth). 

Please, create a follow up issue on `kie-issues` to address my comments and to review the entire BPMN E2E test suite for similar things that need to be addressed.

For this case, I believe creating smaller PRs would be more benefitial instead of creating the entire test suite, as any change request requires to analyze/update many files.

rajalakshmys-27:
Hi @ljmotta, thank you for the approval and the feedback!

I have addressed some of the refactoring based on your last comments and created a follow-up issue on [kie-issues#2322](https://github.com/apache/incubator-kie-tools/issues/3865) to track the remaining improvements and review the entire BPMN E2E test suite for similar patterns.

thiagoelg (approved):
LGTM! We can merge once the build is green. Thanks @rajalakshmys-27

PR #3557, opened 2026-05-04
kie-issues#2291: BPMN Editor: Unable to create completion conditions on Milestones

### LINKED ISSUE #2291: BPMN Editor: Unable to create completion conditions on Milestones
When a new Milestone node is dragged onto the BPMN canvas, it is created without a `Condition` data input assignment, leaving the node in an incomplete state.

**Steps to Reproduce**
1. Open the BPMN editor and create or open a Case Definition
2. Drag and drop a **Milestone** node onto the canvas
3. Open the **Properties Panel** and expand **Data Mapping**
4. Observe that the panel is **empty** — the `Condition` input must be added manually

Image

<br/>

**Expected Behavior**
- The Milestone node should be created with a pre-configured `Condition` input parameter (of type `String`) by default
- The user should be able to directly enter a Drools expression in the Source field of the `Condition` row
- Example expression: `favouriteColour.equals("blue")`

rajalakshmys-27:
Closes : [kie-issues#2291](https://github.com/apache/incubator-kie-issues/issues/2291)

Fixes the issue where a newly created Milestone node did not include a `Condition` data input assignment by default. 
Since a Milestone always requires a condition to track its completion state, the node was being created in an incomplete state, requiring the user to manually add and configure the `Condition` input themselves.

**Changes**

- Added `ioSpecification` with a `Condition` data input to the Milestone node in `produce()`
- Added `onAdded()` callback to find or create a `String` ItemDefinition and link it to the `Condition` input via `itemSubjectRef`
- Initialized `dataInputAssociation` and `dataOutputAssociation` as empty arrays

image


packages/bpmn-editor-envelope/src/customTasks/MilestoneTask.tsx
```diff
@@ -86,12 +86,61 @@ export function getMilestoneTask(i18n: BpmnEditorEnvelopeI18n): CustomTask {
...
+      );
+
+      if (!stringItemDefinition) {
+        stringItemDefinition = {
+          __$$element: "itemDefinition",
+          "@_id": generateUuid(),
+          "@_structureRef": "String",
+        };
+        definitions.rootElement.push(stringItemDefinition);
+      }
```
tiagobento: Can you use `addOrGetItemDefinitions` here instead?

tiagobento: You could also look into `getDataMapping` and `setInputDataMapping`.

packages/bpmn-editor-envelope/src/customTasks/MilestoneTask.tsx
```diff
@@ -86,12 +86,61 @@ export function getMilestoneTask(i18n: BpmnEditorEnvelopeI18n): CustomTask {
...
+        __$$element: "task",
+        "@_id": generateUuid(),
+        "@_drools:taskName": "Milestone",
+        "@_name": i18n.milestone,
+        ioSpecification: {
+          "@_id": generateUuid(),
+          dataInput: [
+            {
+              "@_id": conditionInputId,
+              "@_name": "Condition",
```
tiagobento: Can we please extract the "Condition" string to be a constant? I've been keeping track of all of the extension points we have in here https://github.com/apache/incubator-kie-tools/blob/main/packages/bpmn-marshaller/src/drools-extension.ts.

This would be an entirely new constant.

thiagoelg:
Thanks @rajalakshmys-27!

Something I noticed while testing is that the `TaskName` data input added by the classic editor is not "hidden" in the new editor.

Maybe the new and classic editors are using different data input names for this property?


https://github.com/user-attachments/assets/3da554ef-5434-48d1-95e8-f05c44679b6d

BPMN: https://gist.githubusercontent.com/thiagoelg/136ece1f729e5d18704c10db5c9a41e6/raw/c30a04a008920ed57ee8f00196d0001a8f2e10de/Untitled.bpmn

packages/bpmn-editor-envelope/src/customTasks/MilestoneTask.tsx
rajalakshmys-27: Done! Replaced with `addOrGetItemDefinitions()`.

packages/bpmn-editor-envelope/src/customTasks/MilestoneTask.tsx
rajalakshmys-27: Added `MILESTONE_TASK_IO_SPECIFICATION_DATA_INPUTS_CONSTANTS` with a `CONDITION` entry to `drools-extension.ts` and replaced the hardcoded string in `MilestoneTask.tsx`.

rajalakshmys-27:
Thanks @tiagobento  for the review and suggestions!

Thanks @thiagoelg ! You're right, the new editor wasn't generating the TaskName input at all.
Fixed by adding TaskName data input generation in `produce()`, the corresponding `dataInputAssociation`, and adding it to `dataInputReservedNames` to hide it from the UI. 
Both editors now generate the same structure.

https://gist.githubusercontent.com/rajalakshmys-27/e044f904b6ac19eee31202664d45334b/raw/685107b4a58a7b04a1044ad82f3e72f7690cc767/Milestone.bpmn

packages/bpmn-editor-envelope/src/customTasks/MilestoneTask.tsx
```diff
@@ -86,13 +94,83 @@ export function getMilestoneTask(i18n: BpmnEditorEnvelopeI18n): CustomTask {
...
+        dataType: DEFAULT_DATA_TYPES.OBJECT,
+      });
+
+      if (task.ioSpecification?.dataInput?.[0]) {
+        task.ioSpecification.dataInput[0]["@_itemSubjectRef"] = stringItemDefinition["@_id"];
+      }
+
+      if (task.ioSpecification?.dataInput?.[1]) {
+        task.ioSpecification.dataInput[1]["@_itemSubjectRef"] = objectItemDefinition["@_id"];
+      }
```
thiagoelg: I feel like there's a better way to do this. 🤔 
You can create the inputs in the `onAdded` callback and then call `setInputAndOutputDataMapping` to generate all of the `ioSpecification` automatically. This way will also work when morphing from other nodes.

Also, we have constants for the data types (`DEFAULT_DATA_TYPES.STRING` and `DEFAULT_DATA_TYPES.OBJECT`).

thiagoelg (changes requested):

packages/bpmn-editor-envelope/src/customTasks/MilestoneTask.tsx
rajalakshmys-27: Refactored to use `setInputAndOutputDataMapping` in `onAdded()`, simplified `produce()` to just create the basic task structure, moved I/O logic to `onAdded()`, and replaced hardcoded strings with `DEFAULT_DATA_TYPES` and `MILESTONE_TASK_VALUES` constants. Much cleaner and handles morphing correctly. Thanks!

ljmotta (approved):
@rajalakshmys-27 Thanks for the PR. I've manually tested and I couldn't find any problems with your PR.

thiagoelg (approved):

PR #3562, opened 2026-05-07
NO-ISSUE : BUMP YAML VERSION

AnnJoy23:
fix CVE-2026-33532

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25,6 +25,8 @@ overrides:
   minimatch@^5: 5.1.9
   minimatch@^4: 5.1.9
   undici: ^6.24.0
+  cosmiconfig>yaml: 1.10.3
+  yaml@^2: ^2.8.3
 
 packageExtensionsChecksum: sha256-oxPwESKKSHRelJQnCQTHzgtG1xkcQOHjfgjFdIfqMfg=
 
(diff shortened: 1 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -25,3 +25,8 @@ overrides:
   # CVE-2026-1526, CVE-2026-2229, CVE-2026-1528, CVE-2026-1527, CVE-2026-1525, CVE-2026-2203: Fix security vulnerability in undici
   # Waiting for @openapi-contrib/openapi-schema-to-json-schema to release patched version
   "undici": "^6.24.0"
+  # CVE-2026-33532: Fix security vulnerability in yaml
+  # Override cosmiconfig>yaml@1.10.2 with patched version 1.10.3
+  "cosmiconfig>yaml": "1.10.3"
+  # Override yaml@^2 (including yaml-language-server yaml@2.0.0-11) used by helm charts, monaco-yaml, vite, and langchain dependencies with patched version 2.8.3
+  "yaml@^2": "^2.8.3"
```
thiagoelg, line 32: These comments do not follow the [guidelines](https://github.com/apache/incubator-kie-tools/blob/main/repo/DEPENDENCY_MANAGEMENT.md).

---

- `cosmiconfig` is the dependency of what modules? Can these modules be updated? Also, if you check the lockfile, it's already using version `1.10.3`, so this override is not needed.

- All `yaml` entries in the lockfile for version 2+ are already using `^2.8.3`; this override is also not needed.

thiagoelg (changes requested):
Thanks @AnnJoy23 

I'm not sure if this PR is needed. All dependencies seem to be already updated in the lockfile.

You can check this by running `pnpm why -r yaml`, for example.

pnpm-lock.yaml
thiagoelg, line 1: Except for the first few lines, all other changes to the lockfile are unrelated to the overrides. Is there something different in your environment? Are you using the Devbox setup?

AnnJoy23:
closing the PR as yaml updated in lock file

PR #3569, opened 2026-05-08
kie-tools#3855: Route Kubernetes Dev Deployment fetches through cors-proxy

### LINKED ISSUE #3855: [KIE Sandbox] Dev Deployments runtime fetches bypass CORS proxy on Kubernetes path
When using KIE Sandbox with a Kubernetes (non-OpenShift) Dev Deployments
authentication session, the deployment row displays a persistent error
icon in the dropdown even when the deployed pod is Ready, and the
browser console shows CORS errors against the runtime's ingress URL:

```
Access to fetch at 'https://<runtime-host>/q/health' from origin
'<sandbox-origin>' has been blocked by CORS policy: No
'Access-Control-Allow-Origin' header is present on the requested resource.
```

The Kubernetes code path issues browser→runtime fetches directly,
bypassing the configured cors-proxy. Because the dev-deployment runtime
image does not configure CORS headers for cross-origin requests, the
browser rejects these calls. The OpenShift code path is unaffected
because it already passes `proxyUrl` to its service and routes through
the proxy via the existing `target-url` header convention (also used by
`kubernetes-bridge`'s `ResourceFetcher`).

**Steps to Reproduce**

1. Run KIE Sandbox locally (`pnpm -F online-editor start`) so the
   cors-proxy is reachable at `KIE_SANDBOX_CORS_PROXY_URL`.
2. Configure a Kubernetes Dev Deployments auth session against a
   cluster whose runtime image does not ship CORS headers (e.g. a kind
   cluster running the upstream `dev-deployment-quarkus-blank-app`
   image).
3. Deploy a model from the Sandbox.
4. Once the pod is Ready (`kubectl get pods` shows `1/1 Running`),
   open DevTools → Network. The fetch to the runtime's `/q/health`
   endpoint fails with the CORS error above.
5. Observe the deployment row in the dropdown shows an error icon
   despite the pod being healthy.

**Expected Behavior**

When `KIE_SANDBOX_CORS_PROXY_URL` is configured, every browser→runtime
fetch from the Kubernetes Dev Deployments code path should route
through the cors-proxy via the `target-url` header convention,
matching the OpenShift code path.

ch-lukas:
Closes: https://github.com/apache/incubator-kie-tools/issues/3855

The Kubernetes Dev Deployments code path issued browser→runtime fetches
directly, bypassing the configured cors-proxy. Because the
dev-deployment runtime image does not configure CORS headers for
cross-origin requests, the browser blocked these calls and the
deployment row stayed in an error state even when the pod was Ready.

Brings the Kubernetes path to parity with the OpenShift path, which
already routes through the cors-proxy via the existing `target-url`
header convention (also used by `kubernetes-bridge`'s `ResourceFetcher`).

**Changes**

- `KieSandboxDevDeploymentsService.getHealthStatus` and the upload-status
  / upload calls in `KieSandboxDevDeploymentsUploadAppApi` now route
  through `proxyUrl` via the `target-url` header when one is configured.
- `KieSandboxKubernetesService` is instantiated in
  `DevDeploymentsContextProvider` with `proxyUrl: env.KIE_SANDBOX_CORS_PROXY_URL`,
  matching how the OpenShift service is constructed.
- `KubernetesService.getBaseUrl` preserves the `connection.host`
  pathname when proxying, fixing K8s API URLs that include a path
  prefix (e.g. kind's `localhost/kube-apiserver`).

**Test plan**

- [x] Manual: deploy a model from KIE Sandbox against a kind cluster
  whose runtime image lacks CORS headers; verify `/q/health` succeeds
  via DevTools → Network and the dropdown shows the deployment as Up.
- [x] Verified the OpenShift code path is structurally unchanged (the
  diff in `KieSandboxOpenShiftService` is empty).
- [ ] CI: TypeScript, ESLint, Jest, Playwright on Linux.

### DIFF
--- a/packages/online-editor/src/accounts/kubernetes/ConnectToKubernetesSection.tsx
+++ b/packages/online-editor/src/accounts/kubernetes/ConnectToKubernetesSection.tsx
@@ -48,6 +48,17 @@ export function ConnectToKubernetesSection() {
   const [isLoadingService, setIsLoadingService] = useState(false);
   const selectedKubernetesSession =
     accounts.section === AccountsSection.CONNECT_TO_KUBERNETES ? accounts.selectedAuthSession : undefined;
+  const selectedKubernetesAuthSession: KubernetesAuthSession | undefined =
+    selectedKubernetesSession?.type === "kubernetes" ? selectedKubernetesSession : undefined;
+  // CORS-proxy opt-in for this auth session. Default unchecked: local
+  // kind/minikube clusters are the most common case, are reachable
+  // directly from the browser, and the project's setup instructions
+  // for those already configure cluster-side CORS allowance. Users on
+  // a remote-cluster topology check the box. Legacy sessions without
+  // the field also default to unchecked via the read-side fallback in
+  // DevDeploymentsContextProvider, so existing setups keep working
+  // unchanged.
+  const [useCorsProxy, setUseCorsProxy] = useState<boolean>(selectedKubernetesAuthSession?.useCorsProxy ?? false);
 
   useCancelableEffect(
     useCallback(
@@ -142,6 +153,8 @@ export function ConnectToKubernetesSection() {
               setMode={setMode}
               connection={connection}
               setConnection={setConnection}
+              useCorsProxy={useCorsProxy}
+              setUseCorsProxy={setUseCorsProxy}
               status={status}
               setStatus={setStatus}
               setNewAuthSession={setNewAuthSession}
@@ -155,6 +168,8 @@ export function ConnectToKubernetesSection() {
               setMode={setMode}
               connection={connection}
               setConnection={setConnection}
+              useCorsProxy={useCorsProxy}
+              setUseCorsProxy={setUseCorsProxy}
               status={status}
               setStatus={setStatus}
               setNewAuthSession={setNewAuthSession}
--- a/packages/online-editor/src/accounts/kubernetes/ConnectToKubernetesSimple.tsx
+++ b/packages/online-editor/src/accounts/kubernetes/ConnectToKubernetesSimple.tsx
@@ -20,6 +20,7 @@
 import React, { useCallback, useEffect, useState } from "react";
 import { Alert } from "@patternfly/react-core/dist/js/components/Alert";
 import { Button } from "@patternfly/react-core/dist/js/components/Button";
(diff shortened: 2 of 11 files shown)


thiagoelg:
Hi @ch-lukas, thank you for the PR!

There's one scenario in which this change could fail: when users host their KIE Sandbox and CORS Proxy services remotely while the Kind/Minikube cluster is local.

A remote CORS Proxy service can't reach and proxy requests to "localhost". 

---

Having said that, not proxying requests to a Kubernetes Dev Deployment makes it impossible to use remote Kubernetes clusters.

My alternative solution proposal is to have a checkbox when creating the Kubernetes Auth Session that enables all requests to it to be proxied by the CORS Proxy service, like this:

image

This way, we can support both local and remote Kubernetes clusters.

ch-lukas:
Thanks @thiagoelg, the checkbox makes sense.

To keep scope tight I plan to:

- Add the checkbox to the Kubernetes auth session wizard only. OpenShift
  stays always-proxied. OpenShift clusters are basically never local, so
  the toggle would never be unchecked there and would just add surface
  area.
- Treat legacy stored sessions (no `useCorsProxy` field) as unchecked, so
  existing setups keep working unchanged. New sessions get the wizard.

Open question: for new sessions, should the wizard default to checked or
unchecked? Your screenshot shows checked, happy to go with that if it's
deliberate.

thiagoelg:
Awesome, thank you @ch-lukas 

> Open question: for new sessions, should the wizard default to checked or
unchecked? Your screenshot shows checked, happy to go with that if it's
deliberate.

I would default to unchecked, as running local Kind and Minikube clusters is the most common scenario, I believe. Also, our Kind and Minikube setup instructions already set them up to allow CORS.

ch-lukas:
Thanks for the steer on the default. Pushed 616e7178ef which:

- Adds a `useCorsProxy?: boolean` field on `KubernetesAuthSession`.
- Adds a "Use CORS Proxy" checkbox to both Kubernetes connect flows
  (simple form and step-by-step wizard), with i18n in en and de.
- Defaults new sessions to unchecked, per your call.
- Treats legacy stored sessions (no `useCorsProxy` field) as unchecked,
  so existing setups keep working unchanged.
- Leaves the OpenShift path alone (always proxied as before).

PR #3570, opened 2026-05-10
kie-issues#2301: BPMN Editor: Opening faulty XML will add a <parsererror> element to the top-level

### LINKED ISSUE #2301: BPMN Editor: Opening faulty XML will add a <parsererror> element to the top-level
When opening a faulty BPMN file, making any modification (e.g. moving a node), and exporting the XML, the exported file contains a spurious `<bpmn:parsererror>` element at the top level. This element is injected by the browser's `DOMParser` when it encounters malformed or schema-invalid XML, but instead of being discarded, it is carried through the editor's internal model and serialized back into the exported file — corrupting it further on every save cycle.

A faulty BPMN file  : https://gist.githubusercontent.com/rajalakshmys-27/28e82fd3de60d656a28a0ab654401e32/raw/855f4fc030cd560d66e6b2cb7107134e1518c90b/parseerror.bpmn

**Steps to reproduce**
1. Open a faulty BPMN file (e.g. one where `<bpmn:message>` is incorrectly nested inside `<bpmn:process>` instead of `<bpmn:definitions>`)
2. Make any modification, such as dragging the end node slightly to the right
3. Export the BPMN XML
4. Observe that the exported file contains a `<bpmn:parsererror>` element at the top level

**Expected behavior**
The `<bpmn:parsererror>` element is an internal parser artifact and should never appear in exported XML. The editor should either reject the malformed file with a clear error, or silently filter the element and export only the valid content.

**Actual behavior**
The `<bpmn:parsererror>` element is preserved in the exported XML, permanently corrupting the file on each edit-export cycle.

rajalakshmys-27:
Closes: [kie-issues#2301](https://github.com/apache/incubator-kie-issues/issues/2301)

When a faulty BPMN/DMN/SCESIM file is opened, any modification and export causes a `<parsererror>` element to appear at the top level of the exported XML, corrupting the file on every edit-export cycle.

**Example faulty file:** https://gist.githubusercontent.com/rajalakshmys-27/28e82fd3de60d656a28a0ab654401e32/raw/855f4fc030cd560d66e6b2cb7107134e1518c90b/parseerror.bpmn

### Root Cause
`<parsererror>` is automatically injected by the browser's DOMParser (and jsdom in Node.js) when it encounters schema violations such as elements in wrong locations, invalid attributes, or incorrect nesting. This is standard W3C behavior and cannot be disabled.

### Fix
Filter out `<parsererror>` elements in `xml-parser-ts` immediately after parsing, before the document is returned to any marshaller. This single fix protects all marshallers (BPMN, DMN, SCESIM) automatically with no changes required elsewhere.


jomarko (commented):
Thank you for the fix, From my point of view, the issue kie-issues#2301 may be worth to mention also as a comment in a code. In future, I can imagine a situation, when will be quite difficult to understand, why we need the code added by this PR.

From my point of view, this PR is worth also adding a test, so idealy it should be merged after #3556  so we can add such test?

ljmotta:
@rajalakshmys-27 Thanks for your PR. I'm not sure if this is the correct way to proceed as the XML is not fixed after filtering out the `parsererror` tag. Users will still have a broken BPMN file even after editing it. I believe the correct approach is to fix the parse error if we erase the `parsererror` tag, or we show an error message on the editor. WDYT?
cc @thiagoelg

rajalakshmys-27:

Just to clarify the intent of this fix - the problem it solves is specifically the **re-injection bug**: when a faulty file is opened, the browser's DOMParser automatically adds a `<parsererror>` node, which then gets serialised back into the exported XML on every edit-export cycle, making the corruption worse with each save. Filtering it out at the `xml-parser-ts` level stops that re-injection.

This approach was also what @tiagobento suggested as a fallback after exploring whether the DOMParser behaviour could be customised (it can't, it's standard W3C behaviour).

That said, you raise a completely valid point: the user's file is still broken even after filtering, and silently allowing edits on a malformed file isn't a great experience. I agree we should also surface an error in the editor when a `<parsererror>` is detected on load, something that makes it clear the file is malformed and that edits may not be reliable.

Would it make sense to handle both in this PR, or would you prefer to track the error UI as a separate follow-up? Happy to go either way. 
cc @thiagoelg

rajalakshmys-27:
@jomarko Thanks for the suggestions!

I'll add a comment in the code referencing kie-issues#2301 to make it easier to understand the context behind this change in the future.

As for the e2e test, we'd prefer not to block the bug fix on it, but we'll make sure to add one as a follow-up once this is merged. 
cc @thiagoelg

ljmotta:
@rajalakshmys-27 I understand the problem. We have two times where the `parsererror` appears - when we open the file, and when we save the file. The error looks worse on KIE Sandbox as it auto-save after every new edit. If we prevent new edits, we will not have the problem of adding multiple `parsererrors`.

I agree we should remove the `parsererror` from the XML, but just this is not the complete fix. Can't we remove the `parsererror` and prevent new edits? We already have a screen to show the parser errors. I believe by throwing an error after removing the `parsererror` element will show this page.

Another option is we could fix the XML structure (erasing any element that is in the wrong place), but I'm not sure about the difficulty of doing it.

rajalakshmys-27:
Thanks! I’ve tried out showing a UI error, verified the xml. 

https://github.com/user-attachments/assets/60536ed3-64d7-4839-afb6-93f84b3216ec

ljmotta:
I've tried with the BPMN Editor on the dev webapp and I couldn't make the error screen appear. I didn't see the error screen on VS Code too.


https://github.com/user-attachments/assets/23504732-e6d9-435f-9a8a-fa8be5ec3490


https://github.com/user-attachments/assets/b8f77bdb-3765-4442-bb21-91301f173e8c

rajalakshmys-27:
@ljmotta But I was able to see the Error UI (parse error removed) in both webapp & vscode extension 

https://github.com/user-attachments/assets/db3b1788-8dff-4c8d-86fe-4538b12a183a

thiagoelg (approved):

ljmotta (approved):
I've manually tested and it is working as expected. @rajalakshmys-27 Thanks for your patience.

PR #3571, opened 2026-05-12
kie-issues#2297: BPMN Editor: Creating new edges on existing edges from and to gateway nodes causes issues

### LINKED ISSUE #2297: BPMN Editor: Creating new edges on existing edges from and to gateway nodes causes issues
For a bpmn, if an edge exists between a gateway node and a task, a new edge can be drawn between the two which results in an <unknown> entry in route dropdown.
If we create an edge from a node to a gateway node, where there is already a connection, unspecified error is displayed

aswathitv:
Closes [kie#2297](https://github.com/apache/incubator-kie-issues/issues/2297)

In BPMN editor, creating a new edge between two nodes which already have an edge create duplicate entry in the xml incoming and outgoing tag. As a result gateway node creates two issues:

1. An entry in route dropdown when a new edge is created on top of an existing edge from gateway node to a task node.

Screenshot 2026-05-12 at 9 33 16 AM

2. If an edge is created from a node to a gateway node, where there is already a connection, unspecified error is displayed.
Screenshot 2026-05-12 at 9 33 36 AM

Fix:
Duplicate entries are not allowed in incoming and outgoing tag for all nodes.

### DIFF
--- a/packages/bpmn-editor/src/mutations/addEdge.ts
+++ b/packages/bpmn-editor/src/mutations/addEdge.ts
@@ -80,7 +80,7 @@ export function addEdge({
     );
   }
 
-  const newEdgeId = generateUuid();
+  let newEdgeId = generateUuid();
 
   const { process, diagramElements } = addOrGetProcessAndDiagramElements({ definitions });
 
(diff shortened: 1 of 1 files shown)


rajalakshmys-27 (approved):
I tested the changes in this PR. The Fix is working as expected

thiagoelg (approved):

PR #3573, opened 2026-05-13
kie-issues#2309: BPMN Editor: Implementation of REST Service Call Task 

aswathitv:
Part of [kie#2309](https://github.com/apache/incubator-kie-issues/issues/2309)

An enhanced, Kogito REST WIH will start being fully supported and continue to be available. As part of this following changes are made:

- A custom task node has been added.

- A dedicated properties panel has been included with options for URL, method, protocol, host, port, request timeout, headers, query parameters and access token acquisition strategy.

- An option to emulate testing from a sandbox has been provided.

For the access token acquisition strategy:

* If configured, a REST service call task ID and test token are provided for testing.
* If propagated, a test token is provided for testing.
* None can be tested directly.

image


thiagoelg (changes requested):
Thanks @aswathitv!

I'm still in the review process, but the first things I noticed are:

### 1. The components in the properties panel seem to be outside of the correct container (see video)

https://github.com/user-attachments/assets/3b419131-1978-47da-9681-117a9061c4d2

---

### 2. When are the Protocol/Host/Port fields used? What about the Path? I see that they should be an alternative to the URL, but the URL field is mandatory...

Image

---

### 3. For the headers and query parameters, you should use the same component we use for Metadata to maintain the visual identity.

Image

---

### 4. The dropdown component also seems different:

Image

Image

versus

Image

---

### 5. What about the ContentData Data Input variable interpolation for the request? Users should be able to reference a Data Input variable with `{variableName}`, and then use mocked values when testing the request.

---

### 6. It would be nice to have tooltips or descriptions explaining what each "Access Token Acquisition Strategy" means.

Image

Also, what is the "REST Service Call Task ID"? It should be **optional** when "Access Token Acquisition Strategy" !== `configured`, not **hidden**.

packages/bpmn-vscode-extension/src/channelApi/BpmnEditorChannelApiImpl.ts
```diff
@@ -0,0 +1,107 @@
...
+      return {
+        status: response.status,
+        data,
+        headers,
+      };
+    } catch (error) {
+      throw new Error(`REST request failed: ${(error as Error).message}`);
+    }
+  }
+}
```
thiagoelg, line 72: It would be better to have this in a separate file: `packages/bpmn-vscode-extension/src/channelApi/BpmnEditorChannelApiImpl.ts`.

Then the `activateBpmnEditor` implementation can remain in `extension.ts`.

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,679 @@
...
+};
+
+export type QueryParameter = {
+  id: string;
+  name: string;
+  value: string;
+};
+
+const HTTP_METHODS = ["GET", "POST", "PUT", "PATCH", "DELETE"];
+const AUTH_STRATEGIES = ["propagated", "configured", "none"];
```
thiagoelg: These should be enums or a const Map. The strings displayed in the UI should be translatable.

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,679 @@
...
+      <FormSection style={{ "--pf-v5-c-form__section--Gap": "0.5rem" } as React.CSSProperties}>
+        <FormGroup label={i18n.restService.url} isRequired fieldId="rest-url">
+          <TextInput
+            id="rest-url"
+            value={getValue(RestProperties.Url)}
+            onChange={(_, value) => updateRestProperties(RestProperties.Url, value)}
+            placeholder={i18n.restService.urlPlaceholder}
+          />
+          <div style={{ fontSize: "0.875rem", color: "#6a6e73", marginTop: "0.25rem" }}>{i18n.restService.urlHelp}</div>
+        </FormGroup>
```
thiagoelg: There are many custom styles and classes in these components. This indicates that the Patternfly components are not being used correctly. Please see the implementation in the existing properties panels in the BPMN Editor.

aswathitv:
@thiagoelg URLs can be valid URLs or valid endpoint. If it is url, Protocol,Host,Port fields will throw error in error tab

jomarko (commented):
Thank you for the PR. I agree with points mentioned by the @thiagoelg and I see some of them have been improved. However I am not sure about

### 2. When are the Protocol/Host/Port fields used? What about the Path? I see that they should be an alternative to the URL, but the URL field is mandatory...

The properties panel is more difficult to read when we display both URL and then partiala parts of URL as an alternative way of configuring URL. Furthermore we do not prohibit to configure a conflict. For example, what happends if user configure both URL and host?

### 3. For the headers and query parameters, you should use the same component we use for Metadata to maintain the visual identity.
I see some rendering issue, highlighted using green color.
Screenshot 2026-05-26 095150

### Bug in URL field

Every keypress in the URL field causes the cursor is moved to the end of the field. It is problem especially if you want to update the value.

aswathitv:
@jomarko 
2. When are the Protocol/Host/Port fields used? What about the Path? I see that they should be an alternative to the URL, but the URL field is mandatory...

The error tab shows error if user types full url as well as  host and protocol as separate.
image

Instead path is allowed:
image

3. For the headers and query parameters, you should use the same component we use for Metadata to maintain the visual identity.

Metadata also follows same look and feel:
image

jomarko:
Thank you for comment @aswathitv . Could you please guide me, how to see the messages about using URL, host ... in VS Code?

Adarshvk98 (commented):
Thanks for the PR, @aswathitv .

1. For the Protocol/Host/Port fields, I find it more difficult and confusing to understand the conflict between the full URL field and the Protocol/Host/Port fields. It creates confusion. I suggest we keep only the URL field to simplify the user experience. Additionally, if we could extract the protocol , host  and port (if there is one) from the URL and display them under the URL field, it would provide a better understanding for the user. WDYT ?
---
2. I also suggest combining the method dropdown and the URL field using an InputGroup. This way, both fields would be grouped together. You can refer to this link for more information: [https://v5-archive.patternfly.org/components/input-group#with-dropdown](https://v5-archive.patternfly.org/components/input-group#with-dropdown)

Image

---

3. Regarding the Content data input, there are two syntaxes provided for variable interpolation: #{var} and {{var}}. I believe these are redundant. We could follow one of them, say {{var}}, and remove the #{var} syntax. I noticed that the #{var} syntax is not works properly when adding new datamapping values and referring to those values in the content data input and the corresponding mock input for the new datamapping value is not displayed. 

Image

---
4. I also noticed that the Content data input (`ContentDataInput.tsx`) is designed to expect JSON input. So, could we make the monoco editor more JSON-friendly by changing `const MONACO_LANGUAGE_ID = “json”; instead of “contentDataLang”`? (Probably more changes would be required to adjust, but I’m just checking if such a possibility exists.)

aswathitv:
Validations added for URL field conflict with port/host/protocol
image

jomarko (commented):
Thank you for all updates.

I noticed one more thing. When there is non empty content data

Image

And then I try to change method to **GET**, it fails and automatically select **POST**

Image

aswathitv:
Closing this PR as latest commits are not getting reflected.

PR #3575, opened 2026-05-13
[incubator-kie-issues#2298] Duplicate datatypes for BPMN files of kie examples

### LINKED ISSUE #2298: BPMN Editor: Duplicate datatypes for BPMN files of kie examples
Issue
When a BPMN file was loaded, multiple elements (Data Types) sharing the same structureRef were kept as duplicates. This caused the same Data Type to show up multiple times in the editor.

Fix
In normalize.ts, on load we now condense duplicate elements into a single one:

Group all itemDefinition elements by their @_structureRef.
Keep the first one as the "survivor" and drop the rest.
Update every itemSubjectRef across the model (process properties, dataObjects, task ioSpecification, inputs/outputs, event data inputs/outputs, and nested subProcess/adHocSubProcess /transaction elements) to point at the survivor's id.

Abhitocode:
Issue
When a BPMN file was loaded, multiple elements (Data Types) sharing the same structureRef were kept as duplicates. This caused the same Data Type to show up multiple times in the editor.

Fix
In normalize.ts, on load we now condense duplicate elements into a single one:

Group all itemDefinition elements by their @_structureRef.
Keep the first one as the "survivor" and drop the rest.
Update every itemSubjectRef across the model (process properties, dataObjects, task ioSpecification, inputs/outputs, event data inputs/outputs, and nested subProcess/adHocSubProcess /transaction elements) to point at the survivor's id.

Closes: https://github.com/apache/incubator-kie-issues/issues/2298

### DIFF
--- a/packages/bpmn-editor/src/normalization/normalize.ts
+++ b/packages/bpmn-editor/src/normalization/normalize.ts
@@ -32,20 +32,6 @@ type WithRequiredDeep<T, K extends keyof any> = T extends undefined
         ? { [P in K]-?: NonNullable<WithRequiredDeep<T[P], K>> }
         : T);
 
-// Get the array for `key`, or create an empty one if missing.
-function addOrGetGroup<K, V>(map: Map<K, V[]>, key: K): V[] {
-  let group = map.get(key);
-  if (!group) {
-    group = [];
-    map.set(key, group);
-  }
-  return group;
-}
-
-function getRefToRemap(itemDefinitionIdsToRemap: Map<string, string>, id: string | undefined): string | undefined {
-  return id === undefined ? undefined : itemDefinitionIdsToRemap.get(id) ?? id;
-}
-
 function remapDataItems(
   itemDefinitionIdsToRemap: Map<string, string>,
   items: Array<{ "@_itemSubjectRef"?: string }> | undefined
@@ -55,7 +41,7 @@ function remapDataItems(
   }
   for (const item of items) {
     if (item["@_itemSubjectRef"] !== undefined) {
-      item["@_itemSubjectRef"] = getRefToRemap(itemDefinitionIdsToRemap, item["@_itemSubjectRef"]);
+      item["@_itemSubjectRef"] = itemDefinitionIdsToRemap.get(item["@_itemSubjectRef"]) ?? item["@_itemSubjectRef"];
     }
   }
 }
@@ -68,7 +54,9 @@ function remapProperties(
     return;
   }
   for (const p of properties) {
-    p["@_itemSubjectRef"] = getRefToRemap(itemDefinitionIdsToRemap, p["@_itemSubjectRef"]);
+    if (p["@_itemSubjectRef"]) {
+      p["@_itemSubjectRef"] = itemDefinitionIdsToRemap.get(p["@_itemSubjectRef"]) ?? p["@_itemSubjectRef"];
+    }
   }
 }
 
@@ -82,10 +70,10 @@ function remapFlowElements(
       el.__$$element === "dataObjectReference" ||
       el.__$$element === "dataStoreReference"
     ) {
-      el["@_itemSubjectRef"] = getRefToRemap(itemDefinitionIdsToRemap, el["@_itemSubjectRef"]);
-    }
-
-    if (
+      if (el["@_itemSubjectRef"] !== undefined) {
+        el["@_itemSubjectRef"] = itemDefinitionIdsToRemap.get(el["@_itemSubjectRef"]) ?? el["@_itemSubjectRef"];
+      }
+    } else if (
       el.__$$element === "businessRuleTask" ||
       el.__$$element === "callActivity" ||
       el.__$$element === "serviceTask" ||
@@ -94,29 +82,28 @@ function remapFlowElements(
       remapDataItems(itemDefinitionIdsToRemap, el.ioSpecification?.dataInput);
       remapDataItems(itemDefinitionIdsToRemap, el.ioSpecification?.dataOutput);
       remapProperties(itemDefinitionIdsToRemap, el.property);
-    }
-
-    if (el.__$$element === "endEvent" || el.__$$element === "intermediateThrowEvent") {
+    } else if (el.__$$element === "endEvent" || el.__$$element === "intermediateThrowEvent") {
       remapDataItems(itemDefinitionIdsToRemap, el.dataInput);
       remapProperties(itemDefinitionIdsToRemap, el.property);
-    }
-
-    if (
+    } else if (
       el.__$$element === "startEvent" ||
       el.__$$element === "intermediateCatchEvent" ||
       el.__$$element === "boundaryEvent"
     ) {
       remapDataItems(itemDefinitionIdsToRemap, el.dataOutput);
       remapProperties(itemDefinitionIdsToRemap, el.property);
-    }
-
-    if (el.__$$element === "subProcess" || el.__$$element === "adHocSubProcess" || el.__$$element === "transaction") {
+    } else if (
+      el.__$$element === "subProcess" ||
+      el.__$$element === "adHocSubProcess" ||
+      el.__$$element === "transaction"
+    ) {
       remapDataItems(itemDefinitionIdsToRemap, el.ioSpecification?.dataInput);
       remapDataItems(itemDefinitionIdsToRemap, el.ioSpecification?.dataOutput);
       remapProperties(itemDefinitionIdsToRemap, el.property);
       if (el.flowElement) {
         remapFlowElements(itemDefinitionIdsToRemap, el.flowElement);
       }
+    } else {
     }
   }
 }
@@ -179,21 +166,24 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
     }
   });
 
-  // Merge itemDefinitions that share the same structureRef.
-  deduplicateItemDefinitions(model.definitions);
-
   const normalizedModel = model as Normalized<BpmnLatestModel>;
+
+  // Merge itemDefinitions that share the same structureRef.
+  deduplicateItemDefinitions(normalizedModel.definitions);
 
   return normalizedModel;
 }
 
 // Keep one itemDefinition per structureRef and update references to point to it.
-export function deduplicateItemDefinitions(definitions: BpmnLatestModel["definitions"]): void {
-  const itemDefinitionsGroupedStructureRef = new Map<string, Array<{ "@_id"?: string }>>();
+export function deduplicateItemDefinitions(definitions: Normalized<BpmnLatestModel["definitions"]>): void {
+  const itemDefinitionsGroupedStructureRef = new Map<string, Array<string>>();
   for (const rootElement of definitions.rootElement ?? []) {
     if (rootElement.__$$element === "itemDefinition") {
       const itemDefinitionStructureRef = rootElement["@_structureRef"] ?? "";
-      addOrGetGroup(itemDefinitionsGroupedStructureRef, itemDefinitionStructureRef).push(rootElement);
+      itemDefinitionsGroupedStructureRef.set(itemDefinitionStructureRef, [
+        ...(itemDefinitionsGroupedStructureRef.get(itemDefinitionStructureRef) ?? []),
+        rootElement["@_id"],
+      ]);
     }
   }
 
@@ -203,15 +193,9 @@ export function deduplicateItemDefinitions(definitions: BpmnLatestModel["definit
     if (itemDefinitionsGroup.length <= 1) {
       continue;
     }
-    const survivorId = itemDefinitionsGroup[0]["@_id"];
-    if (!survivorId) {
-      continue;
-    }
+    const survivorId = itemDefinitionsGroup[0];
     for (let i = 1; i < itemDefinitionsGroup.length; i++) {
-      const duplicateId = itemDefinitionsGroup[i]["@_id"];
-      if (duplicateId) {
-        itemDefinitionIdsToRemap.set(duplicateId, survivorId);
-      }
+      itemDefinitionIdsToRemap.set(itemDefinitionsGroup[i], survivorId);
     }
   }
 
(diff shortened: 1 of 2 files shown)


aswathitv:
https://github.com/user-attachments/assets/72f28ba7-e869-4d47-b1c6-c3fde3958bbb

@Abhitocode Thank you for the PR.
I can add the same datatype again. Only on a reload the duplicate gets removed.

Abhitocode:
https://github.com/user-attachments/assets/dbe8322b-8c83-4b7b-a79b-ab196e211ac1

@aswathitv  fixed, could test once more and confirm. Thank you : )

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -31,6 +32,86 @@ type WithRequiredDeep<T, K extends keyof any> = T extends undefined
...
+  if (!group) {
+    group = [];
+    map.set(key, group);
+  }
+  return group;
+}
+
+function remapRef(remap: Map<string, string>, id: string | undefined): string | undefined {
+  return id === undefined ? undefined : remap.get(id) ?? id;
+}
```
thiagoelg: This is more of a getter than a "remap" function.

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +170,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
+  const itemDefGroups = new Map<string, Array<{ "@_id"?: string }>>();
+  for (const rootElement of definitions.rootElement ?? []) {
+    if (rootElement.__$$element === "itemDefinition") {
+      const key = rootElement["@_structureRef"] ?? "";
+      addOrGetGroup(itemDefGroups, key).push(rootElement);
+    }
+  }
+
+  // Map each duplicate id to the id of the first occurrence we keep.
+  const itemDefIdRemap = new Map<string, string>();
```
thiagoelg:
```suggestion
  const itemDefinitionIdsToRemap = new Map<string, string>();
```

I prefer that the variable names be more descriptive. It makes it easier to read and follow the code.

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +170,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
+  deduplicateItemDefinitions(model.definitions);
+
   const normalizedModel = model as Normalized<BpmnLatestModel>;
 
   return normalizedModel;
 }
+
+// Keep one itemDefinition per structureRef and update references to point to it.
+export function deduplicateItemDefinitions(definitions: BpmnLatestModel["definitions"]): void {
+  const itemDefGroups = new Map<string, Array<{ "@_id"?: string }>>();
```
thiagoelg:
```suggestion
  const itemDefinitionsGroupedStructureRef = new Map<string, Array<{ "@_id"?: string }>>();
```

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +170,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
+  for (const rootElement of definitions.rootElement ?? []) {
+    if (rootElement.__$$element === "itemDefinition") {
+      const key = rootElement["@_structureRef"] ?? "";
+      addOrGetGroup(itemDefGroups, key).push(rootElement);
```
thiagoelg:
```suggestion
      const itemDefinitionStructureRef = rootElement["@_structureRef"] ?? "";
      addOrGetGroup(itemDefinitionsGroupedStructureRef, itemDefinitionStructureRef).push(rootElement);
```

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +170,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
+    }
+  }
+
+  // Map each duplicate id to the id of the first occurrence we keep.
+  const itemDefIdRemap = new Map<string, string>();
+  for (const group of itemDefGroups.values()) {
```
thiagoelg:
```suggestion
  for (const itemDefinitionsGroup of itemDefinitionsGroupedStructureRef.values()) {
```

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +170,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
+      }
+    }
+  }
+
+  if (itemDefIdRemap.size === 0) {
+    return;
+  }
+
+  // Remove the duplicate itemDefinitions and update every itemSubjectRef.
+  const idsToRemove = new Set(itemDefIdRemap.keys());
```
thiagoelg:
```suggestion
  const itemDefinitionIdsToRemove = new Set(itemDefIdRemap.keys());
```

packages/bpmn-editor/src/normalization/normalize.ts
thiagoelg:
```suggestion
function getRefToRemap(remap: Map<string, string>, id: string | undefined): string | undefined {
  return id === undefined ? undefined : remap.get(id) ?? id;
}
```

I find it hard to understand what exactly this function is getting.

thiagoelg (changes requested):
Thanks @Abhitocode!

I left some inline suggestions to make the code more readable.

packages/bpmn-editor/src/normalization/normalize.ts
thiagoelg: What is the `remap` Map? A map of what?

Abhitocode: sorry, I realised it and already started renaming it earlier, it's a map from each duplicate itemDefinition id to the surviving id we want references to point to. Renaming it to itemDefinitionIdsToRemap, pushing the changes in a few minutes.

thiagoelg (approved):

aswathitv (approved):

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -31,6 +32,95 @@ type WithRequiredDeep<T, K extends keyof any> = T extends undefined
...
 
+// Get the array for `key`, or create an empty one if missing.
+function addOrGetGroup<K, V>(map: Map<K, V[]>, key: K): V[] {
+  let group = map.get(key);
+  if (!group) {
+    group = [];
+    map.set(key, group);
+  }
+  return group;
+}
```
ljmotta: Please, consider inlining this function. The function is used just one time and adds complexity. Here is an example of how you could do it, but you can use `if/else` block if you think it is better.


```suggestion
if (rootElement.__$$element === "itemDefinition") {
    const itemDefinitionStructureRef = rootElement["@_structureRef"] ?? "";
    itemDefinitionsGroupedStructureRef.set(
        itemDefinitionStructureRef,
        [...(itemDefinitionsGroupedStructureRef.get(itemDefinitionStructureRef) ?? []), rootElement]
    )
}
```

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -31,6 +32,95 @@ type WithRequiredDeep<T, K extends keyof any> = T extends undefined
...
+  itemDefinitionIdsToRemap: Map<string, string>,
+  items: Array<{ "@_itemSubjectRef"?: string }> | undefined
+) {
+  if (!items) {
+    return;
+  }
+  for (const item of items) {
+    if (item["@_itemSubjectRef"] !== undefined) {
+      item["@_itemSubjectRef"] = getRefToRemap(itemDefinitionIdsToRemap, item["@_itemSubjectRef"]);
+    }
```
ljmotta, line 45:
```suggestion
    if (item["@_itemSubjectRef"] !== undefined) {
      item["@_itemSubjectRef"] = itemDefinitionIdsToRemap.get(item["@_itemSubjectRef"]) ?? item["@_itemSubjectRef"];
    }
```

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -31,6 +32,95 @@ type WithRequiredDeep<T, K extends keyof any> = T extends undefined
...
+  if (!group) {
+    group = [];
+    map.set(key, group);
+  }
+  return group;
+}
+
+function getRefToRemap(itemDefinitionIdsToRemap: Map<string, string>, id: string | undefined): string | undefined {
+  return id === undefined ? undefined : itemDefinitionIdsToRemap.get(id) ?? id;
+}
```
ljmotta: I believe we can inline this to make the operation explict (we are hiding somehting important with it). As far I could see, we always want to check if `"@_itemSubjectRef"]` exist before attributing it. If `["@_itemSubjectRef"]` doesn't exist its value will be left as `undefined`.

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -31,6 +32,95 @@ type WithRequiredDeep<T, K extends keyof any> = T extends undefined
...
+function remapProperties(
+  itemDefinitionIdsToRemap: Map<string, string>,
+  properties: Array<{ "@_itemSubjectRef"?: string }> | undefined
+) {
+  if (!properties) {
+    return;
+  }
+  for (const p of properties) {
+    p["@_itemSubjectRef"] = getRefToRemap(itemDefinitionIdsToRemap, p["@_itemSubjectRef"]);
+  }
```
ljmotta, line 60:
```suggestion
  for (const p of properties) {
    if (p["@_itemSubjectRef"]) {
        p["@_itemSubjectRef"] = itemDefinitionIdsToRemap.get(p["@_itemSubjectRef"]) ?? p["@_itemSubjectRef"]
    }
  }
```

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -31,6 +32,95 @@ type WithRequiredDeep<T, K extends keyof any> = T extends undefined
...
+}
+
+function remapFlowElements(
```
ljmotta, line 63: Please, consider using an `if/else if/else` structure to ensure all cases are covered. The `else` block can be empty with a comment informing that does not have an `itemSubjectRef` on the other elements (if it is the case).

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +179,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
+      rootElement["@_itemSubjectRef"] = getRefToRemap(itemDefinitionIdsToRemap, rootElement["@_itemSubjectRef"]);
+    }
+    if (rootElement.__$$element === "process") {
+      remapDataItems(itemDefinitionIdsToRemap, rootElement.ioSpecification?.dataInput);
+      remapDataItems(itemDefinitionIdsToRemap, rootElement.ioSpecification?.dataOutput);
+      remapProperties(itemDefinitionIdsToRemap, rootElement.property);
+      if (rootElement.flowElement) {
+        remapFlowElements(itemDefinitionIdsToRemap, rootElement.flowElement);
+      }
+    }
```
ljmotta, line 235: Please, consider making the same modification here with `if/else if/else` blocks. This list does not take into account the `@_type` property on `correlationProperty` element.

ljmotta: It would be good to include "message" elements too, which will start to use the `itemRef` after https://github.com/apache/incubator-kie-tools/pull/3555

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +179,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
     }
   });
 
+  // Merge itemDefinitions that share the same structureRef.
+  deduplicateItemDefinitions(model.definitions);
+
   const normalizedModel = model as Normalized<BpmnLatestModel>;
```
ljmotta: Consider using the `deduplicateItemDefinition` after the casting, as all elements from the `normalizedModel` have  `@_id` element

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +179,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
 
   return normalizedModel;
 }
+
+// Keep one itemDefinition per structureRef and update references to point to it.
+export function deduplicateItemDefinitions(definitions: BpmnLatestModel["definitions"]): void {
+  const itemDefinitionsGroupedStructureRef = new Map<string, Array<{ "@_id"?: string }>>();
```
ljmotta: By using the normalized model, you can have an `Array<string>`

```suggestion
export function deduplicateItemDefinitions(definitions: Normalized<BpmnLatestModel["definitions"]>): void {
  const itemDefinitionsGroupedStructureRef = new Map<string, Array<string>>();
```

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +179,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
+  // Map each duplicate id to the id of the first occurrence we keep.
+  const itemDefinitionIdsToRemap = new Map<string, string>();
+  for (const itemDefinitionsGroup of itemDefinitionsGroupedStructureRef.values()) {
+    if (itemDefinitionsGroup.length <= 1) {
+      continue;
+    }
+    const survivorId = itemDefinitionsGroup[0]["@_id"];
+    if (!survivorId) {
+      continue;
+    }
```
ljmotta: This should never happen.

packages/bpmn-editor/src/normalization/normalize.ts
```diff
@@ -89,7 +179,63 @@ export function normalize(model: BpmnLatestModel): State["bpmn"]["model"] {
...
+      }
+    }
+  }
+
+  if (itemDefinitionIdsToRemap.size === 0) {
+    return;
+  }
+
+  // Remove the duplicate itemDefinitions and update every itemSubjectRef.
+  const itemDefinitionIdsToRemove = new Set(itemDefinitionIdsToRemap.keys());
```
ljmotta: This isn't necessary, like a Set, a Map does not support two entries with the same key and it implements the `has` method.

ljmotta (changes requested):
@Abhitocode Thanks for this PR. I've gone through the code and it is looking good. Some of my comments are related to the legibility which could be improved. The method could use the `Normalized` model, this way you don't need to perform many `"@_id"` checks. Finally, I saw that the current implementation is not taking into account two rootElements.

Abhitocode:
Thanks for the review @ljmotta, makes sense, will address and push the changes.

ljmotta (approved):
@Abhitocode Thanks for addressing my previous comments. Code is looking great.

thiagoelg (approved):

PR #3581, opened 2026-05-14
kie-issues#2296: BPMN Editor: Process Variables - cannot close tag dropdown to add another tag without reopening panel

### LINKED ISSUE #2296: BPMN Editor: Process Variables - cannot close tag dropdown to add another tag without reopening panel
When editing process variables in the BPMN Editor, adding a tag via the "Type to add custom tag" dropdown works for the first tag, but after selecting a tag, the dropdown does not close or reset properly. This makes it not possible to add a second tag or interact with other tag fields in the same session.

Steps to Reproduce:

Open a BPMN file in the new BPMN Editor
Open Process Variables panel (left sidebar)
Add a process variable (e.g., hr_approval, type Boolean)
Click the Tags dropdown ("Type to add custom tag") and select or type a tag (e.g., internal)
Try to close the dropdown and add another tag, or try to interact with the Tags field of a different variable
The dropdown does not close to add another tag in the same instance. The only workaround is to close the Process Variables panel entirely and reopen it via the sidebar icon, then add the next tag.

------------------------------------------------------

While configuring Data Mapping in the BPMN Editor, the variable dropdown supports searching/filtering by typing part of the variable name.

Although the dropdown correctly filters matching variables, selecting a variable from the filtered results does not apply the selection. Instead, the dropdown resets to its previous state, and the selected variable is not set.

This prevents reliable variable mapping when using the search functionality.

Reproduction Steps:

Open BPMN Editor.
Create a process with multiple variables (e.g., 10+ variables).
Add a task.
Open Data Mapping.
In the Var dropdown, type part of a variable name to filter.
Click on one of the filtered results
Actual Result:

The dropdown resets.
The selected variable is not applied.
No error is shown.

Adarshvk98:
Closes [kie-issues#2296](https://github.com/apache/incubator-kie-issues/issues/2296)

Two bugs in the BPMN Editor's `TypeaheadSelect` component:

1. Process Variables : after selecting a tag, the dropdown would not close or reset, making it impossible to add a second tag without reopening the panel.

1. Data Mapping : typing to filter variables and clicking a result would reset the dropdown without applying the selection (Enter key worked fine).

Both were caused by broken state management in `TypeaheadSelect`. The component was refactored to fix the option filtering logic, close/reset behavior, and the blur/click race condition that occurs when the menu is portaled to document.body.

### DIFF
--- a/packages/bpmn-editor/src/typeaheadSelect/TypeaheadSelect.tsx
+++ b/packages/bpmn-editor/src/typeaheadSelect/TypeaheadSelect.tsx
@@ -35,6 +35,14 @@ import { useBpmnEditorI18n } from "../i18n";
 
 export type TypeaheadSelectOption = SelectOptionProps & { customLabel?: string | React.ReactElement };
 
+const hasRenderableOptionLabel = (option: TypeaheadSelectOption) => {
+  return (
+    !!option.customLabel ||
+    React.isValidElement(option.children) ||
+    (option.children !== undefined && option.children !== null && String(option.children).trim().length > 0)
+  );
+};
+
 // Based on https://v5-archive.patternfly.org/components/menus/select/#typeahead-with-create-option
 export function TypeaheadSelect({
   id,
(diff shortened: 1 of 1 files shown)


aswathitv (approved):
@Adarshvk98 Thank you for the PR.
Changes are working as expected for Tags dropdown issue in process variable panel and Var dropdown issue in data mapping.

rajalakshmys-27 (approved):

thiagoelg (approved):

PR #3582, opened 2026-05-15
NO-ISSUE: BPMN Editor: Auto-generated namespace has an extra "/"

rajalakshmys-27:
When opening a BPMN diagram in the editor and checking the **ID & Namespace** panel, the auto-generated namespace previously contained an extra `/` character before the GUID.  
For example:  
```
https://kie.apache.org/bpmn//_4E6C3E...
```

This has been corrected so that the namespace now follows the expected format:  
```
https://kie.apache.org/bpmn/_<GUID>
```
- Removed redundant `/` in auto-generated namespace.


thiagoelg (approved):

jomarko (approved):
Changes looks good to me.

PR #3587, opened 2026-05-18
NO-ISSUE: BPMN Editor: Classic to New Migration Guide

rajalakshmys-27:
The transition from the Classic BPMN Editor (Stunner/GWT-based, last available in Apache KIE 10.1) to the New BPMN Editor (React/ReactFlow-based, available from Apache KIE 10.2+) introduces significant architectural and UI/UX changes that extend beyond the previous editor's paradigms. 

Adds `packages/bpmn-editor/docs/MIGRATION_GUIDE.md`, a complete migration reference covering UI/UX differences, a step-by-step walkthrough recreating an Employee Onboarding process in both editors, side-by-side comparisons, common migration issues with solutions.


klbowers (changes requested):
Amazing job on this, @rajalakshmys-27 thank you! The guide is very detailed and clear.

One thing I’d suggest is standardizing the terminology to avoid confusion. For example, the document refers to the same UI area as the “left palette,” “Pallete panel,” and “palette.” Using a single term consistently throughout for this and other terminology would make the guide feel more polished and easier to follow.

I’d also try to keep the workflows similar between the Classic and New Editor examples when possible. For instance, instead of describing the Classic Editor flow as expanding “Tasks” in the palette and dragging a User Task to the canvas, you could align both examples around dragging a Task onto the canvas and converting it into a User Task. That makes the comparison between editors a bit clearer.

rajalakshmys-27:
Thanks @kbowers-ibm! About the standard terminology, I've updated the PR with the changes. 
As for the second point, the Classic Editor requires expanding "Tasks" in the palette panel to drag specific types like "User Task" directly, while the New Editor uses drag-then-morph. Since these reflect how each editor actually works, I've kept them documented separately.

klbowers (approved):
Thanks @rajalakshmys-27, LGTM! 👍

The expanding tasks vs. morphing point was just a minor nitpick from my side so it's fine as is. However just so you know for future reference the "morphing/converting" functionality is possible in both editors!
https://github.com/user-attachments/assets/895be788-4baa-4a4b-99b0-876b3de68157

rajalakshmys-27:
Thanks for the clarification @kbowers-ibm! I have updated the PR to reflect the correct drag-and-drop behaviour.

packages/bpmn-editor/docs/images/migration-guide/new-editor-process-properties.png
jomarko, line 1: I think we have fixed the issue of double slashes in namespace already, could we update this screenshot please?

packages/bpmn-editor/docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,669 @@
...
+| Feature             | Classic Editor         | New Editor                           |
+| ------------------- | ---------------------- | ------------------------------------ |
+| **Multi-Selection** | Ctrl+Click or drag box | Drag box                             |
+| **Bulk Operations** | Limited                | Delete, copy, move multiple elements |
+
+**Note**: Properties panel shows "Multiple nodes selected (N)" when multiple elements are selected. Cannot edit properties when both nodes and edges are selected.
+
+---
+
+### 8. Global Properties
```
jomarko: I think we have wrong indexes
5.
8.

should be
5.
6.

probably

packages/bpmn-editor/docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,669 @@
...
+
+**New Editor**: Click Correlations icon (left sidebar, second icon). Use Properties tab for Correlation Properties and Message Bindings. Use Keys tab for Correlation Keys and Subscriptions.
+
+### Defining Reusable Properties
+
+**New Editor**: Click Properties Management icon (left sidebar, third icon). Tabs available for Data Types, Messages, Signals, Escalations, and Errors.
+
+---
+
+## Step-by-Step Walkthrough: Recreating a BPMN Process
```
jomarko, line 189: This step by steps very valuable and has very good readability and structure as it follows the Pattern

- **Classic**
- steps
- picture
- **New**
- steps
- picture


Could we maybe try to unify the comparison also in other sections, somewhere we use a table to compare **Classic** and **New** (for example **Global properties)** somewhere we use text only, but mention only one of ****New**/Classic** (for example **Keyboard shortcuts)**

jomarko (commented):
Thank you for the PR. I put some minor comments inline.

packages/bpmn-editor/docs/images/migration-guide/new-editor-process-properties.png
rajalakshmys-27: sure

packages/bpmn-editor/docs/MIGRATION_GUIDE.md
rajalakshmys-27: sure

packages/bpmn-editor/docs/MIGRATION_GUIDE.md
rajalakshmys-27: Thanks for the suggestion! I agree consistency helps, but for the Step‑by‑Step Walkthrough: Recreating a BPMN Process, screenshots are necessary so readers can actually follow and reproduce the steps. Adding images everywhere would make the doc bulky, so I’d propose keeping screenshots only in walkthroughs, and using tables or concise text for sections like Global properties or Keyboard shortcuts to keep things balanced. WDYT?

jomarko: sure, I am sorry I was not clear enough in previous comment. I totally agree it is not needed to add addition screenshots.

Image

my point is in the attached, on very small area we mix 3 approaches

- pink compares using text **Classic** on left side, **New** on right side
- green compares using tables
- blue does not compare, has only **New**

I think we should compare always either by **Table** for example as we do in 1,2, 5 or 6, or by two text sections. positioned probably like 

```
**Classic**
lorem ipsum text

**New**
Lorem ipsum text
```

is my point more clean now?

rajalakshmys-27: Yes, much clearer thanks! And it's sorted.

thiagoelg (approved):

PR #3588, opened 2026-05-18
NO-ISSUE: Use pom.xml template in maven-base so new Kogito version are propagated to Gradle modules

thiagoelg:
With the initial change in https://github.com/apache/incubator-kie-tools/pull/3541 I introduced the support for overriding the Kogito version without changing the pom.xml. This was enough for Maven-based modules, but it fails for Gradle-based modules, as Gradle can't override properties from transitive dependencies.

This change removes the maven-base pom.xml in favor of a pom.xml.template file. The install script creates the pom.xml from the pom.xml.template and replaces the Quarkus and Kogito versions in it.

The pom.xml is ignored by git, so any changes from forks are not propagated to PRs.


examples/process-security/docker-compose/pgadmin/pgpass
```diff
@@ -1,6 +1,3 @@
-postgres:5432:bamoe1:kie-user-1:kie-pass-1
-postgres:5432:bamoe2:kie-user-2:kie-pass-2
-postgres:5432:bamoe3:kie-user-3:kie-pass-3
```
yesamer, line 3: @thiagoelg Is this related?

thiagoelg: Not really, just some leftovers I found from a while ago

yesamer (commented):
How this is related with Gradle? @thiagoelg

thiagoelg:
> How this is related with Gradle? @thiagoelg

For downstream packages that use Gradle, overriding the `version.org.kie.kogito` version is not sufficient, as POM resolution occurs before any Gradle processing. That means that the `version.org.kie.kogito` value in the kie-tools-maven-base pom.xml is used instead of Maven config overrides (via `-Dversion.org.kie-kogito`) or any other Gradle setting.

I get some error like this:
```
quarkusProdBaseRuntimeClasspathConfigurationCopy1
Could not parse POM ...
```
This suggests the Quarkus Gradle plugin resolves things before Gradle, and Gradle can't do anything about it.

yesamer (approved):

ljmotta (approved):

PR #3604, opened 2026-05-26
kie-issues#118: Upgrade React from v17 to v18 on kie-tools

Kusuma04-dev:
This is part of partial react upgrade task which has react v17 in boxed-expression-component


Kusuma04-dev:
Not needed!

PR #3607, opened 2026-05-27
kie-issues#2309: BPMN Editor: Implementation of REST Service Call Task

aswathitv:
Part of [kie#2309]( https://github.com/apache/incubator-kie-issues/issues/2309)

An enhanced, Kogito REST WIH will start being fully supported and continue to be available. As part of this following changes are made:

A custom task node has been added.

A dedicated properties panel has been included with options for URL, method, protocol, host, port, request timeout, headers, query parameters and access token acquisition strategy.

An option to emulate testing from a sandbox has been provided.

For the access token acquisition strategy:

If configured, a REST service call task ID and test token are provided for testing.
If propagated, a test token is provided for testing.
None can be tested directly.

Continuation of [3573](https://github.com/apache/incubator-kie-tools/pull/3573) 

image


Adarshvk98 (commented):
Hi @aswathitv, I’ve tested it and have a few suggestions. Feel free to consider them.

I noticed that the URL input in the inputGroup overlaps with the method dropdown. Additionally, when there’s no value selected in the method dropdown, it displays `<Undefined>`. Is this behavior expected?

Image

---

I notice a significant amount of space on the right side of the input fields, such as Protocol, Host, Port, and Request Timeout (ms). Additionally, the input values for these fields are relatively short in length. Therefore, I suggest that we consider splitting them into two rows, with each row containing two items. For instance, Protocol and Request Timeout (ms) could be placed in the same row, while Host and Port could be placed in the 2nd row. I believe this layout would reduce the user’s effort in scrolling down the input fields. What are your thoughts on this suggestion? 

Image

---

Regarding the test result, it only returns the test result as HTTPS status. Is this intended? It would be better to display the test response as well. What are your thoughts?

As of now, it only displays the HTTP status code, as shown below.

Image

Image

aswathitv:
Thanks @Adarshvk98 for the suggestions.
1. The dropdown is consistent with other dropdowns in the sandbox.
2. Grouping the inputs make them very much cluttered.
image

3. Added option to show response

jomarko (commented):
Thank you for all the changes. Issues I have reported in previous comments are fixed now I think.

However I agree with the point about the **Method** select dropdown raised by @Adarshvk98 [here](https://github.com/apache/incubator-kie-tools/pull/3607#pullrequestreview-4373971029).

The **Method** is mandatory and click on the clear button is quite difficult to understand. What is meaning of such action? Because it always needs to be some HTTP method when we speak about REST.

I had a look on the implementation https://github.com/apache/incubator-kie-tools/blob/ca1506dbf3187511d1bc0f0a488d94ba0675b958/packages/bpmn-editor/src/typeaheadSelect/TypeaheadSelect.tsx

I think we could extend `TypeaheadSelectOption`, att there some optional `hideClearIcon: boolean` and then we could adapt

```
<TextInputGroupUtilities {...(!inputValue ? { style: { display: "none" } } : {})}>
            <Button variant="plain" onClick={onClearButtonClick} aria-label="Clear input value">
              <TimesIcon aria-hidden />
            </Button>
          </TextInputGroupUtilities>
```

to be 

```
<TextInputGroupUtilities {...((options.hideClearIcon || !inputValue) ? { style: { display: "none" } } : {})}>
            <Button variant="plain" onClick={onClearButtonClick} aria-label="Clear input value">
              <TimesIcon aria-hidden />
            </Button>
          </TextInputGroupUtilities>
```

What do you think?

rajalakshmys-27 (approved):
LGTM! Tested with dummy apis.

packages/bpmn-editor-envelope/src/customTasks/RestServiceCallTask.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import * as React from "react";
+import "@kie-tools/bpmn-marshaller/dist/drools-extension";
```
thiagoelg: What is this importing?

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
thiagoelg, line 1: Image

The placeholders could confuse users. They show real values, and it's a bit hard to tell whether they are filled or not.

Maybe the fields should not have placeholders, and examples should be added to the input description (below the input).

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,1126 @@
...
+                            elementId: task["@_id"],
+                            element: task.__$$element,
+                          });
+                        });
+                      } else {
+                        updateRestProperties(RestProperties.Method, String(value));
+                      }
+                    }
+                    setIsMethodDropdownOpen(false);
+                  }}
```
thiagoelg: Move this to a useCallback

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,1126 @@
...
+                  aria-describedby="rest-content-data-helper"
+                />
+                <FormHelperText>
+                  <HelperText>
+                    <HelperTextItem variant="indeterminate">{i18n.restService.contentDataHelperText}</HelperTextItem>
+                  </HelperText>
+                </FormHelperText>
+              </FormGroup>
+
+              {contentDataVariables.length > 0 && (
```
thiagoelg: I think the Test Variables should be placed right above the "Test Request" button.

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,1126 @@
...
+  const [queryParams, setQueryParams] = React.useState<QueryParameter[]>([]);
+  const [contentDataVariables, setContentDataVariables] = React.useState<ContentDataVariable[]>([]);
+  const [contentDataValue, setContentDataValue] = React.useState<string>("");
+  const headerIdMapRef = React.useRef<Map<string, string>>(new Map());
+  const queryIdMapRef = React.useRef<Map<string, string>>(new Map());
+  const entryStyle = {
+    padding: "4px",
+    margin: "8px",
+    width: "calc(100% - 2 * 4px - 2 * 8px)",
+  };
```
thiagoelg, line 133: This is used in several different inputs. I guess using Patternfly components should let you remove this styling.

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,1126 @@
...
+          <FormGroup label={i18n.restService.requestTimeout} fieldId="rest-timeout">
+            <TextInput
+              id="rest-timeout"
+              type="number"
+              value={localRequestTimeout}
+              onChange={(_, value) =>
+                handleTextInputChange(RestProperties.RequestTimeout, value, setLocalRequestTimeout)
+              }
+              placeholder="30000"
+            />
```
thiagoelg, line 745: What happens if this field is left empty? The runtime will use a default timeout, or will it be set to 0?

packages/bpmn-editor/src/propertiesPanel/dataMapping/DataMappingFormSection.tsx
```diff
@@ -125,7 +125,12 @@ function flaggedDataInputs<T extends { name?: string; "@_name"?: string }>(
         !!customTasks.find(
           (c) =>
             c.matches(element) && !!c.dataInputReservedNames.find((n) => n === (dataInput["@_name"] ?? dataInput.name))
-        )),
+        )) ||
+      // Hide HEADER_ and QUERY_ prefixed parameters for custom tasks
+      (element.__$$element === "task" &&
+        !!customTasks.find((c) => c.matches(element) && c.id === "rest-task") &&
+        (!!(dataInput["@_name"] ?? dataInput.name)?.startsWith("HEADER_") ||
+          !!(dataInput["@_name"] ?? dataInput.name)?.startsWith("QUERY_"))),
```
thiagoelg: I don't think this is the right place for this. The REST Task is implemented in the bpmn-editor-envelope, so it shouldn't be referenced in the BPMN Editor.

An alternative solution would be if the `dataInputReservedNames` supported the `*` wildcard. So you could have `HEADER_*` and `QUERY_*` entries in the `dataInputReservedNames` array, and then you would check for the `*` wildcard and use `startsWith`/`endsWith` comparison.

Line **127** would become something like:
```
c.matches(element) && !!c.dataInputReservedNames.find((n) => {
  dataInputName = dataInput["@_name"] ?? dataInput.name;
  if (n.startsWith("*")) {
    const suffix = n.slice(0, -1);
    return dataInputName.endsWith(suffix);
  } else if (n.endsWith("*")) {
    const prefix = n.slice(1);
    return dataInputName.starsWith(prefix);
  } else {
    return n === dataInputName;
  }
})
```

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,1126 @@
...
+  const [localHost, setLocalHost] = React.useState<string>("");
+  const [localPort, setLocalPort] = React.useState<string>("");
+  const [localRequestTimeout, setLocalRequestTimeout] = React.useState<string>("");
+  const [localRestServiceCallTaskId, setLocalRestServiceCallTaskId] = React.useState<string>("");
+
+  const dataInputVariables = React.useMemo(
+    () =>
+      inputDataMapping
+        ?.filter((data) => data.name && !(REST_PROPERTIES_KEYS as readonly string[]).includes(data.name))
+        ?.map((v) => v?.name)
```
thiagoelg: What is `v`?

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,1126 @@
...
+  variableName: string;
+  variableValue: string | null;
+};
+
+export const RestServiceTaskPropertiesPanel: CustomTask["propertiesPanelComponent"] = ({ task }) => {
+  const i18n = new I18n(bpmnEditorEnvelopeI18nDefaults, bpmnEditorEnvelopeI18nDictionaries).getCurrent();
+  const bpmnEditorStoreApi = useBpmnEditorStoreApi();
+  const { inputDataMapping, outputDataMapping } = useDataMapping(task, () => {});
+  const envelopeContext = useKogitoEditorEnvelopeContext();
+  const channelApi = envelopeContext.channelApi as unknown as MessageBusClientApi<BpmnEditorChannelApi>;
```
thiagoelg:
```suggestion
  const { channelApi } = useKogitoEditorEnvelopeContext();
```

Why is it necessary to do type casting here? `channelApi` should already have the proper type.

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,1126 @@
...
+
+      if (contentData && ["POST", "PUT", "PATCH"].includes(method) && !requestHeaders["Content-Type"]) {
+        const trimmedData = contentData.trim();
+        if (trimmedData.startsWith("{") || trimmedData.startsWith("[")) {
+          requestHeaders["Content-Type"] = "application/json";
+        } else if (trimmedData.startsWith("#{")) {
+          requestHeaders["Content-Type"] = "application/json";
+        } else {
+          requestHeaders["Content-Type"] = "application/json";
+        }
```
thiagoelg:
```suggestion
        requestHeaders["Content-Type"] = "application/json";
```

These comparisons are not needed if the `else` statement always set the Content-Type to "application/json"

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
```diff
@@ -0,0 +1,1126 @@
...
+            headerIdMapRef.current.delete(oldKey);
+          }
+          if (value.trim()) {
+            const newKey = `HEADER_${value}`;
+            headerIdMapRef.current.set(newKey, id);
+          }
+        }
+
+        const updated = prev.map((h) => (h.id === id ? { ...h, [field]: value } : h));
+        updateParameterMapping("HEADER_", updated);
```
thiagoelg: `HEADER_` and `QUERY_` should be constants.

thiagoelg (changes requested):
Thanks @aswathitv! This is looking better!

I did a new round of review and left some comments inline.

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
aswathitv: This is the same style used in Metadata.tsx and Import.tsx. Patternfly is not available

packages/bpmn-editor-envelope/src/customTasks/RestServiceTaskPropertiesPanel.tsx
aswathitv: BPMN will not have an entry for timeout

packages/bpmn-editor-envelope/src/i18n/locales/en.ts
```diff
@@ -56,4 +56,73 @@ export const en: BpmnEditorEnvelopeI18n = {
...
+    httpMethodDelete: "DELETE",
+    host: "Host",
+    hostHelp: "Alternative to full URL (e.g., api.example.com)",
+    hostRequiredError: "Host is required when URL is not complete",
+    port: "Port",
+    portHelp: "Alternative to full URL (e.g., 443)",
+    contentData: "Content Data",
+    contentDataHelp: "Literal, JSON or process variable #{var}",
+    testVariables: "Test Variables",
+    testVariablesHelp: "Provide mock values for process variables used in the content data for the REST endpoint.",
```
thiagoelg:
```suggestion
    testVariablesHelp: "Provide test values for process variables used in the content data for the test request.",
```

thiagoelg (approved):

thiagoelg:
Great work @aswathitv! I appreciate your patience.

rajalakshmys-27 (approved):

yesamer (approved):

PR #3580, opened 2026-05-14
kie-issues#2312: BPMN Editor: Cannot nest Sub-processes within Sub-processes

rajalakshmys-27:
Closes : [kie-issues#2312](https://github.com/apache/incubator-kie-issues/issues/2312)

Drag and drop of subprocesses into other subprocesses was completely broken. Users could not nest subprocesses at any level.

**Root Cause**

The primary issue was that the containment map in `BpmnDiagramDomain.tsx` was missing `NODE_TYPES.subProcess` in the subprocess containment configuration, preventing the editor from recognizing that subprocesses can contain other subprocesses.

Several secondary issues compounded the problem. Mutation functions only searched for subprocesses at the top level of `process.flowElement` rather than traversing nested structures recursively. Diagram rendering used type-based node sorting, which did not guarantee correct parent-child ordering for nested subprocesses. No recursive traversal logic existed to handle multi-level nesting.

**Changes**

Added `NODE_TYPES.subProcess` to the subprocess containment map, allowing subprocesses to contain other subprocesses in accordance with the BPMN 2.0 specification.

Introduced `findSubProcessRecursively()` to locate subprocesses at any nesting depth by traversing the full subprocess tree, and `collectElements()` to gather flow elements from all nesting levels when moving nodes between subprocesses.

Replaced type-based node sorting with depth-based sorting so React Flow always processes parent nodes before their children, regardless of nesting depth.

**Testing**

Verified that subprocesses can be dragged into other subprocesses, that multi-level nesting works at three or more levels, that drag and drop between deeply nested subprocesses functions correctly, that React Flow parent-child relationships are maintained, that BPMN XML document order is preserved, and that no regressions were introduced to existing functionality.

image

image

PR #3609, opened 2026-05-31
NO-ISSUE: Improve and refactor subprocess mutations

rajalakshmys-27:
This is a follow-up to #3580 ([kie-issues#2312](https://github.com/apache/incubator-kie-issues/issues/2312)), addressing review comments. 
Removes redundant `Normalized<>` wrappers from `flowElementsToMove` and `artifactsToMove` array element types in both `moveNodesInsideSubProcess.ts` and `moveNodesOutOfSubProcess.ts`. 
Refactors `collectElements()` to accept the collection array as an explicit parameter instead of relying on closure mutation, and renames the `spliced` variable to `removedArtifacts` for clearer semantic intent.


ljmotta (approved):
@rajalakshmys-27 Thanks for addressing the previous review. LGTM

thiagoelg (approved):

PR #3610, opened 2026-05-31
NO-ISSUE: Fix SCESIM story book due to stricter xml-parser-ts validation

rajalakshmys-27:
After the `xml-parser-ts` fix (#3570) that filters `<parsererror>` elements and throws on malformed XML was merged (as part of the broader BPMN/DMN/SCESIM parser error fix), the SCESIM editor Storybook stories broke in local development.

**Root Cause**

The Storybook stories for the SCESIM editor had two latent bugs that were always present but previously silent:

1. **Malformed XML in the SIMPLE model** — a tag like `<text>Input &gt; 18<text>` was missing the `/` in its closing tag, making it invalid XML.
2. **DevWebApp story passing a filename string instead of XML content** — the story was calling `getMarshaller(emptyFileName).parser.parse()` where `emptyFileName` is `"Untitled.scesim"`, a filename string, instead of actual XML content. `getMarshaller()` expects XML, so the DOMParser tried to parse the filename as XML and silently failed.

These bugs never surfaced before because the browser's `DOMParser` was producing empty `<parsererror>` elements for these inputs, and the old code only threw when `hasRealError` was true (i.e., the parsererror element had non-empty text content). The stricter error handling introduced in the parser fix now correctly detects and throws on these malformed inputs.

### DIFF
--- a/packages/scesim-editor/stories/examples/ExternalDmnModels.ts
+++ b/packages/scesim-editor/stories/examples/ExternalDmnModels.ts
@@ -1012,7 +1012,7 @@ export const SIMPLE = `<?xml version="1.0" encoding="UTF-8" ?>
       <requiredInput href="#_926B9E31-3CEA-4A0E-8A66-97F875EFB428" />
     </informationRequirement>
     <literalExpression id="_1E199E0A-2CC3-4EEE-AA8C-B4F5320EE1DE">
-      <text>Input &gt; 18<text>
+      <text>Input &gt; 18</text>
     </literalExpression>
   </decision>
   <dmndi:DMNDI>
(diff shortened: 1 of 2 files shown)


packages/scesim-editor/stories/examples/ExternalDmnModels.ts
```diff
@@ -1012,7 +1012,7 @@ export const SIMPLE = `<?xml version="1.0" encoding="UTF-8" ?>
       <requiredInput href="#_926B9E31-3CEA-4A0E-8A66-97F875EFB428" />
     </informationRequirement>
     <literalExpression id="_1E199E0A-2CC3-4EEE-AA8C-B4F5320EE1DE">
-      <text>Input &gt; 18<text>
```
thiagoelg, line 1015: Was this XML originally generated by the SCESIM Editor? If it was, then the editor will always generate malformed XML 😕

yesamer: @thiagoelg This is a DMN file not generated by scesim, I guess this is a typo

yesamer (approved):

ljmotta (approved):

thiagoelg (approved):

PR #3611, opened 2026-06-03
NO-ISSUE: update picomatch versions to address CVE-2026-33671 and CVE-2026-3367

Adarshvk98:
Updated picomatch package versions to `3.0.2` and `4.0.4` address security vulnerabilities CVE-2026-33671 and CVE-2026-33672.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -30,8 +30,6 @@ overrides:
   # CVE-2026-41907: Fix security vulnerability in uuid
   # Transitive dependencies (node-notifier@8.0.2, jest-junit@16.0.0) still use uuid@8.3.2
   "uuid": "^11.1.1"
-  # CVE-2026-33671 (High) / CVE-2026-33672 (Medium): picomatch POSIX bracket method injection.
-  # Transitive dep of vite, @angular/build, @parcel/watcher, fdir, tinyglobby (4.x) and pretty-quick (3.x).
-  # The 2.x line already resolves to the patched 2.3.2.
-  "picomatch@3": "3.0.2"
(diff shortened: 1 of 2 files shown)


Kusuma04-dev (approved):

pnpm-workspace.yaml
```diff
@@ -29,3 +30,8 @@ overrides:
   # CVE-2026-41907: Fix security vulnerability in uuid
   # Transitive dependencies (node-notifier@8.0.2, jest-junit@16.0.0) still use uuid@8.3.2
   "uuid": "^11.1.1"
+  # CVE-2026-33671 (High) / CVE-2026-33672 (Medium): picomatch POSIX bracket method injection.
+  # Transitive dep of vite, @angular/build, @parcel/watcher, fdir, tinyglobby (4.x) and pretty-quick (3.x).
+  # The 2.x line already resolves to the patched 2.3.2.
+  "picomatch@3": "3.0.2"
+  "picomatch@4": "4.0.4"
```
thiagoelg: I don't think the override is needed. If you remove these and run `pnpm bootstrap` again, the updated versions should stick to the lockfile since they are only patch-version upgrades.

thiagoelg (changes requested):

pnpm-workspace.yaml
Adarshvk98: Thanks for the review! I tested this by removing the overrides and running `pnpm bootstrap`, and unfortunately the vulnerable versions (3.0.1 and 4.0.3) came back to the lockfile.

I think we need to keep both overrides to prevent the vulnerable versions from sneaking back

thiagoelg (approved):

PR #3633, opened 2026-06-28
NO-ISSUE: Upgrade pnpm to version 10.34.4

thiagoelg:
Upgraded pnpm version from `10.29.2` to `10.34.4`.


porcelli (approved):

thiagoelg:
Build is failing because some artifacts from the Eclipse JDTLS repository were deleted: https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3823

tiagobento:
@yesamer Is there a way for us to work around these failures? It was good while it lasted 😅

yesamer:
@tiagobento Using the snapshot version of 1.41. could be an option. However, in the thread the maintainers said they are going to restore the directory, so I would wait a couple of days before applying a workaround, if possible.

yesamer (approved):

PR #3635, opened 2026-07-01
kie-issues#3634: BPMN Editor: Add processDuration field to the Global Properties

Kusuma04-dev:
closes https://github.com/apache/incubator-kie-tools/issues/3634,

Added processduration to global properties

PR #3640, opened 2026-07-02
kie-issues#2356: BPMN Editor: Subprocess with an Error Start Event causes BPMN code generation to fail

### LINKED ISSUE #2356: BPMN Editor: Subprocess with an Error Start Event causes BPMN code generation to fail
When modelling an Event Subprocess with an Error Start Event using the new BPMN Editor (VS Code Developer Tools), BPMN code generation fails with the following error:

 

```
org.kie.kogito.codegen.process.ProcessCodegenException:

Processes with errors are Invalid process

Caused by: org.jbpm.bpmn2.xml.ProcessParsingValidationException:

Could not find source node for connection:<ErrorStartEventId>
```

If we create the same using classic editor, issue doesn't occur.

SUNILathulya:
Closes: https://github.com/apache/incubator-kie-issues/issues/2356

## Problem
Event Subprocess with Error Start Event fails with "Could not find source node for connection" error.

## Root Causes
1. **Missing Event References**: Event definitions (error, signal, message, escalation) lacked required reference attributes like `errorRef`, causing nodes to be invalid and not registered in the process model.
2. **Wrong Scope**: SequenceFlows were placed at process level instead of inside the subprocess, so runtime couldn't find the connected nodes.

## Fix
- Added proper event definition references in `useEventNodeMorphingActions.tsx`
- Implemented container detection in `addEdge.ts` to place flows in correct scope

## Testing
✅ Canvas tested - no errors
✅ BPMN XML validated - correct structure

### DIFF
--- a/packages/bpmn-editor/src/diagram/nodes/morphing/useEventNodeMorphingActions.tsx
+++ b/packages/bpmn-editor/src/diagram/nodes/morphing/useEventNodeMorphingActions.tsx
@@ -30,6 +30,10 @@ import { BPMN20__tProcess } from "@kie-tools/bpmn-marshaller/dist/schemas/bpmn-2
 import { ElementFilter } from "@kie-tools/xml-parser-ts/dist/elementFilter";
 import { NODE_COLORS } from "../NodeSvgs";
 import { deleteEdge } from "../../../mutations/deleteEdge";
+import { addOrGetErrors } from "../../../mutations/addOrGetErrors";
+import { addOrGetSignals } from "../../../mutations/addOrGetSignals";
+import { addOrGetMessages } from "../../../mutations/addOrGetMessages";
+import { addOrGetEscalations } from "../../../mutations/addOrGetEscalations";
 import { setBpmn20Drools10MetaData } from "@kie-tools/bpmn-marshaller/dist/drools-extension-metaData";
 
 export type Event = Normalized<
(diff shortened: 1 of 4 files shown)


jomarko (commented):
Thank you for the changes.

### Manual review
I created a diagram using this PR. I was able to define error start event

Image

but once I move the start event it is reset to a default one

Image


### Code review
To stay consistent in PRs, I think we should add tests. Same requirement for tests is common also in other PRs, see https://github.com/apache/incubator-kie-tools/pull/3635#pullrequestreview-4619086949

SUNILathulya:
> Thank you for the changes.
> 
> ### Manual review
> I created a diagram using this PR. I was able to define error start event
> 
> Image
> but once I move the start event it is reset to a default one
> 
> Image
> ### Code review
> To stay consistent in PRs, I think we should add tests. Same requirement for tests is common also in other PRs, see [#3635 (review)](https://github.com/apache/incubator-kie-tools/pull/3635#pullrequestreview-4619086949)

Thank You for your review, will check this issue and add required test cases.

SUNILathulya:
@jomarko , Thankyou for the review !

Issue reproduced and fixed. The start event type is now preserved when moving a start event within an event sub-process, and I also added a regression test covering this scenario in addStartEvent.spec.ts. The fix is available in the commit below:
[ae2148a](https://github.com/apache/incubator-kie-tools/pull/3640/commits/ae2148a15741532a49e286020b2f0faf8baa8482)

jomarko (approved):
Thank you for the PR.

Kusuma04-dev (approved):

thiagoelg (approved):

PR #3652, opened 2026-07-08
kie-tools#3845: [BPMN Editor] Re-anchor connections to the facing border side when moving nodes

### LINKED ISSUE #3845: [BPMN Editor] Sequence flow does not re-anchor to the nearest border when a node is moved
**Describe the bug**
In the BPMN Editor, when a node connected by a sequence flow is repositioned, the connection's attachment point on that node is not recalculated. The edge stays anchored to its original spot instead of the border side facing the connected node. It's most visible with events (circular nodes), where the connector clearly meets the shape off-center, but it affects all node types.

**Steps to reproduce**
1. Add a start event and a task, and connect them with a sequence flow.
2. Move the event so it sits below, then above, then beside the task. 
3. Second, move the anchor point on the event to a different position, then try moving it around the connected task again.
4. Watch where the connection meets each shape.

**Expected behavior**
The endpoint anchors to the border side facing the connected node — top/bottom when above/below, left/right when beside — and updates as the node crosses between those zones.

**Actual behavior**
The anchor keeps its original position and does not re-optimize as the node moves, so the connection meets the shape at an awkward, off-center point.

**Screenshot**
Image

ch-lukas:
Closes: https://github.com/apache/incubator-kie-tools/issues/3845

Moving a node in the BPMN editor doesn't recalculate where its sequence flows connect, so the line often ends up meeting the shape at an odd spot (usually a corner of the bounding box) rather than the side facing the other node. It's most obvious on events, since they're circles and the connection clearly lands off-centre.

The fix picks the anchor from where the connected node sits relative to this one (left, right, above or below) and attaches to the middle of that side. That's the same dominant-axis logic the auto-positioned edges already use, so I reused it instead of writing a second copy.

On a move it only re-picks the side when the node actually crosses into a different zone. So if you've dragged an anchor onto a particular side and then nudge the node a little, it stays where you put it; move it far enough that it's now on a different side of the target and it snaps across. It covers all node types, and because the side midpoints fall on an event's circle it lines up on events too.

Roughly where the changes are:

- `xyflow-react-kie-diagram`: the auto-anchored (Center) end uses a small `getBorderMidpointTowards` helper that wraps the existing `getPositionalHandlePosition`.
- `bpmn-editor` `repositionNode`: on a move it compares the node's zone before and after, and if it changed, resets that endpoint back to the node centre (which renders as auto) instead of dragging it along.

To try it:

1. Connect an event to a task and move the event below, above and beside it.
2. Drag an endpoint to pin it, nudge the node a bit (should stay), then move it to another side (should re-optimise).

Happy to take feedback on the zone rule and how the pinned-anchor behaviour feels.

- [x] Linked to a GitHub issue
- [x] `pnpm bootstrap` run (formatting + lockfile)
- [ ] `pnpm -F @kie-tools/bpmn-editor... build:prod` passes
- [ ] Tests pass (unit + E2E)
- [x] License headers on new files (n/a, no new files)

### DIFF
--- a/packages/bpmn-editor/tests-e2e/flowElements/connectionReanchorOnNodeMove.spec.ts
+++ b/packages/bpmn-editor/tests-e2e/flowElements/connectionReanchorOnNodeMove.spec.ts
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { test, expect } from "../__fixtures__/base";
+import { JsonModel } from "../__fixtures__/jsonModel";
+import { NodeType } from "../__fixtures__/nodes";
+
+// 0 = node's left border, 0.5 = centre (auto-anchored), 1 = right border.
+async function getEndpointRelativeX(args: {
+  jsonModel: JsonModel;
+  nodeId: string;
+  direction: "incoming" | "outgoing";
+}): Promise<number> {
+  const flows = await args.jsonModel.getSequenceFlows();
+  const flow =
+    args.direction === "incoming"
+      ? flows.find((f) => f["@_targetRef"] === args.nodeId)
+      : flows.find((f) => f["@_sourceRef"] === args.nodeId);
+  expect(flow).toBeTruthy();
+
+  const elements = (await args.jsonModel.getPlane())?.["di:DiagramElement"] ?? [];
+  const shape = elements.find(
+    (e: any) => e.__$$element === "bpmndi:BPMNShape" && e["@_bpmnElement"] === args.nodeId
+  ) as any;
+  const edge = elements.find(
+    (e: any) => e.__$$element === "bpmndi:BPMNEdge" && e["@_bpmnElement"] === flow!["@_id"]
+  ) as any;
+
+  const bounds = shape["dc:Bounds"];
+  const waypoints = edge["di:waypoint"];
+  const endpoint = args.direction === "incoming" ? waypoints[waypoints.length - 1] : waypoints[0];
+
+  return (endpoint["@_x"] - bounds["@_x"]) / bounds["@_width"];
+}
+
+test.beforeEach(async ({ page, baseURL }) => {
+  await page.goto(`${baseURL}/iframe.html?args=&id=misc-connection-reanchor--start-to-end&viewMode=story`);
+  await expect(page.getByTestId("kie-bpmn-editor--diagram-container")).toBeVisible();
+});
+
+test.describe("Connection - re-anchor on node move", () => {
+  test("should re-optimise both incoming and outgoing connections when the task crosses into a different zone", async ({
+    nodes,
+    jsonModel,
+  }) => {
+    const taskId = await nodes.getId({ name: "First Function" });
+
+    // Nudge within the same zones - pinned anchors are kept (also commits a change so the model reads).
+    const box = await nodes.getNodeBounds({ name: "First Function" });
+    await nodes.dragNodeToPosition({
+      name: "First Function",
+      toPosition: { x: box.x + box.width / 2 + 40, y: box.y + box.height / 2 },
+    });
+    expect(await getEndpointRelativeX({ jsonModel, nodeId: taskId, direction: "incoming" })).toBeLessThan(0.2);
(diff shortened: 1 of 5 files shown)


tiagobento:
@rajalakshmys-27 FYI

rajalakshmys-27:
@ch-lukas, thank you for the PR, this is a great fix and the approach described in the PR makes sense. I'm still working through the full review, but I noticed a few E2E visual regression test failures (9) that should be addressed (https://github.com/apache/incubator-kie-tools/actions/runs/28959843229/job/85987009315?pr=3652)

These failures appear to be related to the anchoring changes introduced in this PR. Could you please review the screenshot diffs and, if the visual changes are expected, update the snapshots accordingly and ensure the tests pass.

jomarko:
image

Thank you for contribution, let me understand more the fix.

When creating an connection, user has 5 options to choose to choose 'magnet' point.

### Originally
we had this behavior, 4 blue points on the boundary of task are 'hard'. The edge target is always at that place. When user move the task node anywhere. However the blue magnet point in middle of the task was intended to render the connection in most optimal (shortest as possible) way.

In the attached below the top task is connected via 'hard' boundary magnet point and the bottom node used the center blue magnet point for connection creation.

image


### Question
- Is your PR changing behavior for all magnet points?
- Is your PR changing behavior for specific magnet point?
- Do we need all existing magnet points after your PR?

ch-lukas:
Thanks @jomarko

I'd like to answer your questions with some background, and a few questions of my own.

**What I'm attempting to fix**

Using an E2E test screenshot from the repo as an example: connection layouts could be optimised, as BAs tend to model BPMN diagrams left-to-right / top-down using horizontal connections where they can. These connections are fixed to their anchor points, so if you move the nodes the layout doesn't auto-optimise, and the only way to tidy it up is to re-anchor or recreate each connection using the centred anchor points.

https://github.com/apache/incubator-kie-tools/blob/main/packages/bpmn-editor/tests-e2e/__screenshots__/Google-Chrome/flowElements/complete-process-with-call-activity.png?raw=1

**Connection layout algorithms**

Three common connection layouts:

Screenshot 2026-07-09 at 11 30 16

**Option 1 - straight segments, no diagonals**
Pretty common for sequence flows in the main process. Clean layout with only horizontal/vertical segments that stick to the grid, though each connection ends up with several bend points.

**Option 2 - facing-side midpoint (recommendation)**
This is already the default in places - e.g. if you create a connected node from the popup quick menu, the connection uses the centred anchor point on both source and target, and always attaches at the midpoint of the facing side. Connections to text annotations and data objects commonly use this too. This is the algorithm the PR would like to standardise across all node types (focusing on the source nodes, and leaving placed connection anchors on target nodes as-is).

**Option 3 - shortest-path**
This is what you get today if you already have both nodes and draw a connection between their centre anchors. Similar to option 2 but allows shorter, diagonal lines, so the anchor point isn't always the facing-side midpoint. Worth noting that both option 2 and option 3 are currently implemented, so it might be good to agree on what the centre anchor point should mean.

**Design questions**

- What should using the centre anchor point mean - always find the best facing side to attach to, or always try to create the shortest connection?
- Should the centre anchor point algorithm be standardised across the different ways of creating connections?
- When moving nodes that have placed anchor points, should those be optimised (facing-side midpoint or shortest-path), and should the optimisation also apply to the connected node's anchor point?

jomarko:
Thank you for a detailed comment @ch-lukas .

I think I now understand properly the changes.

### Before
Screenshot 2026-07-13 092616

If user set the source magnet point to the green one, then moving the source node had no effect to the magnet point. Always the green one was in use.

### Now
Screenshot 2026-07-13 092616-2

Even if user specified originally the green source magnet point, moving the source node will cause any of the orange magnet point can start to be in use to have 'the most optimal' connection.

### Feedback
However, I can not confirm only the source magnet point is recomputed.
image

In the screenshot above, when I start to move **middle** node, both **Edge A** target point and **Edge B** source point are dynamically recomputed to have best connection.

**My personal opinion**
- I like this ✅ , is this acceptable to other KIE members ❓ I will try forward this PR to more engineers to have their opinion

**Doubts**
- Are then the all magnet points (the blue points when however a node) needed anymore?
- Are we able to add a playwright test for this?

jomarko (commented):
My review is basically my comment https://github.com/apache/incubator-kie-tools/pull/3652#issuecomment-4955611620

yesamer:
@ch-lukas Can you please rebase and resolve the conflicts?

ch-lukas:
Thanks @jomarko - you've read it exactly right, including that it's both endpoints of the moved node that recompute, not just the source. When a node moves, the anchors on all of its connected edges get re-evaluated.

**Are all the magnet points still needed?**
Yes, I'd keep all five. They still decide where a connection attaches when you draw it.

**A Playwright test?**
Added. It loads a small prebuilt process and then moves a task around.

Happy to adjust based on what the other engineers make of it.

packages/bpmn-editor/tests-e2e/flowElements/connectionReanchorOnNodeMove.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  return (endpoint["@_x"] - bounds["@_x"]) / bounds["@_width"];
+}
+
+test.beforeEach(async ({ page, baseURL }) => {
+  await page.goto(`${baseURL}/iframe.html?args=&id=misc-connection-reanchor--start-to-end&viewMode=story`);
+  await expect(page.getByTestId("kie-bpmn-editor--diagram-container")).toBeVisible();
+});
+
+test.describe("Connection - re-anchor on node move", () => {
```
jomarko, line 57: The only feedback I have is maybe we could add documentation for the test, some ASCI ART that show what was original diagram and what it was after the drag operation.

jomarko (approved):
Thank you for the PR, I think it works properly. I didn't notice issues during manual check. I tried to drag nodes as in diagram root level also nodes nested in subprocess.

PR #3663, opened 2026-07-15
NO-ISSUE: CVE-2026-13676 fast-uri upgrade to 3.1.3

harshmittalibm:
Summary:

Fixed the CVE-2026-13676 by fast-uri package upgrade to 3.1.3.

CVE Remediated:

[CVE-2026-13676](https://github.com/fastify/fast-uri/security/advisories/GHSA-4c8g-83qw-93j6)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -29,6 +29,7 @@ overrides:
   uuid: ^11.1.1
   picomatch@3: 3.0.2
   picomatch@4: 4.0.4
+  '@cypress/request>form-data': 4.0.6
 
 packageExtensionsChecksum: sha256-oxPwESKKSHRelJQnCQTHzgtG1xkcQOHjfgjFdIfqMfg=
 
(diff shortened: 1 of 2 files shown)


thiagoelg (changes requested):
THanks for the PR @harshmittalibm

Please, follow the intrcutions in [DEPENDENCY_MANAGEMENT.md](https://github.com/apache/incubator-kie-tools/blob/main/repo/DEPENDENCY_MANAGEMENT.md).

Which dependency bring `fast-uri` transitevely? Can it be updated instead? Overrides should be a last resort.

harshmittalibm:
Hi @thiagoelg, 

The fast-uri dependency is brought by ajv package. 
ajv@8.20.0 declares "fast-uri": "^3.0.1" — a range that includes both 3.1.2 (vulnerable) and 3.1.3 (patched). Since ajv itself is already at its latest (8.20.0), upgrading ajv is not an option. 

In summary - 
- The direct consumers (ajv) are already at their latest versions.
- The vulnerability is in a transitive dep whose version range in ajv happens to permit both safe and unsafe versions.
- The override pins it firmly to 3.1.3 across all consumers regardless of how pnpm resolves the range in the future.

thiagoelg:
> The direct consumers (ajv) are already at their latest versions.

Great!

> The vulnerability is in a transitive dep whose version range in ajv happens to permit both safe and unsafe versions.

Which means we can update the transitive dependency to the safe version without overriding.

> The override pins it firmly to 3.1.3 across all consumers regardless of how pnpm resolves the range in the future.

Dependencies can't be easily downgraded once updated in the lockfile. Overriding is not necessary.

As I said, overriding is a last-resort solution.

To update fast-uri locally, I did:

```shell
pnpm update -r fast-uri
pnpm dedupe
pnpm prune
```

The last two commands help reduce the noise generated and keep dependency versions consistent across the lockfile.

harshmittalibm:
Hi @thiagoelg, I have updated the code as per your suggestions. Please take a look now.

PR #3612, opened 2026-06-03
NO-ISSUE: Update qs to 6.15.2 address CVE-2026-8723

Adarshvk98:
Adds a qs override to `6.15.2` to address CVE-2026-8723

PR #3669, opened 2026-07-17
NO-ISSUE: force global qs override to patched 6.15.2 to address CVE-2026-8723

Adarshvk98:
Follow-up to #3612, which added a scoped `@cypress/request@3>qs` override for CVE-2026-8723. That scope only covered one path, but a vulnerable `qs@6.14.2` is also reachable through `@kie-tools/cors-proxy → express@4.22.1 → qs`, which the scoped override doesn't affect.

This replaces the scoped override with a global one so every transitive qs resolves to the patched 6.15.2:

```
- "@cypress/request@3>qs": "6.15.2"
+ "qs": "6.15.2"
```

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14,7 +14,7 @@ overrides:
   minimatch@^3>brace-expansion: 1.1.16
   minimatch@^5>brace-expansion: 2.1.2
   openapi-types: 7.2.3
-  '@cypress/request@3>qs': 6.15.2
+  qs: 6.15.2
   path-to-regexp@^0: 0.1.13
   react-dropzone: ^11.4.2
   superagent: 10.2.2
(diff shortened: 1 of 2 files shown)


thiagoelg:
Thanks @Adarshvk98!

I'm curious about this change. No version was updated in the lockfile, meaning Express was already using `6.15.2`. So how can `6.14.2` be reached?

Adarshvk98:
Thank you @thiagoelg! Ah, I see it now. Both `@kie-tools/cors-proxy` and `@kie-tools/runtime-tools-management-console-webapp` declare express as `^4.22.1`, and that range already resolves to `express@4.22.2` in `pnpm-lock.yaml`, which pulls `qs@6.15.2`. Therefore, this change is unnecessary. I’m closing this PR.

PR #3678, opened 2026-07-23
NO-ISSUE: General dependecy upgrades and Podman support

thiagoelg:
## Summary

This branch updates dependencies across the workspace.

The main updates are:
- `@playwright/test`: from `1.45.2` to `1.62.0`
- `webpack`: from `5.107.1` to `5.108.4`
- `prettier`: from `3.3.2` to `3.8.5`
- `eslint`: from `8.52.0` to `8.57.1`

The file `pnpm-lock.yaml` has many other small updates.

This branch removes two patches. These patches are not necessary now:
- `immer@10.0.3.patch`
- `karma-firefox-launcher@2.1.2.patch`

This branch also updates the `zustand` patch. The new patch version is `4.5.7`.

The new Playwright version resolves actions faster. It also enforces ARIA rules more strictly. These two changes exposed old timing errors in the test code.

Also added support for Podman to be used as a replacement for Docker builds, as long as developers have the Docker Compatibility plugin enabled.

## Dependency changes

- `@playwright/test`: `1.45.2` to `1.62.0`. The file `packages/playwright-base/Containerfile` installs the same Playwright version for the browsers. This branch adds a comment there telling developers to keep the two versions equal.
- `webpack`: `5.107.1` to `5.108.4`
- `prettier`: `3.3.2` to `3.8.5`. This version is now fixed, not a range. `pretty-quick`: `4.0.0` to `4.2.2`
- `eslint`: `8.52.0` to `8.57.1`
- `@nice-move/prettier-plugin-package-json`, `@prettier/plugin-xml`, `@types/node`: small updates
- Removed patches: `immer@10.0.3.patch`, `karma-firefox-launcher@2.1.2.patch`. These packages do not need a patch at the new version.
- Updated patch: `zustand@4.4.2.patch` becomes `zustand@4.5.7.patch`
- Removed `dedupe-peer-dependents=false` from `.npmrc`. This setting is not necessary with the new pnpm versions.
- `pnpm-lock.yaml`: fully regenerated

## Test fixes required by the update

**scesim-editor**: Fixed a real error in the application. The file `TestScenarioDrawerDataSelectorPanel.tsx` set `aria-disabled={true}` on the `TreeView` wrapper at all times. The value did not change with the real state. Playwright checks the `aria-disabled` attribute on parent elements. So all actions inside the panel timed out under the new, stricter check. This PR also updates a set of old webkit screenshots. The screenshots were slightly out of date.

**dmn-editor**: The library `@reactflow/core`, version 11.11, stopped a specific action. In earlier versions, it kept each node's measured width and height between updates. Version 11.11 does not do this. So after each update, every node was briefly hidden. This hid the connection handles and the resize handle. As a result, the actions `dragNewConnectedEdge` and `nodes.resize()` timed out. This PR fixes the error in `computeDiagramData.ts`. The fix sets the width and height on each node directly. This branch also fixes several timing errors in the test code. These errors relate to node rename, drag actions, and the documentation panel.

**bpmn-editor**: Fixed similar timing errors in the test code. It also replaces a set of screenshot checks. The new checks compare the model and the DOM directly. This method is more reliable than a screenshot. This PR removes the old `.png` files for these checks.

**boxed-expression-component**: Fixed several timing errors in the test code for resize, selection, and parameter popover actions. The newer Playwright version dispatches events faster. This exposed the timing errors.

**online-editor**: Updated a set of webkit and chromium screenshots for the file-creation flow. These screenshots were out of date.

**storybook-base**: Added `allowedHosts: true` to the Storybook dev-server settings. The containerized e2e tests reach Storybook through an nginx proxy. The proxy changes the `Host` header. The default Storybook setting rejects this header. The new setting fixes this problem.

## Formatting

Lots of files were updated following the new rules imposed by prettier. A notable example is the package.json `repository.url` property that now requires the `git+` prefix.

## Known errors, not new and not fixed in this PR

- boxed-expression-component, test `resizing.spec.ts`, case "check resize on nested stories": This is an existing error in the reset logic of the application. This PR does not fix or skip this test.
- boxed-expression-component, test `keyboard.spec.ts`, line 75: This test is unstable. The instability is not related to this update. This branch does not investigate this test further.

## Testing performed

- The full e2e suites for `boxed-expression-component`, `bpmn-editor`, `dmn-editor`, `scesim-editor`, and `online-editor` now pass. The tests ran on chromium, webkit, and Google Chrome, in the containerized Playwright setup.
- The command `pnpm bootstrap` regenerated the lockfile and the code format.


.npmrc
```diff
@@ -1,2 +1 @@
-dedupe-peer-dependents=false # Without this, pnpm will always install all packages when `pnpm booststrap:root` runs. See https://github.com/pnpm/pnpm/issues/6300.
```
thiagoelg, line 1: Removed, as the [issue](https://github.com/pnpm/pnpm/issues/6300) was fixed in the current pnpm version.

tiagobento (approved):
Many changed files/lines, but an easy review. Thanks @thiagoelg

Kusuma04-dev (approved):
Thanks @thiagoelg , Great work!

PR #3679, opened 2026-07-23
NO-ISSUE: pin Maven version in mvnw bootstrap via build-dependencies-versions.json

AnnJoy23:
maven-wrapper-plugin:3.3.0 hardcodes Maven 3.9.9 as its default when invoked without -Dmaven=, causing maven-wrapper.properties to be generated with 3.9.9 which ships plexus-utils-3.5.1.jar in its lib/.

packages/maven-base/env/index.js: add mvnw.mavenVersion = '3.9.11'
packages/maven-base/index.js: pass -Dmaven=${env.mvnw.mavenVersion} to maven-wrapper-plugin so it generates maven-wrapper.properties with the pinned version instead of its hardcoded default

### DIFF
--- a/packages/maven-base/env/index.js
+++ b/packages/maven-base/env/index.js
@@ -18,6 +18,7 @@
  */
 
 const { varsWithName, composeEnv, getOrDefault } = require("@kie-tools-scripts/build-env");
+const buildDeps = require("../../../repo/build-dependencies-versions.json");
 
 module.exports = composeEnv([require("@kie-tools/root-env/env")], {
   vars: varsWithName({
@@ -42,6 +43,7 @@ module.exports = composeEnv([require("@kie-tools/root-env/env")], {
       },
       mvnw: {
         version: "3.3.0",
+        mavenVersion: buildDeps.maven,
       },
       maven: {
         deploy: {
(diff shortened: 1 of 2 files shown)


packages/maven-base/env/index.js
```diff
@@ -18,6 +18,7 @@
  */
 
 const { varsWithName, composeEnv, getOrDefault } = require("@kie-tools-scripts/build-env");
+const buildDeps = require("../../../repo/build-dependencies-versions.json");
```
tiagobento: I'm not the biggest fan of this reference here. I understand why you did this, but I guess the env/index.js file should actually be the source of truth. And the `../../../repo/build-dependencies-versions.json` shouldn't exist in the future.

AnnJoy23: @tiagobento 
Thank you for the clarification. Since build-dependencies-versions.json is planned for removal, introducing a new dependency on it would be counterproductive. I've reverted that change and retained the version directly in env/index.js, making it the single source of truth.

tiagobento (approved):

Kusuma04-dev (approved):

PR #3684, opened 2026-07-29
kie-issues#2379:Workflow Runtime: Live reload breaks in Quarkus Dev UI

mukeshsharmaibm:
**Description**
When running a Kogito/jBPM application in Quarkus dev mode, live reload breaks after ~60 seconds of inactivity with EntityManagerFactory is closed.

**Why it happens:** Quarkus dev mode hot-reload does not restart the JVM — it restarts the application inside the same JVM using a new classloader. When the old classloader is torn down, all JDBC connections to the H2 in-memory database are closed. H2's default behaviour is DB_CLOSE_ON_EXIT=TRUE, which registers a JVM shutdown hook that closes the in-memory database engine when it detects zero active connections. Since the JVM is still running, this hook fires asynchronously — roughly 60 seconds later. When the new classloader then tries to boot Hibernate and open a session, the H2 database engine is already gone, causing the crash.

**Why this fix works:** Adding ;DB_CLOSE_ON_EXIT=FALSE disables that JVM shutdown hook entirely. The in-memory database now survives for the full lifetime of the JVM process, across as many classloader restarts as Quarkus dev mode performs. The existing DB_CLOSE_DELAY=-1 prevents closure when the last connection is released (classloader teardown); DB_CLOSE_ON_EXIT=FALSE prevents closure by the exit hook. Both flags are required together for the database to remain stable across live reloads.

Issue link -> https://github.com/apache/incubator-kie-issues/issues/2379

### DIFF
--- a/examples/process-compact-architecture-springboot/src/main/resources/application-dev.properties
+++ b/examples/process-compact-architecture-springboot/src/main/resources/application-dev.properties
@@ -18,7 +18,7 @@
 #
 
 # Development profile: in-memory H2 database, no containers needed.
-spring.datasource.url=jdbc:h2:mem:default;NON_KEYWORDS=VALUE,KEY;DB_CLOSE_DELAY=-1
+spring.datasource.url=jdbc:h2:mem:default;NON_KEYWORDS=VALUE,KEY;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE
 spring.datasource.username=kie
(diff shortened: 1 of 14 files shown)


tiagobento (approved):

tiagobento:
Thanks @mukeshsharmaibm!

martinweiler (approved):

thiagoelg (approved):

PR #3992, opened 2026-09-04
NO-ISSUE: Upgrade qs to 6.16.0 to address CVE-2026-82417 and CVE-2026-82562

Adarshvk98:
## Change

```diff
- "@cypress/request@3>qs": "6.15.2"
+ "qs": "^6.16.0"
```

No `package.json` declares `qs`, so a root override is the only way to change its version. `pnpm-lock.yaml` was regenerated with `pnpm bootstrap:root`.

The previous entry was scoped to `@cypress/request` and set it to 6.15.2, while every other consumer resolved to 6.15.3 on its own. A single bare override moves all twelve to 6.16.0. Adding a bare `"qs"` next to the parent-scoped entry would not have worked, because pnpm applies the most specific matching key and `@cypress/request` would have stayed on 6.15.2.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14,7 +14,7 @@ overrides:
   minimatch@^3>brace-expansion: 1.1.18
   minimatch@^5>brace-expansion: 2.1.4
   openapi-types: 7.2.3
-  '@cypress/request@3>qs': 6.15.2
+  qs: ^6.16.0
   path-to-regexp@^0: 0.1.13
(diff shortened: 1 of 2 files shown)


Kusuma04-dev (approved):

jomarko (approved):
Changes seems reasonable to me. I did no manual testing.

yesamer (approved):

yesamer (approved):

PR #3993, opened 2026-09-04
NO-ISSUE: Upgrade js-yaml to 3.15.2 and 4.3.2 to address CVE-2026-84375

Adarshvk98:
## What

`js-yaml` 3.15.0, 3.15.1 and 4.3.0 are affected by CVE-2026-84375 (High, 7.5). `maxTotalMergeKeys` does not count empty mapping sources, so a small YAML document can cause prolonged CPU consumption. Fixed in 3.15.2 and 4.3.2.

## Change

Four override entries become two:

```diff
- "@istanbuljs/load-nyc-config>js-yaml": "^3.15.0"
- "json-refs>js-yaml": "^3.15.0"
- "xmlbuilder2>js-yaml": "^3.15.0"
- "js-yaml@^4": "^4.2.0"
+ "js-yaml@^3": "^3.15.2"
+ "js-yaml@^4": "^4.3.2"
```

The three parent-scoped entries left `@yarnpkg/parsers` uncovered, and it was resolving 3.15.1. A version-scoped key covers every 3.x consumer, including any added later. Replacing them matters, because pnpm applies the most specific key: keeping the old entries alongside a version-scoped one would have held those three parents at 3.15.0 and 3.15.1.

`xmlbuilder2@3.0.1` needs the override permanently: it declares exactly `3.14.0` and can never resolve higher on its own.


Kusuma04-dev (approved):

jomarko:
There are conflicts.

jomarko:
@Adarshvk98 thank you for PR. Dependabot is repeatedly opening PRs to bump `js-yaml`. Currently there is #3997. However it touches different files. I wonder what PR we need (your vs dependabot or both).

Adarshvk98:
Thanks @jomarko!

I'd suggest going with this PR rather than the Dependabot one, since it covers the CVE on **both major lines**:

- **js-yaml 4.x** — the `"js-yaml@^4": "^4.3.2"` workspace override resolves *all* 4.x copies (including the two direct dependencies #3997 bumps) to the patched 4.3.2.
- **js-yaml 3.x** — the vulnerable 3.x copies pulled in transitively via `@istanbuljs/load-nyc-config`, `json-refs` and `xmlbuilder2` are only pinned by these workspace overrides (`^3.15.2`). #3997 can't address them, since js-yaml 3.x isn't a direct dependency anywhere — merging it alone would still leave those transitive copies on an affected version.

If it helps, #3997 could still land afterwards to keep the two direct `package.json` declarations honest — the only overlap between the PRs is `pnpm-lock.yaml`, which resolves naturally when rebasing on top of the other.

Edit:  conflicts resolved

jomarko (approved):
Thank you

PR #4007, opened 2026-09-14
incubator-kie-tools#4006: [drl-vscode-extension] deprecated syntax lint

### LINKED ISSUE #4006: [drl-vscode-extension] feat: Linter for deprecated syntax patterns
https://github.com/apache/incubator-kie/issues/6220 removed some syntax patterns from the Antlr4 parser. In order to aid migration efforts, the VS Code extension should be able to flag these patterns for replacement, particularly as they may cause issues with some of the features in certain scenarios (see https://github.com/apache/incubator-kie-tools/pull/4005#issuecomment-5665734849 )

drccrd:
Add a lint warning when a syntax has been deprecated in the drools parser.
Closes: #4006


packages/drl-vscode-extension/syntaxes/drl.tmLanguage
```diff
@@ -42,6 +42,14 @@
 			<key>name</key>
 			<string>keyword.control.source.drools</string>
 		</dict>
+		<dict>
+			<key>comment</key>
+			<string>Word operators, built in and custom (##supersetOf), colored like the other DRL keywords</string>
+			<key>match</key>
+			<string>##[A-Za-z_][A-Za-z0-9_]*|\b(matches|contains|excludes|memberOf|soundslike|str|after|before|coincides|during|finishes|finishedby|includes|meets|metby|overlaps|overlappedby|starts|startedby)\b</string>
+			<key>name</key>
+			<string>keyword.control.source.drools</string>
```
drccrd, line 51: an alternative would have been "keyword.wordlike" but the coloring was then only the same as "true", "false", and "null" and not the other operators, so I think "control" fits better into the current scheme.

packages/drl-vscode-extension/syntaxes/drl.tmLanguage
```diff
@@ -12,9 +12,9 @@
 	<array>
 		<dict>
 			<key>comment</key>
-			<string>Mathmatical operators</string>
+			<string>Comparison and logical operators</string>
 			<key>match</key>
-			<string>[&lt;|&gt;|==|&lt;=|&gt;=|.|!=]</string>
+			<string>(==|!=|&lt;=|&gt;=|&lt;|&gt;|&amp;&amp;|\|\|)</string>
```
drccrd, line 17: not strictly part of this PR but they seemed to have been missing - the default VSCode themes don't handle these differently anyway.

packages/drools-lsp/drools-completion/src/main/java/org/drools/completion/CustomOperators.java
```diff
@@ -0,0 +1,141 @@
...
+ * engine. Operators registered from Java rather than configured, and ids a
+ * definition registers beyond its property key, are not seen.
+ */
+public final class CustomOperators {
+
+    private static final Logger logger = Logger.getLogger(CustomOperators.class.getName());
+
+    static final String PROPERTY_PREFIX = "drools.evaluator.";
+
+    static final String[] CONF_FILES = {
```
drccrd, line 56: this was previously missing and was a gap between the runtime compiler and the IDE, so so now syntax highlighting is possible too, and potentially other features in the future

packages/drools-lsp/drools-completion/src/main/java/org/drools/completion/DeprecatedConstructs.java
```diff
@@ -0,0 +1,329 @@
...
+        return found;
+    }
+
+    /**
+     * Names the deprecated construct at {@code index}, or {@code null} when the
+     * tokens there are an ordinary parse failure. Order matters: a {@code &&}
+     * before a relational operator is a half constraint, while the same token
+     * after a closing paren joins two patterns.
+     */
+    private static Finding describe(List<Token> tokens, int index, boolean inWhen,
```
drccrd, line 179: currently hand maintained but I don't think there is a flag for them in the drl-parser other than the warning messages - so better than nothing

packages/drl-vscode-extension/syntaxes/drl.tmLanguage
drccrd: also the built-in comparators are not strictly part of this PR but testing without the syntax highlight was a little annoying...

PR #3985, opened 2026-08-28
kie-issues#2395: DMN Editor: Palette Icon Visibility Across Resolutions

rajalakshmys-27:
Closes : [kie-issues#2395](https://github.com/apache/incubator-kie-tools/issues/3986)

Currently, the palette icons are fully visible only on larger screens (height ≥ 758px).

Fix: Implemented a responsive submenu for the DMN editor palette that dynamically shows/hides icons based on available viewport space, improving usability on smaller screens (height ≥ 430px).

**Submenu Strategy**

The palette uses a flat list of 5 primary icons (`primaryIcons`), all eligible for submenu overflow:

- Icons that fit in the available space remain visible in the palette.
- Overflow icons are hidden behind an ellipsis button (`⋯`).
- The secondary palette groups (Group, Text Annotation, DRG Nodes, External Nodes) always remain visible.

**Changes Made**

*Responsive Calculation*
- Implemented `calculateVisibleIconCount()`.
- Dynamically calculates how many icons fit in available viewport space

*Responsive Behaviour*
- Added window resize listener with `requestAnimationFrame` for smooth updates.

*Submenu UI*
- Ellipsis button (`⋯`) appears when icons overflow.
- Grid-based submenu displays hidden icons.
- Submenu only shown when `visibleIconCount < totalIcons` (derived, no extra state).
- `renderPaletteIcon` callback eliminates icon rendering duplication.

**Testing**
- Verified responsive behaviour across different viewport sizes.
- Submenu opens/closes correctly.
- All icons remain draggable from both palette and submenu.

image

PR #4020, opened 2026-09-21
[NO-ISSUE] Fix CI pipeline failures: stale DMN Editor screenshots and XML attribute order on Chrome 153

nrknithin:
`main` has been failing on Linux CI since Sep 9. This PR fixes both causes.

### 1. Stale WebKit screenshots (DMN Editor)

Three snapshots left stale by the palette layout change in #3985, regenerated.

### 2. Attribute order depends on the browser (`xml-parser-ts`)

Chrome 153 changed `DOMParser` to return `xmlns` declarations sorted alphabetically in
`Element.attributes`. `xml-parser-ts` builds its JSON from that order and serializes back in the
same order, so `getContent()` returned the DMN with its namespaces reordered and the round-trip
test failed.

The parser now takes the attribute order from the XML text rather than from the DOM, falling back
to the DOM order when the two cannot be matched. It sits in the shared parser, so all marshallers
and editors pick it up.

### Testing

- `xml-parser-ts` suite green, including new tests for the ordering.
- Verified in the CI container (Ubuntu + Chrome 153): the previously failing `getContent` tests
  now pass.


packages/xml-parser-ts/tests/attributesSourceOrder.test.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  test("applies to nested elements too", () => {
+    const xml = `<rootElement xmlns="https://kie.apache.org/my-schema-1.0"><childElement xmlns:q="urn:q" data-b="1" xmlns:p="urn:p" data-a="2">text</childElement></rootElement>`;
+    const domdoc = domParser.getDomDocument(xml);
+    reorderAttributes(domdoc.documentElement.firstElementChild!, likeChrome153);
+    expect(Array.from(domdoc.documentElement.firstElementChild!.attributes).map((a) => a.name)).toEqual([
+      "xmlns:p",
+      "xmlns:q",
+      "data-b",
+      "data-a",
```
jomarko, line 131: why `data-b` is before `data-a`?

jomarko (commented):
Single question to code, not sure I understand the sorting we have when reading the test. maybe I am missing something?

packages/xml-parser-ts/tests/attributesSourceOrder.test.ts
nrknithin: Good question @jomarko . Nothing is being sorted there — that block is checking what Chrome 153 gives us, not what our parser produces.

The source XML deliberately has `data-b` before `data-a`, so the test would fail if ordinary attributes were sorted alphabetically.

Chrome only reorders the namespace declarations and moves them to the front. So:

`xmlns:q, data-b, xmlns:p, data-a`

becomes:

`xmlns:p, xmlns:q, data-b, data-a`

The actual assertion on line 140 verifies that our parser recovers the original source order.


Basically its just to test our logic

yesamer (approved):

packages/xml-parser-ts/tests/attributesSourceOrder.test.ts
jomarko: thank you

jomarko (approved):
Thank you for the fix

tiagobento (changes requested):
Sorry but what is the problem with alphabetical sorting of xmlns attributes?

I don't really agree with us rolling out our own XML parser.

nrknithin:
@tiagobento I've dropped the custom parsing — the PR is only the test change now.
And you're right that the sorting itself isn't the problem: the test was comparing byte-for-byte, which is stricter than anything we actually guarantee, so it now ignores attribute order and compares everything else exactly.

tiagobento (approved):
Thanks @nrknithin. A little hacky but since this is testing code we can revisit it in the future. Approving to unblock the CI right now.

